PR #380, opened 2014-12-17
RFC: Fixes required to build for Android 5 on AArch64

pao:
I don't expect these to be in the right shape for merging (except possibly the first one, which I hope is uncontroversial), but since I'm about to be minimally available for over a month I thought I'd at least get the patches out there.

The first patch replaces `index` with `strchr`. The bionic libc defines `index` as part of the [`ndk_cruft.cpp`](https://github.com/android/platform_bionic/blob/d0944d9ed38be0d74c3d8876f0f36b9ddc0ce38c/libc/bionic/ndk_cruft.cpp#L301) file, whose definitions are disabled on 64-bit platforms.

The second patch updates the `lxcmntent` compatability implementations to remove most of them. They are defined by the latest version of bionic (API 21). These would be more properly gated by a feature test--I'm just not sure if there is one. They might also be able to be gated by a particular bionic version, but I haven't looked into how that works, either.

The third patch bluntly hacks in the `-pie -fPIE` flags while compiling the lxc executable files. Starting with Android 5, you can't run non-posiiton-independent binaries.

Even with all this, LXC doesn't quite work--I can't actually mount multiple cgroups to a single mount point, so that part of `prepare.sh` fails.

(cf https://github.com/flar2/ElementalX-N9/pull/1)


stgraber:
Sorry for the massive delay. In preparation for LXC 1.1, I've pushed the first of your patch to master since that one is obviously right. The other two I really need to attempt an arm64 build by myself to get a better idea of how to handle them since as they stand, they clearly break all other architectures.

stgraber:
Marking this pull request as incomplete until we have a solution which works for all architectures.

pao:
Sounds right to me. I'm back from my long travel, and am glad to help play with things and test as needed.

stgraber:
LXC has changed a fair bit since this was first opened. I've not heard of anyone else who's got interest in getting arm64 android working and this branch doesn't apply anymore.

Closing this for now, an updated set of patches for android arm64 would be appreciated, ideally, well split and not breaking other architectures or C libraries in the process.

PR #391, opened 2014-12-30
improve "lxc-create -t debian -h" help text, add --keep option, fix option parsing

tpo:
Unfortunately Github can't create pull requests one on top of the other. So here are the descriptions of my commits, each on top of the previous, implementing a different thing:

72c0e81:

- document environment variables
- add missing --packages switch to command line
- describe how to pass template options to lxc-create (since lxc-create -h doesn't tell you)
- render help text in the same pretty format as lxc-create does

bb120e2:

- allow to keep the cache of the downloaded packages
- With this option, the user doesn't have to download all the required packages over and over again, when the download fails half way through for some reason, or when he wants to lxc-create another VM.

843aebc:

- fix option assignment. AFAIK it was broken. See commit text.
 
Please pull, thanks.

Signed-off-by: Tomáš Pospíšek <tpo_deb@sourcepole.ch>

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -225,8 +225,12 @@ configure_debian_systemd()
 
 cleanup()
 {
-    rm -rf $cache/partial-$release-$arch
-    rm -rf $cache/rootfs-$release-$arch
+    if [ -z "$keep_cache" ]; then
+        rm -rf $cache/partial-$release-$arch
+        rm -rf $cache/rootfs-$release-$arch
+    else
+        echo "kept the cache"
+    fi
 }
 
 download_debian()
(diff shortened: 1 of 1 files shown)


stgraber:
Can you please sign off your commits (per CONTRIBUTE)? Thanks

tpo:
I've added the signed off line at the bottom of my pull request, is that OK/as intended? Thanks,

stgraber:
Ah, that's not the usual way of doing it (the commit text should have it) but that's alright, I can edit your commits when pushing. Removing the incomplete tag.

stgraber:
Based on the comments I left on your commits, I'll be applying the first and last commit and skip the broken one. If I'm wrong and the cache is indeed always wiped somehow, then that's a separate bug and we'll have to look into it.

stgraber:
Actually re-opening as none of the commits appear to apply to current LXC git...

stgraber:
Looking at it closer, your first patch still makes sense, the second as I said earlier, doesn't and the third was already fixed by Michael Adam's patchset.

Would be great if you could update this pull request to only include a refreshed version of your first patch, or just create a new pull request if that's easier!

Thanks

stgraber:
Closing this one as it got superseded and merged.

PR #393, opened 2015-01-08
apparmor: Fix slave bind mounts

martinpitt:
The permission to make a mount "slave" is spelt "make-slave", not "slave", see
https://launchpad.net/bugs/1401619. Also, we need to make all mounts slave, not
just the root dir.

https://launchpad.net/bugs/1350947


hallyn:
Thanks, Martin.

PR #397, opened 2015-01-13
Fix instantiation of multiple vlan interfaces with same id

axn:
Container fail to start with configs (as shown below) where the same
vlan id is used for several type=vlan container interfaces.
Then, during the instantiation of the vlan interfaces, an error occurs
because the lxc code tries to assign the same temporary name to both
of them before it is bound into the container.

> lxc.network.type      = vlan
> lxc.network.flags     = up
> lxc.network.link      = eth1
> lxc.network.vlan.id   = 3842
> lxc.network.name      = iso0
>
> lxc.network.type      = vlan
> lxc.network.flags     = up
> lxc.network.link      = eth2
> lxc.network.vlan.id   = 3842
> lxc.network.name      = iso1

Signed-off-by: Axel Neumann <neumann@cgws.de>

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2979,13 +2979,14 @@ static int instantiate_vlan(struct lxc_handler *handler, struct lxc_netdev *netd
 {
 	char peer[IFNAMSIZ];
 	int err;
+	static uint16_t vlan_cntr = 0;
 
 	if (!netdev->link) {
 		ERROR("no link specified for vlan netdev");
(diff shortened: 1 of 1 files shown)


axn:
The problem and fix has also been discussed in lxc-devel mailing list some time ago :-)
https://lists.linuxcontainers.org/pipermail/lxc-devel/2012-July/002584.html

hallyn:
Thanks for the reminder.  Merging

Issue #409, opened 2015-01-22
ubuntu template fails when keyring file is not installed


Wuodan:
The ubuntu template calls debootstrap which uses:
/usr/share/keyrings/ubuntu-archive-keyring.gpg
If that file is missing, debootstrap defaults to debian mirror thus fails to download the ubuntu release.

LXC is not the cause of this issue, rather the victim :-)
But at least 3 others reported the same issue - seems the problem occured on 3 distros (RedHat, Debian, Gentoo).

Solution proposal 1:
If that file is missing, call debootstrap with a mirror, for ex. "http://archive.ubuntu.com/ubuntu"

Solution proposal 2:
wget that file to /var/cache/lxc/keyrings and call debootstrap with --keyring=file.

Links to similar reports:
https://bugzilla.redhat.com/show_bug.cgi?id=1115728
https://lists.debian.org/debian-boot/2014/01/msg00390.html
http://readthefuckingmanual.net/error/5342/

Output of `lxc-create -t ubuntu -n gaga 2>&1`:

Checking cache download in /var/cache/lxc/precise/rootfs-amd64 ... 
dpkg-query: error: failed to open package info file `/var/lib/dpkg/status' for reading: No such file or directory
Installing packages in template: ssh,vim,language-pack-en
Downloading ubuntu precise minimal ...
which: no qemu-debootstrap in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin:/usr/x86_64-pc-linux-gnu/gcc-bin/4.8.3:/opt/cuda/bin:/usr/sbin:/usr/bin:/sbin:/bin)
I: Keyring file not available at /usr/share/keyrings/ubuntu-archive-keyring.gpg; switching to https mirror https://mirrors.kernel.org/debian
I: Retrieving Release 
E: Failed getting release file https://mirrors.kernel.org/debian/dists/precise/Release
lxc_container: lxccontainer.c: create_run_template: 1136 container creation template for gaga failed
lxc_container: lxc_create.c: main: 280 Error creating container gaga

stgraber:
Marking as incomplete as we're waiting for a sign-off in the pull request.

starkers:
still broken on centos :(

gabri94:
broken also on debian

hallyn:
@Wuodan 

Please see Stéphane's comment - could you please sign off your commit?

benbenolson:
[...]

PR #410, opened 2015-01-22
ubuntu template fails when keyring file is not installed - wrt issue #409

Wuodan:
Please see issue #409 - these are the possible solutions that I proposed there.
Proposal 1 is in commit 1, proposal 2 in commit 2.

Other minor changes:
- `which qemu-debootstrap` is not quiet

- echo "Failed to download the rootfs, aborting."
Was shown depending on the previous if, not on result of debootstrap. (`set -e` prevents it from being displayed anyways ...)


stgraber:
Can you please sign-off your commits (per CONTRIBUTE)?

stgraber:
Been over a month without a reply, closing.

PR #412, opened 2015-01-23
Gentoo template - Add a hwaddr if there is only one veth

vriera:
Reuse the code from the Debian template to associate a hwaddr if there
is only one veth interface in the container's config file.

Signed-off-by: Vicente Olivert Riera <Vincent.Riera@imgtec.com>

### DIFF
--- a/templates/lxc-gentoo.in
+++ b/templates/lxc-gentoo.in
@@ -681,6 +681,13 @@ container_conf()
     #at this point if there
     conf_file="${path}/config"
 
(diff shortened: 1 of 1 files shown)


stgraber:
I think that going forward (LXC 1.2) we'll want to rip all of that code out of the templates and just ensure that all distros have an appropriate /etc/lxc/default.conf containing a MAC address template, but for now, that's not wrong, so applied. Thanks!

PR #415, opened 2015-01-24
Deletion of unnecessary checks before some function calls

elfring:
* [The function "free"](http://stackoverflow.com/questions/18775608/free-a-null-pointer-anyway-or-check-first "Free a null pointer anyway or check first?") is documented in the way that no action shall occur for a passed null pointer. Redundant safety checks can be avoided.
* The following functions perform also input parameter validation.
  * [container_destroy](https://github.com/lxc/lxc/blob/d2cf4c378588cc1d497fe8b2ba3f835d6b03fe38/src/lxc/lxccontainer.c#L2062 "Destroy a container.")
  * [lxc_cgroup_process_info_free_and_remove](https://github.com/lxc/lxc/blob/ec64264d78d4ed608553842ce9e1f07eeab2a032/src/lxc/cgfs.c#L1186 "Free process membership information and remove control groups.")
  * [lxc_cgroup_put_meta](https://github.com/lxc/lxc/blob/ec64264d78d4ed608553842ce9e1f07eeab2a032/src/lxc/cgfs.c#L598 "Put meta.")
  * [toss_list](https://github.com/lxc/lxc/blob/67c660d0aaff5f2854a55da936fe6cd82510494f/src/lxc/lxc_autostart.c#L310 "Toss a list.")

  It is therefore not needed that a function caller repeats a corresponding check.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -108,8 +108,7 @@ static struct lxc_proc_context_info *lxc_proc_get_context_info(pid_t pid)
 		}
 	}
 
-	if (line)
-		free(line);
+	free(line);
 	fclose(proc_file);
 
 	if (!found) {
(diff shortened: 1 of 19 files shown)


stgraber:
Applied, thanks!

PR #437, opened 2015-02-08
Enable fuse filesystems for containers

brauner:
This enables containers to run a fuse filesystems per default. The mount
is designed to be safe. Hence, it can be enabled per default in
common.conf. It will lead to a cleaner boot for some unprivileged
systemd-based containers.

### DIFF
--- a/config/templates/common.conf.in
+++ b/config/templates/common.conf.in
@@ -45,6 +45,7 @@ lxc.cgroup.devices.allow = c 136:* rwm
 
 # Setup the default mounts
 lxc.mount.auto = cgroup:mixed proc:mixed sys:mixed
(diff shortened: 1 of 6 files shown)


stgraber:
So I think this patch ought to do a few more things to properly enable fuse:
 - Remove the same line from the ubuntu and debian configs since they inherit from common anyway
 - Also move the fuse lxc.cgroup.devices.allow lines from the debian/ubnutu configs to the common one

brauner:
Ah, of course. Sorry.

stgraber:
I squashed the two commits together and updated the description a tiny bit, this has now been pushed to git master, closing.

Thanks!

PR #447, opened 2015-02-21
templates/lxc-gentoo.in: Fix stage3 autodetection code

hwoarang:
The latest-stage3-$variant.txt files list both the tarball name
and the released date but we only need to pass the stage3 tarball
name to the subsequent wget command otherwise we end up with 404
errors.

Signed-off-by: Markos Chandras <hwoarang@gentoo.org>


hwoarang:
Oh I see something similar was posted in the mailing list

https://lists.linuxcontainers.org/pipermail/lxc-devel/2015-February/011402.html

So I am closing this pull request

PR #451, opened 2015-02-25
Lxccreate: add a backing store type (bdevtype) to a python create function

pymag09:
The main purpose of this commit is to add a backing store type (bdevtype) to a python create function (like lxc-create -B). If you find it is good enough, please approve my merge request.

Signed-off-by: Bieliaievskyi Sergey magelan09@gmail.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,4 @@ notifications:
   email:
     recipients:
       - lxc-devel@lists.linuxcontainers.org
-  irc:
-    channels:
-      - "chat.freenode.net#lxcontainers"
-    use_notice: true
+  webhooks: https://linuxcontainers.org/webhook-lxcbot/


src/python-lxc/lxc/__init__.py
```diff
@@ -217,22 +217,18 @@ def create(self, template=None, flags=0, args=()):
...
-            return _lxc.Container.create(self, template=template,
-                                         flags=flags,
-                                         args=tuple(template_args))
-        else:
-            return _lxc.Container.create(self, flags=flags,
-                                         args=tuple(template_args))
+                tmp_args.append("--%s" % item[0])
+                tmp_args.append("%s" % item[1])
+        template_args = {}
+        template_args['template'] = template
```
stgraber: This is now always passing template to Create instead of only passing it if it's set.

stgraber:
Besides the comment I left, this looks fine. However before we can include any change in LXC, the commits must contain the Signed-off-by line (as per CONTRIBUTING.md).

I'd also recommend you squash both commits together.

Thanks!

PR #460, opened 2015-03-04
gentoo.common.conf improvements

hwoarang:
Hi,

A couple of patches in order to support /dev/shm as tmpfs which is needed for building python in a Gentoo LXC. This will make the Gentoo LXC more useful for developer (and users) who use it to build and test Gentoo ebuilds.

### DIFF
--- a/config/templates/gentoo.common.conf.in
+++ b/config/templates/gentoo.common.conf.in
@@ -24,3 +24,7 @@ lxc.cgroup.devices.allow = c 10:232 rwm
 
 # /dev/ is created by template
 lxc.autodev = 0
+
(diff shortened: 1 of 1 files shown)


stgraber:
We've been standardizing around using lxc.autodev by default everywhere so that unprivileged containers are guaranteed to work everywhere in a consistent way.

Would dropping your lxc.autodev = 0 change and instead adding create=dir to the dev/shm mount option work?

hwoarang:
I see. Sorry I didn't know about that

Yes create=dir seems to work fine. I will submit a new pull request. Thanks

PR #461, opened 2015-03-04
autostart: Fix bug with containers being started in reverse order

bostjan:
As configuration parameter is called lxc.start.ORDER, one expects that container with order setting "1" will start before container with order setting set to "2", but this is not the case. Containers start in reverse order. If that is the desired behaviour, then I believe "order" should be called "priority" or something similar.

Anyway, I looked into src/lxc/lxc_autostart.c, and found this:

In main() there is this code:
qsort(&containers[0], count, sizeof(struct lxc_container *), cmporder);

In cmporder() the last comparison looks like this:
return (c1_order - c2_order) * -1;

qsort() sorts elements from lower to higher value, using callback cmporder(). Therefore cmporder() should return negative value if container c1 should be started before c2. That "* -1" achieves exactly the oposite result, as it causes containers to be sorted in descending order regarding to their order setting.

This pull request fixes the behaviour. See mailing list for additional discussion about multi-container operation ordering.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,4 @@ notifications:
   email:
     recipients:
       - lxc-devel@lists.linuxcontainers.org
-  irc:
-    channels:
-      - "chat.freenode.net#lxcontainers"
-    use_notice: true
+  webhooks: https://linuxcontainers.org/webhook-lxcbot/


stgraber:
Hi,

Per CONTRIBUTING.md, can you update your commit to include the required Signed-off-by line?

Thanks!

bostjan:
Done, but created another pull request to pull from bugfix/* branch instead of master.
https://github.com/lxc/lxc/pull/465

PR #463, opened 2015-03-09
lxc-top: limit container name to 18 characters

bostjan:
Currently, if container name is fqdn (which frequently exceeds 18 characters), this breaks the output of lxc-top and misaligns the columns. This patch fixes this by limiting container name to 18 characters, which is what lxc-top currently assumes as maximum for container name.

### DIFF
--- a/src/lxc/lxc_top.c
+++ b/src/lxc/lxc_top.c
@@ -303,7 +303,7 @@ static void stats_print(const char *name, const struct stats *stats,
 	size_humanize(stats->blkio, blkio_str, sizeof(blkio_str));
 	size_humanize(stats->mem_used, mem_used_str, sizeof(mem_used_str));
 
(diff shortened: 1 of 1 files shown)


dwengen:
Hi, a much simpler change would be to just change the printf specifier to %18.18s.

bostjan:
It's actually %-18.18, but you are right.

I am not that proficient in printf syntax (nor C for that matter), and after a couple of tries with dot and minus I failed to hit the right syntax, thus I assumed it is not supported to put two modifiers for the same string.

Fixed, pull request is updated.

stgraber:
Hi,

Per CONTRIBUTING.md, can you update your commit to include the required Signed-off-by line?

Thanks!

bostjan:
Done.

stgraber:
Thanks!

bostjan:
Heh, this pinky "incomplete" icon next to message title made me curious why it's not green :)

PR #464, opened 2015-03-10
 Lxccreate: add a backing store type (bdevtype) to a python create function

pymag09:
The main purpose of this commit is to add a backing store type (bdevtype) to a python create function (like lxc-create -B)

Signed-off-by: Bieliaievskyi Sergey magelan09@gmail.com


stgraber:
Thanks!

PR #468, opened 2015-03-12
templates: gentoo.common: Add /dev/shm tmpfs mount entry

hwoarang:
/dev/shm needs to be mounted as tmpfs. It's needed by python
and possibly other packages.

Signed-off-by: Markos Chandras <hwoarang@gentoo.org>


stgraber:
thanks!

PR #469, opened 2015-03-17
Allow veth that is not attached to a bridge on unprivileged container

fajarnugraha:
This PR is a follow up to https://www.mail-archive.com/lxc-users@lists.linuxcontainers.org/msg02624.html

Short summary, I suggest the following enhancement steps for unprivileged containers to allow veth networking with proxyarp/non-bridge setup:

1. Allow empty lxc.network.link with lxc.network.type = veth for unprivileged container, so that user can create a veth interface not attached to any bridge
2. Allow custom veth name on the host side
3. Allow static IP address definition on container config file

This PR only implements (1)

Tested on top of current master on ubuntu 14.04.2 host and container. I use this setup

```
$ cat /etc/lxc/lxc-usernet
# USERNAME TYPE BRIDGE COUNT
user veth lxcbr0 2
user veth none 2

$ grep link ~/.local/share/lxc/u1/config
#lxc.network.link = lxcbr0
```

with lxc.network.link commented-out, ``lxc-user-nic`` will be called with ``none`` as bridge, and a veth interface will be created that is not attached to any bridge.

veth count quota works, normal ``lxc.network.link = lxcbr0`` setup also still works.

Signed-off-by: Fajar A. Nugraha <github@fajar.net>

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2979,6 +2979,7 @@ static int unpriv_assign_nic(struct lxc_netdev *netdev, pid_t pid)
 	int bytes, pipefd[2];
 	char *token, *saveptr = NULL;
 	char buffer[MAX_BUFFER_SIZE];
+	char netdev_link[IFNAMSIZ+1];
 
 	if (netdev->type != LXC_NET_VETH) {
 		ERROR("nic type %d not support for unprivileged use",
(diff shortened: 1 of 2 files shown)


stgraber:
I expect 2 and 3 to be tricky from a security point of view, but 1 is fine and may still be useful on its own, so merging this.

PR #488, opened 2015-04-12
Use the non-obsolete spelling

lu-zero:


stgraber:
Hi,

Per CONTRIBUTING.md, can you update your commit message to include the required Signed-off-by line?

lu-zero:
I just used the quickedit feature from github since it is a single letter change and not creative at all, I'll fold this patch in a set with more consistent stuff hopefully later. Thanks for the pointer!

PR #489, opened 2015-04-13
vgname support for container cloning

brsyuksel:
i have added vgname feature for container cloning, and some api defines has been changed at the same time. i am newbie for lxc internals (and system programming), maybe i have been made some mistake about fssize or fstype, sorry :)

### DIFF
--- a/config/init/common/lxc-net.in
+++ b/config/init/common/lxc-net.in
@@ -1,7 +1,6 @@
 #!/bin/sh -
 
 distrosysconfdir="@LXC_DISTRO_SYSCONF@"
-localstatedir="@LOCALSTATEDIR@"
 varrun="@RUNTIME_PATH@/lxc"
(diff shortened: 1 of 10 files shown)


brsyuksel:
oops, sorry for tests :(

stgraber:
Hi,

A few comments:
  - Per CONTRIBUTING.md, all commits must include the Signed-off-by line
  - Looks like you've merged/rebased your patchset a few times, you may want to rebase it again and merge some of those commits together to make things easier to review
  - You're changing the LXC public API by modifying lxcapi_clone. Please don't do this unless absolutely required as such a change would require us to bump liblxc to version 2, breaking all backward compatibility in the process. It'd be better to try to fit this new use case using the existing arguments, or if not at all possible, then possibly introduce a new API call.

PR #491, opened 2015-04-17
Fix case where we have config like lxc.network.link

syed:
Currently get_config_item works for `network` only if it is the form `lxc.network.0.XXX` but we can have an option like `lxc.network.XXX` which should be interpreted correctly.

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -403,7 +403,8 @@ static int get_network_netdev_idx(const char *key)
 	int ret, idx;
 
 	if (*key < '0' || *key > '9')
-		return -1;
+		return 0; //it is of the form lxc.network.link
+
 	ret = sscanf(key, "%d", &idx);
 	if (ret != 1)
 		return -1;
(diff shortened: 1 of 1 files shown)


caglar10ur:
FYI https://github.com/lxc/go-lxc/issues/37

src/lxc/confile.c
```diff
@@ -403,7 +403,8 @@ static int get_network_netdev_idx(const char *key)
 	int ret, idx;
 
 	if (*key < '0' || *key > '9')
-		return -1;
+		return 0; //it is of the form lxc.network.link
```
stgraber: You probably want a more specific test, because your change also makes "lxc.network.a" a valid string, which it isn't.

stgraber:
I'm not convinced that this change is such a good idea.

I was actually considering changing LXC to no longer allow lxc.network.KEY and always require the interface number.

The reason for this is that get_config_item("lxc.network.type") with the following config will be confusing:

    lxc.network.type = veth
    lxc.network.link = lxcbr0
    
    lxc.network.type = veth
    lxc.network.link = br0

With the current code, you get an error.
With your change, you get the first entry.
However, for every other key, they override each other, so the user would expect the last one.

And all of those behaviors are basically wrong and will not return what the user expects.


So I think the best course forward is to deprecate that order-dependent network config syntax and instead require the interface number.

PR #501, opened 2015-04-25
Start error message fix1

krichter722:
added hint about option to use for foreground mode in start error message and introduced necessary string literal reusage (not too elegant for instance, but no need to change for all literals as long as there's no reusage - I can do it if you think it's valuable (open an issue and reference or assign me)).

### DIFF
--- a/src/lxc/lxc_start.c
+++ b/src/lxc/lxc_start.c
@@ -54,6 +54,15 @@
 #define OPT_SHARE_IPC OPT_USAGE+2
 #define OPT_SHARE_UTS OPT_USAGE+3
 
+/*
+define option names as constants in case they're reused
+*/
+// do this more elegant if http://stackoverflow.com/questions/29868286/how-to-reuse-a-literal-in-a-char-and-a-one-character-string-variable
+// is clearified
+#define FOREGROUND_OPTION_NAME 'F'
+#define FOREGROUND_OPTION_NAME_STRING "F"
+#define FOREGROUND_OPTION_NAME_LONG "foreground"
+
 lxc_log_define(lxc_start_ui, lxc);
 
 static struct lxc_list defines;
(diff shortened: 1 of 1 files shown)


stgraber:
Hi,

Your last commit is missing the Signed-off-by line.

stgraber:
Also the switch to constants is making things much harder to read and almost impossible for somebody to grep, so I'd prefer we don't do that.

Would you mind updating your branch to only updating the error message and dropping those constants?

Thanks!

krichter722:
Do you really want to maintain string literals in two places? I offer to convert everything to constants in order to increase readability by systematic.

stgraber:
Yes, I'd much rather have to update strings in two places if we ever change them (and we won't as that'd break every existing script out there) than break the ability of people to grep for strings in our code.

krichter722:
The constants are still grepable.

stgraber:
No they're not.
Before your change I could grep for "-F, --foreground", after it, I can't and I have to more cleverly grep for part of the string I see on my screen.

krichter722:
I added this as a comment. The code maintainability shouldn't suffer from such a rare (if not strange) use case for developers.

stgraber:
Apparently I wasn't clear enough...

I'm not going to merge this branch until those constants are gone.
That code change makes the code uglier, requires thinking to replace those constants with their value in your head and that for no gain as we're not going to change those arguments and even if we were, we'd still be grepping through the whole project to make sure we haven't missed anything.

krichter722:
> requires thinking to replace those constants with their value in your head
No. If you use constants correctly/consequently, you don't - that's the purpose of constants.

> we'd still be grepping through the whole project to make sure we haven't missed anything.
That sounds like a very inefficient, unreliable way of developing.

> I'm not going to merge this branch until those constants are gone.
I strongly recommend to get rid of the mentioned development techniques. Then we'll talk again. On the long term I could contribute such changes, but only after you're convinced they're an improvement. github.com offers gitter chat. We can move this conversation there if you want.

PR #506, opened 2015-04-28
macvlan: add 'passthru' mode

regit:
In setup where we want to sniff with an IDS from inside a container
we can use the 'passthru' mode of macvlan. This was not accessible
from the config and this patch fixes the issue.

Signed-off-by: Eric Leblond <eric@regit.org>


stgraber:
Looks good.

Can you also update the lxc.container.conf manpage to list the new option?

Thanks!

regit:
Replaced by #533

PR #512, opened 2015-05-07
Call /lib/apparmor/profile-load directly instead of the wrapper

martinpitt:
AppArmor ships /lib/apparmor/profile-load. /lib/init/apparmor-profile-load is
merely a wrapper which calls the former, so just call it directly to avoid the
dependency on the wrapper.

LP: #1432683


hallyn:
Thanks, Martin.

hallyn:
Thanks, Martin.

PR #518, opened 2015-05-09
lxc-debian.in: Fixed errors if dbus is not installed

cybit:
The lxc-debian template debootstraps a minimum debian system which does not contain dbus.
If systemd is used this will result in getty-static.service to be used instead of getty@ .
The systemd default files uses 6 tty's instead of the 4 the script creates.
This will lead to repeated error messages in the systemd journal.

 Signed-off-by: Cyril Bitterich <cebit-ghsignoff@gunnet.de>

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -200,6 +200,11 @@ configure_debian_systemd()
     mkdir -p ${rootfs}/{lib,etc}/systemd/system
     mkdir -p ${rootfs}/etc/systemd/system/getty.target.wants
 
+    # Fix getty-static-service as debootstrap does not install dbus
+    if [ -e $rootfs//lib/systemd/system/getty-static.service ] ; then
+        sed 's/ getty@tty[5-9].service//g' $rootfs/lib/systemd/system/getty-static.service |  sed 's/\(tty2-tty\)[5-9]/\14/g' > $rootfs/etc/systemd/system/getty-static.service
+    fi
+
     # This function has been copied and adapted from lxc-fedora
     rm -f ${rootfs}/etc/systemd/system/default.target
     chroot ${rootfs} ln -s /dev/null /etc/systemd/system/udev.service


cybit:
Long explanation as to why i created this pull request:

Neither debootstrap nor lxc.debian.in install the dbus-package.
Therefore /lib/systemd/system/getty-static.service comes into play with
`ConditionPathExists=!/lib/systemd/system/dbus.service`

Unfortunately this calls 
`ExecStart=/bin/systemctl --no-block start getty@tty2.service getty@tty3.service getty@tty4.service getty@tty5.service getty@tty6.service`

Which leads to messages in logctl:
`agetty[111]: /dev/tty5: No such file or directory`
`agetty[112]: /dev/tty6: No such file or directory`

stgraber:
I agree this is an issue but I don't agree with the fix.

Modifying the systemd unit in that way will only work until the next systemd update, at which point the unit will be overriden with the original one, loosing the change in the process.

I can think of two other solutions for this problem:
 - Make the lxc-debian template install dbus (preferred since systemd pretty much depends on it anyway)
 - Use the same check but instead bump lxc.tty to 6

stgraber:
Oh and also, please sign-off your commits (Signed-off-by line per CONTRIBUTING.md).

cybit:
Actually this is my first pull-request so I'm sorry if I miss stuff. Hope I fixed the Sign off to the right place.

My patch will not be overwritten by the next systemd-update because I do not change the original file but create a new one in /etc/systemd . And it's the reason why I do delete ttys 5 to 9 and not only 5 and 6.

Why didn't I use one of two solutions you mentioned:
Solution 1:
If for some reason someone does not want to install dbus or dbus is deinstalled because something else comes up the patch would be more intrusive.

Solution  2:
I would not need to have the check if I bumped up the ttys to 6.
But I assumed there to be a reason why the original commit has only 4 ttys and I don't feel comfortable questioning that.

stgraber:
Somehow missed the fact that you were writting it as an override in /etc, so yeah, that's fine.

stgraber:
You added the Signed-off-by to the github pull request description, not the actual commit, but that's fine, I'll just manually commit this for you and edit your commit message.

In the future, just pass -s to git commit, that'll add the needed line.

Thanks!

stgraber:
Applied

PR #540, opened 2015-05-27
config: miscellaneous signals for lxc.*signal

ysbnim:
I found that there is no SIGPWR in configuration.
I added miscellaneous signals such as SIGPWR, SIGURG for lxc.haltsignal, lxc rebootsignal, lxc.stopsignal.


stgraber:
do we actually expect any of those signals to be undefined? (wondering if we actually need all the ifdefs)

ysbnim:
Those signals can be undefined. it depends on kernel version and architecture.
There is a similar code in util-linux and they also use ifdefs.

Please refer to :
https://git.kernel.org/cgit/utils/util-linux/util-linux.git/tree/misc-utils/kill.c (line 88 ~ line 171)

PR #543, opened 2015-05-29
The yum in Centos 5.11 does not know about '--release'. The release v…

bogen:
The yum in CentOS 5.11 does not know about '--releasever'. The release version only needs to be set in the outer bootstrap, not the inner one.

Since '--releasever' is used by yum in both the outer and inner bootstrap calls, the inner call fails when bootstrapping CentOS 5.11.

With this patch an lxc-create bootstrap of CentOS 5.11 completes enough to be usable. CentOS 5.11 containers can be created, started, stopped, and networking works.

### DIFF
--- a/templates/lxc-centos.in
+++ b/templates/lxc-centos.in
@@ -413,10 +413,12 @@ download_centos()
 
     # download a mini centos into a cache
     echo "Downloading centos minimal ..."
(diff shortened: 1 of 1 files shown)


stgraber:
Can you edit your commit message to include the required Signed-off-by line (see CONTRIBUTING.md)?

Thanks!

bogen:
I ammended my commit message to include the Signed-off-by-line.

PR #545, opened 2015-05-30
The yum in Centos 5.11 does not know about '--releasever', which is u…

bogen:
…sed by: lxc-create ... -- release=VERSION

The release version only needs to be set in the outer bootstrap, not the inner one.
With this change an lxc-create bootstrap of CentOS 5.11 completes enough to be usable.
CentOS 5.11 containers can be created, started, stopped, and networking works.
Signed-off-by: Dwight Schauer <das@teegra.net>


bogen:
This addresses issue https://github.com/lxc/lxc/issues/546

PR #547, opened 2015-05-31
overlayfs: support multiple lower layers

martinolsen:
overlayfs supports multiple lower layers[1]. This change enables mounting these in lxc.

Example usage:

    lxc.rootfs = overlayfs:/foo:/bar:/baz

Will give an overlay mounted container root with /bar at the lowest layer and /baz at the highest.

  [1] https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt

### DIFF
--- a/src/lxc/bdev.c
+++ b/src/lxc/bdev.c
@@ -2194,7 +2194,7 @@ static int overlayfs_mount(struct bdev *bdev)
 	strcpy(dup, bdev->src);
 	if (!(lower = strchr(dup, ':')))
 		return -22;
-	if (!(upper = strchr(++lower, ':')))
+	if (!(upper = strrchr(++lower, ':')))
 		return -22;
 	*upper = '\0';
 	upper++;


stgraber:
Hi, per CONTRIBUTING.md, could you sign-off your commit (Signed-off-by line)?

Thanks!

stgraber:
Closing due to lack of response.

benofbrown:
Is there anything that can be done to revive this other than getting the original author to add the Signed-off-by line?

stgraber:
Not really, we can't use code which wasn't signed off, so it'd have to be reimplemented by someone who hasn't seen this proposal's code.

brauner:
Well, there is the function `ovl_get_rootfs()` in `lxcoverlay.{c,h}` which we use in order to get the rootfs path for the container. It is currently only called in `ovl_mkdir()` in order to check that `ovl_mkdir()` does not create `work` and `delta` directories under the containers rootfs. It currently also assumes that overlay rootfs has the form `overlayfs:/lower:/upper`. The function could easily be generalized to return an array of two strings containing as first entry a string of all the lower directories and as second entry the upper directory. Then we can conveniently call that function in `ovl_mount()` to retrieve the upper and lower directories and we don't need the modification suggested here at all.

dceee:
General question since this is an old discussion: 
Are multiple lower layers supported in LXC by now or is this here the last status?

brauner:
Multiple lower layers are supported.

dceee:
Thanks! I am currently observing an issue with this but I will better open a new ticket.

PR #552, opened 2015-06-05
Wait on correct container name

lrowe:
I haven't managed to get past the lxc-checkpoint without failing, but the lxc-wait should be on $name.


stgraber:
Thanks for the fix!

Could I just ask you to edit your commit message to add the necessary Signed-off-by line (see CONTRIBUTING.md)?

Thanks!

PR #555, opened 2015-06-09
Update lxc-debian.in

fanyeren:
fix

Failed to read /proc/cmdline. Ignoring: No such file or directory
invoke-rc.d: policy-rc.d denied execution of start.

Signed-off-by: feng xiahou xiahoufeng@yahoo.com

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -133,7 +133,14 @@ EOF
 
         rm -f $rootfs/etc/ssh/ssh_host_*key*
 
+        cp -f $rootfs/var/lib/dpkg/info/openssh-server.postinst $rootfs/var/lib/dpkg/info/openssh-server.postinst.bak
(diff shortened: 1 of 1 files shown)


stgraber:
What version of Debian is having problems with the existing code?

We already do policy-rc.d + turn off the upstart job, which is all we should ever have to do per-policy. If ssh somehow still tries to start, that's likely to be a bug in Debian.

fanyeren:
jessie

stgraber:
Is that actually causing any fatal error?

"invoke-rc.d: policy-rc.d denied execution of start." is very precisely what we expect as we write a policy-rc.d file which makes any invoke-rc.d call return in a non-fatal way.

That's the standard way of dealing with daemons in chroots as done by debootstrap and other tools, so if the container is created succesfuly with those messages printed, then the template is in fact doing precisely what it's supposed to and there's nothing to fix.

PR #558, opened 2015-06-10
Added container-cache option to templates

cloudnull:
This change adds in the container-cache option within several of the the mainline
default lxc templates. The purpose of the change is to allow a template to pull from 
a cache location that may not be `@LOCALSTATEDIR@/cache/lxc`

Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>


cloudnull:
As more background into this PR we're looking to add this functionality due to a work around being added into our upstream project to fulfill this PR by modifying the on disk template. The review for that work can be found here: https://review.openstack.org/#/c/190048/

stgraber:
Hi,

I think I'd prefer to use an environment variable for this rather than an argument, would that work for you? Maybe use LXC_CACHE_PATH or something along those lines.

Also, in the change to the download template, you're only overriding the path for the privileged case, not for the unprivileged case. I think it'd make sense for the option to apply to both.

cloudnull:
@stgraber  thanks for the review it makes a lot of sense to be able to configure it with an environment variable, would you also like to keep the flag or simply convert this option to strictly and environment variable? I will update the download template with these new bits in a subsequent patch to ensure that if an override is set it forces the unprivileged containers to the same location.

stgraber:
environment only would be best I think

cloudnull:
Ok will refactor and resubmit shortly.

cloudnull:
@stgraber  the initial pr has been refactored to now only use the $LXC_CACHE_PATH environment variable. Thanks for the review and hope this is on the right track.

stgraber:
Looks good, thanks!

PR #562, opened 2015-06-11
Fix creation of dev/mqueue and dev/shm on Gentoo

devurandom:
The dev/mqueue and dev/shm directories do not exist when using lxc.autodev, thus they have to be created upon mount.


stgraber:
Hi,

Thanks for the fix! Can you just update your commit message to include the required Signed-off-by line (see CONTRIBUTING.md)?

devurandom:
See #563

PR #563, opened 2015-06-11
Fix creation of dev/mqueue and dev/shm on Gentoo

devurandom:
The dev/mqueue and dev/shm directories do not exist when using lxc.autodev, thus they have to be created upon mount.

Signed-off-by: Dennis Schridde <devurandom@gmx.net>


devurandom:
Could this please also be merged into stable-1.1?

devurandom:
Could this please also be merged into stable-1.1? (@hwoarang)

PR #564, opened 2015-06-11
Adopt capability (lxc.cap.drop) documentation from other distros in Gentoo config, drop setpcap and sys_nice

devurandom:
sys_admin and sys_resources are needed for systemd, setpcap is required for journald.

Also include the explanatory blurb that can be found in the other distro's configs.

See-Also: https://bugs.gentoo.org/show_bug.cgi?id=551792

Signed-Off-By: Dennis Schridde <devurandom@gmx.net>

### DIFF
--- a/config/templates/gentoo.moresecure.conf.in
+++ b/config/templates/gentoo.moresecure.conf.in
@@ -12,9 +12,26 @@ lxc.mount.entry=mqueue dev/mqueue mqueue rw,nodev,noexec,nosuid,create=dir 0 0
 lxc.mount.entry=shm dev/shm tmpfs rw,nosuid,nodev,noexec,relatime,create=dir 0 0
 lxc.mount.entry=run run tmpfs rw,nosuid,nodev,relatime,mode=755 0 0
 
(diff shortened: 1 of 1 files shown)


stgraber:
Does Gentoo ship with systemd by default nowadays or is it the recommended default?

Just wondering as last I heard it had its own init system and so wouldn't need those caps.

hwoarang:
That is correct. Gentoo comes with OpenRC as the default init so you don't need these capabilities. I'd say that this pull request needs to be rejected.

It's best if we put such details to the Gentoo wiki for those interested to run gentoo containers with systemd.

devurandom:
Naturally I disagree with rejecting this pull request. It also aligns the way lxc.cap.drop is documented with the day other distributions do it, and adds sys_nice to the drop list.

I would hence suggest to modify the pullrequest to add back sys_admin and sys_resources to the drop list, and leave the documentation about potential systemd breakage as it is in my current proposal.

devurandom:
Commit replaced, commit message expanded, issue summary adjusted.

PR #569, opened 2015-06-16
use `hostname` for DHCP_HOSTNAME in ifcfg-eth0

LenzGr:
Updated centos/fedora/oracle templates to use `hostname` for DHCP_HOSTNAME in
/etc/sysconfig/network/ifcfg-eth0, so the container's host name is propagated
to the host's DHCP server (e.g. dnsmasq, which also acts as the DNS server).
This resolves lxc/lxd#756

Signed-off-by: Lenz Grimmer <lenz@grimmer.com>


stgraber:
Thanks!

PR #574, opened 2015-06-18
Debian template: randomized root password

major:
By default, the Debian template sets root's password to 'root' during the build of a new container.  This commit sets a randomized password based on a six-character random string and the container's name.

At the end of the LXC build, the output looks like this:

    $ lxc-create -n debian-test -t debian
    --- SNIPPED OUTPUT ---
    Root password is 'Root-debian-test-U2wgIi', please change !

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -157,8 +157,9 @@ EOF
         echo "Timezone in container is not configured. Adjust it manually."
     fi
 
-    echo "root:root" | chroot $rootfs chpasswd
-    echo "Root password is 'root', please change !"
+    root_password=$(mktemp -u Root-${name}-XXXXXX)
+    echo "root:${root_password}" | chroot $rootfs chpasswd
+    echo "Root password is '${root_password}', please change !"
 
     return 0
 }


stgraber:
So while we certainly should have something like that, I don't think I agree with the implementation.

I'd much rather see a single pull request which introduces a shared shell file that can be sourced by all templates and provides generic functions to deal with passwords.

That should ideally be based on something like what the fedora/centos templates are doing so that the user can alter the behavior based on environment variables.

The main issue with your implementation, besides deviating the behavior of the templates even more from each other than they are today is that if somebody is scripting LXC, there's no easy way for them to get to that password.

I think ideally, I'd like for:
 - All templates to default to no password at all (no an empty password)
 - All templates to support a common set of environment variables or/and arguments to have passwords generated for them or to use passwords provided by the user
 - Have a way (possibly optional) for those credentials to be written down into a text file in the container's directory (for use by scripts).
 - Print a generic message to the user, advising them of any credential that was generated and that they can use lxc-attach to interact with the container without them.

I'd also strongly recommend that this happens at the same time as we remove sshd by default from all templates, so our containers don't have any remote exposure by default (outside of the dhcp client).

major:
I agree with your ideal goals completely as well as the removal of ssh.  That might be a challenge because each template uses different methods to build each type of instance and there may need to be some agreement between distributions on how to do that.

 _(And now I see why this problem hasn't been solved before now.). ;)_

I'll go back to the drawing board and see what I can propose.  Do you have a preference where the shared shell script could be located?

stgraber:
We already have a lxc.functions script in /usr/share/lxc that could probably include those new functions. Alternatively, you could make a new template.functions script in the same location.

stgraber:
Being discussed on mailing-list and elsewhere.

stgraber:
I still strongly feel that the right answer there is a serious cleanup of our templates to share a common core, that core would support a common set of arguments and environment variables, leading in predictable, consistent behavior across templates.

Closing this pull request as we really should discuss the overall rework of templates on the list and once we agree, implement things centrally.

(Not sure how much progress will happen for 2.0 as we'll be tagging in the coming months.)

PR #577, opened 2015-06-30
Add instanced systemd service

CameronNemo:


stgraber:
Can you edit your commit message to contain the required Signed-off-by string (per CONTRIBUTING.md)?

Thanks!

CameronNemo:
Done. Sorry about that :)

stgraber:
so shouldn't there be something in the main lxc.service unit to trigger the lxc@ unit for all auto-started containers?

stgraber:
@CameronNemo ping

CameronNemo:
Sorry, meant to get back then it slipped my mind.

My intention was to support both separately so someone could use the lxc-autostart interface or just enable lxc@foo.service with systemd. Having lxc.service call out to lxc@.service would be a little harder, especially since I know it is not recommended to run systemctl start, stop, etc. as part of a systemd service because it is prone to deadlocks...

I think that this is ok to merge for now, and if somebody wants to do the work to make lxc.service call lxc@.service then that can be done later on.

If you do not want to merge without that, then I can just close the PR and re-open when I find time to test it out thoroughly to ensure there are no deadlock issues.

PR #580, opened 2015-07-03
/proc/$$/loginuid doesn't work under user namespace

fanyeren:
which prevents logins from working

Signed-off-by: feng xiahou <xiahoufeng@yahoo.com>


stgraber:
I fixed that in PAM upstream a while ago, I'd much prefer we get Debian to include the fix in their PAM package rather than have LXC sed the pam config files with the risk that the next package update reverts it all.

fanyeren:
good job!

BTW, Systemd in Centos7 has same problem, but i cannot use such dirty sed* to fix it.

PR #581, opened 2015-07-03
Use /dev/loop-control if it exists

Blub:
Loop devices can be added on the fly when needed, they're
not always created beforehand. The loop-control device can
be used to find and allocate the next available number
instead of going through the /dev directory contents (which
is now only a fallback mechanism).

### DIFF
--- a/src/lxc/bdev.c
+++ b/src/lxc/bdev.c
@@ -1868,7 +1868,7 @@ static int loop_detect(const char *path)
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


Blub:
Just in case this was an issue (due to the sign-off line) I just linked my work email address (which I used for this commit) to my github account. (As I noticed github otherwise shows an "unknown author" icon.)

stgraber:
Thanks!

Issue #417, opened 2015-01-26
lxc-execute failed to exec a command in unprivileged mode on Ubuntu 14.04


sartrevan:
I'm trying to use lxc to start a java process in unprivileged mode following the guide:https://www.stgraber.org/2014/01/17/lxc-1-0-unprivileged-containers/

I've managed to start an lxc ubuntu guest os but I'd rather use lxc to start a java process of my own as an application container. Is it possible? I've googled for 2 days, still no luck.

![image](https://cloud.githubusercontent.com/assets/4744114/5898826/6251befa-a58e-11e4-8aa6-b7730db888bf.png)

stgraber:
What distro, kernel and version of LXC is that?

sartrevan:
**Distro**
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.1 LTS"
NAME="Ubuntu"
VERSION="14.04.1 LTS, Trusty Tahr"
**Kernel**
3.13.0-32-generic
**LXC Version**
1.0.7

ptoomey3:
I ran into what seems like a directly related issue when playing with `lxc-execute` and unprivileged containers. It seems that `lxc-execute` assumes that you are either running a privileged container or are starting an unprivileged container where root is mapped to some subordinate IDs. It looks like `lxc-execute` always does a setuid(0) in https://github.com/lxc/lxc/blob/3b6aebceaf331dd4967f86af0632a3b1cde0120c/src/lxc/start.c#L672 when you invoke a new namespace. It isn't clear to me that calling `lxc-execute`, as the original submitter did  above (without a uid/gid mapping) automatically creates a new user namespace or if that wouldn't happen unless you passed an explicit uid/gid map config.

Regardless of the above, it seems like `lxc-execute` doesn't work as one would think it should when you are not running as root. Two quick examples:

```
lxc-execute -n foobar -- whoami
lxc-execute: start.c: must_drop_cap_sys_boot: 554 failed to clone (0x20000011): Operation not permitted
lxc-execute: namespace.c: lxc_clone: 67 failed to clone (0x6c020000): Operation not permitted
lxc-execute: start.c: lxc_spawn: 897 Operation not permitted - failed to fork into a new namespace
lxc-execute: start.c: __lxc_start: 1087 failed to spawn 'foobar'
```

[...]

PR #585, opened 2015-07-09
Add support for specified UID/GID for lxc-execute in a private user namespace 

ptoomey3:
In #417 some people noted that `lxc-execute` didn't work as expected when not running as root. The primary thing I noticed was that, no matter what, the process that was spawned was executed as container UID 0. This is a direct artifact of lxc assuming that, whatever container is launched, is started using "init" with UID 0. For lxc-start that makes a ton of sense, as we are trying to simulate a full distribution and init always runs as root. But, for lxc-execute, we aren't really trying to launch a full system. We are only trying to launch a process in some sort of default containment. So, it doesn't make much sense for something like `lxc-execute whoami` to return `root` unless that user is actually running as root.

I'v modified the "always become root" logic to check to see whether we are being executed via lxc-execute or not. If we are, then we don't become root. This doesn't make `lxc-execute` work 100% the way I would hope, but it does seem much closer. 

For example, with this change, I still can't do something like:

```
ptoomey3@ubuntu:/mnt/hgfs/lxc/lxc$ lxc-execute -n foobar  whoami
lxc-execute: start.c: must_drop_cap_sys_boot: 568 failed to clone (0x20000011): Operation not permitted
lxc-execute: namespace.c: lxc_clone: 67 failed to clone (0x6c020000): Operation not permitted
lxc-execute: start.c: lxc_spawn: 974 Operation not permitted - failed to fork into a new namespace
lxc-execute: start.c: __lxc_start: 1173 failed to spawn 'foobar' 
```

I'm not 100% sure what I would even hope the above would do, but crashing is probably not it :smile:. But, so long as we pass in a reasonable UID/GID remapping we can do this:

```
ptoomey3@ubuntu:/mnt/hgfs/lxc/lxc$ whoami
ptoomey3

ptoomey3@ubuntu:/mnt/hgfs/lxc/lxc$ cat ~/.config/lxc/default.conf
lxc.id_map = u 1000 1000 1
lxc.id_map = u 0 100000 1
lxc.id_map = g 1000 1000 1
lxc.id_map = g 0 100000 1

ptoomey3@ubuntu:/mnt/hgfs/lxc/lxc$ lxc-execute -n foobar -f ~/.config/lxc/default.conf whoami
init.lxc: initutils.c: mount_fs: 36 failed to mount /proc : Operation not permitted
init.lxc: initutils.c: mount_fs: 36 failed to mount /dev/shm : Operation not permitted
init.lxc: initutils.c: mount_fs: 36 failed to mount /dev/mqueue : Operation not permitted
ptoomey3
```

Before this patch, the above `whoami` would have returned `root` since the container was always switched to `root` for any unprivileged container.

So, while I'm not 100% sure this fixes all of #417, it does seem to at least allow unprivileged containers to execute a command as the original user. 

One last thing, as you can see, `mount_fs` seems to have a bit of an issue remounting when we don't switch our UID/GID to 0. This seems to be a non-fatal warning. So, I'd love it if someone with more LXC background could comment on the above warnings. Do they need to be "fixed" or can we simply add some logic to not attempt remounting for some situations?

/cc @sartrevan - since you opened the original `lxc-execute` issue.

### DIFF
--- a/src/lxc/lxc_execute.c
+++ b/src/lxc/lxc_execute.c
@@ -59,7 +59,9 @@ static int my_parser(struct lxc_arguments* args, int c, char* arg)
 {
 	switch (c) {
 	case 'f': args->rcfile = arg; break;
-	case 's': return lxc_config_define_add(&defines, arg);
+	case 's': return lxc_config_define_add(&defines, arg); break;
+	case 'u': args->uid = atoi(arg); break;
+	case 'g': args->gid = atoi(arg);
 	}
 	return 0;
 }
@@ -67,6 +69,8 @@ static int my_parser(struct lxc_arguments* args, int c, char* arg)
 static const struct option my_longopts[] = {
 	{"rcfile", required_argument, 0, 'f'},
 	{"define", required_argument, 0, 's'},
+	{"uid", required_argument, 0, 'u'},
+	{"gid", required_argument, 0, 'g'},
 	LXC_COMMON_OPTIONS
 };
 
@@ -81,7 +85,9 @@ and execs COMMAND into this container.\n\
 Options :\n\
   -n, --name=NAME      NAME for name of the container\n\
   -f, --rcfile=FILE    Load configuration file FILE\n\
-  -s, --define KEY=VAL Assign VAL to configuration variable KEY\n",
+  -s, --define KEY=VAL Assign VAL to configuration variable KEY\n\
+  -u, --uid=UID Execute COMMAND with UID inside the container\n\
+  -g, --gid=GID Execute COMMAND with GID inside the container\n",
 	.options  = my_longopts,
 	.parser   = my_parser,
 	.checker  = my_checker,
@@ -138,6 +144,12 @@ int main(int argc, char *argv[])
 
 	if (lxc_config_define_load(&defines, conf))
 		return 1;
+
+	if (my_args.uid)
+		conf->init_uid = my_args.uid;
+
+	if (my_args.gid)
+		conf->init_gid = my_args.gid;
 
 	ret = lxc_execute(my_args.name, my_args.argv, my_args.quiet, conf, my_args.lxcpath[0], false);
 
--- a/doc/lxc.container.conf.sgml.in
+++ b/doc/lxc.container.conf.sgml.in
@@ -247,6 +247,39 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
     </refsect2>
 
     <refsect2>
+      <title>Init ID</title>
+      <para>
+        Sets the UID/GID to use for the init system, and subsequent command, executed by lxc-execute.
+
+        These options are only used when lxc-execute is started in a private user namespace.
+
+        Defaults to: UID(0), GID(0)
+      </para>
+      <variablelist>
+        <varlistentry>
+          <term>
+            <option>lxc.init_uid</option>
+          </term>
+          <listitem>
+            <para>
+              UID to use within a private user namesapce for init.
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>lxc.init_gid</option>
+          </term>
+          <listitem>
+            <para>
+              GID to use within a private user namesapce for init.
+            </para>
+          </listitem>
+        </varlistentry>
+      </variablelist>
+    </refsect2>
+
+    <refsect2>
       <title>Network</title>
       <para>
         The network section defines how the network is virtualized in
--- a/src/lxc/arguments.h
+++ b/src/lxc/arguments.h
@@ -88,6 +88,10 @@ struct lxc_arguments {
 	char *lvname, *vgname, *thinpool;
 	char *zfsroot, *lowerdir, *dir;
 
+	/* lxc-execute */
+	uid_t uid;
+	gid_t gid;
+
 	/* auto-start */
 	int all;
 	int ignore_auto;
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2604,6 +2604,11 @@ struct lxc_conf *lxc_conf_init(void)
 	for (i = 0; i < LXC_NS_MAX; i++)
 		new->inherit_ns_fd[i] = -1;
 
+	/* if running in a new user namespace, init and COMMAND
+	 * default to running as UID/GID 0 when using lxc-execute */
+	new->init_uid = 0;
+	new->init_gid = 0;
+
 	return new;
 }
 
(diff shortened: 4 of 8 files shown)


stgraber:
This seems like a bad idea, unless we setuid/setgid, the process will not be running inside the container's user namespace and so would run as the unmapped uid -1 with no right whatsoever against the resources of the container but instead with rights against resources outside of it. This seems like a security problem waiting to happen...

stgraber:
I however wouldn't be against having flags for lxc-execute to specify what uid and gid the process is meant to be started as, then have LXC drop to that uid/gid before executing the binary.

Well, so long as that's done without breaking the ABI that is.

ptoomey3:
> This seems like a bad idea, unless we setuid/setgid, the process will not be running inside the container's user namespace and so would run as the unmapped uid -1 with no right whatsoever against the resources of the container but instead with rights against resources outside of it. This seems like a security problem waiting to happen...

Ah, I didn't know about the requirement to call `setuid`/`setgid` after creating the new namespace. 

> I however wouldn't be against having flags for lxc-execute to specify what uid and gid the process is meant to be started as, then have LXC drop to that uid/gid before executing the binary.

Do we even need to get that fancy for now? It seems like the common scenario would be to either set uid/gid to 0 for `lxc-start` or set it to the parent uid/gid for `lxc-execute`. I pushed up https://github.com/lxc/lxc/commit/bd6f4d2880fa5db78711df55dd06dcc542fce799 with that change. Thoughts?

ptoomey3:
bump. @stgraber - Does is seem reasonable to start with a much simpler approach and just assume that non-root `lxc-execute` should `setuid`/`setgid` to the same ID as the caller?

ptoomey3:
The relevant commit for that change is https://github.com/ptoomey3/lxc/commit/bd6f4d2880fa5db78711df55dd06dcc542fce799.

stgraber:
With unprivileged containers, the parent uid/gid as seen outside the container may not exist in the container and even if it did exist in the container, it'd most likely resolve to a completely different uid/gid outside of the container.

ptoomey3:
> With unprivileged containers, the parent uid/gid as seen outside the container may not exist in the container 

If we are using an unprivileged container isn't there an implicit assumption that we have provided a valid UID/GID map? So, wouldn't the existence/non-existence be entirely up to the caller? 

> and even if it did exist in the container, it'd most likely resolve to a completely different uid/gid outside of the container.

Is that a problem? If I run `lxc-execute` as UID 1000 I can then map inside container UID 1000 to subordinate ID 100000. That seems fine, right? The main idea is that calling `id` outside/inside the container will return the same value. 

At least to me, the thing that feels consistent is that whatever UID outside the container executes `lxc-execute` is the same UID inside container that is running the containerized process. Currently if I run `lxc-execute` as UID 0 then the container process will run UID 0 and it up to the UID/GID map to map that to a different UID outside the container. It feels like the same should be true for non-root too. If I run `lxc-execute` as UID 1000 then the container process will run as UID 1000 and it is up to me to provide a UID/GID map to map that to a subordinate ID (or not).

ptoomey3:
> If we are using an unprivileged container isn't there an implicit assumption that we have provided a valid UID/GID map? So, wouldn't the existence/non-existence be entirely up to the caller?

I'm probably a bit confused here. Can you distinguish the difference bewteen `lxc-execute -n foobar whoami` and `lxc-start -n foobar whoami`? I was generally thinking of `lxc-execute` to spawn a command in a temporary container with default isolation. In that circumstance I was thinking that the UID of the caller would always exist inside the container. Is that not true? I can see how for a "full container" that the UID outside the container may not exist inside.

stgraber:
A valid map may not cover the uid/gid used outside the container.

For example, my current uid/gid on my laptop is 205102 / 205115, containers only have 65536 uids and gids mapped, so with your suggested change, setuid/setgid will fail as you're trying to change to a uid and gid outside the range provided to the container.

ptoomey3:
@stgraber - Gotcha. So, would you be agreeable to allowing an explicit UID to switch to as a `lxc-execute` flag?

stgraber:
Yes, allowing the user to pass a uid and gid would be fine.

ptoomey3:
@stgraber - I took a first pass at adding UID/GID support via a flag in https://github.com/ptoomey3/lxc/commit/471c33ff869034691d973cab248ae7a364d7c117. Let me know what you think when you get a chance.

src/lxc/lxc_execute.c
```diff
@@ -139,6 +145,12 @@ int main(int argc, char *argv[])
 	if (lxc_config_define_load(&defines, conf))
 		return 1;
 
+	if (my_args.uid)
+		conf->init_uid = my_args.uid;
```
ptoomey3: Passing these as additions the existing `conf` seemed like the most trivial way to implement this. If you would prefer that they get passed explicitly as args it will require updating a number of call sites. I don't feel strongly at all and will defer to you. If we do decide to go with using `conf`, I could update [lxc_conf_init](https://github.com/ptoomey3/lxc/blob/1f885925ced89360bb84f199632d7101e6f7e469/src/lxc/conf.c#L2553) to default these values to 0.

ptoomey3: I convinced myself that adding these via configuration file would be useful. So, I added support in https://github.com/ptoomey3/lxc/commit/8084bd518bd7da606e21462c71af78f6c69bf31d and updated the English docs in https://github.com/ptoomey3/lxc/commit/14dcbac128bf975822d90cd986694db8e7534084. I'll leave it up to you whether you would like this set:

* Only via command line flags
* Only via the configuration file
* Both

I kind of vote both since this setting is unique to `lxc-execute`, but could also be useful to store in a configuration file for repeated use.

ptoomey3:
Friendly bump. @stgraber - Are there any changes you would like to see (ex. whether the UID/GID should/should not be settable via a config file option, etc)?

stgraber:
You seem to be missing the get_config_item code, needed for a complete config option addition.

Once done, please check that the following works:

set_config_item("lxc.init_uid", "100")
set_config_item("lxc.init_gid", "100")
get_config_item("lxc.init_uid")
get_config_item("lxc.init_gid")
save_config("/tmp/blah")
load_config("/tmp/blah")


That should validate that you've not missed any of the usual bits.

ptoomey3:
Ok @stgraber I added support for `get_config_item` in https://github.com/ptoomey3/lxc/commit/cba2726b44965dafb276c98624dbe8b03e8baaf2. I wrote some tests to validate basic behavior. And, I manually tested the full set/get/save/load process you noted above.

stgraber:
Looks all good. One last thing, your commits aren't signed-off. Could you edit the commit messages to add the required Signed-off-by line (see CONTRIBUTING)?

Thanks!

ptoomey3:
@stgraber - Ok, I've force pushed up a new batch of commits with the appropriate Signed-off line. Thanks for taking the time to guide me through the process!

stgraber:
Thanks!

PR #586, opened 2015-07-10
Split config_mount() into separate functions

brauner:
Currently config_mount() handles lxc.mount, lxc.mount.entry and lxc.mount.auto.
This commit splits config_mount() into config_mount_fstab() which handles
lxc.mount, config_mount_kernelfs() which handels lxc.mount.auto (Regarding the
name, there already is a config_mount_auto() function.), and
config_mount_entry() which handles lxc.mount.entry. They receive separate
entries in the lxc_config_t struct. Note that the order in which they are
inserted into lxc_config_t is important here! The original config_mount() first
checked for lxc.mount.entry, lxc.mount.auto and finally for lxc.mount. Now they
appear in the same order in lxc_config_t as they were orignally tested in
config_mount().

Signed-off-by: Christian Brauner <christianvanbrauner@gmail.com>

### DIFF
--- a/src/lxc/seccomp.c
+++ b/src/lxc/seccomp.c
@@ -259,6 +259,7 @@ static int parse_config_v2(FILE *f, char *line, struct lxc_conf *conf)
 	uint32_t default_policy_action = -1, default_rule_action = -1, action;
 	enum lxc_hostarch_t native_arch = get_hostarch(),
 			    cur_rule_arch = native_arch;
+	uint32_t compat_arch = SCMP_ARCH_NATIVE;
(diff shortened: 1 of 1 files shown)


stgraber:
What issue is this fixing?

I'm a bit wary of merging this if all it does is split them into separate functions as this will make things harder to backport and for seemingly not gain.

brauner:
It was listed as TODO. And I think its a good idea since the functions are concerned with different things in mounting and it will be easier to maintain or change them in the future. But it's your call really.

Issue #609, opened 2015-08-01
Alpine template is broken


rcarmo:
On Ubuntu Trusty, with `lxc(-templates) 1.0.7`I get the following errors when trying to create an Alpine container:

<pre>
$ sudo lxc-create --name builder -t alpine
Determining the latest release... v3.2
Using static apk from http://wiki.alpinelinux.org/cgi-bin/dl.cgi/v3.2/main/x86_64
Downloading alpine-mirrors-3.1.3-r2.apk
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
Downloading apk-tools-static-2.6.3-r0.apk
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
Downloading alpine-keys-1.1-r0.apk
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
alpine-devel@lists.alpinelinux.org-4a6a0840.rsa.pub: OK
Verification Failure
Failed to download a valid static apk
lxc_container: lxccontainer.c: create_run_template: 1125 container creation template for builder failed
lxc_container: lxc_create.c: main: 271 Error creating container builder
</pre>

Using the current template on git tip yields identical results.

xor-gate:
On Arch Linux host I also get the same error.

```
[root@shulgin jerry]# lxc-create --name tst02 -t alpine
which: no apk in (/usr/local/sbin:/usr/local/bin:/usr/bin:/data/files/scripts:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/usr/sbin:/usr/bin:/sbin:/bin)
Determining the latest release... v3.2
Using static apk from http://wiki.alpinelinux.org/cgi-bin/dl.cgi/v3.2/main/x86_64
Downloading alpine-mirrors-3.1.3-r2.apk
[...]

PR #611, opened 2015-08-03
Fix issue #609 - Alpine Linux template apk static binary check

xor-gate:
See #609


stgraber:
Closing this in favor of @ncopa's.

PR #619, opened 2015-08-05
Fresh CONTRIBUTING

alkino:
Refresh CONTRIBUTING file.


stgraber:
You forgot to sign-off your commit, which is pretty hilarious considering that you're updating the very document that describes this :)

alkino:
Ahah. I'm confuse about it ;) I never use signature so I forgot. Be sure I will put a pre-push hook in this directory specially ;)

PR #621, opened 2015-08-06
Refactor lxc-snapshot, lxc-clone, make LXC_CLONE_KEEPNAME work and add option to destroy container with all snapshots to lxc-destroy

brauner:
lxc-snapshot:
- lxc_snapshot.c lacked necessary members in the associated lxc_arguments struct
  in arguments.h. This commit extends the lxc_arguments struct to include
  several parameters used by lxc-snapshot which allows a rewrite that is more
  consistent with the rest of the lxc-* executables.
- All tests have been moved beyond the call to lxc_log_init() to allow for the
  messages to be printed or saved.
- Some small changes to the my_args struct. (The enum task is set to
  SNAP (for snapshot) per default and variables illustrating the usage of the
  command line flags are written in all caps.)
- arguments.h has been extended to accommodate a rewrite of lxc-clone

lxc-clone:
- This commit adapts lxc-clone to be similiar in usage and feel to the other
  lxc-* executables. It builds on the previous extension of the lxc_argument
  struct and now uses the default lxc_arguments_parse() function.
- Options which were not used have been removed.
- The LXC_CLONE_KEEPNAME flag was not respected in the previous version of
  lxc-clone. The culprit is a missing if-condition in lxccontainer.c. As this
  requires a change in one of the API functions in lxccontainer.c it will be
  addressed in a follow-up commit.

LXC_CLONE_KEEPNAME:
Passing the LXC_CLONE_KEEPNAME flag to do_lxcapi_clone() was not respected. We
wrap clear_unexp_config_line() and set_config_item_line() in an appropriate
if-condition.

lxc-destroy
- This enables the user to destroy a container with all its snapshots without
  having to use lxc-snapshot first to destroy all snapshots.
- Some unification regarding the usage of exit() and return has been done.

### DIFF
--- a/src/lxc/arguments.h
+++ b/src/lxc/arguments.h
@@ -94,6 +94,22 @@ struct lxc_arguments {
 	int list;
 	char *groups;
 
+	/* lxc-snapshot and lxc-clone */
+	enum task {
+		DESTROY,
+		LIST,
+		RESTORE,
+		SNAP,
+		RENAME,
+	} task;
+	int print_comments;
+	char *commentfile;
+	char *newname;
+	char *newpath;
+	char *snapname;
+	int keepname;
+	int keepmac;
+
 	/* remaining arguments */
 	char *const *argv;
 	int argc;
--- a/src/lxc/lxc_attach.c
+++ b/src/lxc/lxc_attach.c
@@ -146,7 +146,7 @@ static struct lxc_arguments my_args = {
 Execute the specified COMMAND - enter the container NAME\n\
 \n\
 Options :\n\
-  -n, --name=NAME   NAME for name of the container\n\
+  -n, --name=NAME   NAME of the container\n\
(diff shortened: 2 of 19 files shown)


hallyn:
Thank you, this is very nice.

hallyn:
Sorry, you are changing the command line options.  I.e. 'lxc-clone -o old -n new' or 'lxc-clone -s old new' no longer work.  We can't change that ,too many people use this.  Reverting this set for now.

brauner:
I'll try to change the flags back. Though this will be hard for lxc-clone since -n is actually reserved for the name of an existing container defined by LXC_COMMON_OPTIONS. I'll think of something.

brauner:
Updated in #633 to retain old behaviour of lxc-snapshot. (lxc-clone will be addresses separately.)

PR #627, opened 2015-08-13
Mark unmodified strings as const in public API

jacky405:
That commit fixes the public API in such a way that strings which are not modified by the LXC library are marked as const. Currently, users of the library typically need to cast pointers to constant strings into pointers to modifiable strings, in order to be able to call the LXC functions, which is something that should be avoided as much as possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,10 @@
+
+===
+Please note: This is a working copy of LXC used for patches.
+If you need the real upstream, please see:
+https://linuxcontainers.org/
+===
+
 Please see the COPYING file for details on copying and usage.
 Please refer to the INSTALL file for instructions on how to build.
 
--- a/README.md
+++ b/README.md
@@ -1,8 +1,7 @@
 
 ===
 Please note: This is a working copy of LXC used for patches.
-If you need the real upstream, please see:
-https://linuxcontainers.org/
+If you need the real upstream, please see: https://linuxcontainers.org/
 ===
 
 Please see the COPYING file for details on copying and usage.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
 
 ===
+
 Please note: This is a working copy of LXC used for patches.
 If you need the real upstream, please see: https://linuxcontainers.org/
+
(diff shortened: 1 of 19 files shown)


stgraber:
Hmm, so I don't think this change is acceptable until we bump the ABI to 2.0 as this change would require code changes to any existing API user, therefore breaking our otherwise stable interface.

stgraber:
Also, your commit isn't signed-off (see CONTRIBUTING).

jacky405:
The commit is now signed-off.

Concerning the potential API change, it could indeed break some existing code, if that code is compiled with the "-Werror" compiler option. The reason is that the GCC C compiler generates a warning when implicitly casting a "char**" into a "const char**", which is really a weird behavior, IMHO. The GCC C++ compiler does not generate such a warning since the C++ specification is clear about the fact that this type of implicit cast is perfectly safe.

I did try to disable the "-Werror" flag and could build the LXC package itself without having to make any change to the code relying on the library API (the "lxc_*.c" files, and the tests).
I assume the same thing will happen with existing packages which rely on the library API (unless they use the "-Werror" flag of course), so I guess it is fine to not consider that change as an API breakage, and introduce it before the version 2.0.

IMHO, it is very important to have a clean API, which does not force people to introduce some weirdnesses such as const casts (casting constness away) in their code.

stgraber:
Looking at our existing downstreams (language bindings, orchestration tools, ...), most of them support multiple LXC releases, typically everything which implements the 1.0 ABI (so currently 1.0, 1.1 and master). They also, at least in Linux distributions, all build with -Werr, so this change would break them, requiring changes that would then make their own backward compatibility break.

I absolutely agree that it'd be nice to apply your change from an API cleanliness point of view, but since we're going to remain perfectly compatible with the 1.0 ABI in LXC 2.0, we'll not be merging this change for quite some time.

As a result, I'm closing this pull request. I will however ensure that any major ABI rework (whenever we actually work on the 2.0 ABI), is more carefully designed wrt to const identifiers.

PR #633, opened 2015-08-14
Refactor lxc-snapshot, make LXC_CLONE_KEEPNAME work, add option to destroy container with all snapshots to lxc-destroy

brauner:
Refactor lxc-snapshot

- lxc_snapshot.c lacked necessary members in the associated lxc_arguments struct
  in arguments.h. This commit extends the lxc_arguments struct to include
  several parameters used by lxc-snapshot which allows a rewrite that is more
  consistent with the rest of the lxc-* executables.
- All tests have been moved beyond the call to lxc_log_init() to allow for the
  messages to be printed or saved.
- Some small changes to the my_args struct. (The enum task is set to SNAP (for
  snapshot) per default and variables illustrating the usage of the command line
  flags are written in all caps.)
- arguments.h has been extended to accommodate a future rewrite of lxc-clone
- Traditional behaviour of the executable has been retained in this commit.

lxc-destroy: Remove container with all snapshots

- This enables the user to destroy a container with all its snapshots without
  having to use lxc-snapshot first to destroy all snapshots. (The enum values
  DESTROY and SNAP from the previous commit are reused here again.)
- Some unification regarding the usage of exit() and return has been done.

Make LXC_CLONE_KEEPNAME work

- Passing the LXC_CLONE_KEEPNAME flag to do_lxcapi_clone() was not respected and
  let to unexpected behaviour for e.g. lxc-clone. We wrap
  clear_unexp_config_line() and set_config_item_line() in an appropriate
  if-condition.

Fix grammar in some of the executables

"NAME for name of the container" becomes "NAME of the container"

### DIFF
--- a/src/lxc/arguments.h
+++ b/src/lxc/arguments.h
@@ -94,6 +94,22 @@ struct lxc_arguments {
 	int list;
 	char *groups;
 
+	/* lxc-snapshot and lxc-clone */
+	enum task {
+		DESTROY,
+		LIST,
+		RESTORE,
+		SNAP,
+		RENAME,
+	} task;
+	int print_comments;
+	char *commentfile;
+	char *newname;
+	char *newpath;
+	char *snapname;
+	int keepname;
+	int keepmac;
+
 	/* remaining arguments */
 	char *const *argv;
 	int argc;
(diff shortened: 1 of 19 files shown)


stgraber:
This has been manually merged by @hallyn.

PR #641, opened 2015-08-23
Write path and name of clone to lxc-snapshots file

brauner:
If we currently create clone-snapshots via lxc-clone only the plain total
number of the containers it serves as a base-container for is written to the file
"lxc-snapshots". This commit modifies mod_rdep() so it will store the paths and
names to the containers that are clone-snapshots (similar to the "lxc_rdepends"
file for the clones). **Users which still have containers that have a non-empty
(with a number > 0 as an entry) "lxc-snapshots" file in the old format are not
affected by this change. It will be used until all old clones have been
deleted!** For all others, the "lxc_snapshots" file placed under the original
container now looks like this:

      /var/lib/lxc
      bb
      /var/lib/lxc
      cc
      /opt
      dd

This is an example of a container that provides the base for three
clone-snapshots bb, cc, and dd. Where bb and cc both are placed in the usual
path for privileged containers and dd is placed in a custom path.

- Add additional argument to function that takes in the clone-snapshotted
  lxc_container.
- Have mod_rdep() write the path and name of the clone-snapshotted container the
  file lxc_snapshots of the original container.
- If a clone-snapshot gets deleted the corresponding line in the file
  lxc_snapshot of the original container will be deleted and the file updated
  via mmap() + memmove() + munmap() + ftruncate().
- Adapt has_fs_snapshots().
- **If an lxc-snapshot file in the old format is found we'll keep using it.**

The advantage is that the dependencies between the clones and the original container are clear now. Furthermore, lxc-destroy can be changed to destroy a clone with all snapshots: those located in the snaps/ folder and clone-snapshots.

(P.S.: These commits are based on a different branch from me which is also staged for commit. If you don't find the last two commits suitable but do find the others suitable just merge the other branch and leave a brief note here.)

### DIFF
--- a/src/lxc/arguments.h
+++ b/src/lxc/arguments.h
@@ -94,6 +94,22 @@ struct lxc_arguments {
 	int list;
 	char *groups;
 
+	/* lxc-snapshot and lxc-clone */
+	enum task {
+		DESTROY,
+		LIST,
+		RESTORE,
+		SNAP,
+		RENAME,
+	} task;
+	int print_comments;
+	char *commentfile;
+	char *newname;
+	char *newpath;
+	char *snapname;
+	int keepname;
+	int keepmac;
+
 	/* remaining arguments */
 	char *const *argv;
 	int argc;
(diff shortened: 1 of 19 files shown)


hallyn:
Merged, thanks.

PR #651, opened 2015-09-06
Destroy on signal

brauner:
These commits enable a container to be destroyed on shutdown by setting `lxc.ephemeral = 1` in its configuration file. This will allow us a consistent implementation of `lxc-copy = lxc-clone + lxc-start-ephemeral` without having to resort to hacks.

Small changes:

- `lxc_delete_string_in_array()` has beed added to `utils.c`. It replaces the manual method for deleting a container entry in `lxc_snapshots` files  in `mod_rdep()` in `lxccontainer.c`.
- static `do_bdev_destroy()` has been moved to `bdev.c` and `bdev.h` and made public because it will be reused in `start.c`.
- The "shutdown-handler" `lxc_fini()` has been modified to call the new static function `lxc_destroy_container_on_signal()` to destroy containers which have `lxc.ephemeral = 1` set in their config file. Currently only `overlayfs` and `aufs` backed containers can be destroyed in this way because this is so far the only case we care about (`lxc-start-ephemeral` part of `lxc-copy`) but the function works for all bdev types supported by lxc.

### DIFF
--- a/src/lxc/bdev.c
+++ b/src/lxc/bdev.c
@@ -3614,3 +3614,21 @@ bool rootfs_is_blockdev(struct lxc_conf *conf)
 		return true;
 	return false;
 }
+
+bool bdev_destroy(struct lxc_conf *conf)
+{
+	struct bdev *r;
+	bool ret = false;
+
+	r = bdev_init(conf, conf->rootfs.path, conf->rootfs.mount, NULL);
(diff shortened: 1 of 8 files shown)


brauner:
These patches have also been send to lxc-devel list. Closing this in favour of lxc-devel.

PR #660, opened 2015-09-24
Add a slackware template.

Ponce:
Requires pkgtools and slackpkg (from the slackware-current tree).

Signed-off-by: ponce <matteo.bernardini@gmail.com>


Ponce:
I'll send another with my name in the committer info.

PR #664, opened 2015-09-28
Add a slackware template.

Ponce:
Requires pkgtools and slackpkg (from the slackware-current tree).

Signed-off-by: Matteo Bernardini <ponce@slackbuilds.org>


stgraber:
Hi,

Thanks for this!

Any chance you can make this look like modern templates, using the configs in config/templates/, only setting the config keys which are different from the common file?

Ponce:
yes, sorry I stayed a little behind on that!
I'll check the examples and adapt the template accordingly.

PR #665, opened 2015-09-28
add --force-arch to skip host related errors

szaszg:
It will be nice, if we can override: "can't create $arch container on $hostarch" errors. (lxc-debian template)

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -469,7 +469,6 @@ Options :
   -p, --path=PATH        directory where config and rootfs of this VM will be kept
   -a, --arch=ARCH        The container architecture. Can be one of: i686, x86_64,
                          amd64, armhf, armel, powerpc. Defaults to host arch.
-  --force-arch           Ignore architecture specific errors.
   -r, --release=RELEASE  Debian release. Can be one of: squeeze, wheezy, jessie, sid.
                          Defaults to current stable.
   --mirror=MIRROR        Debian mirror to use during installation. Overrides the MIRROR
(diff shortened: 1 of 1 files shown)


stgraber:
I'd really rather we fix the check to support any other architecture with personality support, or is that a case where you're attempting to use qemu-user-static or something?

stgraber:
Also, your commit isn't signed-off (Signed-off-by line missing)

szaszg:
I have a mixed i386/amd64 debian system with a x86_64 kernel. I try to lxc-create .... -t debian -- -a amd64 and i got a not too clever error message: "can't create amd64 container on i386" .. and i cannot 
persuade lxc to give up this pessimistic idea...

So, i track down until found that:

```bash
if which dpkg > /dev/null 2>&1 ; then
    arch=$(dpkg --print-architecture)
else
    arch=$(uname -m)
    if [ "$arch" = "i686" ]; then
        arch="i386"
    elif [ "$arch" = "x86_64" ]; then
        arch="amd64"
    elif [ "$arch" = "armv7l" ]; then
        arch="armhf"
    fi
fi
hostarch=$arch

```
This is quite strange because if found dpkg, than we assume that: the host kernel architecture is the same as the dpkg architecture. On other systems we just directly check the kernel arch... We may check dpkg arch and kernel arch... and allow amd64 on i386/x86_64 or something similar, but IMHO more easy and elegant if we allow users to override the decision of template:lxc-debian about such a ''harmful'' things :-)

stgraber:
Indeed checking the dpkg architecture doesn't make much sense here, we should only check the kernel architecture. A fix in that direction would be welcome.

szaszg:
Hmm..
How can i sign-off a commit?

stgraber:
Use the -s flag to git commit or just manually write the "Signed-off-by: firstname lastname \<email\>" line at the end of the commit message

szaszg:
Huhh... some git magic... i'm afraid i loose somewhere... I made a new commit (or 2 or 3) with kernel arch check only... ?

stgraber:
Quite a messy branch, anyway, I see you've signed-off at least some of the commits and the change seems fine, so I'll squash that stuff all together and push it myself :)

stgraber:
In my local branch, will be pushed along some other stuff in the next hour, closing this.

Thanks!

PR #670, opened 2015-10-05
Add Ceph RBD backingstore.

ksperis:
I propose to add the management of Rados Block Device (http://docs.ceph.com/docs/master/rbd/rbd/)
Currently managing snapshots is not implemented. And the rbd devices are not remap in /dev/* when physical host restart.
Do you have any suggestions?

### DIFF
--- a/src/lxc/lxccontainer.h
+++ b/src/lxc/lxccontainer.h
@@ -857,11 +857,11 @@ struct bdev_specs {
 		char *lv; /*!< LVM Logical Volume name */
 		char *thinpool; /*!< LVM thin pool to use, if any */
 	} lvm;
+	char *dir; /*!< Directory path */
 	struct {
 		char *rbdname; /*!< RBD image name */
 		char *rbdpool; /*!< Ceph pool name */
 	} rbd;
-	char *dir; /*!< Directory path */
 };
 
 /*!
--- a/src/lxc/lxccontainer.h
+++ b/src/lxc/lxccontainer.h
@@ -857,6 +857,10 @@ struct bdev_specs {
 		char *lv; /*!< LVM Logical Volume name */
 		char *thinpool; /*!< LVM thin pool to use, if any */
 	} lvm;
+	struct {
+		char *rbdname; /*!< RBD image name */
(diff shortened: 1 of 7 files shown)


stgraber:
Looks okay to me, except for your branch looking a bit odd with that merge in there. Could you rebase you branch on git master, that should clean things up.

@hallyn since you're the bdev master, mind giving this a quick review?

ksperis:
Thanks. I rebased my branch.

hallyn:
Sorry, I'll take a look soon

ksperis:
ok, thank you. there is no emergency :)

hallyn:
Hi,

You said:
"About this, is that "rbd" should not be before "dir" ?"

I don't understand.  Can you please rephrase?

Overall looks good to me, thanks much.

src/lxc/lxccontainer.h
```diff
@@ -841,6 +841,10 @@ struct bdev_specs {
 		char *lv; /*!< LVM Logical Volume name */
 		char *thinpool; /*!< LVM thin pool to use, if any */
 	} lvm;
+	struct {
+		char *rbdname; /*!< RBD image name */
+		char *rbdpool; /*!< Ceph pool name */
+	} rbd;
```
stgraber: This here is an API break, only append to existing struct, never insert new members in the middle!

stgraber:
@hallyn you may want to re-review after the recent changes. It looks okay to me except for the part where this breaks the API right now, so that part needs fixing.

hallyn:
The only other thing I see is that the mkdir of the bdev>dest *could* result in -EEXIST which should probably be ignored rather than raise an error.

otherwise lgtm.

stgraber:

stgraber:
Hi, any chance you can update your branch to address the comments @hallyn and I made?

Other than those it looks fine and it'd be great to merge this.

Thanks!

ksperis:
Hi, I updated my branch and fix according to your comments.

stgraber:
jenkins: test this please

stgraber:
Thanks!

PR #672, opened 2015-10-07
Set supplementary groups too when attaching.

roadrunner2:


stgraber:
Hi,

You commit isn't signed-off, can you please edit your commit message to include the required Signed-off-by line (see CONTRIBUTING)?

Thanks

roadrunner2:
Sorry, missed that. Done.

src/lxc/attach.c
```diff
@@ -997,6 +1016,8 @@ static int attach_child_main(void* data)
 	int fd;
 	uid_t new_uid;
 	gid_t new_gid;
+	gid_t new_groups[50];
```
stgraber: Why 50?

stgraber:
Also wondering, when exactly does an init pid ever have additional groups?

src/lxc/attach.c
roadrunner2: I just  picked a number that I figured was going to be ample for practically all cases. Could of course go an count how would be needed too, but that seemed like more work than was warranted.

roadrunner2:
When the init process decides to acquire more (calls setgroups())? This patch is a port from one we were using a while ago on an older version of lxc (certainly before c476bdce46f5fd8), where init did indeed have supplementary groups set. It's not the case anymore (for us at least), but while upgrading I saw the TODO and figured I might as well port it and send it in, in case anybody else runs into this issue with their init.

stgraber:

To have this branch tested by Jenkins, use the "ok to test" command.
To have a one time test done, use the "test this please" command.

Commands are simple Github comments of the format: "jenkins: COMMAND"

roadrunner2:
jenkins: ok to test

src/lxc/attach.c
stgraber: Is there a relevant kernel limit that would be better than the 50 here?

roadrunner2: sysconf(_SC_NGROUPS_MAX)

stgraber: That's 65k so probably not a sane default :)

I guess you could go with an empty array and re-alloc it when adding stuff to it, that'd get you rid of the arbitrary limit and not require you to init the array at 65536.

stgraber:
This branch is wrongly targeted at the stable-1.1 branch, code should only ever be submitted against the master branch.

Closing this pull request, please re-open one against the master branch, thanks!

PR #673, opened 2015-10-07
Added a new 'cgroup' hook.

roadrunner2:
This hook runs in the host after all host-side initialization is done,
right before triggering the final init in the child. As such it runs
shortly before the 'start' hook.

The motivation/need for this hook is because we need to fix up the
cgroups as our containers have some special requirements regarding
ownership and permissions of some of the entries. The hook as
presented here is actually more general in the sense that it runs as
almost the last thing on the host side before the container is booted,
but I couldn't think of a good name for it (since start and pre-start are
already taken) - suggestions welcome.

Lastly, the Japanese translation is left as FIXME right now, but if the
rest of this patch is ok I can provide one.


stgraber:
Hi,

You commit isn't signed-off, can you please edit your commit message to include the required Signed-off-by line (see CONTRIBUTING)?

Thanks

roadrunner2:
Done.

stgraber:
@hallyn thoughts on that?

As for the japanese and korean translations, you can just skip the files entirely, our translators are pretty quick to pick up new features and update the manpages.

hallyn:
LGTM.  Do you have an example of the hook you need to use?  Should you be passing a LXC_CGROUP_PATH env variable into the hook or something, so the hook knows where to be making changes?

stgraber:

To have this branch tested by Jenkins, use the "ok to test" command.
To have a one time test done, use the "test this please" command.

Commands are simple Github comments of the format: "jenkins: COMMAND"

roadrunner2:
For our containers, we (need to) run something like the following:
  chmod 775 {$cpu_cg,$acct_cg}/{,tasks,cgroup.procs}
  chown $uid:$gid {$cpu_cg,$acct_cg}/{,tasks,cgroup.procs}
(where $uid and $gid are very particular id's, and $cpu_cg and $acct_cg are the cpu and cpuacct cgroups for the container).

Yes, an LXC_CGROUP_PATH or something with the path of the actual cgroup hierarchy for the container would be really useful, as right now we use this slightly hacky command:

roadrunner2:
jenkins: ok to test

roadrunner2:
Regarding the LXC_CGROUP_PATH, please ignore what I said above, I was confusing something. I just realized that `/proc/<pid>/cgroup` actually has the relative paths, and so a LXC_CGROUP_PATH that points to the root for those relative paths (e.g. `/sys/fs/cgroup`) would indeed be useful.

hallyn:
So do you use lxc-info -n $1 -p to get the pid, then /proc/pid/cgroup to get the cgroups?

Just trying to gauge how much effort it is to get the info and if it's worth suppling the LXC_CGROUP_PATH

hallyn:
jenkins: ok to test

hallyn:
Thanks for this patch, I'll merge when jenkins finishes.

Actually one thing -this is bikeshedding, but do you think the hook
should be called 'lxc.hook.cgroup' instead of 'cgroups'?  (Or maybe
either one should be accepted in confile.c)  cgroups makes more sense
when I actually think about it, but I have a feeling most ppl will just
type 'cgroup'.

stgraber:
I had to force remove this as it was wrongly aimed at the stable-1.1 branch and not at master.

This is a new feature and therefore unsuitable for stable releases.

Unfortunately this branch also doesn't merge cleanly into master, so you'll have to re-submit it targeting the master branch instead.

roadrunner2:
Unfortunately trying to run `lxc-info -n $1 -p` in this new cgroups hook hangs (and ends up hanging other lxc commands), and running it in the mount hook gives a "doesn't exist" error. Furthermore, it's a bit tricky to get that pid via some `pgrep`. For this reason I'm currently using the above mentioned `ls` instead. Probably the cleanest solution would actually be to pass the pid as another parameter to the hooks.

roadrunner2:
Regarding the hook name, sure, I can have it recognize both 'cgroup' and 'cgroups' - I'll do that for the next pull. And I'll base it off master.

PR #683, opened 2015-10-30
arch template: Fix systemd-sysctl service

Kubuxu:
The systemd-sysctl service includes condition that `/proc/sys/` has to be read-write.
In lxc only `/proc/sys/net/` is read-write which causes the condition to fail and service not to run.
This patch changes the check to `/proc/sys/net/` and makes the service apply only rules that are in `net` tree.

### DIFF
--- a/templates/lxc-archlinux.in
+++ b/templates/lxc-archlinux.in
@@ -103,6 +103,10 @@ sed -e 's/^ConditionPathExists=/# ConditionPathExists=/' \
     -e 's/After=dev-%i.device/After=/' \
     < /lib/systemd/system/getty\@.service \
     > /etc/systemd/system/getty\@.service
+# fix systemd-sysctl service
+sed -e 's/^ConditionPathIsReadWrite=\/proc\/sys\/$/ConditionPathIsReadWrite=\/proc\/sys\/net\//' \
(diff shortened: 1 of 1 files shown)


CameronNemo:
We need to do this in all of the templates TBH. Or fix it upstream, which I have suggested. http://lists.freedesktop.org/archives/systemd-devel/2015-October/034776.html

lisaev:
But before, can you please explain the rationale?

For example, sysctl is for kernel parameters, right? So why do we even need systemd-sysctl on a system w/o a kernel?

Also, what you suggest is anyway a bad solution because of
* it will cause errors in logs when _all_ parameters in /usr/lib/sysctl.d/50-default.conf are applied;

* the patch modifies a stock file that will be overwritten upon update (which in arch means in a couple of weeks).

I don't mind modifying arch template simply because it's largely irrelevant, as arch provides rootfs images anyway. But other distros aren't that convenient, so please don't spread this mess there.

Kubuxu:
sysctl is used for network configuration, that is why it is applied. With `--perfix net` it won't cause any errors but instead it will function as expected. I was surprised when my config for IPv6 forwarding wasn't applied due to this.

PR #692, opened 2015-11-09
lxc-ls: use /usr/bin/env to find an appropriate python3 to run

fwilson42:
Previously, lxc-ls would fail on systems without a python3 in /usr/bin:

`-bash: /usr/local/bin/lxc-ls: /usr/bin/python3: bad interpreter: No such file or directory`

Use `env` instead in order to find an appropriate interpreter in $PATH.

Signed-off-by: Fox Wilson <2016fwilson@tjhsst.edu>

### DIFF
--- a/config/apparmor/lxc-generate-aa-rules.py
+++ b/config/apparmor/lxc-generate-aa-rules.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 import sys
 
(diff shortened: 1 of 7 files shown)


stgraber:
Can you also update the other python3 scripts?

fwilson42:
Sure, I'll update this later today.

fwilson42:
Alright, that looks like all of the scripts with `#!/usr/bin/python3`.

stgraber:
Looks good, though your branch commit history does contain an extra merge commit that shouldn't be there.

Can you rebase your branch to have that commit go away? then I'll merge it.

fwilson42:
I **think** this is right.

PR #693, opened 2015-11-10
Fetch Debian archive GPG keyrings when they're not available

hsoft:
When running the debian template on a non-debian host, it's usual not to
have debian-archive-keyring.gpg. When that happens, we skip the
signature checking of the release, which is dangerous because it's made over
HTTP.

This commit adds automatic fetching of Debian release keys.

Strongly related to #409


stgraber:
Can you sign-off your commit? (see CONTRIBUTING)

Thanks!

hsoft:
Oh, sorry, I hadn't read that part. Done.

PR #718, opened 2015-12-11
Return immediately in save_phys_nics if not run as root

liqiu:
Physical nic is not instantiated in lxc_create_network if not run as root


stgraber:

stgraber:
Is the only benefit from that that we're saving a tiny bit of time looping through the interfaces or is there an actual bug that this is fixing?

stgraber:
Also, can you sign-off your commit? (see CONTRIBUTING)

Thanks

PR #724, opened 2015-12-12
split bdev into modules

brauner:
*I'm currently working on splitting bdev.{c,h} into modules per block device for the liblxc 2.0 release. If we agree that this is worth it I can push the other modules as well. We should probably take the opportunity to rename/rewrite some of the functions.*

With this commit we start to split bdev.{c,h} into modules located in the subfolder bdev. This will ultimately lead to a cleaner design and will be easier to manage. We start by creating a module for overlay: overlay.{c,h}.

- The functions:

	- overlayfs_detect()
	- overlayfs_mount()
	- overlayfs_umount()
	- overlayfs_clonepaths()
	- overlayfs_destroy()
	- overlayfs_create()

  move from bdev.{c,h} to overlay.{c,h}. The only thing that remains in bdev.c
  is the static definition of

	- static const struct bdev_ops overlayfs_ops

- The functions:

	- update_ovl_paths()
	- overlay_getlower()

  move from lxccontainer.c to overlay.{c,h}. update_ovl_paths() is used to
  update absolute paths for overlay lxc.mount.entry entries but it seems to fit
  more here than into lxccontainer.c.
  The Function overlay_getlower() is used to extract the lower directory for
  overlay (and aufs) rootfs. It should at some point become a common helper.

- The functions:

	  - do_rsync()

	  - dir_new_path()

   remain in bdev.c for now but become extern. We declare them extern in
   overlay.c to be able to call them. As the comment to them correctly notices,
   they should at some point become common helpers and probably move
   to utils.{c,h} or some other more appropriate place.

- The structs:

	- struct bdev; /* defined in bdev.h */

	- struct bdev_specs; /* defined in lxccontainer.h */

	- struct lxc_conf; /* defined conf.h */

  are forward declared/put as incomplete types in overlay.h so that the
  functions have access to it.

- The header overlay.h is *not* included in bdev.h but only in bdev.c so that
  when bdev.h is included the public functions in overlay.h cannot be accessed,
  i.e. if an implementation wants to call functions from overlay.h they need to
  explicitly include it. (As is e.g. done in the case of lxccontainer.c.)

- The header

	- lxc-btrfs.h

  also moves to the bdev subfolder.


stgraber:

stgraber:
Looks like this needs rebasing

brauner:
@stgraber rebased onto current master. Would be nice if you could leave a short comment whether I should proceed with splitting bdev.{c,h} into modules.

stgraber:
I'm happy with the concept, yes. The bdev stuff is getting a bit difficult to read right now :)

@hallyn ^

hallyn:
haven't looked at the patch, but in general iiuc it's something i definately want.

brauner:
Good, removing the [WIP] then. I'll wait for this to be merged before sending another module since the changes for each are quite substantial in terms of quantity and reviewing all together would be a pain.

@hallyn
@stgraber

hallyn:
Jenkins: ok to test

this branch looks good to me, thanks.

hallyn:
jenkins: test this please

brauner:
This Jenkins testing is pretty cryptic. Is it expected to fail and if not how does one find out what needs to be fixed?

stgraber:
You need to go dig into other jenkins jobs to figure out the actual result and yeah, that's not ideal, I've got a todo item to improve that by setting up our jobs in the same way we've done for LXD (that is, one big matrix job).

stgraber:
@hallyn I'll leave it up to you to merge this if you think it's ready. I'm certainly happy with the concept but haven't reviewed the implementation.

PR #733, opened 2015-12-20
debootstrap now uses --variant=minbase

kvaidas:
Added --variant=minbase to the debootstrap command which results in fewer packages being installed by default: 141 now vs the 287 before.

Added sudo and isc-dhcp-client packages so that the container starts correctly and networking works out of the box.

Signed-off-by: Vaidas Kascėnas <vaidas@kascenas.lt>

### DIFF
--- a/templates/lxc-ubuntu.in
+++ b/templates/lxc-ubuntu.in
@@ -331,7 +331,7 @@ download_ubuntu()
     arch=$2
     release=$3
 
-    packages_template=${packages_template:-"ssh,vim"}
+    packages_template=${packages_template:-"ssh,vim,sudo,isc-dhcp-client"}
 
     # Try to guess a list of langpacks to install
     langpacks="language-pack-en"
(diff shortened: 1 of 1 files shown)


stgraber:

stgraber:
Hi,

Two things:
 - Your commit isn't signed-off. See CONTRIBUTING
 - This looks to me like a pretty major change in behavior for anyone currently depending on the Debian template. This change would also apply to everyone using the download template and in that case, for all LXC releases. I don't think it's a good idea to push such a major change to all users, instead, it may make a lot more sense to allow for the variant to be passed as an argument to the template, keeping the default as it is today.

stgraber:
Mean Ubuntu template in my earlier comment, sorry.

fanyeren:
I think isc-dhcp-client is not needed in many company

kvaidas:
I added it because the template configures the network interface to be configured with DHCP by default - so that the user still had an out-of-the-box experience.

kvaidas:
Rewrote to be an optional parameter: #740

PR #734, opened 2015-12-21
split bdev into modules: btrfs + rsync

brauner:
I'm currently working on splitting bdev.{c,h} into modules per block device for the liblxc 1.2 release. I'll push one or two modules at a time since the changes are quite substantial in terms of quantity and reviewing them would otherwise be a pain. We should probably take the opportunity to rename/rewrite some of the functions.


stgraber:

brauner:
Note: The branch has already been marked as "ok to test" by Serge.

brauner:
Even though rsync is not a bdev module it makes sense to take it out of bdev.{c,h} since it makes it obvious that there is al level of abstraction over all but also independent of all bdevs associated with it.

hallyn:
Hi - iiuc, patch 5/6 (new naming for modules) should not exist - just update the previous commits to use the final names.  (unless there is a good reason not to due to the patch flow)

Otherwise basics look good, thanks.

brauner:
@hallyn, done. It did involve some serious rebasing-skills. :)

hallyn:
jenkins: test this please

hallyn:
Manually merged (and fixed the spaces in Makefile)

brauner:
I thought I fixed them already this morning. Hm, sorry.

hallyn:
Oh - you might have.  I used the .patch I had fetched yesterday.

Issue #377, opened 2014-12-10
ls --fancy by default


daniel-baumann:
similar to lxc-start with the -d option, lxc commands should default to the convenient mode, therefore.. please make lxc-ls have the --fancy option implicit default and for those who want to use lxc-ls in scripts or otherwise require the short option, have them use lxc-ls --short.

stgraber:
Seems reasonable to me, maybe for 1.2.

lyrixx:
:+1:

Kentoseth:
@brauner:

Is #748 a solution to this issue? I speed-read through it and understood that quite a few other changes were added to that issue/feature.

brauner:
@Kentoseth, ultimately we did decide against fancy by default to not break expectations of existing users. Afaik, I don't think that has changed (@stgraber).

stgraber:
Right, a lot of things seem to wrap lxc-ls, so changing the default output would break people.

PR #748, opened 2016-01-02
Reimplement lxc-ls in C

brauner:
This is a reimplementation of `lxc-ls` in C. It supports all features previously supported by `lxc-ls` except `--nesting` (for now). I need to put more thought into the nesting part in since we fork when calling  `lxc_attach()`. A recursive implementation seems to be the most intuitive one. Thoughts on that appreciated.

All flags and parameters have the same name as before except when the user specifies a regex to filter container names by. In the Python implementation the regex was passed without a parameter flag. The new C-implementation has the parameter flag `-r/--regex` for this.

### DIFF
--- a/src/lxc/Makefile.am
+++ b/src/lxc/Makefile.am
@@ -215,6 +215,7 @@ bin_PROGRAMS = \
 	lxc-execute \
 	lxc-freeze \
 	lxc-info \
+	lxc-ls \
 	lxc-monitor \
 	lxc-snapshot \
 	lxc-start \
@@ -250,6 +251,7 @@ init_lxc_SOURCES = lxc_init.c
 lxc_monitor_SOURCES = lxc_monitor.c
 lxc_monitord_SOURCES = lxc_monitord.c
 lxc_clone_SOURCES = lxc_clone.c
+lxc_ls_SOURCES = lxc_ls.c
 lxc_copy_SOURCES = lxc_copy.c
 lxc_start_SOURCES = lxc_start.c
 lxc_stop_SOURCES = lxc_stop.c
--- a/src/lxc/arguments.h
+++ b/src/lxc/arguments.h
@@ -21,11 +21,14 @@
  * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
+
 #ifndef __LXC_ARGUMENTS_H
 #define __LXC_ARGUMENTS_H
 
 #include <getopt.h>
+#include <stdbool.h>
 #include <stdint.h>
+#include <sys/types.h>
 
 struct lxc_arguments;
 
@@ -116,6 +119,18 @@ struct lxc_arguments {
 	int keepdata;
 	int keepname;
 	int keepmac;
+
+	/* lxc-ls */
+	char *ls_fancy_format;
+	char *ls_groups;
+	char *ls_regex;
+	bool ls_active;
+	bool ls_fancy;
+	bool ls_frozen;
+	bool ls_line;
+	bool ls_nesting;
+	bool ls_running;
+	bool ls_stopped;
 
 	/* remaining arguments */
 	char *const *argv;
--- a/src/lxc/lxc_ls.c
+++ b/src/lxc/lxc_ls.c
@@ -1,0 +1,1159 @@
+/*
+ *
+ * Copyright © 2016 Christian Brauner <christian.brauner@mailbox.org>.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2, as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#define _GNU_SOURCE
+#include <getopt.h>
+#include <regex.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+
+#include "arguments.h"
+#include "conf.h"
+#include "config.h"
+#include "confile.h"
+#include "log.h"
+#include "lxc.h"
+#include "lxccontainer.h"
+#include "utils.h"
+
+lxc_log_define(lxc_ls, lxc);
+
+#define LINELEN 1024
+#define LS_FROZEN 1
+#define LS_STOPPED 2
+#define LS_ACTIVE 3
+#define LS_RUNNING 4
+#define LS_NESTING 5
+
+/* Store container info. */
+struct ls {
+	char *name;
+	char *state;
+	char *groups;
+	char *interface;
+	char *ipv4;
+	char *ipv6;
+	unsigned int nestlvl;
+	pid_t init;
+	double ram;
+	double swap;
+	bool autostart;
+	bool running;
+};
+
+/* Keep track of field widths for printing. */
+struct lengths {
+	unsigned int name_length;
+	unsigned int state_length;
+	unsigned int groups_length;
+	unsigned int interface_length;
+	unsigned int ipv4_length;
+	unsigned int ipv6_length;
+	unsigned int init_length;
+	unsigned int ram_length;
+	unsigned int swap_length;
+	unsigned int autostart_length;
+};
+
+static int ls_deserialize(int rpipefd, struct ls **m, size_t *len);
+static void ls_field_width(const struct ls *l, const size_t size,
+		struct lengths *lht);
+static void ls_free(struct ls *l, size_t size);
+static void ls_free_arr(char **arr, size_t size);
+/*
+ * This is a very lenient function. It tries to gather as much information about
+ * a container as it can thereby skipping over possible failures (e.g. failed
+ * non-fatal strdup() or malloc() failures in functions it calls etc.). It will
+ * only bail if the minimum amount of information (name and state of the
+ * container) cannot be retrieved.
+ */
+static int ls_get(struct ls **m, size_t *size, const struct lxc_arguments *args,
+		struct lengths *lht, const char *basepath, const char *parent,
+		unsigned int lvl);
+static char *ls_get_cgroup_item(struct lxc_container *c, const char *item);
+static char *ls_get_config_item(struct lxc_container *c, const char *item,
+		bool running);
+static char *ls_get_groups(struct lxc_container *c, bool running);
+static char *ls_get_ips(struct lxc_container *c, const char *inet);
+struct wrapargs {
+	const struct lxc_arguments *args;
+	struct lengths *lht;
+	int pipefd[2];
+	size_t *size;
+	const char *parent;
+	unsigned int nestlvl;
+};
+/*
+ * Takes struct wrapargs as argument.
+ */
+static int ls_get_wrapper(void *wrap);
+/*
+ * To calculate swap usage we should not simply check memory.usage_in_bytes and
+ * memory.memsw.usage_in_bytes and then do:
+ *	swap = memory.memsw.usage_in_bytes - memory.usage_in_bytes;
+ * because we might receive an incorrect/negative value.
+ * Instead we check memory.stat and check the "swap" value.
+ */
+static double ls_get_swap(struct lxc_container *c);
+static unsigned int ls_get_term_width(void);
+static char *ls_get_interface(struct lxc_container *c);
+static bool ls_has_all_grps(const char *has, const char *must);
+static struct ls *ls_new(struct ls **ls, size_t *size);
+/*
+ * Print user-specified fancy format.
+ */
+static void ls_print_fancy_format(struct ls *l, struct lengths *lht,
+		size_t size, const char *fancy_fmt);
+/*
+ * Only print names of containers.
+ */
+static void ls_print_names(struct ls *l, struct lengths *lht,
+		size_t ls_arr, size_t termwidth);
+/*
+ * Print default fancy format.
+ */
+static void ls_print_table(struct ls *l, struct lengths *lht,
+		size_t size);
+/*
+ * id can only be 79 + \0 chars long.
+ */
+static int ls_read_and_grow_buf(const int rpipefd, char **save_buf,
+		const char *id, ssize_t nbytes_id,
+		char **read_buf, ssize_t *read_buf_len);
+static int ls_serialize(int wpipefd, struct ls *n);
+static int ls_write(const int wpipefd, const char *id, ssize_t nbytes_id,
+		const char *s);
+static int my_parser(struct lxc_arguments *args, int c, char *arg);
+
+static const struct option my_longopts[] = {
+	{"line", no_argument, 0, '1'},
+	{"fancy", no_argument, 0, 'f'},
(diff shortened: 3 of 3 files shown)


stgraber:

brauner:
@hallyn, @stgraber, would be good to hear your thoughts on the second commit, especially about using `socketpair()` to pass nested containers from the child to the parent struct. If you're fine with it I can squash them.

hallyn:
The socketpair looks fine.

What should we do about versioning?

We don't want mixed versions of lxc in host/container to confuse each other.  We also want older lxc to be useable in a child container for lxc-ls --nested.

brauner:
I'll get back to you today.

brauner:
On Fri, Jan 08, 2016 at 07:04:39PM -0800, Serge Hallyn wrote:
> The socketpair looks fine.
> 
> What should we do about versioning?
(1) We can simply give another name to lxc-ls in C (lxc-list) although lxc-ls
    seems to be the best choice and I disprefer this option.
(2) We could rename Python lxc-ls to lxc-pyls and C lxc-ls to lxc-cls and make
    lxc-ls a symbolic link to the new C version for new lxc versions. This would
    look like this:

        /usr/bin/lxc-pyls
        /usr/bin/lxc-ls -> lxc-cls
        /usr/bin/lxc-cls

> 
> We don't want mixed versions of lxc in host/container to confuse each other.
That should be resolved by choosing on of the options above.

> We also want older lxc to be useable in a child container for lxc-ls --nested.
I'm not sure I quite understand. If we have a newer version of lxc on the host
with lxc-ls in C and an older version of lxc with lxc-ls in Python in a
container A and in a container B that is nested in A why should the old version
of lxc-ls not work? In fact this was one of my test cases. But I might have
misunderstood you.

hallyn:
On Sat, Jan 09, 2016 at 09:01:32AM -0800, Christian Brauner wrote:
> On Fri, Jan 08, 2016 at 07:04:39PM -0800, Serge Hallyn wrote:
> > The socketpair looks fine.
> > 
> > What should we do about versioning?
> (1) We can simply give another name to lxc-ls in C (lxc-list) although lxc-ls
>     seems to be the best choice and I disprefer this option.

Oh I didn't mean that.

I meant that when, inevitably, a change is made to the stuff
we want to pass over the socketpair, how do we handle that
when the nested container is version n, parent container n+1,
and host n+2?

stgraber:
So the way I was doing it in python3 is that I'd never depend on binaries from the container, instead lxc-ls would exec itself into the container's namespace. It may well have been massively easier to do so in python than it is in C though but it was a nice trick.

It was the rough equivalent of having the parent lxc-ls fork, make sure everything it needs loaded is loaded (library wise), then setns and run the function which exports serialized container state and writes it to the shared fd.

That thing was basically recursive and so was working no matter how many level of nesting you had.

brauner:
On Sat, Jan 09, 2016 at 10:36:07AM -0800, Stéphane Graber wrote:
> So the way I was doing it in python3 is that I'd never depend on binaries from
> the container, instead lxc-ls would exec itself into the container's
> namespace. It may well have been massively easier to do so in python than it
> is in C though but it was a nice trick.
> 
> It was the rough equivalent of having the parent lxc-ls fork, make sure
> everything it needs loaded is loaded (library wise), then setns and run the
> function which exports serialized container state and writes it to the shared
> fd.

I assume you are referring to this part, Stéphane:

            # Nested containers
            if args.nesting:
                if container.running:
                    # Recursive call in container namespace
                    temp_fd, temp_file = tempfile.mkstemp()
                    os.remove(temp_file)

                    container.attach_wait(get_containers, temp_fd)

Here you call the Python wrapper for lxc_attach() if I'm not mistaken? This
is similar to what I do:

        int status = c->attach(c, ls_get_wrapper, &wargs, &aopt, &out);

is it not?

The relevant code-lines are:

		/* get nested containers */
		if (args->ls_nested && running) {
			struct wrapargs wargs= (struct wrapargs){.args = NULL};

			int ret = socketpair(PF_LOCAL, SOCK_STREAM | SOCK_CLOEXEC, 0, wargs.pipefd);
			if (ret == -1)
				goto next;

			wargs.args = args;
			wargs.lht = lht;

			pid_t out;

			lxc_attach_options_t aopt = LXC_ATTACH_OPTIONS_DEFAULT;
			aopt.env_policy = LXC_ATTACH_CLEAR_ENV;

			/* we fork here */
			int status = c->attach(c, ls_get_wrapper, &wargs, &aopt, &out);
			/* close the socket */
			close(wargs.pipefd[1]);

			if (status == 0)
				if (ls_deserialize(wargs.pipefd[0], &m, size) == -1)
					goto next;

			lxc_wait_for_pid_status(out);

			/* We've done all the communication we need so shutdown
			 * the socket and close it. */
			shutdown(wargs.pipefd[0], SHUT_RDWR);
			close(wargs.pipefd[0]);
		}

> 
> That thing was basically recursive and so was working no matter how many level
> of nesting you had.
The solution I'm currently proposing should also work independent of how many
levels of nesting we have.

brauner:
On Sat, Jan 09, 2016 at 10:30:58AM -0800, Serge Hallyn wrote:
> On Sat, Jan 09, 2016 at 09:01:32AM -0800, Christian Brauner wrote:
> > On Fri, Jan 08, 2016 at 07:04:39PM -0800, Serge Hallyn wrote:
> > > The socketpair looks fine.
> > > 
> > > What should we do about versioning?
> > (1) We can simply give another name to lxc-ls in C (lxc-list) although lxc-ls
> >     seems to be the best choice and I disprefer this option.
> 
> Oh I didn't mean that.
> 
> I meant that when, inevitably, a change is made to the stuff
> we want to pass over the socketpair, how do we handle that
> when the nested container is version n, parent container n+1,
> and host n+2?
Hm, I see you mean if we break compatibility in the API at some point... I don't
have an immediate solution. Do you have a suggestion? (And does the current
Python implementation suffer from the same problem?)

brauner:
Ah, to be explicit: My solution is not dependent on any executables only on the liblxc library. This is equivalent to the Python implementation of lxc-ls.

hallyn:
On Sat, Jan 09, 2016 at 11:13:31AM -0800, Christian Brauner wrote:
> Ah, to be explicit: My solution is not dependent on any executables only on the liblxc library. This is equivalent to the Python implementation of lxc-ls.

Right - Stéphane's comment reminded me that we can do that :)

Excellent.

stgraber:
How is it dependent on the version of liblxc in the container?

The python version made sure to load the lxc module on the host before switching into the container (well, python kinda did that for us), so you never even needed liblxc available in the container for things to work.

brauner:
On Sat, Jan 09, 2016 at 12:50:14PM -0800, Stéphane Graber wrote:
> How is it dependent on the version of liblxc in the container?
> 
> The python version made sure to load the lxc module on the host before
> switching into the container (well, python kinda did that for us), so you
> never even needed liblxc available in the container for things to work.
Sorry, I think I was mixing things up. I was thinking about the fact that in
order for a container to have other nested containers in it the container needs
to have lxc installed and hence have liblxc available. This of course is
orthogonal to our lxc_ls discussion. As far as I see it neither in the Python
nor in the C implementation does lxc-ls need any specific version of liblxc in
the nested container to be present.

brauner:
Btw, expect more changes. I'm implementing getting nested containers when the parent container is not running.

hallyn:
Please comment here when you're done making those changes, thanks.

brauner:
I think I'm done. All other things are minor and can be sent as additional PR.

stgraber:
jenkins: test this please

brauner:
@hallyn, done so far. The executable will not gain any more command line flags from this point on. We need to settle two things however:

1. I renamed one command line flag: `--nesting` in the Python implementation of `lxc-ls` became `--nested` in the C implementation. That happened by accident because I thought the flag was called like that in the previous implementation. But that can be easily changed.

2. I remember that someone once requested (#377) that `-f/--fancy` should be the default output of `lxc-ls`. Do we want that?

brauner:
3. Should I squash any commits?

hallyn:
On Tue, Jan 12, 2016 at 08:45:43AM -0800, Christian Brauner wrote:
> @hallyn, done so far. The executable will not gain any more command line flags from this point on. We need to settle two things however:
> 
> 1. I renamed one command line flag: `--nesting` in the Python implementation of `lxc-ls` became `--nested` in the C implementation. That happened by accident because I thought the flag was called like that in the previous implementation. But that can be easily changed.

Personally I'd much rather not change it.  If the command name is staying the same, scripts will get confused.

@stgraber - please comment. 

> 2. I remember that someone once requested (#377) that `-f/--fancy` should be the default output of `lxc-ls`. Do we want that?

Heh, i'll leave for @stgraber again.  I'd far prefer that behavior but don't think we can change it from under people.  but maybe the switch to v2.0 makes it ok.

hallyn:
On Tue, Jan 12, 2016 at 09:15:27AM -0800, Christian Brauner wrote:
> 3. Should I squash any commits?

Yeah - in this case it'll be easier to just proofread the final code, since it's
all new anyway.

brauner:
On Tue, Jan 12, 2016 at 09:25:15AM -0800, Serge Hallyn wrote:
> On Tue, Jan 12, 2016 at 08:45:43AM -0800, Christian Brauner wrote:
> > @hallyn, done so far. The executable will not gain any more command line flags from this point on. We need to settle two things however:
> > 
> > 1. I renamed one command line flag: `--nesting` in the Python implementation of `lxc-ls` became `--nested` in the C implementation. That happened by accident because I thought the flag was called like that in the previous implementation. But that can be easily changed.
> 
> Personally I'd much rather not change it.  If the command name is staying the same, scripts will get confused.
> 
> @stgraber - please comment. 
I changed it from --nested to --nesting so it's the same for the Python and C
version.
> 
> > 2. I remember that someone once requested (#377) that `-f/--fancy` should be the default output of `lxc-ls`. Do we want that?
> 
> Heh, i'll leave for @stgraber again.  I'd far prefer that behavior but don't think we can change it from under people.  but maybe the switch to v2.0 makes it ok.
Ok.

brauner:
On Tue, Jan 12, 2016 at 09:26:17AM -0800, Serge Hallyn wrote:
> On Tue, Jan 12, 2016 at 09:15:27AM -0800, Christian Brauner wrote:
> > 3. Should I squash any commits?
> 
> Yeah - in this case it'll be easier to just proofread the final code, since it's
> all new anyway.
Squashed. Review me into pieces, please. :)

brauner:
Changelog 2016-01-13T00:29:

- Use

```
for (i = 0, n = m; i < len; i++, n++) {
        if (ls_serialize(...))
        goto out;
}
```
- When the call to `regcomp()` or `regexec()` fails `goto next_no_put` instead of `goto out`.
- Immediately return when the call to `ls_get_cgroup_item()` in `ls_get_swap()` returns `NULL` to avoid `SIGSEV`.

brauner:
Changelog 2016-01-13T00:29 (continued):

- Call `dir_exists()` first to check if `path` exists and immediately `goto out` to avoid unnecessary error messages.

brauner:
Changelog 2016-01-13T01:37:
- Use `continue` when `regcomp()` or `regexec()` fails.
- Use more descriptive `goto` label `put_and_next`.
- set `num = 0` and `goto out` when `list_*_containers()` fails to `free()` `path` on exit.

hallyn:
(I should probably just merged and discuss later, but)

you are still doing 'goto out' if the lxc-container_new() failed.  That means if one container at that level has a bad config file, you'll stop reading right there.  Is there a reason to prefer that over doing 'continue' there as well?

brauner:
(I like the approach of being as critical as possible before the merge.)

No, we can just `continue` there as well. I just took it to be one of the few critical points where we should bail.

brauner:
I'd like to bring one small, maybe pointless thing. Since we're doing (non-tail) recursion here there is the inherent danger that we might blow the stack at some level of nesting. At least in the case where we do not attach to the namespace of a running container but rather append paths to get nested containers. Although the argument `lvl` in `ls_get()` was originally just to keep track of the nesting level for pretty-printing we could also use it as a bootstrap condition to prevent this: When we reach a certain level of nesting which we deem to be too deep we return to the original caller thereby unwinding the stack.

brauner:
Changelog 2016-01-13:
- if `lxc_container_new()` fails in `ls_get()` do not bail immediately by going to `goto out` rather just `continue`.

stgraber:
jenkins: test this please

brauner:
Changelog 2016-01-13:
- Make sure `stat` is `free()`d in `ls_get_swap()`.

brauner:
Changelog 2016-01-13:
- put set `ret = 0` after the end of the `for`-loop in `ls_get()` and not in the `for`-loop.

brauner:
Changelog 2016-01-13:
- When we did not find any containers and hence, there is nothing to print set `ret = EXIT_SUCCESS` so that we do not irritate the user by reporting an error to him.

hallyn:
> No, we can just `continue` there as well. I just took it to be one of the few critical points where we should bail.

I think bailing on any memory allocation would make more sense.

Now most cases of lxc_container_new failing are also memory allocation erors, but not all.

If you wanted to (in a future commit :) you could set the lxc_last_error or whatever it's called to ENOMEM in lxc_container_new, then detect that here to decide whehter to bail or continue.  But for now I think it's best to continue in this case, and bail on local memory allocation failures.

hallyn:
On Wed, Jan 13, 2016 at 02:22:07AM -0800, Christian Brauner wrote:
> I'd like to bring one small, maybe pointless thing. Since we're doing (non-tail) recursion here there is the inherent danger that we might blow the stack at some level of nesting. At least in the case where we do not attach to the namespace of a running container but rather append paths to get nested containers. Although the argument `lvl` in `ls_get()` was originally just to keep track of the nesting level for pretty-printing we could also use it as a bootstrap condition to prevent this: When we reach a certain level of nesting which we deem to be too deep we return to the original caller thereby unwinding the stack.
> 

How do we set a reasonable limit?

Allow overriding it on the command line?

brauner:
On Wed, Jan 13, 2016 at 09:06:03AM -0800, Serge Hallyn wrote:
> On Wed, Jan 13, 2016 at 02:22:07AM -0800, Christian Brauner wrote:
> > I'd like to bring one small, maybe pointless thing. Since we're doing (non-tail) recursion here there is the inherent danger that we might blow the stack at some level of nesting. At least in the case where we do not attach to the namespace of a running container but rather append paths to get nested containers. Although the argument `lvl` in `ls_get()` was originally just to keep track of the nesting level for pretty-printing we could also use it as a bootstrap condition to prevent this: When we reach a certain level of nesting which we deem to be too deep we return to the original caller thereby unwinding the stack.
> > 
> 
> How do we set a reasonable limit?
Although rather arbitrary I would say 5 should be more than enough for most
purposes.
Thoughts?

> 
> Allow overriding it on the command line?
Definitely. If someone wants to try her/his luck she/he should be able to do so.
I'd say we let

        --nesting

take an optional (unsigned int) argument (specifying the maximum nesting level).
Thoughts?

brauner:
On Wed, Jan 13, 2016 at 09:02:53AM -0800, Serge Hallyn wrote:
> > No, we can just `continue` there as well. I just took it to be one of the few critical points where we should bail.
> 
> I think bailing on any memory allocation would make more sense.
> 
> Now most cases of lxc_container_new failing are also memory allocation erors, but not all.
> 
> If you wanted to (in a future commit :) you could set the lxc_last_error or
> whatever it's called to ENOMEM in lxc_container_new, then detect that here to
I thought about that too. The problem with this approach is that when
lxc_container_new() fails than there will be no lxc_last_error variable
(actually its called error_num, I think :)) which we can set but POSIX and susv
align themselves in the future (from what I can gather from
http://pubs.opengroup.org/onlinepubs/9699919799/) in setting errno to ENOMEM:

        Issue 6

        Extensions beyond the ISO C standard are marked.
        The following new requirements on POSIX implementations derive from alignment
        with the Single UNIX Specification:

        In the RETURN VALUE section, the requirement to set errno to indicate an error
        is added.

        The [ENOMEM] error condition is added.

So we can simply add an errno = ENOMEM memory check right after
lxc_container_new in lxc_ls.c and goto out when it fails and if errno is not set
simply continue. Can't hurt.

PR #757, opened 2016-01-10
Add openSUSE Leap release in opensuse template

tsdmgz:
Adding release Leap 42.1. Repo locations were moved one level deeper with it, explaining why $DISTRO is made to "leap/42.1". I couldn't think of a cleaner way to do it.


hallyn:
jenkins: test this please

hallyn:
Thanks - looks good to me.  Jenkins run looks like a waste of watts, but i've seen weirder things than a commit like this causing failure, so...

will merge in a bit.

tsdmgz:
Thank you!
Quick question: is it okay to set 42.1 instead of 13.1 as default? Sorry, forgot to ask that earlier.

hallyn:
On Fri, Jan 15, 2016 at 08:01:29AM -0800, TSDominguez wrote:
> Thank you!
> Quick question: is it okay to set 42.1 instead of 13.1 as default? Sorry, forgot to ask that earlier.

What are the implications either way?

tsdmgz:
That's what I'm concerned about. With limited usage I've done so far, the 42.1 image gets built and runs without issue.

PR #759, opened 2016-01-11
btrfs: snapshot by default

hallyn:
Always try a btrfs snapshot first.  If that fails and the user
did not request -s, then try a copy-clone.

See lp #1532125


hallyn:
No, this is no good.

hallyn:
new approach.

hallyn:
jenkins: test this please

PR #777, opened 2016-01-28
cgroup ns: move the check for whether cgns is supported

hallyn:
We have to do it before we join the container's mntns so we have
the host's procdir.


stgraber:
jenkins: test this please

stgraber:
jenkins: test this please

PR #798, opened 2016-02-03
Comment the lxc_rootfs structure

hallyn:
Comment rootfs.path and rootfs.mount so people can better figure
out which to use.

Remove the unused pivotdir argument from setup_rootfs_pivot_root().
Remove the unused pivot member of the lxc_rootfs struct.  And just
return 0 (success) when someone passes a lxc.pivotdir entry.  One
day we'll turn that into an error, but not yet...


brauner:
Needs a rebase.

PR #799, opened 2016-02-03
move and rename mount_entry_create_aufs_dirs()

brauner:
- The function mount_entry_create_aufs_dirs() moves from conf.c to
  lxcaufs.{c,h} where it belongs.
- In accordance with the "aufs_" prefix naming scheme for functions associated
  with lxcaufs.{c,h} mount_entry_create_aufs_dirs() becomes aufs_mkdir().

### DIFF
--- a/src/lxc/bdev/lxcaufs.c
+++ b/src/lxc/bdev/lxcaufs.c
@@ -30,6 +30,7 @@
 
 #include "bdev.h"
 #include "log.h"
+#include "lxcaufs.h"
 #include "lxcrsync.h"
 #include "utils.h"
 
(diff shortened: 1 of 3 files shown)


hallyn:
Hi - looks good, but the duplicate aufs_get_rootfs() worries me.  Do you think it will diverge from ovl_get_rootfs()?  If not, should they be merged into a stackfs_get_rootfs() in utils.c?

Happy to merge this anyway and wait for a future patch consolidating, but leaving this open while we discuss.

brauner:
I thought that maybe - at some point - we should give each block device type /storage type a small function that returns the rootfs for that block device type / storage type. And then have some sort of wrapper function (e.g. `get_rootfs()`) that hands us the rootfs for each block device type / storage type. When the mount syntax changes we can then simply update the individual helpers while `get_rootfs()` stays exactly the same.

Concerning this commit specifically, can we assume that `lxc.rootfs` entries for aufs and overlay can be parsed the same way? I was worried that the syntax might change at some point. But if you prefer we can just create one common helper right now and put it into `bdev.{c,h}`.

hallyn:
Sounds like keeping them separate like you have them is the better way, thanks.

PR #800, opened 2016-02-03
no rootfs => mounts always relative to host's /

brauner:
All lxc.mount.entry entries will be relative to the host's / when a container
does not have a rootfs.


hallyn:
Thanks.  Unfortunately github has confused itself with my merge of your last patch and thinks it is now unmergable.  'git am' happily applies it locally, but @stgraber has the tree locked down so I can't push :)

Can you please rebase?

brauner:
Done. :)

PR #801, opened 2016-02-03
allow overlay lxc.mount.entry with no rootfs

brauner:
Allow lxc.mount.entry entries for containers without a rootfs.


stgraber:
Needs rebasing

PR #808, opened 2016-02-08
apparmor: don't fail if current aa label is given

hallyn:
Ideally a container configuration will specify 'unchanged' if
it wants the container to use the current (parent) profile.  But
lxd passes its current label.  Support that too.

Note that if/when stackable profiles exist, this behavior may
or may not be what we want.  But the code to deal with aa
stacking will need some changes anyway so this is ok.

With this patch, I can create nested containers inside a
lxd xenial container both using

lxc launch x2

and unprivileged

lxc-start -n x2

### DIFF
--- a/src/lxc/lsm/apparmor.c
+++ b/src/lxc/lsm/apparmor.c
@@ -146,16 +146,15 @@ static bool aa_stacking_supported(void) {
 	return false;
 }
 
-/* are we in a confined container? */
-static bool in_aa_confined_container(void) {
-	char *p = apparmor_process_label_get(getpid());
-	bool ret = false;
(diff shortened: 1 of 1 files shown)


brauner:
I think the test suite failed due to network errors but I'm not completely sure so holding off the merge.

brauner:
Nope, I was wrong, lxc-test-apparmor-mount fails.

hallyn:
jenkins: test this please

hallyn:
Previous patch had a flaw in its logic.  This one locally passes lxc-test-apparmor for me.  Which doesnt' mean I didn't break another testcase.

PR #811, opened 2016-02-10
lxc_destroy: be quiet if asked

hallyn:
As per https://bugs.launchpad.net/bugs/1543016.

### DIFF
--- a/src/lxc/lxc_destroy.c
+++ b/src/lxc/lxc_destroy.c
@@ -33,6 +33,7 @@
 lxc_log_define(lxc_destroy_ui, lxc);
 
 static int my_parser(struct lxc_arguments* args, int c, char* arg);
+static bool quiet;
 
 static const struct option my_longopts[] = {
 	{"force", no_argument, 0, 'f'},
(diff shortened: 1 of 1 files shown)


martinpitt:
Thanks Serge for fixing this! But I wonder if that was a bit overzealous? The usual meaning of `--quiet` is to not show logging and success messages, but error messages like "no such container" or "insufficient privileges" should still be shown.

hallyn:
On Wed, Feb 10, 2016 at 02:32:00PM -0800, Martin Pitt wrote:
> Thanks Serge for fixing this! But I wonder if that was a bit overzealous? The usual meaning of `--quiet` is to not show logging and success messages, but error messages like "no such container" or "insufficient privileges" should still be shown.

Yeah I wondered that.  I'm open to changing it, kinda felt like shutting it up completely.

It will still return -1 on error.  If you're doing 'lxc-destroy -q -n ab || true' or 'if lxc-destroy -q -n ab; then echo ab existed' you don't want even error messages.

brauner:
On Thu, Feb 11, 2016 at 09:15:56AM -0800, Serge Hallyn wrote:
> On Wed, Feb 10, 2016 at 02:32:00PM -0800, Martin Pitt wrote:
> > Thanks Serge for fixing this! But I wonder if that was a bit overzealous? The usual meaning of `--quiet` is to not show logging and success messages, but error messages like "no such container" or "insufficient privileges" should still be shown.
> 
> Yeah I wondered that.  I'm open to changing it, kinda felt like shutting it up completely.
Hm, most users will probably take "--quiet" to simply mean "does not print
stuff".

PR #813, opened 2016-02-10
lxc-destroy: deal with ephemeral containers

brauner:
- Ephemeral containers are destroyed on shutdown so we do not destroy them.
- Destroy ephemeral containers with clones: first destroy all the clones, then
  destroy the container.
- Ephemeral containers with snapshots cannot be easily handled but we can
  probably trust that no one will try to make snapshots of an ephemeral
  container.

### DIFF
--- a/src/lxc/lxc_destroy.c
+++ b/src/lxc/lxc_destroy.c
@@ -58,13 +58,13 @@ Options :\n\
 	.task     = DESTROY,
 };
 
-static int do_destroy(struct lxc_container *c);
-static int do_destroy_with_snapshots(struct lxc_container *c);
+static bool do_destroy(struct lxc_container *c);
+static bool do_destroy_with_snapshots(struct lxc_container *c);
 
 int main(int argc, char *argv[])
 {
 	struct lxc_container *c;
-	int ret;
+	bool bret;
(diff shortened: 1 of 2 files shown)


hallyn:
jenkins: test this please

hallyn:
Ok, so the old lxc-start-ephemeral python script may end up racy, but (a) that's going away and (b) it was never a good idea to destroy those.  lxc-copy ephemeral scripts are not racy as lxc_fini won't return until it is destroyed.

Looks good - thanks!

brauner:
lxc-test-apparmor-mount which shouldn't be affected by this patch at all. Any idea why?

brauner:
Regarding the Python lxc-start-ephemeral racing possibility: It can't race. I fixed the script some time ago right when I implemented ephemeral containers. All it does now is set `lxc.ephemeral = 1` in the containers config before starting it. The post-stop removal script it used I've taken out. So lxc-start-ephemeral should be safe as well.

hallyn:
jenkins: test this please

hallyn:
@brauner  yeah i'm pretty sure that's a transient error on the server, those failure don't make sense.

PR #818, opened 2016-02-13
Make lxc-config use the lxc parser

### LINKED ISSUE #259: Have all C commands use the same code
Currently, most C commands use arguments.c and the common parser, but there are a few that do their own things and so don't respect common options like --version, those should be converted to using the same argument parser as the rest.

brauner:
closes #259

This ist the last executable we needed to convert. Now all lxc executables use the common lxc parser.

`lxc-unshare` and `lxc-usernsexec` do not count since they are not solely concerned with the lxc suite.

Note, that this change requires a change to a command line flag: `lxc-config` uses `-l` without an argument to list all containers. The parser however uses this for the logfile. Hence, I changed `lxc-config` to use `-L/--list`. I think this change is acceptable since we're probably releasing `lxc 2.0` at some point and we're changing quite a bit anyway .

### DIFF
--- a/src/lxc/arguments.h
+++ b/src/lxc/arguments.h
@@ -99,7 +99,7 @@ struct lxc_arguments {
 	/* auto-start */
 	int all;
 	int ignore_auto;
-	int list;
+	int list; /* also used by lxc-config */
 	char *groups; /* also used by lxc-ls */
 
 	/* lxc-snapshot and lxc-clone */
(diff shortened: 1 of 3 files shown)


hallyn:
This doesn't seem to work.  'lxc-config -l' now prints an error, and 'lxc-config' with no arguments doesn't.

I appreciate the idea of commonizing the code, but in this case I think it complicates a very simple piece of code for no benefit.  If we were planning on complicating lxc-config, it'd be worth it, but it doens't seem to me that this is worth it.

(Please do argue if you feel I'm wrong :)

brauner:
Yeah, changing the command-line flag would unfortunately be necessary. I'm not going to argue for something trivial; just my passion to finally close #259.

PR #835, opened 2016-02-21
fix android build

brauner:
bionic libc doesn't know _Exit(). Replace it with _exit().


stgraber:
And Android is happy again! Thanks!

PR #836, opened 2016-02-22
2016 02 21/cgns.aa

hallyn:
Apparmor policy support for user namespaces.


stgraber:
Looking in git I also see a number of tests which directly check for the lxc-container-default string, those same tests will fail when run on a cgns enabled kernel with this branch, so they need to be updated.

stgraber:
The lxc.container.conf manpage should probably be updated too

PR #846, opened 2016-02-25
cgfs: do not automount if cgroup namespaces are supported

hallyn:
In that case containers will be able to mount cgroup filesystems
for themselves as they do on a host.

This fixes inability to start systemd based containers on cgns-enabled
kernels with cgmanager not running.

I've tested debianjessie, busybox, ubuntu trusty and xenial, all of
which booted ok.  However if there are some setups which require
premounted cgroupfs (i.e. they don't mount if they detect being in
a container), this may cause trouble.


stgraber:
Can you update lxc.container.conf.sgml to mention this behavior?

hallyn:
Yup, done.

PR #849, opened 2016-02-25
lxc-attach: rework pty allocation

brauner:
Previously we implemented two ways to get a pty for lxc-attach:
	1. get a pty in the container
	2. get a pty on the host

Where 1. was the default and 2. was only tried after 1. failed.
For safety and simplicity reasons, we remove 1. and just keep 2. around.

### DIFF
--- a/doc/lxc-attach.sgml.in
+++ b/doc/lxc-attach.sgml.in
@@ -80,20 +80,17 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
     </para>
     <para>
     Previous versions of <command>lxc-attach</command> simply attached to the
-    specified namespaces of a container and ran a shell or the specified
-    command without allocating a pseudo terminal. This made them vulnerable to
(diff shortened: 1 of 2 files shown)


hallyn:
Hi,

just fwiw this looks fine, but it doesn't look to me like the current order should hurt anything, because we're not dereferencing anything from 'c'.

brauner:
It will matter when someone else tries to fiddle with the container while we're still waiting, no?

hallyn:
Sorry, can be more precise?  Note that 'lxc_container_put/get' is only about locking the in-memory structure among threads, and we (lxc_attach.c) are not threaded.  You'd need the disk_lock to lock among processes.  We aren't worried about that.

But maybe I'm missing your concern.

brauner:
Dammit, I always mix those two up. Anyway, I got some weird behavior before and thought I traced it back to the ordering `waitpid()` and `lxc_container_put()` but maybe that were still side effects from `lxc-attach()` behaving badly when the container exits unexpectedly.

PR #851, opened 2016-02-26
start: setsid in init task as well

tych0:
If we don't do this, we'll leak the parent's session id to the container,
which maybe doesn't matter, but it still seems better to set it anyway.

Also, it breaks CRIU for containers that don't call setsid themselves.

Signed-off-by: Tycho Andersen <tycho.andersen@canonical.com>

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -851,6 +851,8 @@ static int do_start(void *data)
 		goto out_warn_father;
 	}
 
+	setsid();
+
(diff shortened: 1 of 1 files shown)


stgraber:
Can there be a case where someone would intentionally want to use the same session as the parent?
Thinking about application containers kind of use maybe.

hallyn:
(discussed on irc, we're all pretty sure it's the right thing and won't break anything)

PR #859, opened 2016-02-29
lxc-attach: always allocate current controlling pty

### LINKED ISSUE #858: Segmentation fault on lxc-attach
Hi.

Ubuntu Xenial Xerus (development branch)
4.4.0-7-generic #22-Ubuntu SMP Thu Feb 18 20:50:38 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

    # lxc-attach --version
    2.0.0.rc4

Fresh install. What was done:

    # apt-get install lxc libnl-3-dev criu pkg-config thin-provisioning-tools libprotobuf-dev protobuf-c-compiler libcap-dev protobuf-compiler
    # # get latest criu from git and compile
    # lxc-create --template ubuntu --name goodicon.ru -- --release xenial
    # lxc-start -n goodicon.ru
    # lxc-attach -n goodicon.ru
    Segmentation fault (core dumped)

Strace:

[out.txt](https://github.com/lxc/lxc/files/150627/out.txt)

brauner:
closes #858 

lxc-attach uses lxc_console_create() to allocate a pty on the host.
lxc_console_create() in turn calls lxc_console_peer_default() which either
makes the current controlling pty our controlling pty for the container, or
uses whatever the user gave us (e.g. /dev/tty2 etc.). For lxc-attach we always
want the current controlling pty to be used. This commit ensures that we're in
fact always using the current controlling pty. The commit also fixes a segfault
when the user specified lxc.console.path = none.

### DIFF
--- a/src/lxc/lxc_attach.c
+++ b/src/lxc/lxc_attach.c
@@ -253,10 +253,27 @@ static int get_pty_on_host(struct lxc_container *c, struct wrapargs *wrap, int *
 	}
 
 	conf = c->lxc_conf;
+	free(conf->console.log_path);
+	conf->console.log_path = NULL;
+
+	/* In the case of lxc-attach our peer pty will always be the current
+	 * controlling terminal. We clear whatever was set by the user for
+	 * lxc.console.path here and set it to "/dev/tty". Doing this will (a)
+	 * prevent segfaults when the container has been setup with
+	 * lxc.console = none and (b) provide an easy way to ensure that we
+	 * always do the correct thing. strdup() must be used since console.path
+	 * is free()ed when we call lxc_container_put(). */
+	free(conf->console.path);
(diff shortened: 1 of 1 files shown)


hallyn:
jenkins: test this please

hallyn:
Nicely spotted.

My only concern here is that the lxc.console.log_path is meant to be a second copy of the console output.  So the container configuration could point to a log_path in /var/log, and lxc-attach ought to send output both to that file and tty.

Merging this anyway as it fixes a segfault which is more important.  We can decide what to do about the log_output later.

brauner:
lxc-attach does not seem to be able to send output to that log path file. If we want this feature we need to figure out a way to have lxc-attach be able to talk to send output to that file.

PR #872, opened 2016-03-04
Execute script lxc-devsetup also with sysvinit and upstart.

clopez:
* This script sets /dev/.lxc which is needed for autodev containers.
  * Previously was only executed with systemd. Execute it also with
    the other init systems (sysvinit and upstart)


config/init/upstart/Makefile.am
```diff
@@ -1,4 +1,4 @@
-EXTRA_DIST = lxc.conf lxc-instance.conf lxc-net.conf.in
+EXTRA_DIST = lxc.conf.in lxc-instance.conf lxc-net.conf.in
```
brauner: This is `lxc.conf` without the `.in`-prefix on purpose; it doesn't need to be processed before.

brauner:
I'll leave this one to @hallyn.

hallyn:
Heh, in general I leave stuff like this to @stgraber .  If he likes I'll look, but this is his domain :)

stgraber:
Oh, I forgot that this feature even still existed, never used it myself but the change seems reasonable.

jenkins: test this please

PR #873, opened 2016-03-04
lxc-attach: fix redirection - stderr

brauner:
So that we can do things like:

	lxc-attach -n a -- sh -c 'echo ERR >&2' > /dev/null

There seems to be no easy way to discern when we need to write to stderr
instead of stdout when we receive an event on the master fd of an allocated
pty. So we're using a "trick"/"hack". We write to STDOUT_FILENO if it refers to
a pty. If STDOUT_FILENO does not refer to a pty we check whether STDERR_FILENO
refers to a pty and if so write to it.

### DIFF
--- a/src/lxc/lxc_attach.c
+++ b/src/lxc/lxc_attach.c
@@ -273,7 +273,39 @@ static int get_pty_on_host(struct lxc_container *c, struct wrapargs *wrap, int *
 	if (lxc_console_create(conf) < 0)
 		return -1;
 	ts = conf->console.tty_state;
-	ts->stdoutfd = STDOUT_FILENO;
+	/*
+	 * We need to make sure that the ouput that is produced inside the
+	 * container is received on the host. Suppose we want to run
+	 *
+	 *	lxc-attach -n a -- /bin/sh -c 'hostname >&2' > /dev/null
+	 *
+	 * This command produces output on stderr inside the container. On the
+	 * host we close stdout by redirecting it to /dev/null. But stderr is,
+	 * as expected, still connected to a tty. We receive the output produced
+	 * on stderr in the container on stderr on the host.
+	 *
+	 * For the command
+	 *
+	 *	lxc-attach -n a -- /bin/sh -c 'hostname >&2' 2> /dev/null
+	 *
+	 * the logic is analogous but because we now have closed stderr on the
+	 * host no output will be received.
+	 *
+	 * Finally, imagine a more complicated case
+	 *
+	 *	lxc-attach -n a -- /bin/sh -c 'echo OUT; echo ERR >&2' > /tmp/out 2> /tmp/err
+	 *
+	 * Here, we produce output in the container on stdout and stderr. On the
+	 * host we redirect stdout and stderr to files. Because of that stdout
+	 * and stderr are not dup2()ed. Thus, they are not connected to a pty
+	 * and output on stdout and stderr is redirected to the corresponding
+	 * files as expected.
+	 */
+	if (!isatty(STDOUT_FILENO) && isatty(STDERR_FILENO))
+		ts->stdoutfd = STDERR_FILENO;
+	else
+		ts->stdoutfd = STDOUT_FILENO;
+
 	conf->console.descr = &descr;
 
 	/* Shift ttys to container. */
--- a/src/tests/lxc-test-lxc-attach
+++ b/src/tests/lxc-test-lxc-attach
@@ -1,0 +1,135 @@
+#!/bin/sh
+
+# lxc: linux Container library
+
+# Authors:
+# Christian Brauner <christian.brauner@mailbox.org>
+#
+# This is a test script for the lxc-attach program. It tests whether I/O
+# redirection works correctly.
+
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+set -e
+
+FAIL() {
+	echo -n "Failed " >&2
+	echo "$*" >&2
+        lxc-destroy -n busy -f
+	exit 1
+}
+
+# Create a container, start it and wait for it to be in running state
+lxc-create -t busybox -n busy || FAIL "creating busybox container"
+lxc-start -n busy -d || FAIL "starting busybox container"
+lxc-wait -n busy -s RUNNING || FAIL "waiting for busybox container to run"
+
+# Test if simple attach is working
+# - stdin  --> attached to pty
+# - stdout --> attached to pty
+# - stderr --> attached to pty
+attach=$(lxc-attach -n busy -- hostname) # expect output
+if [ "$attach" != "busy" ]; then
+        FAIL " simple attach"
+fi
+
+# Test if we still receive output when stdin does not refer to a tty
+# - stdin  --> /dev/null
+# - stdout --> attached to pty
+# - stderr --> attached to pty
+attach=$(lxc-attach -n busy -- hostname < /dev/null) # expect output
+if [ "$attach" != "busy" ]; then
+        FAIL " with stdin redirection < /dev/null"
+fi
+
+# Test if we are silent when stdout does not refer to a tty
+# - stdin  --> attached to pty
+# - stdout --> /dev/null
+# - stderr --> attached to pty
+attach=$(lxc-attach -n busy -- hostname > /dev/null) # expect no output
+if [ -n "$attach" ]; then
+        FAIL " with stdout redirection > /dev/null"
+fi
+
+# Test if we are silent when stdout and stdin do not refer to a tty
+# - stdin  --> /dev/null
+# - stdout --> /dev/null
+# - stderr --> attached to pty
+attach=$(lxc-attach -n busy -- hostname > /dev/null < /dev/null) # expect no output
+if [ -n "$attach" ]; then
+        FAIL " with stdin and stdout redirection > /dev/null < /dev/null"
+fi
+
+# Test if we still receive output when stdin and stderr do not refer to a tty
+# - stdin  --> /dev/null
+# - stdout --> attached to pty
+# - stderr --> /dev/null
+attach=$(lxc-attach -n busy -- hostname 2> /dev/null < /dev/null) # expect output
+if [ "$attach" != "busy" ]; then
+        FAIL " with stdin and stderr redirection 2> /dev/null < /dev/null"
+fi
+
+# - produce output on stderr in container
+# - redirect stdout on host to /dev/null
+# - output on host should be received on stderr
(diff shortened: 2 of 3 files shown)


stgraber:
Can we add a simple test for this?

Just creating and starting a busybox container and then doing attach with the various modes I listed here: http://paste.ubuntu.com/15282891/ should prevent any regression from happening.

brauner:
Yup, I'm on this. I don't know if I'll have time to finish it this weekend though. I'll try to squeeze it in.

src/tests/lxc-test-lxc-attach
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+# Test if we still receive output when stdin does not refer to a tty
+attach=$(lxc-attach -n busy -- hostname < /dev/null) # expect output
+if [ "$attach" != "busy" ]; then
+        FAIL " with stdin redirection < /dev/null"
+fi
+
+# Test if we are silent when stdout does not refer to a tty
+attach=$(lxc-attach -n busy -- hostname > /dev/null) # expect no output
```
martinpitt: This does not actually test lxc-attach.  `$(anything >/dev/null)` will *always* be empty. Maybe you meant to test that you don't get any stderr here?

src/tests/lxc-test-lxc-attach
```diff
@@ -0,0 +1,71 @@
...
+if [ -n "$attach" ]; then
+        FAIL " with stdout redirection > /dev/null"
+fi
+
+# Test if we are silent when stdout and stdin do not refer to a tty
+attach=$(lxc-attach -n busy -- hostname > /dev/null < /dev/null) # expect no output
```
martinpitt: Same here.

src/tests/lxc-test-lxc-attach
brauner: Ah, you probably want your test: `sudo lxc-attach -n adt-virt-lxc-lbmmpf -- sh -c 'echo ERR>&2' >/dev/null`?

src/lxc/lxc_attach.c
```diff
@@ -273,7 +273,12 @@ static int get_pty_on_host(struct lxc_container *c, struct wrapargs *wrap, int *
 	if (lxc_console_create(conf) < 0)
 		return -1;
 	ts = conf->console.tty_state;
-	ts->stdoutfd = STDOUT_FILENO;
+	/* If stdout is not a tty but stderr is, we write to stderr. */
+	if (!isatty(STDOUT_FILENO) && isatty(STDERR_FILENO))
+		ts->stdoutfd = STDERR_FILENO;
+	else
+		ts->stdoutfd = STDOUT_FILENO;
+
```
martinpitt: The comment and intent is highly unclear -- why would we *ever* redirect the container command's stdout to stderr? It shouldn't matter what input redirection is being done on the host. This looks like very unexpected and untypical Unix behaviour and magic?

src/tests/lxc-test-lxc-attach
martinpitt: Adding that would be useful too indeed (as a test for this regression). But I also meant the other way round:

    attach=$(lxc-attach -n busy -- hostname 2>&1)
    if [ -n "$attach" ]; then
        FAIL "unexpected stderr output witih stdout redirection"

src/lxc/lxc_attach.c
hallyn: Martin, are you joking?  Your example in the description is:

lxc-attach -n a -- sh -c 'echo ERR >&2' > /dev/null

redirecting container's stdout to stderr.

hallyn: Oh, I see.  Martin, this is because the 'ts' only has stdin and stdout, no stderr.  The container's output is combined into the ts->stdout fd.

In your example you redirected container's stdout to /dev/null;  we need to make sure to let any valid output from container reach the console, so we send it to fd 2.

martinpitt: This does *not* redirect container's stdout to stderr. `sh -c 'echo ERR >&2'` is merely a synthetic reproducer command (running in the container) that writes something to stderr. But lxc-attach apparently now redirects stderr of the command in the container to stdout of lxc-attach under some conditions -- that's the regression. In case it's clearer, you can also use this one:

    lxc-attach -n a -- yes --foo > /dev/null

yes' error message about the unknown option also goes to stderr, so you should see it. But lxc-attach redirects it to stdout now, therefore hiding it. (It was still working correctly until rc1 at least)

martinpitt: > we need to make sure to let any valid output from container reach the console, so we send it to fd 2.

That's the bit that I don't understand. If a command (run in the container) is only outputting stuff to its stdout, and lxc-attach's stdout is redirected into a file, that command stdout output should go into that file, and nowhere else. Why should it suddenly go to the host's (i. e. lxc-attach's) stderr? Similarly, the container command's stderr data should go (only) to lxc-attach's stderr, nowhere else.

That's why "If stdout is not a tty but stderr is, we write to stderr" sounds plain wrong. Maybe the intent is something different here?

Out of interest, what is `ts`? If that's something that multiplexes the command's stdout and stderr, how does that get demultiplexed on the other end?

src/tests/lxc-test-lxc-attach
```diff
@@ -0,0 +1,71 @@
...
+if [ -n "$attach" ]; then
+        FAIL " with stdin and stdout redirection > /dev/null < /dev/null"
+fi
+
+# Test if we still receive output when stdin and stderr do not refer to a tty
+attach=$(lxc-attach -n busy -- hostname 2> /dev/null < /dev/null) # expect output
+if [ "$attach" != "busy" ]; then
+        FAIL " with stdin and stderr redirection 2> /dev/null < /dev/null"
+fi
```
martinpitt: Would also be useful to do combined testing:

    lxc-attach-n busy -- sh -c 'echo OUT; echo ERR >&2' > /tmp/out 2>/tmp/err

and ensuring that `/tmp/out` contains exactly `OUT`  and `/tmp/err` contains exactly `ERR` (just with proper temp file names of course)

hallyn: Yes this test would be good.

src/lxc/lxc_attach.c
hallyn: Agreed it is confusing, and I'm only paying half attention so should probably let Christian explain.
The fundamental reason this had to change was that, for security reasons, we needed to interject a new pty and insulate lxc-attach's caller's tty from the container.
If any caller's fd 0-2 is not tty, then it is *not* redirected.  If any is, then a new pty is dup'ed over it.
So when you lxc-attach -n x1  ... < a > b, then stderr will point to a new pty, but 0 and 1 will point to a and b, respectively.  if you lxc-attach -n x1 ... < a, then 1 and 2 will both be dups of the new pty, and in that case stdout and stderr are combined.

src/tests/lxc-test-lxc-attach
brauner: Maybe I'm mistaken, but wouldn't you expect output in this case? You're redirecting stderr to stdout.

hallyn: Quoting Christian Brauner (notifications@github.com):
> > +lxc-wait -n busy -s RUNNING || FAIL "waiting for busybox container to run"
> > +
> > +# Test if simple attach is working
> > +attach=$(lxc-attach -n busy -- hostname) # expect output
> > +if [ "$attach" != "busy" ]; then
> > +        FAIL " simple attach"
> > +fi
> > +
> > +# Test if we still receive output when stdin does not refer to a tty
> > +attach=$(lxc-attach -n busy -- hostname < /dev/null) # expect output
> > +if [ "$attach" != "busy" ]; then
> > +        FAIL " with stdin redirection < /dev/null"
> > +fi
> > +
> > +# Test if we are silent when stdout does not refer to a tty
> > +attach=$(lxc-attach -n busy -- hostname > /dev/null) # expect no output
> 
> Maybe I'm mistaken, but wouldn't you expect output in this case? You're redirecting stderr to stdout.

lxc-attach -n busy -- hostname > /dev/null - expect no output
lxc-attach -n busy -- hostname 2>&1 - expect output
lxc-attach -n busy -- "hostname 2>&1" - expect output
lxc-attach -n busy -- "hostname 1>&2" > /dev/null - expect output

(compare to ssh)

src/tests/lxc-test-lxc-attach
brauner: Test added.

brauner:
jenkins: test this please

brauner:
jenkins: test this please

src/lxc/lxc_attach.c
brauner: This is indeed confusing. So maybe a longer explanation is in order. As Serge correctly pointed out we have to allocate a new pty for `lxc-attach` when any of the standard file descriptors refers to a `pty`. This is to avoid providing an attack surface for input faking via a `TIOCSTI` `ioctl()` call.

The idea is to create a new pty master/slave pair on the host via `openpty()`. We then pass the file descriptors for the master/slave pair to our `lxc_attach()` function. `lxc_attach()` attaches to the (specified) namespaces of a running container and runs a command in it. Before, this was done via either one of two callback functions `lxc_attach_run_shell()` or `lxc_attach_run_command()`, depending on whether the user wants an interactive shell or simply run a command he specifies. In the new implementation we provide an additional wrapper around these two functions. This wrapper receives the master/slave pair file descriptors we created on the host and basically does what `login_tty()` would do. We only skip the `fork()` step since `lxc_attach()` will fork anyway. To go into a little more detail: Right after the fork we close the master file descriptor of the pty since we do not need it anymore. Afterwards we do `setsid()` to become session leader and `ioctl(TIOCSTTY)` on the slave file descriptor of the pty to become the controlling terminal of the session. Finally, and this is the crucial step here, we `dup2()` any standard file descriptors (stdin, stdout, and stderr) that refer to a pty on the host. It is important to note that this is an iff: If-and-only-if the standard file descriptor in question refers to a pty on the host, we `dup2()` it. If not we leave it untouched. This means redirection should stay intact. In case all file descriptors refer to a pty stdout and stderr are mixed and there is no way to separate them (This however, should not be a problem since that simply means that no redirection occurs in most cases. If I'm not mistaken this is similar to an ssh session that allocates a pty.) However, if an fd like stderr does not refer to a pty it is not `dup2()`ed. Thus, in an example like:

```
lxc-attach -n a -- sh -c "hostname >&2" > /dev/null
```

the following happens:
- We create a new pty on the host and receive a master/slave fd pair.
- We pass the master/slave fd pair to our wrapper and send it to `lxc_attach()` to execute in the containers namespaces.
- The wrapper checks which standard file descriptors refer to a pty and `dup2()`s those. For our example this means: The wrapper will find that stdin and stderr refer to a terminal an `dup2()`s them. Since stdout, however, is redirected and does not refer to a pty it will not be `dup2()` and refer to `/dev/null`. Within the container the synthetic reproducer created via `/bin/sh -c "hostname >&2"` generates output on stderr. We also want to receive that ouput on stderr on the host. The little trick in the code we pull is to check if stdout refers to a pty and if so we set `ts->stdoutfd` to stdout. If however, we detect that stdout does not refer to a pty we test whether stderr still does and if so set `ts->stdoutfd` to stderr. This will allow us to receive output on the correct standard file descriptor as the command in the container sent output to.

Now, with regards to `ts`. `ts` is a structure we use to keep all information regarding pty setup to `epoll()` on in a mainloop. This mainloop has two i/o handlers: `lxc_console_cb_tty_stdin()` and `lxc_console_cb_tty_master()`. `*_tty_stdin()` will wait for input on stdin (refered to via `ts->stdinfd` and write to `ts->masterfd`. `*_cb_tty_master()` will then receive the input on `ts->masterfd` and write to `ts->stdoutfd` which we have setup appropriately before (to stdout or stderr). In case no redirection occurs at all, `ts->stdoutfd` can be safely set to stdout since the stdout and stderr streams are mixed anyway. I hope this helps to clarify a little.

PR #875, opened 2016-03-05
cgfsng: next generation filesystem-backed cgroup implementation

hallyn:
This makes simplifying assumptions:  all usable cgroups must be
mounted under /sys/fs/cgroup/controller or /sys/fs/cgroup/contr1,contr2.

Currently this will only work with cgroup namespaces, because
lxc.mount.auto = cgroup is not implemented.  So cgfsng_ops_init()
returns NULL if cgroup namespaces are not enabled.

### DIFF
--- a/src/lxc/cgfsng.c
+++ b/src/lxc/cgfsng.c
@@ -1324,53 +1324,6 @@ static int cgfsng_set(const char *filename, const char *value, const char *name,
 }
 
 /*
- * Check whether a container already has a particular rule, as otherwise
- * may end up with spurious permission errors.
- */
-static bool cgroup_devices_has_allow_or_deny(struct cgfsng_handler_data *d,
-					     char *v, bool for_allow, char *path)
-{
-	FILE *devices_list;
-	char *line = NULL;
-	size_t sz = 0;
-	bool ret = !for_allow;
-
-	/* if it's a deny rule and container has all devices, then it doesn't
-	 * yet have the deny rule */
-	if (!for_allow && strcmp(v, "a") != 0 && strcmp(v, "a *:* rwm") != 0)
-		return false;
-
-	devices_list = fopen_cloexec(path, "r");
-	if (!devices_list) {
-		free(path);
-		return false;
-	}
-
-	while (getline(&line, &sz, devices_list) != -1) {
-		size_t len = strlen(line);
-		if (len > 0 && line[len-1] == '\n')
-			line[len-1] = '\0';
-		if (strcmp(line, "a *:* rwm") == 0) {
-			/* if container has all access and we're adding allow rule,
-			 * then already has it; if it has all access and we're
-			 * adding a deny rule, then it does not. */
-			ret = for_allow;
-			goto out;
-		} else if (for_allow && strcmp(line, v) == 0) {
-			/* if the line is there verbatim and it is an
-			 * allow rule, then it already has it */
-			ret = true;
-			goto out;
-		}
-	}
-
-out:
-	fclose(devices_list);
-	free(line);
-	return ret;
-}
-
-/*
  * Called from setup_limits - here we have the container's cgroup_data because
  * we created the cgroups
  */
@@ -1425,12 +1378,6 @@ static bool cgfsng_setup_limits(void *hdata, struct lxc_list *cgroup_settings,
 		cg = iterator->elem;
 
 		if (do_devices == !strncmp("devices", cg->subsystem, 7)) {
-			if (strcmp(cg->subsystem, "devices.deny") == 0 &&
-					cgroup_devices_has_allow_or_deny(d, cg->value, false, listpath))
-				continue;
-			if (strcmp(cg->subsystem, "devices.allow") == 0 &&
-					cgroup_devices_has_allow_or_deny(d, cg->value, true, listpath))
-				continue;
 			if (lxc_cgroup_set_data(cg->subsystem, cg->value, d)) {
 				if (do_devices && (errno == EACCES || errno == EPERM)) {
 					WARN("Error setting %s to %s for %s",
--- a/src/lxc/cgfsng.c
+++ b/src/lxc/cgfsng.c
@@ -648,6 +648,10 @@ static bool parse_hierarchies(struct cgfsng_handler_data *d)
 	char **klist = NULL, **nlist = NULL;
 	size_t len = 0;
 
+	/*
+	 * Root spawned containers escape the current cgroup, so use init's
+	 * cgroups as our base in that case.
+	 */
 	if (geteuid())
 		basecginfo = read_file("/proc/self/cgroup");
 	else
@@ -922,7 +926,7 @@ static void remove_path_for_hierarchy(struct hierarchy *h, char *cgname)
 }
 
 /*
- * Try to create the same cgrou pin all hierarchies.
+ * Try to create the same cgroup in all hierarchies.
  * Start with cgroup_pattern; next cgroup_pattern-1, -2, ..., -999
  */
 static inline bool cgfsng_create(void *hdata)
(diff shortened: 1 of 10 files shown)


src/lxc/cgfsng.c
```diff
@@ -0,0 +1,1466 @@
...
+#include "log.h"
+#include "cgroup.h"
+#include "utils.h"
+#include "commands.h"
+
+lxc_log_define(lxc_cgfsng, lxc);
+
+static struct cgroup_ops cgfsng_ops;
+
+//#define EXTRADEBUG 1
```
brauner: on purpose?

src/lxc/cgfsng.c
```diff
@@ -0,0 +1,1466 @@
...
+ */
+static char **get_controllers(char **klist, char **nlist, char *line)
+{
+	// the fourth field is /sys/fs/cgroup/comma-delimited-controller-list
+	int i;
+	char *p = line, *p2, *tok, *saveptr = NULL;
+	char **aret = NULL;
+
+	for (i = 0; i < 4; i++) {
+		p = index(p, ' ');
```
brauner: trivial point: `index()` is deprecated in favor of `strchr()`. (I think there are 10 references overall to `index()` instead of `strchr()`: `s/index(/strchr(/g`.)

src/lxc/cgfsng.c
```diff
@@ -0,0 +1,1466 @@
...
+	if (clist) {
+		int i;
+
+		for (i = 0; clist[i]; i++)
+			free(clist[i]);
+		free(clist);
+	}
+}
+
+/* Re-alllocate a pointer, do not fail */
```
brauner: Why would you not want to fail on `ENOMEM`? Wouldn't this basically hang indefinitely until memory is available?

src/lxc/cgfsng.c
```diff
@@ -0,0 +1,1466 @@
...
+		return false;
+	}
+	len = strlen(tmp) + 5; // leave room for -NNN\0
+	cgname = must_alloc(len);
+	strcpy(cgname, tmp);
+	free(tmp);
+	offset = cgname + len - 5;
+
+again:
+	if (idx == 1000)
```
brauner: A `DEBUG()` saying that too many cgroups have been created or something could be nice.

src/lxc/cgfsng.c
hallyn: Quoting Christian Brauner (notifications@github.com):
> > +	char *container_cgroup; // cgroup we created for the container
> > +	char *name; // container name
> > +};
> > +
> > +static void free_string_list(char **clist)
> > +{
> > +	if (clist) {
> > +		int i;
> > +
> > +		for (i = 0; clist[i]; i++)
> > +			free(clist[i]);
> > +		free(clist);
> > +	}
> > +}
> > +
> > +/* Re-alllocate a pointer, do not fail */
> 
> Why would you not want to fail on `ENOMEM`? Wouldn't this basically hang indefinitely until memory is available? 

Nah, oom will kick in, which I suspect will happen anyway.

This shouldn't be done when reading less-privileged files or doing
huge allocations, but for a 20 byte string this seems to me better
than trying to cleanly handle the failure up the stack just to
get killed by oom a few seconds later anyway.

I'm open to the idea that I'm overdoing it here, but I liked
this idea in libnih, and I'm liking it here...

brauner:
Looks great, +1.

src/lxc/cgfsng.c
hallyn: Quoting Christian Brauner (notifications@github.com):
> > +/*
> > + * Get the controllers from a mountinfo line
> > + * There are other ways we could get this info.  For lxcfs, field 3
> > + * is /cgroup/controller-list.  For cgroupfs, we could parse the mount
> > + * options.  But we simply assume that the mountpoint must be
> > + * /sys/fs/cgroup/controller-list
> > + */
> > +static char **get_controllers(char **klist, char **nlist, char *line)
> > +{
> > +	// the fourth field is /sys/fs/cgroup/comma-delimited-controller-list
> > +	int i;
> > +	char *p = line, *p2, *tok, *saveptr = NULL;
> > +	char **aret = NULL;
> > +
> > +	for (i = 0; i < 4; i++) {
> > +		p = index(p, ' ');
> 
> trivial point: `index()` is deprecated in favor of `strchr()`. (I think there are 10 references overall to `index()` instead of `strchr()`: `s/index(/strchr(/g`.)

Hm, deprecated isn't right, but not supported by bionic may be.
I thought that was only rindex, but I can't keep it straight.

src/lxc/cgfsng.c
hallyn: Quoting Christian Brauner (notifications@github.com):
> > +		return false;
> > +	}
> > +
> > +	tmp = lxc_string_replace("%n", d->name, d->cgroup_pattern);
> > +	if (!tmp) {
> > +		ERROR("Failed expanding cgroup name pattern");
> > +		return false;
> > +	}
> > +	len = strlen(tmp) + 5; // leave room for -NNN\0
> > +	cgname = must_alloc(len);
> > +	strcpy(cgname, tmp);
> > +	free(tmp);
> > +	offset = cgname + len - 5;
> > +
> > +again:
> > +	if (idx == 1000)
> 
> A `DEBUG()` saying that too many cgroups have been created or something could be nice.

Hmmm, there will be 1000 'failed to create' messages before that :)  Keeping
this less noisy was one of my goals.  But I added this.

I'm considering rate-limiting the 'failed to create' messages. Not sure of
the best way.

src/lxc/cgfsng.c
brauner: Ah, you're using EXTRADEBUG further down in the code.

hallyn: Quoting Christian Brauner (notifications@github.com):
> > +#include <errno.h>
> > +#include <sys/types.h>
> > +#include <unistd.h>
> > +#include <dirent.h>
> > +#include <grp.h>
> > +
> > +#include "log.h"
> > +#include "cgroup.h"
> > +#include "utils.h"
> > +#include "commands.h"
> > +
> > +lxc_log_define(lxc_cgfsng, lxc);
> > +
> > +static struct cgroup_ops cgfsng_ops;
> > +
> > +//#define EXTRADEBUG 1
> 
> on purpose?

Yeah, I don't want to always do that, and I realize this
isn't something I can ask users to easily turn on...  but
until I think of a good way to switch it on when needed I
figure it's doing no harm.

src/lxc/cgfsng.c
brauner: Brute force would probably be to keep a simple count somewhere: `was_warned` and then do `if (was_warned > some_arbitrary_value) do not warn`.

hallyn: > Brute force would probably be to keep a simple count somewhere: `was_warned` and then do `if (was_warned > some_arbitrary_value) do not warn`.

Yeah, others are to cache the last error and only print it if we fail;
or warn when count % 100 == 0;  or somesuch.

The pain of course is that you don't want it except when you want it :)

But printing the last error only when we fail might be best.  After
all if you care enough you can just check which index was actually
used, i.e. if you had 94 failures, you can check the container :)

stgraber:
This looks so much cleaner than the old cgfs backend! I can even understand what the code does!

The only bit that seems potentially fragile though I haven't found anything obviously wrong with it is the code to check whether a given device is already in the allow/deny list.
This code looks fragile to me due to the different ways of globbing those devices, making it not obvious to spot overlaps.

hallyn:
On Mon, Mar 07, 2016 at 09:55:57AM -0800, Stéphane Graber wrote:
> This looks so much cleaner than the old cgfs backend! I can even understand what the code does!
> 
> The only bit that seems potentially fragile though I haven't found anything obviously wrong with it is the code to check whether a given device is already in the allow/deny list.
> This code looks fragile to me due to the different ways of globbing those devices, making it not obvious to spot overlaps.

Realistically we can probably drop this now.  I think this was mainly there so that unpriv users who had lxc.cgroup.devices lines but couldn't write to devices cgroup files.  We work around that now with different include files for user namespaced containers.

PR #890, opened 2016-03-10
cgfsng: get_cgroup_path: return the cgroup path not full mounted path

hallyn:
Add a temporary workaround for talking to containers started with the
buggy monitor.  We can remove it sometime after 2.0 release.


tych0:
+1 works for me.

PR #897, opened 2016-03-16
Prevent access to pci devices

hallyn:
Prevent privileged containers from messing with the host's pci devices
directly.  Refuse access under /proc/bus, and drop cap_sys_rawio.  Some
containers may need to re-enable cap_sys_rawio (i.e. if they run an
X server).

It may be desirable to break some of this stuff into files which can be
separately included (or not included), but this patch isn't the right
place for that.


stgraber:
You forgot to re-generate the final aparmor profiles (see README).

PR #898, opened 2016-03-16
2016 03 16/cgroupauto

hallyn:
Not sure about the second patch.  Do we want cgmanager to be used if available and compiled in, or not?

### DIFF
--- a/src/lxc/cgfsng.c
+++ b/src/lxc/cgfsng.c
@@ -1022,9 +1022,6 @@ static void cgfsng_destroy(void *hdata, struct lxc_conf *conf)
 
 struct cgroup_ops *cgfsng_ops_init(void)
 {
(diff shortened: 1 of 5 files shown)


brauner:
What behavior do we expect for systems that do not handout writeable cgroups for crucial subsystems on login? If having cgmanager available will make it easier to deal with the aforementioned case than we should keep it in. If not let's definitely kick it; smaller codebase => easier maintenance. :)

stgraber:
@hallyn gcc looks unhappy

stgraber:
I think we need to keep cgmanager around for a bit just for backport's and so long as we keep the code around, I think it makes sense to use it over cgfsng when it's available and LXC was built with it.

brauner:
The `c99` trap with `gcc`; `clang` always allows for-loop initial declarations, `gcc` doesn't.

PR #900, opened 2016-03-17
autodev: don't always create /dev/console

tych0:
In particular, only create /dev/console when it is set to "none".
Otherwise, we will bind mount a pts device later, so let's just leave it.

Also, when bind mounting the pts device, let's create /dev/console if it
doesn't exist, since it may not already exist due to the above :)

Signed-off-by: Tycho Andersen <tycho.andersen@canonical.com>

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1130,7 +1130,7 @@ static const struct lxc_devs lxc_devs[] = {
 	{ "console",	S_IFCHR | S_IRUSR | S_IWUSR,	       5, 1	},
 };
 
-static int fill_autodev(const struct lxc_rootfs *rootfs)
+static int fill_autodev(const struct lxc_rootfs *rootfs, bool mount_console)
 {
 	int ret;
 	char path[MAXPATHLEN];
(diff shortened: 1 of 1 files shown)


src/lxc/conf.c
```diff
@@ -1404,9 +1407,14 @@ static int setup_dev_console(const struct lxc_rootfs *rootfs,
 		return -1;
 	}
 
-	if (access(path, F_OK)) {
-		WARN("rootfs specified but no console found at '%s'", path);
-		return 0;
+	fd = open(path, O_CREAT, S_IXUSR | S_IXGRP | S_IXOTH);
+	if (fd < 0) {
+		if (errno != EEXIST) {
+			SYSERROR("failed ot create console");
```
stgraber: typo

src/lxc/conf.c
```diff
@@ -1152,6 +1152,10 @@ static int fill_autodev(const struct lxc_rootfs *rootfs)
 	cmask = umask(S_IXUSR | S_IXGRP | S_IXOTH);
 	for (i = 0; i < sizeof(lxc_devs) / sizeof(lxc_devs[0]); i++) {
 		const struct lxc_devs *d = &lxc_devs[i];
+
+		if (!strcmp(d->name, "console") && !mount_console)
```
stgraber: hmm, shouldn't that be strcmp(d->name, "console") == 0 && !mount_console ?

tych0: On Thu, Mar 17, 2016 at 10:31:21AM -0700, Stéphane Graber wrote:
> > @@ -1152,6 +1152,10 @@ static int fill_autodev(const struct lxc_rootfs *rootfs)
> >  	cmask = umask(S_IXUSR | S_IXGRP | S_IXOTH);
> >  	for (i = 0; i < sizeof(lxc_devs) / sizeof(lxc_devs[0]); i++) {
> >  		const struct lxc_devs *d = &lxc_devs[i];
> > +
> > +		if (!strcmp(d->name, "console") && !mount_console)
> 
> hmm, shouldn't that be strcmp(d->name, "console") == 0 && !mount_console ?

there's a ! there, but I can change it to == 0 if you want.

PR #917, opened 2016-03-24
Fix daemonized containers without autodev and/or /dev mount

### LINKED ISSUE #910: Container fails to start daemonized. But starts successfully in foreground.
I have a container that starts successfully both in background and foreground with LXC 1.0.6 (from Debian Jessie repositories).

When I upgrade to LXC 1.1.5 (I've rebuilt source package from Sid), or try to start exactly the same container on Ubuntu 16.04 (LXC 2.0-rc), it fails to start daemonized.
It starts successfully only when I pass -F to lxc-start.

Log files:
- [fg.txt](https://github.com/lxc/lxc/files/186120/fg.txt) is the log file for foreground mode (successful start)
- [bg.txt](https://github.com/lxc/lxc/files/186119/bg.txt) is for background mode (failure).

I see following errors when I try to start the container in background:

      lxc-start 20160323131017.938 ERROR    lxc_sync - sync.c:__sync_wait:51 - invalid sequence number 1. expected 4
      lxc-start 20160323131017.938 ERROR    lxc_start - start.c:__lxc_start:1286 - failed to spawn 'mycontainer'

What may this "invalid sequence number" mean?

Also, looks like it fails to clean up:

      lxc-start 20160323131017.941 WARN     lxc_cgfsng - cgfsng.c:cgroup_rmdir:964 - cgroup_rmdir: failed to delete /sys/fs/cgroup/systemd//lxc/mycontainer: No such file or directory
      lxc-start 20160323131017.941 ERROR    lxc_cgfsng - cgfsng.c:recursive_destroy:999 - Error destroying /sys/fs/cgroup/systemd//lxc/mycontainer
      lxc-start 20160323131017.943 WARN     lxc_cgfsng - cgfsng.c:cgroup_rmdir:964 - cgroup_rmdir: failed to delete /sys/fs/cgroup/cpu//lxc/mycontainer: No such file or directory
      lxc-start 20160323131017.943 ERROR    lxc_cgfsng - cgfsng.c:recursive_destroy:999 - Error destroying /sys/fs/cgroup/cpu//lxc/mycontainer
      lxc-start 20160323131017.946 WARN     lxc_cgfsng - cgfsng.c:cgroup_rmdir:964 - cgroup_rmdir: failed to delete /sys/fs/cgroup/cpuset//lxc/mycontainer: No such file or directory
      lxc-start 20160323131017.946 ERROR    lxc_cgfsng - cgfsng.c:recursive_destroy:999 - Error destroying /sys/fs/cgroup/cpuset//lxc/mycontainer

amezin:
I have a container whose /dev is populated by its init, so I don't need autodev hook.

Recent versions of LXC (starting from 1.1, I think) try to open /dev/null from container's rootfs when starting daemonized. And the container fails to start, because there's no /dev/null yet.

I changed the code to open /dev/null before fs root switch, so now container's init will be connected to host's /dev/null. I think it shouldn't break anything or cause a security problem.

Also, the error message was a bit confusing ("unexpected sequence number"). I tried to improve it here too.

This PR fixes #910.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -868,9 +868,9 @@ static int do_start(void *data)
 	handler->ops->start(handler, handler->data);
 
 out_warn_father:
-	/* we want the parent to know something went wrong, so any
-	 * value other than what it expects is ok. */
-	lxc_sync_wake_parent(handler, LXC_SYNC_POST_CONFIGURE);
+	/* we want the parent to know something went wrong, so we return a special
+	 * error code. */
+	lxc_sync_wake_parent(handler, LXC_SYNC_ERROR);
 	return -1;
 }
 
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -710,6 +710,7 @@ static int do_start(void *data)
 {
 	struct lxc_list *iterator;
 	struct lxc_handler *handler = data;
+	int devnull_fd = -1;
 
 	if (sigprocmask(SIG_SETMASK, &handler->oldmask, NULL)) {
 		SYSERROR("failed to set sigprocmask");
@@ -788,6 +789,13 @@ static int do_start(void *data)
 	}
 	#endif
 
+	if (handler->backgrounded) {
+		devnull_fd = open_devnull();
+
+		if (devnull_fd < 0)
+			goto out_warn_father;
+	}
+
 	/* Setup the container, ip, names, utsname, ... */
 	if (lxc_setup(handler)) {
 		ERROR("failed to setup the container");
@@ -796,7 +804,7 @@ static int do_start(void *data)
 
 	/* ask father to setup cgroups and wait for him to finish */
 	if (lxc_sync_barrier_parent(handler, LXC_SYNC_CGROUP))
-		return -1;
+		goto out_error;
 
 	/* Set the label to change to when we exec(2) the container's init */
 	if (lsm_process_label_set(NULL, handler->conf, 1, 1) < 0)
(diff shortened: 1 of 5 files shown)


brauner:
@stgraber, I'll leave this to you.

stgraber:
That feels wrong to me. If I remember that code well, it means any error which would usually be printed by LXC during that whole init sequence now won't be printed to the terminal.

stgraber:
@hallyn

amezin:
@stgraber Yes, seems you're right... Then I think the right solution is to open /dev/null at the beginning (where I moved `null_stdfds()`), and do `dup2()` at the end, where `null_stdfds()` currently is.

amezin:
@stgraber Fixed

stgraber:
That looks better, still would like @hallyn to take a look though.

stgraber:
jenkins: test this please

hallyn:
Thanks - looks good to me.

stgraber:
Just a note that this is regressing CRIU. @tych0 is taking a look.

Basically CRIU cannot restore open fds where the source is outside the container.

Current though is to restrict this code change to only trigger if the container has autodev disabled AND /dev/null doesn't exist inside it, keeping the code unmodified (open /dev/null from inside the container) for all other cases.

amezin:
But in this case CRIU will still fail to restore the fd when the container doesn't have /dev/null.

It's better to simply revert https://github.com/lxc/lxc/commit/c44de748afb0080271f6a68bc5defb2c74516859, I think... I can pre-create /dev/null in the container fs, after all.

stgraber:
@amezin Sure but the combination of daemonized + no-autodev + no /dev/null is pretty rare so that's fine.

hallyn:
Is someone working on that patch then?

Also, couldn't/shouldn't criu special case this for silly things like /dev/null, /dev/zero?

stgraber:
@hallyn @tych0 fixed this already with #927

hallyn:
probably - though I still think it would make sense for criu to specially mark open fds to /dev/null and /dev/zero, since that's a very common special case.

PR #919, opened 2016-03-25
Fix installation of out-of-tree (VPATH) builds

amezin:
Currently, LXC successfully builds out-of-tree (at least it looks so), but `make install` fails, because it tries to install some non-generated files without full path. Prepending "$(srcdir)/" fixes it, and also doesn't break in-source build.


stgraber:
jenkins: test this please

PR #920, opened 2016-03-25
Timezone inside the container is not the same as the host

thtanaka:
In OL6, it is missing /etc/sysconfig/clock

Tested-by: Deepak Patel <deepak.patel@oracle.com>
Signed-off-by: Thomas Tanaka <thomas.tanaka@oracle.com>


brauner:
jenkins: test this please

PR #922, opened 2016-03-26
use httpredir.debian.org as the default Debian mirror

evgeni:
http.debian.net is an alias anyways and httpredir.debian.org is the official name of the service


brauner:
jenkins: test this please

Issue #625, opened 2015-08-12
lxc-debian should not enable contrib and non-free by default


daniel-baumann:
contrib and non-free by definition are not part of the official debian distribution, lxc-debian should not enable it by default.

nehaljwani:
I think commit c2a85d043b35d82087b85f8bfb75ae578f3d9ac3 fixes this issue.

PR #925, opened 2016-03-26
only enable Debian's main repository by default

evgeni:
This inverts the logic done in c2a85d0 to default to "main only" and allow enabling contrib and non-free on user request.

Closes: #625

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -512,7 +512,7 @@ Options :
   --packages=PACKAGE_NAME1,PACKAGE_NAME2,...
                          List of additional packages to install. Comma separated, without space.
   -c, --clean            only clean up the cache and terminate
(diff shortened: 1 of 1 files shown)


brauner:
We might end up confusing users who rely on the current default behavior of the script. (ping @stgraber)

evgeni:
@brauner then including this in the 1.1 → 2.0 switch is a good time, no? :)

I agree that this is a behavior switch, but I do not think people actually need non-free in containers (no, I have no numbers) as non-free is a lot about docs, drivers and firmware -- stuff you usually do not need in a container.

stgraber:
I think the other flag was introduced recently too (not in any stable release) so changing it shouldn't really break anyone either, the only possible breakage is the change to the resulting sources.list but as @evgeni said, it's unlikely anyone actually needed contrib and non-free.

PR #926, opened 2016-03-27
add funs to mmap() files to \0-terminated strings

brauner:
In order to do this we make use of the MAP_FIXED flag of mmap(). MAP_FIXED
should be safe to use when it replaces an already existing mapping. To this
end, we establish an anonymous mapping that is one byte larger than the
underlying file. The pages handed to us are zero filled.  Now we establish a
fixed-address mapping starting at the address we received from our anonymous
mapping and replace all bytes excluding the additional \0-byte with the file.
This allows us to use normal string-handling function.  The idea implemented
here is similar to how shared libraries are mapped.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -2164,7 +2164,7 @@ static bool mod_rdep(struct lxc_container *c0, struct lxc_container *c, bool inc
 
 			if (fbuf.st_size != 0) {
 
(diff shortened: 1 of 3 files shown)


stgraber:
@hallyn

hallyn:
Thanks - the patch looks good.  I'm probably over-thinking this, but would it be worthwhile to rename this lxc_string_mmap or something like it to make clear what it is doing when you just see it being called?

brauner:
@hallyn, thanks. Renamed.

hallyn:
Thanks

PR #929, opened 2016-03-30
use smarter error handling for lxc_strmmap()

brauner:
Additionally, fix the comment for lxc_strmmap() and cleanup #includes in
utils.{c,h}.


hallyn:
Hi,

regarding the re-ordering of the includes, in general we want to keep 'sys/' includes after others.  So unistd.h should come before sys/*.

PR #932, opened 2016-03-30
fix possible buffer overflow

n-eiling:
strncat only returns its first argument and not the end of the written string.
Thus "buf-pos" is always 0 and consquently no range check is performed.
Also the terminating null character was not being accounted for.

Signed-off-by: Niklas Eiling <niklas.eiling@rwth-aachen.de>

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -364,10 +364,13 @@ static void exec_criu(struct criu_opts *opts)
 
 	buf[0] = 0;
 	pos = 0;
+
 	for (i = 0; argv[i]; i++) {
-		strncat(buf, argv[i], sizeof(buf) - pos - 1);
-		strncat(buf, " ", sizeof(buf) - pos - 1);
-		pos += strlen(argv[i]);
+		ret = snprintf(buf + pos, sizeof(buf) - pos, "%s ", argv[i]);
+		if (ret < 0 || ret >= sizeof(buf) - pos)
+			goto err;
+		else
+			pos += ret;
 	}
 
 	INFO("execing: %s", buf);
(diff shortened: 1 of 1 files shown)


stgraber:
jenkins: test this please

src/lxc/criu.c
```diff
@@ -363,10 +363,11 @@ static void exec_criu(struct criu_opts *opts)
 	argv[argc] = NULL;
 
 	buf[0] = 0;
-	pos = buf;
+	pos = 0;
 	for (i = 0; argv[i]; i++) {
-		pos = strncat(buf, argv[i], buf + sizeof(buf) - pos);
-		pos = strncat(buf, " ", buf + sizeof(buf) - pos);
+		strncat(buf, argv[i], sizeof(buf) - pos - 1);
```
brauner: Just out of curiosity, did you test that code on a small program or something? It looks like it would segfault when `pos += strlen(argv[i])` is larger than `sizeof(buf)`. Your code and the original code would work safely if it included a check like:
```
size_t len = 0;
for (...) {
        len += strlen(argv[i]);
        if (len + 1 > sizeof(buf))
                break; // or something similar
}
```

hallyn: Also every strncat and strlen is walking the length of that string.  It would
be nice to just do something like:

	size_t pos = 0;
	for (i = 0; i < arv[i]; i++) {
		ret = snprintf(buf + pos, sizeof(buf) - pos, "%s", argv[i]);
		if (ret < 0 || ret + pos >= sizeof(buf))
			EOVERFLOW
		pos += ret;
	}

n-eiling: I agree. Using snprintf is much nicer.

stgraber:
jenkins: test this please

PR #934, opened 2016-03-31
Fix buffer overflow in do_start()

amezin:
It's unlikely to happen, but still better check for overflow before writing, not after.


brauner:
jenkins: test this please

Issue #935, opened 2016-04-02
lxc-destroy kaput for unprivileged btrfs-backed containers with kernel >= 4.2


overlayfs:
To reproduce the bug on ubuntu 14.04, 15.10 or 16.04, create an unprivileged, btrfs-backed container and then attempt to destroy it:
``lxc-create -B btrfs -t downloads -n test_unprivileged_btrfs -- -d ubuntu -r wily -a amd64``
``lxc-destroy -n test_unprivileged_btrfs``
``sudo lxc-destroy -P /home/ubuntu/.local/share/lxc -n test_unprivileged_btrfs``
-  **sudo lxc-destroy**:  always succeeds.
- **lxc-destroy**: succeeds for kernels <= 4.1 if the container only has a single btrfs subvolume; otherwise it fails.
- **lxc-destroy**:  always fails for kernels >= 4.2 (tested up to 4.6rc1).

The error message for kernel >= 4.2 is:
_lxc-destroy: bdev/lxcbtrfs.c: btrfs_recursive_destroy: 619 Error: can't perform the search under /home/ubuntu/.local/share/lxc/test_unprivileged_btrfs/rootfs_

overlayfs:
**Details**
The error is unaffected by:
- the version of lxc (regular repositories or ppa:ubuntu-lxc/daily)
- the kernel variant (regular repositories or mainline ppa).

overlayfs:
**Straces**

**strace when lxc-destroy succeeds: (kernel <= 4.1)**
     open("/home/ubuntu/.local/share/lxc/test_unprivileged_btrfs/rootfs", O_RDONLY) = 4
     ioctl(4, BTRFS_IOC_INO_LOOKUP, 0x7ffe3ae4bd80) = -1 EPERM (Operation not permitted)
     close(4)                          = 0
     open("/home/ubuntu/.local/share/lxc/test_unprivileged_btrfs", O_RDONLY) = 4
     ioctl(4, BTRFS_IOC_SNAP_DESTROY, 0x7ffc6d5f7970) = 0

**strace when lxc-destroy fails: (kernel >= 4.2)**
     open("/home/ubuntu/.local/share/lxc/test_unprivileged_btrfs/rootfs", O_RDONLY) = 4
     ioctl(4, BTRFS_IOC_INO_LOOKUP, 0x7fff204b69d0) = 0
[...]

PR #936, opened 2016-04-02
fix btrfs_recursive_destroy

overlayfs:
A change in kernel 4.2 caused btrfs_recursive_destroy to
fail to delete unprivileged containers.  This patch restores
the pre-kernel-4.2 behaviour.  Ref: Issue #935 .

Signed-off-by: Oleg Freedhom <overlayfs@gmail.com>

### DIFF
--- a/src/lxc/bdev/lxcbtrfs.c
+++ b/src/lxc/bdev/lxcbtrfs.c
@@ -616,8 +616,13 @@ static int btrfs_recursive_destroy(const char *path)
 		ret = ioctl(fd, BTRFS_IOC_TREE_SEARCH, &args);
 		if (ret < 0) {
 			close(fd);
-			ERROR("Error: can't perform the search under %s\n", path);
 			free_btrfs_tree(tree);
+			if (errno == EPERM || errno == EACCES) {
+				WARN("Warn: can't perform the search under %s. Will simply try removing", path);
+				goto ignore_search;
+			}
+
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

brauner:
@hallyn, I don't see a problem with this but I haven't tried to reproduce the issue on a > 4.2 kernel and probably won't be able to do this today.

src/lxc/bdev/lxcbtrfs.c
```diff
@@ -616,8 +616,13 @@ static int btrfs_recursive_destroy(const char *path)
 		ret = ioctl(fd, BTRFS_IOC_TREE_SEARCH, &args);
 		if (ret < 0) {
 			close(fd);
-			ERROR("Error: can't perform the search under %s\n", path);
 			free_btrfs_tree(tree);
+			if (errno == EPERM || errno == EACCES) {
```
stgraber: errno could be modified by the close() or free_btrfs_tree() calls, you probably want it saved right after the ioctl and then look at the saved copy.

hallyn: True but same is true in the other place where we goto ignore_search.  Would you mind fixing both here?

overlayfs:
errno is now saved immediately after the ioctl.

brauner:
jenkins: test this please

PR #940, opened 2016-04-03
switch user facing name to lxc.rootfs.backend

hallyn:


stgraber:
Can you add a doc entry?

PR #949, opened 2016-04-06
netlink_open: close socket on error

Blub:
All uses of netlink_open() assume that on error the
nl_handler doesn't need to be closed, but some error cases
happen after the socket was opened successfully and used to
simply return -errno.


brauner:
jenkins: test this please

PR #951, opened 2016-04-07
lxc-attach: attach even without sigwinch handler

brauner:
lxc_console_create() calls lxc_console_peer_default() which in turn calls
lxc_console_sigwinch_init() which sets up the lxc_tty_state struct for the
current pty. Prior to this commit lxc_console_sigwinch_init() would consider
failures to install a SIGWINCH handler fatal and and return NULL. This commit
makes failures to install a SIGWINCH handler non-fatal. In such cases the
lxc_tty_state struct will still be set up but the sigfd member, which contains
the fd which receives SIGWINCH events, will be set to -1. (This also entails
that the sigset_t oldmaks field is meaningless.) Callers of
lxc_console_sigwinch_init() and lxc_console_sigwinch_fini() should thus make
sure that sigfd >= 0 or sigfd != -1 before trying to register a SIGWINCH handler
in e.g. an lxc_mainloop or resetting the sigmask (cf. lxc_attach.c).

These changes also imply that lxc_console_sigwinch_init() only fails with
ENOMEM. Thus, all cases where lxc_console_sigwinch_init() returns NULL are to be
considered fatal. This wasn't the case before this commit.

### DIFF
--- a/src/lxc/console.c
+++ b/src/lxc/console.c
@@ -118,9 +118,9 @@ struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
 		return NULL;
 
 	memset(ts, 0, sizeof(*ts));
-	ts->stdinfd  = srcfd;
+	ts->stdinfd = srcfd;
 	ts->masterfd = dstfd;
-	ts->sigfd    = -1;
+	ts->sigfd = -1;
 
 	/* add tty to list to be scanned at SIGWINCH time */
 	lxc_list_add_elem(&ts->node, ts);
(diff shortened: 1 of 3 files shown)


brauner:
rebasing this before merge

src/lxc/console.c
```diff
@@ -129,26 +129,19 @@ struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
 	sigemptyset(&mask);
 	sigaddset(&mask, SIGWINCH);
 	if (sigprocmask(SIG_BLOCK, &mask, &ts->oldmask)) {
-		SYSERROR("failed to block SIGWINCH");
-		goto err1;
+		SYSERROR("failed to block SIGWINCH.");
+		ts->sigfd = -1;
 	}
 
 	ts->sigfd = signalfd(-1, &mask, 0);
```
hallyn: Should this only be done if the sigprocmask didn't fail?

PR #959, opened 2016-04-10
2016 04 08/lxc attach simplify and log

hallyn:
Do not merge - testing jenkins.


brauner:
@hallyn, the bug lies in the fact that the standard file descriptors do not refer to a pty on jenkins. The pty logging option however only works in this case. I sent a new pr for this (#958). Thanks for helping debug this. Closing this pr.

PR #961, opened 2016-04-11
Use size_t as an iteration point over int

staticfox:
This cleans up some sign-compare warnings as well as avoids any
possibilities of unintended signed offsets for indices during
iteration.

Signed-off-by: Matt Ullman <staticfox@staticfox.net>

### DIFF
--- a/src/lxc/parse.c
+++ b/src/lxc/parse.c
@@ -66,7 +66,7 @@ int lxc_file_for_each_line(const char *file, lxc_file_cb callback, void *data)
 
 int lxc_char_left_gc(const char *buffer, size_t len)
 {
-	int i;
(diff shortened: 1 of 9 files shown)


src/lxc/parse.c
```diff
@@ -78,7 +78,7 @@ int lxc_char_left_gc(const char *buffer, size_t len)
 
 int lxc_char_right_gc(const char *buffer, size_t len)
 {
-	int i;
+	size_t i;
```
brauner: That should either remain `int i` or you turn `size_t len` and `int i` into `ssize_t`.

brauner:
jenkins: test this please

PR #966, opened 2016-04-12
nicer date format and support for SOURCE_DATE_EPOCH in LXC_GENERATE_DATE

evgeni:
Using $(date) for LXC_GENERATE_DATE has various flaws:
* formating depends on the locale of the system we execute configure on
* the output is not really a date but more a timestamp

Let's use $(date --utc '+%Y-%m-%d') instead.

While at it, also support SOURCE_DATE_EPOCH [1] to make the build
reproducible

[1] https://reproducible-builds.org/specs/source-date-epoch/


brauner:
jenkins: test this please

PR #969, opened 2016-04-13
lxcaufs: use correct branch syntax for aufs mounts

### LINKED ISSUE #967: aufs mount syntax with "br:" not supported
In the [aufs manpage](https://sourceforge.net/p/aufs/aufs-util/ci/aufs4.x-rcN/tree/aufs.in.5) of the current version of aufs-utils, only the br: syntax for branch options is documented, the br= syntax is not mentioned. But lxc only supports the br= syntax, and it fails without any explanation if you try to use the other syntax.

Most mount errors are accompanied at least by an explanatory WARN level log message, but none of the "goto err" cases in aufs_mkdir() or ovl_mkdir() result in a log message, either directly or in the caller mount_entry_create_dir_file(). So all you get is a vague message like

lxc-start: conf.c: lxc_setup: 3920 failed to setup the mount entries for 'container-name'

If the mount is optional, you don't even get that much.

brauner:

That should close #967.

Someone more familiar with aufs should take a look whether `br:` is the only valid branch syntax or whether we also need to support `br=` (@hallyn?).

### DIFF
--- a/src/lxc/bdev/lxcaufs.c
+++ b/src/lxc/bdev/lxcaufs.c
@@ -248,7 +248,7 @@ int aufs_mount(struct bdev *bdev)
 		return -22;
 
 	//  separately mount it first
-	//  mount -t aufs -obr=${upper}=rw:${lower}=ro lower dest
+	//  mount -t aufs -o br:${upper}=rw:${lower}=ro lower dest
 	dup = alloca(strlen(bdev->src)+1);
 	strcpy(dup, bdev->src);
 	/* support multiple lower layers */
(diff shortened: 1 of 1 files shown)


tenforward:
The README of aufs says that both syntax are equivalent.
https://github.com/sfjro/aufs4-linux/tree/aufs4.x-rcN/Documentation/filesystems/aufs (at "4. Usage")

brauner:
Sigh...

hallyn:
@brauner - note the louder complaint in that issue was about the lack of error logging on failure.

It sounds like switching to br: *should* be safe, but if they're supposed to be equivalent anyway then I think I'd rather leave it as is (to not break any current users) and add a good error message.

I do *not* want fallback code, bc this code is convoluted enough as it is :)  Unless some reorg can simplify the flow.

brauner:
As discussed on irc: closing this.

Issue #310, opened 2014-08-21
sysvinit uses invalid lsb functions


ghost:
the sysvinit script uses "action" function which doesn't exist, rather than log_daemon_msg.

evgeni:
@stgraber can you please close this? https://github.com/lxc/lxc/commit/3083782cf871b7278f4643414bc060830659f6f4 is fixing it, but the closes line in the commit failed :(

Issue #309, opened 2014-08-21
sysvinit has wrong lsb function default locations


ghost:
by default, lsb functions are located in /lib/lsb/init-functions, not /etc/rc.d/init.d/functions.

PR #978, opened 2016-04-15
don't make sysv init scripts dependant on distribution specifics

evgeni:
- /etc(/rc.d)?/init.d/functions does not exist on all distributions
- LSB does not define a message function without an explicit status
- Debian-derived systems add a log_daemon_msg for that

lets define an own log_daemon_msg as echo and try to load LSB init
functions afterwards, which might overload it with a nicer version

that way the init scripts should work on any system, without hard
dependencies on neither LSB nor /etc/init.d/functions


brauner:
@stgraber.

ar45:
Related to #309 and #310

PR #980, opened 2016-04-18
Switch allocated pty to full raw mode.

LynxChaus:
Switch allocated pty to full raw mode. Without this some ncurses-based programs incorrectly display output.

Signed-off-by: Andrey. Jr. Melnikov <temnota.am@gmail.com>


brauner:
Hi,

two minor things:
- You haven't signed off your commits. This is something we require. (https://github.com/lxc/lxc/blob/master/CONTRIBUTING)
- All those "Merge branch" commits should not be present. You should simply rebase your feature branch onto current master.

Thanks.

PR #982, opened 2016-04-19
use raw settings of ssh for pty

brauner:
Let's be on the safe side and mimic ssh.

### DIFF
--- a/src/lxc/console.c
+++ b/src/lxc/console.c
@@ -260,16 +260,21 @@ int lxc_setup_tios(int fd, struct termios *oldtios)
 
 	newtios = *oldtios;
 
-	/* Remove the echo characters and signal reception, the echo
-	 * will be done with master proxying */
-	newtios.c_iflag &= ~(IGNBRK|ISTRIP|INLCR|IGNCR|ICRNL|IXON|IXANY|IXOFF);
-	newtios.c_iflag &= BRKINT;
-	newtios.c_lflag &= ~(ECHO|ICANON|ISIG|IEXTEN|ECHOE|ECHOK|ECHONL);
+	/* We use the same settings that ssh does. */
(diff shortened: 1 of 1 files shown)


src/lxc/console.c
```diff
@@ -260,16 +260,21 @@ int lxc_setup_tios(int fd, struct termios *oldtios)
 
 	newtios = *oldtios;
 
-	/* Remove the echo characters and signal reception, the echo
-	 * will be done with master proxying */
-	newtios.c_iflag &= ~(IGNBRK|ISTRIP|INLCR|IGNCR|ICRNL|IXON|IXANY|IXOFF);
-	newtios.c_iflag &= BRKINT;
```
hallyn: Do you know why the BRKINT was there originally?

What about using 'cfmakeraw' ?

hallyn: Do you know why the BRKINT was there originally?

What about using 'cfmakeraw' ?

brauner: Nope, not really nothing of this is documented here very well. But this is a bitwise and assignment meaning (more telling this to myself than you) that only when c_iflag already has BRKINT set to 1 will it be set to 1 otherwise zero.
My reason for changing this is that now we just keep randomly adding flags without a proper reason. ssh gives us a valid argument to set these flags. cfmakeraw turns off fewer options than ssh and I think the ssh settings make more sense. But I'm fine with using cfmakeraw if you'd prefer that.

hallyn: Thanks - not sure what I prefer;  I'd like the cfmakeraw for less cryptic code (and less chance of accidental breakage later as we're setting fewer of our ownoptions), but ssh is a nice standard to go by.

I'll merge as is - thanks.

PR #988, opened 2016-04-25
add vagrant for lxc

akshaykarle:
* uses 14.04 official image
* sets up the dependencies required to compile lxc
* shares the local development directory from host to /home/vagrant/lxc

Signed-off-by: Akshay Karle <akshay.a.karle@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,3 +145,5 @@ TAGS
 
 doc/api/doxygen_sqlite3.db
 doc/api/*.tmp
+
+.vagrant
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,15 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+  config.vm.box = "ubuntu/trusty64"
+
+  config.vm.synced_folder ".", "/home/vagrant/lxc"
+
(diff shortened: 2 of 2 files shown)


brauner:
Hi,

Thanks for the patch. But I think we would prefer to keep this out of our tree. You can always git clone lxc in the Vagrantfile itself though.

akshaykarle:
Hi @brauner thanks for your response.

I'm working off an OSX machine, so I need to run a VM to compile and test my changes. So I do my development on my host (which is OSX) and then test my changes on vagrant. Also, for developers running a linux host, it would be beneficial as you won't mess up your host machine incase you do a `make install`. And you can always start clean, do whatever mess you want and blow the VM away.

I'm okay if you don't want to include this in master anyway, but could you please help me understand the reasoning behind not including the Vagrantfile in master?

brauner:
Hi,

Vagrant is a separate upstream project dealing with a lot of virtualization solutions. It would not be a good idea to have all upstream projects Vagrant deals with provide a config file. Furthermore we might set a precedence which would invite people asking us to include their config files as well (juju etc.). Thanks again though!

akshaykarle:
That makes sense. Thanks for the clarification, I'll just keep a copy of the Vagrantfile on my dev machine.

PR #996, opened 2016-05-01
Added `type` to keys in lxc_list_nicconfigs

ar45:
I wanted to have network.type exposed in lxc-python. I'm not sure if we want this key to be part of the network config, if not here I would modify lxc-python to allow such access `Container('ct').network[0].type`.

Signed-off-by: Aron Podrigal <aronp@guaranteedplus.com>


brauner:
jenkins: test this please

PR #1012, opened 2016-05-09
Added OR for ID = rhel in RHEL 7+

hartnellfoster:
Discovered a bug when attempting to create CentOS containers on RHEL 7.2, where lxc-create would fail due to ID in /etc/os-release being 'rhel' rather than 'redhat'.


hartnellfoster:
Which is the correct branch into which to merge this PR?

brauner:
Hi, the branch target is correct. However, you're missing a signed-off-by line which is necessary for us to merge the commit (cf. https://github.com/lxc/lxc/blob/master/CONTRIBUTING). You can sign-off your commit with `git commit -s`.

PR #1014, opened 2016-05-09
Unshare netns after setting the userns mappings

hallyn:
so that there is a root uid mapping for the /proc/net files.

This is a fix for lxd issue #1978.


brauner:
jenkins: test this please

PR #1015, opened 2016-05-10
Allow configuration file values to be quoted

stewartbrodie:
If the value starts and ends with matching quote characters, those
characters are stripped automatically.   Quote characters are the
single quote (') or double quote (").  The quote removal is done after
the whitespace trimming.

This is needed particularly in order that lxc.environment values may
have trailing spaces.  However, the quote removal is done for all values
in the parse_line function, as it has non-const access to the value.

Signed-off-by: Stewart Brodie <stewart@metahusky.net>

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -1964,6 +1964,14 @@ static int parse_line(char *buffer, void *data)
 	value += lxc_char_left_gc(value, strlen(value));
 	value[lxc_char_right_gc(value, strlen(value))] = '\0';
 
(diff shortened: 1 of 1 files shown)


hallyn:
jenkins: ok to test

hallyn:
Thanks, you're right there's currently no way of doing this.  There are two open questions about this patch,

1. is anyone currently relying on the fact that quotes are not being removed?
2. get_config_value will return the processed value (without quotes) - I'm not sure - is that what we want?

hallyn:
I think it's all good.

PR #1021, opened 2016-05-11
Add apt-transport-https to minbase variant packages_template

odyssey4me:
In many environments the preference is to configure containers with
apt mirrors that are SSL-secured.

When building containers using the download template this can't be
done unless an insecure mirror is first used to install the
apt-transport-https package, then the sources reconfigured to
use the https URL.

When building containers without using the download template this
can't be done unless the container creator specifically includes
this package in the package list at build time.

It seems more intuitive to me to have the package installed by
default. This patch includes the required package for the minbase
variant only as this is the default.

### DIFF
--- a/templates/lxc-ubuntu.in
+++ b/templates/lxc-ubuntu.in
@@ -359,7 +359,7 @@ download_ubuntu()
         debootstrap_parameters="$debootstrap_parameters --variant=$variant"
     fi
     if [ "$variant" = 'minbase' ]; then
-        packages_template="${packages_template},sudo,ifupdown,isc-dhcp-client"
+        packages_template="${packages_template},sudo,ifupdown,isc-dhcp-client,apt-transport-https"
     fi
 
     echo "Installing packages in template: ${packages_template}"


odyssey4me:
@stgraber Wow, I didn't expect such a quick merge. What would the process be for getting this change into stable branches?

stgraber:
That changed qualifies for stable, I'll cherry-pick it when I do my next (very long) review of all pending commits.

odyssey4me:
Excellent, thanks @stgraber 👍

odyssey4me:
@stgraber Do the builds on https://images.linuxcontainers.org/ rely on the 2.x.x release tags? I'd like to know when this will be included in the images there. Is there any expected timeframe for the next 2.x.x release?

stgraber:
@odyssey4me the images are built from git master, so the change should already have made it to some of them

odyssey4me:
@stgraber On Ubuntu Trusty if I execute

  lxc-create --name=download1 --template=download -- --dist ubuntu --release trusty --arch amd64
  chroot /var/lib/lxc/download1/rootfs dpkg-query --list | grep apt

I'm not seeing the ``apt-transport-https`` package installed. Am I doing something wrong, or is the build process not using the minbase variant as I thought?

stgraber:
@odyssey4me just looked into this. The reason is simply because the download template images aren't using the minbase variant so aren't hitting your change.

odyssey4me:
@stgraber Oh dear, I seem to have misread the flow of the build then. :/ If it's not using minbase then I expect that it's not specifying a variant? If so, does it make better sense to include the ``apt-transport-https`` in the base ``packages_template`` in https://github.com/lxc/lxc/blob/master/templates/lxc-ubuntu.in#L345 ?

stgraber:
@odyssey4me yep

odyssey4me:
@stgraber Thanks for confirming - I've created https://github.com/lxc/lxc/pull/1040 to do exactly that.

Issue #392, opened 2014-12-30
python3-lxc on PyPi


run2fail:
Hi,

as of today the lxc Python module that is part of the official lxc source tree is not yet available via PyPi. Instead there are multiple alternative modules that try to do (nearly) the same, for example:
- [pylxc](https://github.com/cloud9ers/pylxc)
- [pylxclite](https://github.com/Unixlike/pylxclite)
- [lxc4u](https://github.com/ravenac95/lxc4u)

It would be great if the lxc project team would consider an upload to PyPi because...
- you cannot specify the official lxc module as dependency in `setup.py` or `requirements.txt` for automatic installation
- this makes bulding/testing on CI systems problematic as the lxc module cannot be installed via the normal processes (and a distribution specific package may not be available/installable)
- it would make the official module better known and avoid that developers "reinvent the wheel". I only learned about the lxc module's existence via [Stéphane's blog posts](https://www.stgraber.org/2012/09/28/introducing-the-python-lxc-api/)

Kind regards,

BB

stgraber:
Is maintaining python3-lxc (and possibly python2-lxc) on pypi something you'd be interested in?

Also, can pypi support multiple major versions in parallel? So we can have both 1.1 and 1.0 in there? (otherwise we should only publish 1.0).

run2fail:
It would certainly interest me to support the lxc project by bringing ``python*-lxc`` to pypi. Unfortunately, I currently lack experience because I was planning to bring my first Python application ([Locker](https://github.com/run2fail/locker) a fig-like frontend for lxc containers based on your excellent bindings) to pypi once all dependencies can be downloaded via ``pip``.

Take the following info with a grain of salt due to my lack of experience:
* You can provide different versions via pypi, i.e., the older versions remain available. This is sometimes even required, if a particular Python application specifies an explicit version dependency in ``setup.py``.
* Both versions (for Python 2 and 3) can be managed separately, if their name differs - at least that seems to be the case with a other packages like [protobu](https://pypi.python.org/pypi?%3Aaction=search&term=protobuf&submit=search).
* In general I prefer distribution provided packages as probably anyone can upload anything to pypi. Unfortunately, many available services on the web support only installation of Python dependencies via ``pip``. For example, travis-ci prevents the use of Python packages installed via Debian packages inside ``virtualenv`` environments.

tebanep:
[...]

PR #1031, opened 2016-05-27
python-lxc: enable standalone builds

amezin:
This PR makes it possible to build python-lxc bindings without the rest of LXC source and build trees.
Only public header and compiled library are required.
Builds driven by automake should still work as they did before.

This is the first step towards #392


src/python-lxc/lxc.c
```diff
@@ -77,6 +80,36 @@ static int lxc_wait_for_pid_status(pid_t pid)
     return status;
 }
 
+/* Copied from lxc/confile.c, with HAVE_SYS_PERSONALITY_H check removed */
+signed long lxc_config_parse_arch(const char *arch)
+{
+    struct per_name {
```
amezin: I've removed HAVE_SYS_PERSONALITY_H check here, and just copied necessary constants from sys/personality.h. Not sure if it's the right way, but at least it doesn't require custom configure step, which is rather hard to implement correctly in distutils.

brauner:
jenkins: test this please

brauner:
lgtm. @stgraber?

src/python-lxc/lxc.c
```diff
@@ -25,14 +25,91 @@
...
+#ifndef CLONE_NEWPID
+#  define CLONE_NEWPID            0x20000000
+#endif
+#ifndef CLONE_NEWNET
+#  define CLONE_NEWNET            0x40000000
+#endif
+
+/* From sys/personality.h */
+#define PER_LINUX 0x0000
+#define PER_LINUX32 0x0008
```
stgraber: Have you confirmed we actually need those? ^

I believe we have them as ifndef in the LXC code tree because of the Android C library not defining those, but that shouldn't be a problem on any system shipping python.

amezin: At least on Arch, sys/personality.h is a part of glibc package, and it is not shipped by any other libc implementation. So this will not compile with any other libc (not sure if it is actually a problem).
PER_LINUX* were added to glibc in 2002: https://sourceware.org/git/?p=glibc.git;a=commit;f=sysdeps/unix/sysv/linux/sys/personality.h;h=61ae62f27037d4c38acca2533a06dcd17fbc2ca2 so they should be present in all supported distros.

But for CLONE_* definitions everything is different:
$ grep -R 'CLONE_NEWCGROUP' /usr/include/
prints nothing. I don't have CLONE_NEWCGROUP defined in any system header, on a up-to-date Arch Linux installation. Usually it means these definitions aren't available in any released glibc version.

stgraber: Fair enough. I'm not sure how well python builds with other C libraries, but the comment about some CLONE flags not being in glibc yet is absolutely correct. CLONE_NEWCGROUP is very very new (as in 4.6 upstream kernel) so it makes sense it hasn't made it to glibc yet.

PR #1034, opened 2016-05-28
python-lxc: change distribution name _lxc -> lxc

amezin:
Distribution name starting with an underscore is considered invalid by
many tools. For example, you can't list such name in
install_requires in your setup.py.

error: Invalid distribution name or version syntax: -lxc-0.1

### DIFF
--- a/src/python-lxc/setup.py
+++ b/src/python-lxc/setup.py
@@ -63,7 +63,7 @@ class LxcBuildExtCommand(BuildExtCommand):
         super(LxcBuildExtCommand, self).build_extensions()
 
 
-setup(name='_lxc',
+setup(name='lxc',
       version='0.1',
       description='LXC',
       packages=['lxc'],


stgraber:
That particular change will break everything.

We have two modules:
 - _lxc is the C extension module
 - lxc is the python wrapper

That's the same arrangement as say _socket and socket in python core.

Renaming _lxc to lxc, will cause a module conflict which will mask one or the other and break everything.

stgraber:
You should never depend on _lxc, it's an internal dependency of the lxc module and is never meant for anyone to use directly.

amezin:
@stgraber 
It doesn't break anything.
name= passed to setup() is used in metadata only.
_lxc extension is still there.

stgraber:
Ok, should be fine then

stgraber:
Side effect of this change seems to be a broken testsuite: https://jenkins.linuxcontainers.org/view/All/job/lxc-github-commit-master/236/arch=amd64,compiler=gcc,restrict=lxc-priv/console

amezin:
@stgraber More likely it's caused by distutils->setuptools switch.

_lxc extension is built:

    running build_ext
    building '_lxc' extension
    creating /build/src/python-lxc/build/temp.linux-x86_64-3.4
    x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fPIC -I/build/src -I/usr/include/python3.4m -c lxc.c -o /build/src/python-lxc/build/temp.linux-x86_64-3.4/lxc.o
    x86_64-linux-gnu-gcc -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-Bsymbolic-functions -Wl,-z,relro -g -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 /build/src/python-lxc/build/temp.linux-x86_64-3.4/lxc.o -L/build/src/lxc -llxc -o /build/src/python-lxc/build/lib.linux-x86_64-3.4/_lxc.cpython-34m.so

I'll try to figure out what's wrong

amezin:
build 231 (this PR exactly) was successful https://jenkins.linuxcontainers.org/view/LXC/view/LXC%20builds/job/lxc-github-commit-master/231/
so it's definitely setuptools

PR #1035, opened 2016-05-30
python-lxc: use setuptools instead of distutils

amezin:
setuptools is recommended by Python Packaging Guide
https://python-packaging-user-guide.readthedocs.io/en/latest/current/

It contains some useful extensions like 'develop' command. Also it
is required for building wheels AFAIK.

The only downside is an extra build-time dependency. setuptools are
packaged in both Debian and Ubuntu (python3-setuptools), as well as
other major distros, so it shouldn't be an issue, I think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ compiler:
 before_install:
  - sudo add-apt-repository ppa:ubuntu-lxc/daily -y
  - sudo apt-get update -qq
- - sudo apt-get install -qq libapparmor-dev libcap-dev libseccomp-dev python3-dev docbook2x libgnutls-dev liblua5.2-dev libselinux1-dev libcgmanager-dev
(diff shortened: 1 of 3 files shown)


stgraber:
You could make it a conditional import which wouldn't break anyone.

try:
    import the new stuff
except ImportError:
    import the things we have today

amezin:
@stgraber I thought about it.

It can create two different configurations, with, possibly, different bugs. So I thought it's better to depend on setuptools unconditionally. I think, anybody who develops Python code has pip and setuptools installed.

If an extra dependency is an issue, however, I can add distutils fallback. But I'd prefer not doing it.

stgraber:
Shouldn't be a problem for new releases, this change however would become unsuitable for backporting to the stable branches.

amezin:
I'm not sure this change should be backported. It is a feature, not a fix.

rpm/deb installations won't benefit from it at all. And python-lxc from master branch builds successfully against stable library version.

If backport will be necessary, distutils fallback can be added as part of it. So that only stable branches will have the fallback.

stgraber:
jenkins: test this please

PR #1042, opened 2016-06-06
python-lxc: follow global rpath setting for autotools builds

amezin:
When LXC is configured with --enable-rpath, I expect Python bindings
to be able to find the library in a non-standard location, just like
LXC command-line tools.


brauner:
jenkins: test this please

PR #1052, opened 2016-06-18
lxc-debian: make sure init is installed

terceiro:
init 1.34 is not "Essential" anymore, in order to make it not required
on minimal chroots, docker containers, etc. Because of that we now need
to manually include it on systems that are expected to boot.


terceiro:
This patch has been included in the Debian release 1:2.0.1-2

PR #1066, opened 2016-06-30
You can specify additional parameters for lxc.script.up/down

vel21ripn:
This can be useful when working with openvswitch.

Config example:

```
lxc.network.type = veth
lxc.network.veth.pair = wfctl.v22
lxc.network.script.args = br-ovs0 vlan22
lxc.network.script.up = /etc/lxc/ovs
lxc.network.script.down = /etc/lxc/ovs
lxc.network.name = vlan0022

lxc.network.type = veth
lxc.network.veth.pair = wfctl.v23
lxc.network.script.args = br-ovs0 vlan23
lxc.network.script.up = /etc/lxc/ovs
lxc.network.script.down = /etc/lxc/ovs
lxc.network.name = vlan0023
```

Example script for setting openvswitch see in doc/example/ovs.sh

### DIFF
--- a/doc/examples/ovs.sh
+++ b/doc/examples/ovs.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+#
+#lxc.network.script.args = br10 vlan52
+# or
+#lxc.network.script.args = br10 trunk53
+
+BRIDGE="${6}"
+VLAN="${7}"
+
+case "$3" in
+up)
+	[ -n "$BRIDGE" ] || exit 1
+
+	ovs-vsctl --may-exist add-br $BRIDGE
+	ovs-vsctl --if-exists del-port $BRIDGE $5
+	vcfg=""
+	if [ -n "$VLAN" ]; then
+		if [ "${VLAN#vlan}" != "${VLAN}" ]; then
+			vcfg="tag=${VLAN#vlan}"
+		fi
+		if [ "${VLAN#trunk}" != "${VLAN}" ]; then
+			vcfg="trunks=${VLAN#trunk}"
+		fi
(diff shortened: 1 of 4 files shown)


stgraber:
Wouldn't it make more sense to allow for the arguments to be passed directly on the lxc.network.up and lxc.network.down lines? Similar to what's done for lxc.init_cmd?

vel21ripn:
> allow for the arguments to be passed directly on the lxc.network.up and lxc.network.down lines

"lxc.network.up = script arg1 arg2" equivalent to running "script arg1 arg2 utsname net up ifname ...".
It would contradict the documentation describing order of the standard arguments of the script.
In my case there is no need to duplicate the arguments for both scripts.
It would be useful to have "lxc.network.updown" if  using the same script for up/down the interface.

brauner:
Hm, I'm not so sure about this. The arguments specified in `lxc.network.script.args` would necessarily be the same for `lxc.network.script.{up,down}`. Aside from you use-case not requiring separate arguments, generally it seems to make more sense to be able to specify separate args that get passed to `lxc.network.script.{up,down}`. I think we'd like to avoid introducing additional keys for this. As @stgraber noted, it might be possibe to simply allow the arguments to be passed to `lxc.network.script.{up,down}` directly.

doc/examples/ovs.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
brauner: Fwiw, we try to ship POSIX-compliant shell scripts only.

stgraber:
Closing due to inactivity.

PR #1072, opened 2016-07-04
c/r: drop in-flight connections during CRIU dump

adrianreber:
Shortly after CRIU 2.3 has been released a patch has been added to skip
in-flight TCP connections. In-flight connections are not completely
established connections (SYN, SYN-ACK). Skipping in-flight TCP
connections means that the client has to re-initiate the connection
establishment.

This patch stores the CRIU version detected during version check, so
that during dump/checkpoint options can be dynamically enabled depending
on the available CRIU version.

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -50,6 +50,8 @@
 #define CRIU_GITID_VERSION	"2.0"
 #define CRIU_GITID_PATCHLEVEL	0
 
+#define CRIU_IN_FLIGHT_SUPPORT	"2.4"
+
 lxc_log_define(lxc_criu, lxc);
 
 struct criu_opts {
@@ -75,6 +77,9 @@ struct criu_opts {
 	 * different) on the target host. NULL if lxc.console = "none".
 	 */
 	char *console_name;
+
+	/* The detected version of criu */
+	char *criu_version;
 };
 
 static int load_tty_major_minor(char *directory, char *output, int len)
@@ -263,6 +268,10 @@ static void exec_criu(struct criu_opts *opts)
 		ret = snprintf(log, sizeof(log), "/sys/fs/cgroup/freezer/%s", freezer_relative);
 		if (ret < 0 || ret >= sizeof(log))
 			goto err;
+
+		if (!opts->user->disable_skip_in_flight &&
+				strcmp(opts->criu_version, CRIU_IN_FLIGHT_SUPPORT) >= 0)
+			DECLARE_ARG("--skip-in-flight");
 
 		DECLARE_ARG("--freeze-cgroup");
 		DECLARE_ARG(log);
@@ -511,11 +520,11 @@ version_error:
 
 /* Check and make sure the container has a configuration that we know CRIU can
  * dump. */
-static bool criu_ok(struct lxc_container *c)
+static bool criu_ok(struct lxc_container *c, char **criu_version)
 {
 	struct lxc_list *it;
 
-	if (!criu_version_ok(NULL))
+	if (!criu_version_ok(criu_version))
 		return false;
 
 	if (geteuid()) {
@@ -573,7 +582,7 @@ out_unlock:
 
 // do_restore never returns, the calling process is used as the
 // monitor process. do_restore calls exit() if it fails.
-void do_restore(struct lxc_container *c, int status_pipe, struct migrate_opts *opts)
+void do_restore(struct lxc_container *c, int status_pipe, struct migrate_opts *opts, char *criu_version)
 {
 	pid_t pid;
 	char pidfile[L_tmpnam];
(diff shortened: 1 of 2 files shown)


brauner:
jenkins: test this please

src/lxc/criu.c
```diff
@@ -256,6 +260,10 @@ static void exec_criu(struct criu_opts *opts)
 		if (ret < 0 || ret >= sizeof(log))
 			goto err;
 
+		if (criu_version >= CRIU_IN_FLIGHT_SUPPORT) {
+			DECLARE_ARG("--skip-in-flight");
```
adrianreber: I guess adding it to migrate_opts also means that it will be a command-line option to lxc-checkpoint, right?

tych0: On Tue, Jul 05, 2016 at 07:19:17AM -0700, Adrian Reber wrote:
> > @@ -256,6 +260,10 @@ static void exec_criu(struct criu_opts *opts)
> >  		if (ret < 0 || ret >= sizeof(log))
> >  			goto err;
> >  
> > +		if (criu_version >= CRIU_IN_FLIGHT_SUPPORT) {
> > +			DECLARE_ARG("--skip-in-flight");
> 
> I guess adding it to migrate_opts also means that it will be a command-line option to lxc-checkpoint, right?

If you like; we haven't added e.g. predump_dir, or the pageserver
addresses or preserves_inodes flags to lxc-checkpoint, mostly because
I think everyone is using liblxc via the API to migrate stuff, so
that's really the only place we need it.

src/lxc/criu.c
```diff
@@ -458,8 +466,12 @@ static bool criu_version_ok()
 		if (fgetc(f) != '\n')
 			goto version_error;
 
-		if (strcmp(version, CRIU_VERSION) >= 0)
+		if (strcmp(version, CRIU_VERSION) >= 0) {
+			float tmp = 0;
+			sscanf(version, "%f", &tmp);
+			criu_version = (int)(tmp*100);
```
tych0: On Tue, Jul 05, 2016 at 07:42:22AM -0700, Adrian Reber wrote:
> > @@ -458,8 +466,12 @@ static bool criu_version_ok()
> >  		if (fgetc(f) != '\n')
> >  			goto version_error;
> >  
> > -		if (strcmp(version, CRIU_VERSION) >= 0)
> > +		if (strcmp(version, CRIU_VERSION) >= 0) {
> > +			float tmp = 0;
> > +			sscanf(version, "%f", &tmp);
> > +			criu_version = (int)(tmp*100);
> 
> Setting `criu_version` only makes sense if the base CRIU version, which is defined with the string `CRIU_VERSION`. If the detected CRIU version is lower we will abort anyway.

Oh, sorry, I misread this as 'goto version_error', you're right that
it's only on version match.

> The float is needed to convert it to an integer value, which sounded easier to me than string comparisons.
> 
> Returning an integer instead of bool in the check function with the CRIU versions sounds like a good idea. It would then be part of criu_opts.
> 
> It is not clear how the conversion from the string with a dot (e.g., "2.3") to an integer should be done without using a float.

`sscanf("%d.%d", &major, &minor)` I think will work for this.

adrianreber:
Okay, let me rework the patch.

brauner:
@adrianreber, fwiw we merged https://github.com/lxc/lxc/pull/1073 which means you can now let `criu_version_ok()` allocate a version string for you which you can use to check against.

adrianreber:
@brauner, thanks for letting me know. I think, however, that the solution proposed by @tych0, returning the version as an integer, would provide a simpler interface. Especially for the case when using a CRIU version from a GIT checkout. It would also remove the necessity to handle memory allocation of the version string correctly and would make comparisons easier if it is an integer.

I have something ready locally which I need to test a bit more with different CRIU versions, but I am open to use your new string based version.

I am not really sure what is more fragile: converting the version string using `sscanf()` to a number or carrying the version string around which is then evaluated with `strcmp()`.

brauner:
I think @tych0 and I discussed this on irc and thought that a version string might make more sense because otherwise you'd have to deal with stuff like `2.10` and converting back and forth and what happens if they decide to append stuff like `2.10-2`. But I'm open to new solutions. :)

adrianreber:
Yes, without knowing how the CRIU versioning will look in the future, string comparison is probably the safest. I had a look at CRIU's RPC interface, but getting the version via RPC is probably overkill and also not supported. I will rework this PR to use your new version check.

adrianreber:
I updated this PR to optionally disable the skipping of in-flight connections via `struct migrate_opt` and now that criu-2.4 has been released I was able to test it without any workarounds.

I am not happy with passing around the criu_version pointer (and the pointer of the pointer).  I guess this could be implemented nicer, but I am not sure how. Using a static global variable would be simpler, but then again global. Let me know how this could be improved.

brauner:
Looks good to me, but @tych0 is more familiar with this. :)

tych0:
+1, LGTM too!

brauner:
jenkins: test this please

adrianreber:
Would it be possible to get this change also on stable-2.0 branch. This would need the following commits:

 * 5407e2a
 * b9ee664
 * ebb088e
 * f195450

These commits apply cleanly without any changes.

brauner:
Backporting is in @stgraber's ballpark. :)

stgraber:
CRIU changes usually get included in stable-2.0 whenever I refresh it

PR #1073, opened 2016-07-05
store criu version

brauner:
- If version != NULL criu_version_ok() stores the detected criu version in
  version. Allocates memory for version which must be freed by caller.
- If version == NULL criu_version_ok() will return true when the version
  matches, false in all other cases.

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -402,8 +402,11 @@ err:
  *
  * The intent is that when criu development slows down, we can drop this, but
  * for now we shouldn't attempt to c/r with versions that we know won't work.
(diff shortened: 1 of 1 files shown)


src/lxc/criu.c
```diff
@@ -452,16 +455,20 @@ static bool criu_version_ok()
 			return false;
 		}
 
-		if (fscanf(f, "Version: %1023[^\n]s", version) != 1)
+		tmp = malloc(1024);
+		if (!tmp)
+			return false;
```
hallyn: Leaking pipes[0]?

PR #1077, opened 2016-07-11
c/r: initialize migrate_opts properly

adrianreber:
The commit "c/r: add support for CRIU's --action-script" breaks
lxc-checkpoint on the command-line. It produces errors like:

 sh: $'\260\366\b\001': command not found

and then it fails. src/lxc/criu.c expects migrate_opts->action_script to
be either NULL, then it is ignored, or to actually contain the name of
an action scripts.

As the struct migrate_opts has not static storage is has to be explicitly
initialized or the value of the structure's members is indeterminate.


adrianreber:
@tych0 like this?

tych0:
On Mon, Jul 11, 2016 at 01:06:18PM -0700, Adrian Reber wrote:
> @tych0 like this?

Yes, exactly. I think this is the simplest thing to be defensive.
Thanks, and sorry for the breakage!

brauner:
jenkins: test this please

PR #1082, opened 2016-07-15
c/r: make local function static

adrianreber:
This is a minimal commit which makes the function 'do_restore()' static
as it is not used anywhere else in the code. This also removes a
trailing space my editor complained about.


brauner:
jenkins: test this please

Issue #1081, opened 2016-07-12
Migration from lxc-start-ephemeral to lxc-copy


beanieboi:
Hey,

As mentioned in #1080. I want to move from `lxc-start-ephemeral` to `lxc-copy`.
The problem with `lxc-copy` is that I wasn't able to reproduce `lxc-start-ephemeral` behavior.
We kind of implicitly rely on the feature that `lxc-start-ephemeral` uses a tmpfs to back the overlayfs and keeps everything in memory. I tried to move to lxc-copy but even with fast SSD's we saw a major performance hit. Kind of expected if you compare memory vs SSD but it was at a level that wasn't acceptable to us at this time and would need some major changes on our infrastructure to allow this kind of "slowness". Is there a way to reproduce the tmpfs other than managing the tmpfs on my own?
I'm not sure how complicated it is to add such a backing storage to `lxc-copy` or was this intentionally left out?

Thanks for your input
Best,
ben

brauner:
Should be closed via #1084.

PR #1084, opened 2016-07-16
lxc-copy: allow snapshots to be placed on tmpfs

brauner:
Place an ephemeral container started with -e flag on a tmpfs. Restrictions are
that you cannot request the data to be kept while placing the container on a
tmpfs, that either overlay or aufs backing storage must be used, and that the
storage backend of the original container must be a directory.

For ephemeral snapshots backed by overlay or aufs filesystems, a fresh tmpfs
is mounted over the containers directory if the user requests it. This should
be the easiest options. Anything else would require us to change the current
mount-layout of overlay and aufs snapshots. (A standard overlay or aufs
snapshot clone currently has the layout:

	/var/lib/lxc/CLONE_SNAPSHOT/delta0      <-- upperdir
	/var/lib/lxc/CLONE_SNAPSHOT/rootfs
	/var/lib/lxc/CLONE_SNAPSHOT/olwork
	/var/lib/lxc/CLONE_SNAPSHOT/olwork/work <-- workdir

with the lowerdir being

	/var/lib/lxc/CLONE_PARENT/rootfs

The fact that upperdir and workdir are not placed in a common subfolder under
the container directory has the consequence that we cannot simply mount a fresh
tmpfs under upperdir and workdir because overlay expects them to be on the same
filesystem.)

Because we mount a fresh tmpfs over the directory of the container the updated
/etc/hostname file created during the clone residing in the upperdir (currently
named "delta0" by default) will be hidden. Hence, if the user requests that the
old name is not to be kept for the clone, we recreate this file on the tmpfs.
This should be all that is required to restore the exact behaviour we would get
with a normal clone.
NOTE: If the container is rebooted all changes made to it are lost. This is not
easy to prevent since each reboot remounts the rootfs again.


brauner:
Should close #1081.

stgraber:
@hallyn can you take this one? I think after @brauner you're the one with the best understanding of our bdev/snapshot code :)

hallyn:
Patch itself looks fine.

Losing all changes on reboot seems like an unexpected behavior that'll cause a lot of heartburn.  Don't know that we can do better, but can we print a concise warning ("Unsafe to reboot") in the help output for the option?

brauner:
@hallyn, `lxc-start-ephemeral` itself couldn't allow for reboots but was affected by the same problem. The problem is that the rootfs is remounted on each reboot which also means that the tmpfs is mounted and remounted and all changes are lost. That seems difficult to avoid without changing how reboot works.

hallyn:
Agreed, I'm not asking you to fix it right now :)  (At some point
maybe we can brainstorm on ways to support it)

Just looking for ways we can warn the user.  Help text isn't ideal
as ppl may see it the first time they look for it only.  But maybe
it's better than nothing.  (Maybe not)

brauner:
Ah, that came out wrong. :) I totally agree about the warning I was just trying to give a little more insight into why I think this is currently happening.

brauner:
@hallyn, updated. Prints warning now.

src/lxc/lxc_copy.c
```diff
@@ -423,6 +451,10 @@ static int do_clone_ephemeral(struct lxc_container *c,
 	if (!my_args.quiet)
 		printf("Created %s as clone of %s\n", arg->newname, arg->name);
 
+	if (arg->tmpfs)
+		printf("Container is placed on tmpfs.\nRebooting will cause "
+		       "all changes made to it to be lost!");
```
brauner: Print's warning here.

Issue #1085, opened 2016-07-18
Please use lxc.haltsignal = SIGRTMIN+3 for systemd containers


mbiebl:
Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=831691

Hi,

he had an interesting discussion today on #pkg-systemd IRC about what
modifications are currently done for LXC containers to work properly
with systemd. Ideally we'd like to make a default systemd installation work
ootb.

Atm, the Debian template hooks up
/etc/systemd/system/sigpwr.target -> /lib/systemd/system/halt.target.

At some point, Martin added a workaround to the systemd package [1].

Systemd upstream is pretty clear though, that (mis)using SIGPWR to
initiate an ordered shutdown is not the preferred way under systemd [2].
If the container manager sends SIGRTMIN+3 to the init process following
the container spec [3], then systemd will shut down nicely.

I therefor suggest, that lxc should use haltsignal = SIGRTMIN+3 for
systemd based containers.

I would set that in the container config file during lxc-create if the
template is for a systemd-based distro.

Regards,
Michael

[1] https://anonscm.debian.org/cgit/pkg-systemd/systemd.git/commit/?id=44c3369f9
[2] https://lists.freedesktop.org/archives/systemd-devel/2016-July/037166.html
[3] https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/

brauner:
Should be closed by #1086. Please reopen, if you experience issues.

PR #1086, opened 2016-07-18
[RFC]: lxccontainer: detect if we should send SIGRTMIN+3

brauner:
This is required by systemd to cleanly shutdown. Other init systems should not
have SIGRTMIN+3 in the blocked signals set.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -1604,8 +1604,16 @@ static bool do_lxcapi_shutdown(struct lxc_container *c, int timeout)
 	pid = do_lxcapi_init_pid(c);
 	if (pid <= 0)
 		return true;
+
+	/* Detect whether we should send SIGRTMIN + 3 (e.g. systemd). */
+	if (task_blocking_signal(pid, (SIGRTMIN + 3)))
+		haltsignal = (SIGRTMIN + 3);
+
 	if (c->lxc_conf && c->lxc_conf->haltsignal)
 		haltsignal = c->lxc_conf->haltsignal;
+
+	INFO("Using signal number '%d' as halt signal.", haltsignal);
+
 	kill(pid, haltsignal);
 	retv = do_lxcapi_wait(c, "STOPPED", timeout);
 	return retv;
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -1838,3 +1838,41 @@ int lxc_strmunmap(void *addr, size_t length)
 {
 	return munmap(addr, length + 1);
 }
+
+/* Check whether a signal is blocked by a process. */
+bool task_blocking_signal(pid_t pid, int signal)
+{
+	bool bret = false;
+	char *line = NULL;
+	long unsigned int sigblk = 0;
+	size_t n = 0;
+	int ret;
+	FILE *f;
+
+	/* The largest integer that can fit into long int is 2^64. This is a
+	 * 20-digit number. */
+	size_t len = /* /proc */ 5 + /* /pid-to-str */ 21 + /* /status */ 7 + /* \0 */ 1;
+	char status[len];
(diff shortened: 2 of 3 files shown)


brauner:
There's an RFC about this: #1085.

brauner:
I think this was the original thread that lead to the adoption of `SIGPWR` as default shutdown signal. This [RFC] is basically trying to retain the default behavior.
Please note, that we want to avoid having to make any hard assumptions about what init system is running inside the container or in general, what is running inside the container. So detecting whether systemd is running or not would be the dispreferred solution here. Suggestions welcome!

evgeni:
@mbiebl @martinpitt what are your thoughts?

basically this is what you suggested, but without explicitly setting it in the config (which I disliked), win win? :)

mbiebl:
Shouldn't the explicitly configured haltsignal from the conftainer config take precedence?

Assuming systemd is used in the majority of cases (which I think is fair, given that all major distros have switched), don't you think making SIGRTMIN+3 the new default and falling to SIGPWR in case the signal is blocked or the proc file can't be read would be safer?

brauner:
@mbiebl, systemd seems to be the only init system I found that uses `SIGRTMIN+3` as its shutdown signal so this is easier to detect than the other way around.

src/lxc/lxccontainer.c
```diff
@@ -1604,8 +1605,19 @@ static bool do_lxcapi_shutdown(struct lxc_container *c, int timeout)
...
+
+	/* Detect whether we should send SIGRTMIN + 3 (e.g. systemd). */
+	if (haltsignal == SIGPWR)
+		if (check_sigblk(pid, (SIGRTMIN + 3)))
+			haltsignal = (SIGRTMIN + 3);
+
 	if (c->lxc_conf && c->lxc_conf->haltsignal)
 		haltsignal = c->lxc_conf->haltsignal;
+
+	sigstring = strsignal(haltsignal);
```
hallyn: strsignal is new to me, but according to the manpage it is not threadsafe.

src/lxc/utils.c
```diff
@@ -1838,3 +1838,39 @@ int lxc_strmunmap(void *addr, size_t length)
...
+/* Check whether we need to send SIGRTMIN+3 to init. (e.g. used by systemd to
+ * cleanly shutdown container.
+ */
+bool check_sigblk(pid_t pid, int signal)
+{
+	bool bret = false;
+	char *line = NULL;
+	long unsigned int sigblk = 0;
+	size_t n = 0;
+	char status[MAXPATHLEN];
```
hallyn: Maybe I'm being silly, but that's kind of big.

src/lxc/utils.h
```diff
@@ -296,4 +296,9 @@ int open_devnull(void);
 int set_stdfds(int fd);
 int null_stdfds(void);
 int lxc_count_file_lines(const char *fn);
+
```
hallyn: Both here and above the real function, you describe how we use it (checking for SIGRTMIN+3), but the fn itself is generic.  Better to leave these comments just saying "check for signal", as it may grow other users.

hallyn:
Thanks, it's a cool idea.

src/lxc/utils.h
hallyn: In fact do you mind renaming it task_blocking_signal(pid, signal)?

brauner:
Changelog 2016-07-19:
- rename check_sigblk() to task_blocking_signal()
- shorten comment
- use smaller buffer
- remove strsignal

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
(network failure)

brauner:
jenkins: test this please

PR #1092, opened 2016-07-20
simple enhancements of lxc-debian-template-script

LicoMonch:
Nothing special in these commits (imho). Just a few changes explained by the commit-messages.

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -244,7 +244,8 @@ configure_debian_systemd()
 
     # Fix getty-static-service as debootstrap does not install dbus
     if [ -e "$rootfs//lib/systemd/system/getty-static.service" ] ; then
-        local tty_services=$(for i in $(seq 2 "$num_tty"); do echo -n "getty@tty${i}.service "; done; )
+        local tty_services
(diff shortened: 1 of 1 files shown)


brauner:
Could you please prefix your commits with "templates:" e.g. "templates: avoid noisy perl warnings caused by missing locales" etc. These are a lot of small changes. Did you test that the template still works?

LicoMonch:
The commit messages were changed as requested.

The template still works (tested with 2.0.3-0ubuntu1~ubuntu16.04.1) in any combination of parameters set.
The greatest danger is the quoting of variables here and I focused on path-names and some minor other vars.
I also checked that variables like "$packages" are passed without quotes if needed (eg ```apt-get install "$packages"``` would not work,, so I didn't modify that)

brauner:
jenkins: test this please

LicoMonch:
I'm not familiar with jenkins, but to me the failed lxc-test-usernic relates to ubuntu-containers and not to debian.

Locally this test also passes here (lxc-tests 2.0.3-0ubuntu1~ubuntu16.04.1)

PR #1097, opened 2016-07-23
lxc-alpine: mount tmpfs under /run

jirutka:
When running under grsecurity kernel or userns, it can't be mounted from inside.

### DIFF
--- a/config/templates/alpine.common.conf.in
+++ b/config/templates/alpine.common.conf.in
@@ -18,3 +18,6 @@ lxc.cap.drop = sys_resource
 lxc.cap.drop = sys_tty_config
 lxc.cap.drop = syslog
 lxc.cap.drop = wake_alarm
+
+# Mount tmpfs under /run.
+lxc.mount.entry=run run tmpfs rw,nodev,relatime,mode=755 0 0


brauner:
I'm not sure about this. Running a userns alpine container creates and mounts `/run` just fine for me. I don't know about grsecurity and whether we should accommodate this in our templates or leave that for the user to configure in the container configuration. /cc @stgraber

jirutka:
You can find the same setting also in [gentoo.moresecure.conf](https://github.com/lxc/lxc/blob/5429ec248a1090edde6285e4bc729641f6f79b57/config/templates/gentoo.moresecure.conf.in#L13).

brauner:
Thanks, good to know! Then we're good to go.

brauner:
Why this specific size for the tmpfs? Wouldn't it be better to just leave it unspecified as in the gentoo example?

jirutka:
To be honest, I don’t know. When I look at output of `mount` on my Alpine system, there’s the size specified.

brauner:
Just leave it empty as in the gentoo example. Per default the maximum allowed size of a tmpfs will usually be half your ram size.

jirutka:
Okay, I’ve updated the PR.

PR #1101, opened 2016-07-26
Add a prefix to the lxc.pc

sergiusens:
This allows installing to different locations and using
the lxc.pc to build using the generated includedir and
libdir.

Signed-off-by: Sergio Schvezov <sergio.schvezov@ubuntu.com>


brauner:
jenkins: test this please

PR #1102, opened 2016-07-27
setns instead of unshare in lxc-attach

chenhaiq:
The process created by lxc-attach has incorrect cgroup ns. It fails lxc-checkpoint as reported in [criu issue](https://github.com/xemul/criu/issues/174).
A real example for this issue, which is used to update ip address:
  1. start lxc instance with debian image
  2. lxc-attach -n <lxc name> -- service networking restart
  3. lxc-checkpoint -n <lxc name> -D <dump dir>

lxc-checkpoint will fail because dhclient process has incorrect cgroup ns. It needs to use setns instead of unshare to set cgroup ns. "setns" already supports cgroup ns in kernel of ubuntu 16.04.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -217,10 +217,10 @@ static int lxc_attach_to_ns(pid_t pid, int which)
 	 * the file for user namepsaces in /proc/$pid/ns will be called
 	 * 'user' once the kernel supports it
 	 */
-	static char *ns[] = { "user", "mnt", "pid", "uts", "ipc", "net" };
+	static char *ns[] = { "user", "mnt", "pid", "uts", "ipc", "net", "cgroup" };
 	static int flags[] = {
 		CLONE_NEWUSER, CLONE_NEWNS, CLONE_NEWPID, CLONE_NEWUTS, CLONE_NEWIPC,
-		CLONE_NEWNET
+		CLONE_NEWNET, CLONE_NEWCGROUP
 	};
 	static const int size = sizeof(ns) / sizeof(char *);
 	int fd[size];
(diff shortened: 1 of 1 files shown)


src/lxc/attach.c
```diff
@@ -932,7 +931,7 @@ int lxc_attach(const char* name, const char* lxcpath, lxc_attach_exec_t exec_fun
 	}
 
 	if (options->attach_flags & LXC_ATTACH_MOVE_TO_CGROUP && cgns_supported())
-		unshare_cgns = true;
+		options->namespaces = options->namespaces | CLONE_NEWCGROUP;
```
brauner: Could you please rewrite this to:
`options->namespaces |= CLONE_NEWCGROUP;`

src/lxc/attach.c
```diff
@@ -932,7 +931,7 @@ int lxc_attach(const char* name, const char* lxcpath, lxc_attach_exec_t exec_fun
 	}
 
 	if (options->attach_flags & LXC_ATTACH_MOVE_TO_CGROUP && cgns_supported())
```
brauner: This is just style preference but I'd like this to have extra brackets so that the intentions are clear: 
`if ((options->attach_flags & LXC_ATTACH_MOVE_TO_CGROUP) && cgns_supported())`

brauner:
Oh, and please name your commit with `attach: setns instead of unshare in lxc-attach`. Thanks! :)

src/lxc/attach.c
chenhaiq: applied

src/lxc/attach.c
chenhaiq: applied

stgraber:
jenkins: test this please

PR #1105, opened 2016-07-28
Add syslog support

bapt:
Add a syslog logger activable via the configuration file.
While here add the VM name in the log lines to help debugging

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -38,6 +38,7 @@
 #include <net/if.h>
 #include <time.h>
 #include <dirent.h>
+#include <syslog.h>
 
 #include "parse.h"
 #include "config.h"
@@ -105,6 +106,7 @@ static int config_haltsignal(const char *, const char *, struct lxc_conf *);
 static int config_rebootsignal(const char *, const char *, struct lxc_conf *);
 static int config_stopsignal(const char *, const char *, struct lxc_conf *);
 static int config_start(const char *, const char *, struct lxc_conf *);
+static int config_syslog(const char *, const char *, struct lxc_conf *);
 static int config_monitor(const char *, const char *, struct lxc_conf *);
 static int config_group(const char *, const char *, struct lxc_conf *);
 static int config_environment(const char *, const char *, struct lxc_conf *);
@@ -184,6 +186,7 @@ static struct lxc_config_t config[] = {
 	{ "lxc.init_uid",             config_init_uid             },
 	{ "lxc.init_gid",             config_init_gid             },
 	{ "lxc.ephemeral",            config_ephemeral            },
+	{ "lxc.syslog",               config_syslog               },
 };
 
 struct signame {
@@ -268,6 +271,23 @@ static const struct signame signames[] = {
 #ifdef SIGSYS
 	{ SIGSYS,    "SYS" },
 #endif
+};
+
+struct syslog_facility {
+	const char *name;
+	int facility;
+};
+
+static const struct syslog_facility syslog_facilities[] = {
+	{ "daemon",	LOG_DAEMON },
+	{ "local0",	LOG_LOCAL0 },
+	{ "local1",	LOG_LOCAL1 },
+	{ "local2",	LOG_LOCAL2 },
+	{ "local3",	LOG_LOCAL3 },
+	{ "local4",	LOG_LOCAL4 },
+	{ "local5",	LOG_LOCAL5 },
+	{ "local6",	LOG_LOCAL6 },
+	{ "local7",	LOG_LOCAL7 },
 };
 
 static const size_t config_size = sizeof(config)/sizeof(struct lxc_config_t);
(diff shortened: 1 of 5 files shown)


brauner:
I want to test this first. :)

hallyn:
lgtm, but yeah this stuff has proven easy to break, so heavy testing is recommended.

brauner:
- There currently doesn't seem to be any difference in specifying one of ` {daemon,local[0-7]}` for `lxc.syslog` in the containers config file. Right?

- I'd really appreciate it if you'd elaborate a little more on the chosen design: My impressions is that you establish a mapping between `LXC` log flags and `syslog` log flags. So the `LXC` log output is only seen in the `syslog`  when two conditions are met:
	1. `lxc.syslog = {daemon, local{0-7}}` is set
	2. `-l/--logpriority` is specified when starting the container

- It would be super-nice if you could simultaneously add documentation for `lxc.syslog` to https://github.com/lxc/lxc/blob/master/doc/lxc.container.conf.sgml.in to explain how this works.

src/lxc/log.c
```diff
@@ -43,23 +44,78 @@
...
+static int log_append_syslog(const struct lxc_log_appender *appender,
+			     struct lxc_log_event *event)
+{
+	char *msg;
+	int rc;
+
+	if (!syslog_enable)
+		return 0;
+
+	rc = vasprintf(&msg, event->fmt, *event->vap);
```
brauner: Please avoid using any GNU extensions. We try to be POSIX-compatible and although we've used non-GNU extensions before we try to avoid adding new ones. (We also build on Android so one should also try to keep in mind that even some standard POSIX features might be not available. Though I guess, `vasprintf()` is.  :)

bapt: I used *asprintf because it is present in all the libc I'm aware of, but if you feel like it is an issue I can replace it with a malloc+vsnprintf

bapt:
>     There currently doesn't seem to be any difference in specifying one of {daemon,local[0-7]} for lxc.syslog in the containers config file. Right?
It makes a difference on the facility used when logging, to allow one to filter the messages properly in the syslog implementation used, there are multiple facility predefined, I decided to only go for daemon or the various local, as the first is where daemons are supposed to used for logging and the local[0-7] are reserved one for local use (on production at gandi we use the one of the local facility for example)
> 
>     I'd really appreciate it if you'd elaborate a little more on the chosen design: My impressions is that you establish a mapping between LXC log flags and syslog log flags. So the LXC log output is only seen in the syslog when two conditions are met:
>         lxc.syslog = {daemon, local{0-7}} is set
>         -l/--logpriority is specified when starting the container

Sure I'm mapping the lxc log priority to the syslog ones they pretty much matches.
so yes to be able to use syslog one should define in the configuration file:
lxc.syslog = "thefacility"

Then the log level mechanism is working exactly the same way as the current logging mechanism so you play with -l decide which level/priority you want

>     It would be super-nice if you could simultaneously add documentation for lxc.syslog to https://github.com/lxc/lxc/blob/master/doc/lxc.container.conf.sgml.in to explain how this works.
> 
I will look into it

bapt:
btw asprintf is used in many places, which is another reason why I chose to use it

brauner:
On Fri, Jul 29, 2016 at 07:13:01AM -0700, Baptiste Daroussin wrote:
> btw asprintf is used in many places, which is another reason why I chose to use it

Yeah, I know, that's why I said we try to avoid adding new GNUisms. If it's not
too much trouble I'd like to ask you to replace it. :)

src/lxc/confile.c
```diff
@@ -270,6 +273,23 @@ static const struct signame signames[] = {
 #endif
 };
 
+struct facilityname {
```
brauner: I think I'd prefer if this was named `syslog_facility` so that it's nice and easy to spot what this is for. :)

brauner:
If you don't mind could you squash the first two commits into one aka merge `Add a new logger: syslog
` and `Print the VM name in the logs` so that we only have two commits: the implementation and the documentation.

brauner:
Thanks for this! And +1 one for adding the documentation. :)

bapt:
All request change done (if I haven't missed one :))

src/lxc/log.c
```diff
@@ -43,23 +46,85 @@
...
+		return 0;
+
+	va_copy(args, *event->vap);
+	len = vsnprintf(NULL, 0, event->fmt, args) + 1;
+	va_end(args);
+	msg = malloc(len * sizeof(char));
+	if (msg == NULL)
+		return 0;
+	rc = vsnprintf(msg, len, event->fmt, *event->vap);
+	if (rc == -1) {
```
brauner: `if (rc < 0 || rc >= len)`
The final lament from my side. :)

bapt:
Done I was also missing a free() due to the change from vasprintf to malloc + vsnprintf, which is now fixed

brauner:
jenkins: test this please

bapt:
tar: var/log/apt/history.log: Cannot write: No space left on device

That is not my fail :)

brauner:
Yeah. :) We'll wait for it to fail and restart once it is fixed. I'm running a local test here again after all the updates. But from my side lgtm, once jenkins is fine we should be good to go.

brauner:
jenkins: test this please

brauner:
Thanks!

PR #1109, opened 2016-07-31
partially reorganize repo

brauner:
This is part of an ongoing effort to restructure parts of our repo to improve orientation for new contributers.


brauner:
Nope. :) If that causes too much trouble we can abandon this. I just though at some point it might be a good idea to reorganize because it's a mystery for most people to know what's going on in this repo. :)

stgraber:
So since that only applies to the commands and it's reasonably rare that we change those in stable releases, it may not be too bad. Basically what happens with your change is that we won't be able to cleanly cherry-pick any change to any of the files that got renamed, those will need to manually be applied.

stgraber:
The cgroup bits will be the most annoying in stable I suspect, but those are pretty small, self-contained changes usually so I don't mind having to re-apply those by hand too much.

PR #1114, opened 2016-08-04
lxc-attach: add -f option & cleanup

Blub:
Attaching to a container with a configuration file in a non standard path used to work. We used it for template building. With the pty allocation changes this stopped because that part of the code uses the configuration struct, and has a new check which then errors with `container XY is not defined`. This patch copies over the `-f` switch from `lxc-start`. Alternatively it might be possible to restore support for the old config-less variant, but it seems to make more sense to just provide the same config path as was used with `lxc-start`.


brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

stgraber:
jenkins: test this please

PR #1116, opened 2016-08-07
templates: add support for new arch on Alpine Linux

clandmeter:
We are adding support for both armv7 and aarch64 architectures to Alpine Linux. We do not have official repositories yet but they will come with next Alpine release.


stgraber:
What's the difference between armv7 and armhf in alpine?

Debian and Ubuntu's definition of armhf is armv7 hard-float. So I'm guess that the reason for you to define a separate armv7 arch is that your existing armhf is actually armv6 hard-float (compatible with 1st gen raspberry pi for example).

clandmeter:
Yes you are right, most probably the only use case for armhf will be raspberry gen1.

cc @fabled

PR #1118, opened 2016-08-09
tree-wide: replace readdir_r() with readdir()

### LINKED ISSUE #1117: LXC does not compile with glibc 2.24
LXC compiles with -Werror and uses readdir_r().

readdir_r() has been deprecated in glibc 2.24 and when you compile code that uses readdir_r() a warning about this is generated.  This warning is converted into an error with -Werror in LXC and thus LXC does not compile with glibc 2.24.

brauner:
To clarify the motivation behind this PR: `glibc 2.24` has deprecated `readdir_r()` (https://lwn.net/Articles/696469/) and some users already report that they can't compile the `LXC` sources because of that. So we should replace `readdir_r()` with `readdir()`. The problem however is that POSIX currently doesn't require that `readdir()` be thread-safe but it seems all of the relevant `libc`s actually provide a version of `readdir()` that is thread-safe. At least when multiple threads read from different `DIR` streams.

closes #1117.

### DIFF
--- a/src/lxc/bdev/lxcloop.c
+++ b/src/lxc/bdev/lxcloop.c
@@ -269,7 +269,7 @@ static int do_loop_create(const char *path, uint64_t size, const char *fstype)
 
 static int find_free_loopdev_no_control(int *retfd, char *namep)
 {
-	struct dirent dirent, *direntp;
+	struct dirent *direntp;
 	struct loop_info64 lo;
 	DIR *dir;
 	int fd = -1;
@@ -279,7 +279,7 @@ static int find_free_loopdev_no_control(int *retfd, char *namep)
 		SYSERROR("Error opening /dev");
 		return -1;
 	}
-	while (!readdir_r(dir, &dirent, &direntp)) {
+	while ((direntp = readdir(dir))) {
 
 		if (!direntp)
 			break;
--- a/src/lxc/cgroups/cgfs.c
+++ b/src/lxc/cgroups/cgfs.c
@@ -157,7 +157,7 @@ static struct cgroup_ops cgfs_ops;
 
 static int cgroup_rmdir(char *dirname)
 {
-	struct dirent dirent, *direntp;
+	struct dirent *direntp;
 	int saved_errno = 0;
 	DIR *dir;
 	int ret, failed=0;
@@ -169,7 +169,7 @@ static int cgroup_rmdir(char *dirname)
 		return -1;
 	}
 
-	while (!readdir_r(dir, &dirent, &direntp)) {
+	while ((direntp = readdir(dir))) {
 		struct stat mystat;
 		int rc;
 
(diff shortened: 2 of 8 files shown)


brauner:
jenkins: test this please

brauner:
jenkins: test this please

hallyn:
No.

readdir is not thread-safe.  assuming that glibc has made readdir thread-safe
and therefore deprecated readdir_r, then we have to work harder to be safe
backward-compatible.

brauner:
Yup.

brauner:
POSIX seems to require that multiple calls to `readdir()` on different directory streams do not overwrite the returned buffer:

> The returned pointer, and pointers within the structure, might be invalidated or the structure or the storage areas might be overwritten by a subsequent call to readdir() on the same directory stream. They shall not be affected by a call to readdir() on a different directory stream. (http://pubs.opengroup.org/onlinepubs/9699919799/)

This only seems to happen when multiple threads read from the same directory stream. If this cannot be the case we could probably just switch to `readdir()`.

mbachry:
Here's what `readdir_r` man page says under Fedora 24:

       This function is deprecated; use readdir(3) instead.

       The  readdir_r()  function was invented as a reentrant version of readdir(3).  It reads the next directory entry from the directory stream dirp, and returns it in the caller-allo‐
       cated buffer pointed to by entry.  For details of the dirent structure, see readir(3).

       A pointer to the returned buffer is placed in *result; if the end of the directory stream was encountered, then NULL is instead returned in *result.

       It is recommended that applications use readdir() instead of readdir_r().  Furthermore, since version 2.24, glibc deprecates readdir_r().  The reasons are as follows:

       *  On systems where NAME_MAX is undefined, calling readdir_r() may be unsafe because the interface does not allow the caller to specify the length  of  the  buffer  used  for  the
          returned directory entry.

       *  On  some  systems, readdir_r() can't read directory entries with very long names.  When the glibc implementation encounters such a name, readdir_r() fails with the error ENAME‐
          TOOLONG after the final directory entry has been read.  On some other systems, readdir_r() may return a success status, but the returned d_name field may not be null terminated
          or may be truncated.

       *  In  the  current  POSIX.1  specification (POSIX.1-2008), readdir(3) is not required to be thread-safe.  However, in modern implementations (including the glibc implementation),
          concurrent calls to readdir(3) that specify different directory streams are thread-safe.  Therefore, the use of readdir_r() is generally unnecessary in multithreaded  programs.
          In  cases where multiple threads must read from the same directory stream, using readdir(3) with external synchronization is still preferable to the use of readdir_r(), for the
          reasons given in the points above.

       *  It is expected that a future version of POSIX.1 will make readdir_r() obsolete, and require that readdir() be thread-safe when  concurrently  employed  on  different  directory
          streams.

brauner:
C libraries that ship a thread-safe version of `readdir()` (at least when they read from different `DIR` streams):
  - bionic (at least ~2012 likely older): https://github.com/android/platform_bionic/blob/bca25411b882abc6ab0bf07d25266ba9f8e4d398/libc/bionic/dirent.cpp
  - glibc (since ~1996): https://fossies.org/dox/glibc-2.24/sysdeps_2posix_2readdir_8c_source.html

(Looking at others.)

brauner:
CONTINUED: C libraries that ship a thread-safe version of `readdir()` (at least when they read from different `DIR` streams):
  - musl (since it was created in 2011): http://git.musl-libc.org/cgit/musl/tree/src/dirent/readdir.c

brauner:
I honestly don't know what to make of `dietlibc`s `readdir()` and `readdir_r()` implementations.

hallyn:
Ok, thanks.  We never re-use the same DIR* in multiple threads so this should be fine.

PR #1134, opened 2016-08-16
bdev: use correct overlay module name

brauner:
- Assume that the module name is "overlay" per default and not "overlayfs".
- Assume that the overlay version we are using requires a workdir.
- When we mount an overlay filesystem and we fail with ENODEV retry once with
  the module name we haven't already used.

### DIFF
--- a/src/lxc/bdev/lxcoverlay.c
+++ b/src/lxc/bdev/lxcoverlay.c
@@ -39,6 +39,7 @@
 lxc_log_define(lxcoverlay, lxc);
 
 static char *ovl_name;
+static char *ovl_version[] = {"overlay", "overlayfs"};
 
 /* defined in lxccontainer.c: needs to become common helper */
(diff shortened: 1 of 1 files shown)


hallyn:
This stuff is fragile (in the kernel);  please make sure it works on your current distro, on Ubuntu 14.04 (with backports) and on latest Ubuntu.  If it does, I see no problems, please go ahead and merge.  Thanks!

stgraber:
The reason for this change is to fix a breakage in Ubuntu as the overlayfs module is being removed from the new kernel.

stgraber:
So that was tested by Jenkins on a 4.4 kernel, which happens to match Ubuntu 16.04 and 16.10, as well as the backports kernel for Ubuntu 14.04.

The only other supported kernel we care about as far as Ubuntu is concerned is the 3.13 kernel.

I'll let this in now, wait for PPA builds and then test on a 14.04 system.

PR #1136, opened 2016-08-17
Rename physical devices earlier during shutdown [WIP

Blub:
The second commit probably requires some additional auditing. I've gone through the involved git commit history and am not entirely convinced the current code is necessary.
I've com across #649 while going through the history and saw that the call added there already happens during a regular shutdown anyway a few lines below (and already did back then).
We've had some reports of interface deletions not always working apparently due to the fact that netlink is "not a reliable protocol", and said patch looks to me like a simple retry attempt reducing the probability of failures but not completely ruling them out.
So I'm wondering if maybe we should add a retry loop to the delete netlink events?

And finally: Should we also rename interfaces to temporary names before moving them to the container? Eg. if you want to assign eth0 to a container as eth1, and start with a veth named eth0 inside the container, then moving eth0 will fail as the veth is already named eth0 in the container's namespace.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2885,9 +2885,6 @@ void lxc_delete_network(struct lxc_handler *handler)
 		netdev = iterator->elem;
 
 		if (netdev->ifindex != 0 && netdev->type == LXC_NET_PHYS) {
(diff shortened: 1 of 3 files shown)


Blub:
Opened updated #1269

PR #1145, opened 2016-08-20
add PLD Linux support to upstream repo

glensc:
from pld-linux repo:
https://github.com/pld-linux/lxc/commits/auto/th/lxc-2.0.4-1/lxc-pld.in.sh
https://github.com/pld-linux/lxc/commits/auto/th/lxc-2.0.4-1/lxc-pld.patch

they are mature enough to upstreamed (not much changes recently to those files) and it's always better if these are present in upstream as well so upstream could make common changes to all distro scripts.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,6 +57,7 @@ if test "z$with_distro" = "z"; then
 	AC_CHECK_FILE(/etc/mandriva-release,with_distro="openmandriva")
 	AC_CHECK_FILE(/etc/pardus-release,with_distro="pardus")
 	AC_CHECK_FILE(/etc/altlinux-release,with_distro="altlinux")
+	AC_CHECK_FILE(/etc/pld-release,with_distro="pld")
 fi
 with_distro=`echo ${with_distro} | tr '[[:upper:]]' '[[:lower:]]'`
 
@@ -72,7 +73,7 @@ case $with_distro in
 		distroconf=default.conf.lxcbr
 		distrosysconf="$sysconfdir/default"
 		;;
-	redhat|centos|fedora|oracle|oracleserver|sparclinux|altlinux|suse|opensuse*|plamo)
+	redhat|centos|fedora|oracle|oracleserver|sparclinux|altlinux|suse|opensuse*|plamo|pld)
 		distroconf=default.conf.lxcbr
 		distrosysconf="$sysconfdir/sysconfig"
 		;;
@@ -107,6 +108,9 @@ case "$with_init_script" in
 				;;
 			ubuntu)
 				init_script=upstart,systemd
+				;;
+			pld)
+				init_script=sysvinit,upstart,systemd
 				;;
 			*)
 				echo -n "Linux distribution init system unknown."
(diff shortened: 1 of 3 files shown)


brauner:
Thanks for the patch. Could you please ensure that the template outputs the following error message when people try to create an unprivileged container:

```
This template can't be used for unprivileged containers.
You may want to try the "download" template instead.
```

This is the behavior all other template scripts currently have. For example, trying to create an unprivileged `Fedora` container without the download template outputs:

```
chb@conventiont|~
> lxc-create -n fed -t fedora
This template can't be used for unprivileged containers.
You may want to try the "download" template instead.
lxc-create: fed: lxccontainer.c: create_run_template: 1290 container creation template for fed failed
lxc-create: fed: tools/lxc_create.c: main: 318 Error creating container fed
```

brauner:
Oh, and please prefix your commit with `templates:`:

```
templates: add PLD Linux
```

Thank you. :)

brauner:
Thanks for adding the prefix to your commit. However, the template is still not aligned with the behavior all other templates show when trying to create an unprivileged container. As I've requested above, could you please make sure that the template outputs:

```
This template can't be used for unprivileged containers.
You may want to try the "download" template instead.
```

when users try to create an unprivileged container with it. Essentially all you need to do to achieve this behavior is to put this:

```
# Detect use under userns (unsupported)
for arg in "$@"; do
    [ "$arg" = "--" ] && break
    if [ "$arg" = "--mapped-uid" -o "$arg" = "--mapped-gid" ]; then
        echo "This template can't be used for unprivileged containers." 1>&2
        echo "You may want to try the \"download\" template instead." 1>&2
        exit 1
    fi
done
```

right after your license header in `lxc-pld.in`. After that has been done we're good to merge.

glensc:
yes. in my work queue :)

brauner:
Ah, cool. Thanks! :)

glensc:
https://github.com/lxc/lxc/pull/1145#issuecomment-241202501
i tried on my system and i do not get such error with command given on lxc 2.0.4: http://sprunge.us/hFAC

so i'll just add the changes you asked

brauner:
> https://github.com/lxc/lxc/pull/1145#issuecomment-241202501
i tried on my system and i do not get such error with command given on lxc 2.0.4: http://sprunge.us/hFAC

That's because you're running `lxc-create` as root as evident from the file you attached: `root[load: 0.64]@haproxy`. If you run this command as a normal user you should see the error.

> so i'll just add the changes you asked

Thanks but unfortunately you forgot to sign-off your commit. This is required for us to include the patch (see https://github.com/lxc/lxc/blob/master/CONTRIBUTING). And please either squash the two commits into one single commit or prefix your second commit with `templates:` as you did with the first. Thanks! :)

glensc:
i assumed sign off needed for first commit only. anyway squashed.

brauner:
jenkins: test this please

brauner:
Thanks again!

glensc:
i wonder why this is not appearing in 2.0.5 release?

PR #1146, opened 2016-08-20
use fd 9 instead of 200

glensc:
mksh can't redirect fd 200, use 9
this PR is to catch up mksh changes from 17abf27


brauner:
jenkins: test this please

brauner:
Could you please prefix your commit with `templates:`. :)

PR #1151, opened 2016-08-26
tools: use boolean for ret in lxc_device.c

brauner:

As spotted by @hallyn.


hallyn:
Thanks, looks good.

PR #1162, opened 2016-08-31
templates: rm halt.target -> sigpwr.target symlink

brauner:
Given commit 330ae3d350e060e5702a0e5ef5d0faeeeea8df6e:

    lxccontainer: detect if we should send SIGRTMIN+3

    This is required by systemd to cleanly shutdown. Other init systems should not
    have SIGRTMIN+3 in the blocked signals set.

we should stop symlinking halt.target to sigpwr.target for systemd.


brauner:
Can you please specify what version of lxc you're using and the commands you used to create the container? Is the problem reliably reproducable? Most importantly, please try to reproduce and append the logs you got from doing `lxc-stop -n c -l DEBUG -O c.log`?

PR #1163, opened 2016-09-01
templates: remove creation of bogus directory in Debian templates

mabes:
An incorrect quoting introduced in bf39edb caused a /{lib,etc} folder to
appear in Debian templates

The very next line :
    mkdir -p "${rootfs}/etc/systemd/system/getty.target.wants

makes creating ${rootfs}/etc/systemd/system/ unnecessary in the first
place


brauner:
Thanks!

PR #1166, opened 2016-09-03
implement PR_SET_NO_NEW_PRIVS in liblxc

brauner:
start: set PR_SET_NO_NEW_PRIVS when requested
- Set no_new_privs after setting the lsm label. If we do set it before we aren't
allowed to change the label anymore.

attach: use PR_SET_NO_NEW_PRIVS
- When we detect that the container, we want to attach to, has been stared with
  PR_SET_NO_NEW_PRIVS we attach with PR_SET_NO_NEW_PRIVS as well. (We might
  relax this restriction later but let's be strict for now.)
- When LXC_ATTACH_NO_NEW_PRIVS is set in the flags passed to
  lxc_attach()/attach_child_main() then we set PR_SET_NO_NEW_PRIVS irrespective
  of whether the container was started with PR_SET_NO_NEW_PRIVS or not.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -593,6 +593,10 @@ AM_CONDITIONAL([IS_BIONIC], [test "x$is_bionic" = "xyes"])
 # Some systems lack PR_CAPBSET_DROP definition => HAVE_DECL_PR_CAPBSET_DROP
 AC_CHECK_DECLS([PR_CAPBSET_DROP], [], [], [#include <sys/prctl.h>])
 
+# Some systems lack PR_{G,S}ET_NO_NEW_PRIVS definition => HAVE_DECL_PR_{G,S}ET_NO_NEW_PRIVS
+AC_CHECK_DECLS([PR_SET_NO_NEW_PRIVS], [], [], [#include <sys/prctl.h>])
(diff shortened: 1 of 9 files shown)


brauner:
Not sure why it fails on Jenkins currently; tests run fine locally.

brauner:
TODO: 
- `PR_SET_NO_NEW_PRIVS` should probably be applied before `seccomp`.

brauner:
- `PR_SET_NO_NEW_PRIVS` should also be called before applying `selinux` and `apparmor`.

brauner:
Changelog 2016-09-03:
- `start.c`: Set no_new_privs before setting the lsm label. If we do set it before we aren't
allowed to change the label anymore and the container will fail to start.
- `attach.c`: Set no_new_privs before lsm and seccomp. We probably don't want attach() to
  be able to change the lsm or seccomp policy if the container was started with
  PR_SET_NO_NEW_PRIVS enabled.

brauner:
I wonder if it's ok to set no new privs after we changed into the profile we intended. If it is not than no new privs is basically useless as soon as apparmor is enabled.

stgraber:
Should be fine, so long as apparmor does something reasonable when you turn on namespacing and stacking. So letting you load and switch to profiles within a namespace as those won't ever give you additional privileges.

brauner:
ping @hallyn. :)

PR #1167, opened 2016-09-03
console: use correct log name

brauner:
lxc_console is used with lxc_console.c


brauner:
jenkins: test this please

PR #1168, opened 2016-09-04
configure: add --disable-werror

jerome-pouiller:
-Werror may break builds on some scenarios with trivialities
(especially during developments).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -152,6 +152,11 @@ if test "x$with_systemdsystemunitdir" != "xno"; then
   AC_SUBST([SYSTEMD_UNIT_DIR], [$with_systemdsystemunitdir])
 fi
 
+AC_ARG_ENABLE([werror],
+	[AC_HELP_STRING([--disable-werror],
+	[do not treat warnings as errors])],
+	[], [enable_werror=yes])
+
 # Allow enabling deprecated executables
 AC_ARG_ENABLE([deprecated],
 	[AC_HELP_STRING([--enable-deprecated],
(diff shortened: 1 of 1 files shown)


brauner:
Not a fan of `--disable-werror` but since you make `Werror` the default I can live with it. I let @stgraber call it. :)

stgraber:
Not a huge fan either, I'd much prefer people just fix whatever causes the build failure since we won't be accepting any code which requires --disable-werror anyway.

brauner:
Thank you for the patch. Given that we prefer to do everything to ensure that only code makes it here that passes `-Werror` I think we won't include this specific change.

tpetazzoni:
Including ```-Werror``` is definitely good for developers, but doesn't necessarily work well for users. Indeed, depending on the gcc version, different warnings are raised, some of them not necessarily being that important (especially when using ```-Wall``` on very recent gcc versions). This causes all sort of build breakage that are not useful.

As an example, OpenEmbedded has to explicitly disable some warnings (http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/tree/recipes-containers/lxc/lxc_2.0.0.bb?h=master#n50) and ditto in Buildroot (which is where the patch from @jerome-pouiller is coming from).

So please, please, provide a way to disable -Werror, or we will have to keep a non-upstream patch forever in Buildroot. Thanks!

yann-morin-1998:
If I may chime in...

What causes grievance with `-Werror` is when a new gcc version comes out and one builds an older lxc **release**: any new gcc version will add new warnings which did not previously exist, and then the old lxc **release** becomes unbuildable in this case.

There is in this case no point in wanting to fix the warnings in the relase: it is a relase, not a development snapshot.

Yes, development versions (git tree, snapshots) should enable -Werror by default. But a release should not, or at least offer the option to disable it. See these [in-depth explanations](https://blog.flameeyes.eu/2009/02/future-proof-your-code-dont-use-werror) (a bit oldish now, granted).

stgraber:
jenkins: test this please

yann-morin-1998:
@stgraber Thanks for reconsidering! :-)

I have no idea why the testsuite is failing: it can't checkout a sha1, and indeed it does not exist in the tree. Not sure where it gets that sha1 from... :-/

brauner:
jenkins: test this please

PR #1170, opened 2016-09-06
debian: Make systemd honor SIGPWR

fanyeren:
without this symbolic（sigpwr） link， systemd-shutdownd hung in stopping container, so i think it's not a good idea to remove it!

Signed-off-by: Xiahou Feng <xiahoufeng@yahoo.com>

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -253,12 +253,12 @@ configure_debian_systemd()
 
     # This function has been copied and adapted from lxc-fedora
     rm -f "${rootfs}/etc/systemd/system/default.target"
+    rm -f "${rootfs}/etc/systemd/system/sigpwr.target
     chroot "${rootfs}" ln -s /dev/null /etc/systemd/system/udev.service
     chroot "${rootfs}" ln -s /dev/null /etc/systemd/system/systemd-udevd.service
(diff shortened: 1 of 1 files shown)


brauner:
I'm copying and pasting this verbatim from your comment on #1162:

This shouldn't happen anymore as we are taking care to send the correct shutdown signal to `systemd` now. I'ts very like that we're not going to this before we had a chance to properly debug this: 
- Can you please specify what version of `LXC` you're using and the commands you used to create the container?
- Is the problem reliably reproducable?
- Most importantly, please create a `Debian` container which runs `systemd` as its init system. Then start it and stop it with `lxc-stop -n c -l DEBUG -O c.log` and append `c.log` as a file here.

fanyeren:
@brauner 

LXC version is 1.1.5 with some critical patches, and container is Debian Jessie with systemd

brauner:
Ah, ok. So `LXC 1.1.5` neither has the PR https://github.com/lxc/lxc/pull/1162 you're referencing nor the required PR https://github.com/lxc/lxc/pull/1086. So any issue you might experience with containers hanging is very likely not related to these two patches. That means the symlinks you're trying to re-introduce in master are actually present in your version of `LXC`.

brauner:
Closing this because this doesn't seem to be an issue affecting current git master. :)

fanyeren:
@brauner  thank you, i'll check the patches

PR #1171, opened 2016-09-06
improve detect_ramfs_rootfs() and add test

brauner:


brauner:
ping @hallyn. :)

brauner:
ping @hallyn, @stgraber.

hallyn:
Looks good, thanks.  (Only comment would be I see no point in returning to the initial namespace at the end of the testcase :)

brauner:
> Looks good, thanks. (Only comment would be I see no point in returning to the initial namespace at the end of the testcase :)

In case other tests we might add later require a non-overmounted `/proc/self/mountinfo`. :)

PR #1172, opened 2016-09-06
issue: 1164 fixed broken cron link in alpine image

melato:
tested with LXD for alpine versions 3.4, 3.2, 3.1, edge.  Did not test with plain LXC.
To test:
1) Verify that the /etc/runlevels/default/ links are not broken.
2) exec "ash" in the container, and run "crontab -e" and add the line "* * * * * /bin/date > /root/date.log"
Make sure there is an empty line at the end.
Verify that the file /root/date.log is created in the next minute.

### DIFF
--- a/templates/lxc-alpine.in
+++ b/templates/lxc-alpine.in
@@ -364,8 +364,13 @@ setup_services() {
 	done
 
 	# default runlevel
-	for svc_name in networking cron; do
-		ln -s /etc/init.d/$svc_name etc/runlevels/default/$svc_name
+	for svc_name in networking cron crond; do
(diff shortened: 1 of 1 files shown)


brauner:
Hi, you need to sign-off your commit. Otherwise we are not able to include your commit. :) This is as simple as doing `git commit -s ...`. If your unsure about the format just take a look at one of the commits included.
Also, can you please squash your commits into a single commit and rename it to: 

```
templates: use correct cron version in alpine template
```

Thank you. :)

brauner:
Since you sent an updated branch in #1173 I'll close this one.

PR #1178, opened 2016-09-10
exit with error if no controlling tty is available

elpollodiablo:
Avoids a segfault in lxc_attach.c: get_pty_on_host when executed with su: https://bugs.launchpad.net/ubuntu/+source/lxc/+bug/1567037

### DIFF
--- a/src/lxc/console.c
+++ b/src/lxc/console.c
@@ -527,6 +527,11 @@ int lxc_console_create(struct lxc_conf *conf)
 
 	lxc_console_peer_default(console);
 
+	if (console->peer < 0) {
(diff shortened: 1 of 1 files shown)


elpollodiablo:
jenkins: test this please

brauner:
Does this still happen with `LXC 2.0.4`?

elpollodiablo:
Yes, this happens in both, 2.0.4 (taken from debian backports) and master. Provided you have a setup with unprivileged containers, you should be able to test it with

su -c "lxc-attach -n unprivileged_uts -- echo Hello World" unprivileged_user

Issue #1183, opened 2016-09-12
lxc-snapshot errors with zfs backing


chris-sanders:
lxc-snapshot doesn't properly create or delete zfs backed snapshots.

Trying two different ways to create a snapshot provides an error but does create and mount snapshots.
```
# lxc-create -t ubuntu -n zfsbug -B zfs
# lxc-ls
zfsbug 

# zfs list
NAME                                                                                USED  AVAIL  REFER  MOUNTPOINT
rpool/lxc/zfsbug                                                                    160M  69.4G   160M  /var/lib/lxc/zfsbug/rootfs

# lxc-snapshot --version
2.0.4

# lxc-snapshot -n zfsbug 
could not find any snapshots to destroy; check snapshot names.

# lxc-snapshot -n zfsbug -c bugtest
could not find any snapshots to destroy; check snapshot names.

# zfs list
NAME                                                                                USED  AVAIL  REFER  MOUNTPOINT
rpool/lxc/snap0                                                                    37.5K  69.4G   160M /var/lib/lxc/zfsbug/snaps/snap0/rootfs
rpool/lxc/snap1                                                                    37.5K  69.4G   160M  /var/lib/lxc/zfsbug/snaps/snap1/rootfs
rpool/lxc/zfsbug                                                                    160M  69.4G   160M  /var/lib/lxc/zfsbug/rootfs

# zfs list -t snapshot
NAME                                                                                         USED  AVAIL  REFER  MOUNTPOINT
rpool/lxc/zfsbug@snap0                                                                          0      -   160M  -
rpool/lxc/zfsbug@snap1                                                                          0      -   160M  -
```
Next we'll try and delete the image, which will error due to the snapshots
```
lxc-destroy -n zfsbug
Destroying zfsbug failed: zfsbug has snapshots.
```

Try to remove the snapshots
```
# lxc-snapshot -L -n zfsbug
snap1 (/var/lib/lxc/zfsbug/snaps) 2016:09:12 12:40:19
snap0 (/var/lib/lxc/zfsbug/snaps) 2016:09:12 12:40:08

# lxc-snapshot -n zfsbug -d snap0
# lxc-snapshot -L -n zfsbug
snap1 (/var/lib/lxc/zfsbug/snaps) 2016:09:12 12:40:19
# lxc-snapshot -n zfsbug -d snap1
# lxc-snapshot -L -n zfsbug
No snapshots

# lxc-destroy -n zfsbug
[...]

PR #1184, opened 2016-09-13
lxczfs: small fixes

brauner:
- We expect destroy to fail in zfs_clone() so try to silence it so users are
  not irritated when they create zfs snapshots.
- Add -r recursive to zfs_destroy(). This code is only hit when a) the
  container has no snapshots or b) the user calls destroy with snapshots. So
  this should be safe. Without -r snapshots will remain.


brauner:
Closes #1183.

Issue #1156, opened 2016-08-27
checkpoint failed for debain image 


chenhaiq:
This is another error from #1150 . I have lxc using master code, which fixed error from restoring trusty instance, but debain still fails.

To reproduce:
```
lxc-create -t download -n d1 -- -r wheezy -d debian -a amd64
# edit config to support checkpoint
lxc-start -n d1
lxc-checkpoint -n d1 -v -D /tmp/dump6 -s
lxc-checkpoint -n d1 -v -D /tmp/dump6 -r -o /tmp/dump6/lxc.log -l TRACE
```

log files:
https://pek3a.qingstor.com/oc-storage/dump.log
https://pek3a.qingstor.com/oc-storage/lxc.log
https://pek3a.qingstor.com/oc-storage/restore.log

chenhaiq:
```
lxc.include = /usr/share/lxc/config/debian.common.conf
lxc.arch = x86_64

# Container specific configuration
lxc.rootfs = /var/lib/lxc/d1/rootfs
lxc.rootfs.backend = dir
lxc.utsname = d1

# Network configuration
lxc.network.type = veth
lxc.network.link = lxcbr0
lxc.network.flags = up
lxc.network.hwaddr = 00:16:3e:7f:90:dd


lxc.console = none
lxc.tty = 0
lxc.cgroup.devices.deny = c 5:1 rwm
lxc.mount.auto = cgroup:mixed
lxc.autodev = 1
lxc.kmsg = 0
lxc.include = /usr/share/lxc/config/common.conf.d/00-lxcfs.conf
```
LXC is master code. The same env with #1150

chenhaiq:
debain/jessie also fail in restore, but has different log:
https://pek3a.qingstor.com/oc-storage/dump_8027.log
https://pek3a.qingstor.com/oc-storage/lxc_dump_0827.log
https://pek3a.qingstor.com/oc-storage/lxc_restore_0827.log
https://pek3a.qingstor.com/oc-storage/restore_0827.log

brauner:
This seems to be fixed with above PRs. Please reopen if the issue still persists with newer versions of `LXC`.

PR #1194, opened 2016-09-16
Cgroup root on dump

tych0:
This is a fix for one of the issues mentioned in #1156; see the 4th patch in the series for the bulk of the fix.  The CRIU commit in question went into criu-dev today, and should be in 2.7.

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -140,7 +140,7 @@ static void exec_criu(struct criu_opts *opts)
 
 	/* The command line always looks like:
 	 * criu $(action) --tcp-established --file-locks --link-remap \
-	 * --manage-cgroups=full action-script foo.sh -D $(directory) \
+	 * --manage-cgroups=full --action-script foo.sh -D $(directory) \
 	 * -o $(directory)/$(action).log --ext-mount-map auto
 	 * --enable-external-sharing --enable-external-masters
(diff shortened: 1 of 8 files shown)


src/lxc/criu.c
```diff
@@ -191,6 +191,8 @@ static void exec_criu(struct criu_opts *opts)
 		return;
 	}
 
+	static_args += 2 * cgroup_num_hierarchies();
```
brauner: Stupid question, but why `2 * `? :)

stgraber: Because for each he's passing --cgroup-root <root>

tych0: We pass them on the cli as `--cgroup-root /lxc/foo`, so we need two array elements.

brauner: Ah right, I thought this was only passed once.

hallyn: I don't see where you're checking for cgroup_num_hierarchie() == -1.  I do see that inside the while loop you'll try get_hierarchies() and get false, except you won't get there since the loop won't ever run since it goes from i = 0; i < -1.

tych0: Indeed, whoops. I've just fixed that, thanks!

hallyn:
Thanks - looks good to me, will wait on the testsuite.

PR #1201, opened 2016-09-21
tools: do not add trailing spaces on lxc-ls -1

### LINKED ISSUE #1200: [2.0.4] "lxc-ls --active -1" adding trailing spaces
Hi,

Say you have 2 containers named "bazcontainer" (12 chars) and "foobar" (6 chars) and you try to determine in a script if "foobar" is running, then the following command will never match/work : 
```
lxc-ls --active -1 | grep -E "^foobar$"
```
Because "lxc-ls" is columning the output, printf %s style with a fixed width : The largest container name. In this case "bazcontainer". The real output is in fact : 
```
lxc-ls --active -1 | sed -r -e 's/^(.+)$/|\1|/g'
|bazcontainer|
|foobar      |
```

This is not desirable of course :-)

brauner:

Closes #1200.

### DIFF
--- a/src/lxc/tools/lxc_ls.c
+++ b/src/lxc/tools/lxc_ls.c
@@ -144,7 +144,7 @@ static void ls_print_fancy_format(struct ls *l, struct lengths *lht,
  * Only print names of containers.
  */
 static void ls_print_names(struct ls *l, struct lengths *lht,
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

PR #1206, opened 2016-09-25
Fix null derefence if attach is called without access to any tty

cwoac:
Discovered while trying to invoke lxc through salt stack. If ```lxc_console_peer_default``` fails to allocate the peer (```src/lxc/console.c +432```) then``` ts``` is unset and will segfault on close. 

Signed-off-by: Oliver Matthews <oliver@codersoffortune.net>


brauner:
jenkins: test this please

PR #1211, opened 2016-09-26
utils: lxc_deslashify() prevent fallthrough

brauner:
lxc_string_join() should not be reached when !*path, otherwise we'll risk a
segfault.


brauner:
jenkins:

PR #1224, opened 2016-10-08
mark the python examples as having utf-8 encoding

evgeni:
this allows running them also under Python2, which otherwise
would choke on Stéphane's name and error out with
 SyntaxError: Non-ASCII character '\xc3' in file …


brauner:
jenkins: test this please

PR #1225, opened 2016-10-11
log: sanity check the returned value from snprintf()

jiazhang0:
The returned value from snprintf() should be checked carefully.

This bug can be leveraged to execute arbitrary code through carefully
constructing the payload, e.g,

lxc-freeze -n `python -c "print 'AAAAAAAA' + 'B'*959"` -P PADPAD -o /tmp/log

This command running on Ubuntu 14.04 (x86-64) can cause a segment fault.

Signed-off-by: Lans Zhang <jia.zhang@windriver.com>

### DIFF
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -170,10 +170,13 @@ static int log_append_logfile(const struct lxc_log_appender *appender,
 		     event->locinfo->file, event->locinfo->func,
 		     event->locinfo->line);
 
-	n += vsnprintf(buffer + n, sizeof(buffer) - n, event->fmt,
-		       *event->vap);
+	if (n < 0)
+		return n;
 
-	if (n >= sizeof(buffer) - 1) {
+	if (n < sizeof(buffer) - 1)
+		n += vsnprintf(buffer + n, sizeof(buffer) - n, event->fmt,
+			       *event->vap);
+	else {
 		WARN("truncated next event from %d to %zd bytes", n,
 		     sizeof(buffer));
 		n = sizeof(buffer) - 1;


brauner:
jenkins: test this please

src/lxc/log.c
```diff
@@ -170,10 +170,13 @@ static int log_append_logfile(const struct lxc_log_appender *appender,
 		     event->locinfo->file, event->locinfo->func,
 		     event->locinfo->line);
 
-	n += vsnprintf(buffer + n, sizeof(buffer) - n, event->fmt,
-		       *event->vap);
+	if (n < 0)
+		return n;
 
-	if (n >= sizeof(buffer) - 1) {
+	if (n < sizeof(buffer) - 1)
```
brauner: It maybe smarter to only insert the check for `n < 0` but then still call `vsnprintf()` unconditionally as it is done now and let it truncate the string. This way we provide a little more information but also warn the user that happened.

jiazhang0: That doesn't work. Actually this patch resolves the condition "n >= sizeof(buffer) - 1". If this condition is met, vsnprintf() should not be called at all. In this situation, the first parameter of vsnprintf() will point to the address ">= buffer + sizeof(buffer) - 1", and the second parameter will >=1, allowing vsnprintf() to overwrite any address beyond the address of buffer if n is big enough. This is a typical security vulnerability, allowing carefully constructing an attack payload to overwrite any contents, e.g, the returned address of current function log_append_logfile().

Fortunately, the known applications leveraging log_append_logfile() seem to be not related to escalation of privileges (e.g, set-uid-root), so this issue is not a real CVE bug.

jiazhang0: Note that the synopsis of snprintf():
int snprintf(char *str, size_t size, const char *format, ...);

If n is bigger than sizeof(buffer), sizeof(buffer) - n is still positive. This is why we cannot call vsnprintf() unconditionally.

PR #1234, opened 2016-10-14
tools: better error reporting for lxc-start

brauner:

Closes https://bugs.launchpad.net/ubuntu/+source/lxc/+bug/1302053.

### DIFF
--- a/src/lxc/tools/lxc_start.c
+++ b/src/lxc/tools/lxc_start.c
@@ -86,7 +86,7 @@ static int ensure_path(char **confpath, const char *path)
 			goto err;
 		}
 	}
-	err = 0;
+	err = EXIT_SUCCESS;
 
 err:
 	free(fullpath);
@@ -231,6 +231,12 @@ int main(int argc, char *argv[])
 		exit(err);
 	lxc_log_options_no_override();
 
+	if (access(my_args.lxcpath[0], O_RDWR) < 0) {
+		if (!my_args.quiet)
+			fprintf(stderr, "You lack access to %s\n", my_args.lxcpath[0]);
+		exit(err);
+	}
+
 	const char *lxcpath = my_args.lxcpath[0];
 
 	/*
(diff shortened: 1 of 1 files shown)


src/lxc/tools/lxc_start.c
```diff
@@ -231,6 +231,14 @@ int main(int argc, char *argv[])
 		exit(err);
 	lxc_log_options_no_override();
 
+	if (geteuid()) {
```
hallyn: Note that if running lxc-start nested inside an unprivileged container, geteuid() may not help.

src/lxc/tools/lxc_start.c
```diff
@@ -275,9 +283,19 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	if (!c->may_control(c)) {
```
hallyn: The ->may_control should probably come after ->is_defined.

src/lxc/tools/lxc_start.c
brauner: It's in all other tools and it shouldn't get in the way I expect.

hallyn: Maybe, but since the goal is better error reporting, is there any point in hiding this check behind uid != 0?  Is there any case where the access check fails when getuid returns 0, where we would *not* want to give a clear error message and exit now?

brauner:
Addressed.

hallyn:
Hm, note I did not mean for that other check to be removed - i simply wanted it also done if getuid() == 0.

brauner:
Hm, I'm slow today.

brauner:
@hallyn, added the `getuid()` check.

hallyn:
Hi,

what I've been trying to say is that the access() check might fail even if you are root, in which case container start will still fail later on, so we may as well stop early in that case as well.  Do you see any problem with github.com/hallyn/lxc #2016-10-18/br2 ?  (github is trying to tell me how to do something cool to link to the branch here but I'm failing)

brauner:
Nope, all good.

hallyn:
Cool, please update and I'll hit merge.  Sorry this took so long :)

brauner:
I updated. Sorry I was so dense. :D

PR #1241, opened 2016-10-18
lxc-alpine: do not drop setfcap

jirutka:
@brauner I’m not entirely sure about this, could you please answer me two questions?

1. Am I right that `setfcap` is not a security risk, i.e. it cannot be abused to escape from container to the host system?
2. Is it true that `setcap` cannot work in unprivileged containers (i.e. containers that uses user namespace) anyway, only in privileged ones?

### DIFF
--- a/config/templates/alpine.common.conf.in
+++ b/config/templates/alpine.common.conf.in
@@ -8,7 +8,6 @@ lxc.devttydir =
 lxc.cap.drop = audit_write
 lxc.cap.drop = ipc_owner
 lxc.cap.drop = mknod
-lxc.cap.drop = setfcap
 lxc.cap.drop = setpcap
 lxc.cap.drop = sys_nice
 lxc.cap.drop = sys_pacct


brauner:
Regarding 1.: I don't see a problem with not dropping `setfcap`. A lot of other templates also do not drop it. If you wanted you could add a `alpine.moresecure.conf` similar to `gento.moresecure.conf` and drop `setfcap` there. Users who wanted additional security could then use the `alpine.moresecure.conf` config. That's extra effort though.

Regarding 2.: Correct, `setcap` will not work in user namespaces. @hallyn, will at some point however render the preceding statement false. (A little temporal logic on the side here. :))

jirutka:
@brauner Thanks, so I guess you can merge it now. :)

I’m interested in getting `setcap` work in unprivileged container, so I’ll appreciate more info about it.

hallyn:
Not dropping setfcap allows a privileged user in the non-user-namespaced container who also has a nonroot account on the host to get full privileges on the host.

brauner:
Yeah, that's out of the question, but we're essentially not dropping it in other templates, including ubuntu so we have a case for merging. :)

hallyn:
Yes, sorry, I was mainly answering "Am I right that setfcap is not a security risk" as a separate sentence.  For clarity, in regular lxc usage we do *not* expect that a untrusted user be given the ability (through filecaps or setuid-root) to create and start privileged containers.  And we don't allow unprivileged users to even descend into privileged containers' rootdirs.  So for this template it's ok.  Sorry to sound alarmist above :)

jirutka:
@hallyn thanks a lot for your comment.

I’m in doubt now. I know that giving untrusted users access to root in privileged containers is not a good idea, but still… Maybe it would be better to move `lxc.cap.drop = setfcap` from `alpine.common.conf` to the template for container-specific config in the lxc-alpine template script, the same as `lxc.cap.drop = sys_admin` ([here](https://github.com/lxc/lxc/blob/514120e227384f42130f318ab974275b475dffad/templates/lxc-alpine.in#L392-L398)), so it can be easily disabled when needed…

LXC doesn’t allow to override `lxc.cap.drop` from included config in the container’s config, does it? So if user wants to keep some capability that has been dropped in `alpine.common.conf`, the only sane option is to copy `alpine.common.conf` somewhere else, modify it and include that into the container’s config (or just copy it to the container’s config directly), right?

brauner:
Yes. What you can do is not drop a cap in common.conf and but then drop it in the containers config. The other options is the aforementioned `*.moresecure.conf` template.

hallyn:
FWIW you can put

lxc.cap.drop =

in your container config file to completely clear it out and start over.  You can't just drop one capability from the set though.

ncopa:
I think it would be better to use whitelist. Drop all capabilities by default, enable the minimal set for alpine.common.conf that are strictly needed, and then add `lxc.cap.keep = ...` for every additional needed capability.

PR #1242, opened 2016-10-20
seccomp: check libseccomp version for s390x

brauner:
If libseccomp version is < 2.3 we do not have appropriate s390x support.

### DIFF
--- a/src/lxc/seccomp.c
+++ b/src/lxc/seccomp.c
@@ -33,6 +33,12 @@
 #include "log.h"
 #include "lxcseccomp.h"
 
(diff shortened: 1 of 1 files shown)


stgraber:
Turns out libseccomp in Xenial has a s390x patch which adds those defines. So it's perfectly correct for us to check for the defines, what we need to figure out is what's missing in that patched libseccomp and add whatever's missing.

brauner:
Is someone looking into this or should I?

stgraber:
@brauner I'm off today so feel free to take this :)

PR #1243, opened 2016-10-20
s390x: Fix seccomp handling of personalities

stgraber:
There are no personalities for s390x, so don't list itself as one.


stgraber:
@brauner that should do the trick. I'm just confused as to how this ever worked for you.

The error we were getting "File exists" makes sense though as we were effectively applying the same policy twice on s390x

PR #1245, opened 2016-10-21
Setup libtool

stgraber:


stgraber:
jenkins: test this please

stgraber:
Looks like the python part needs a bit more tweaking.

stgraber:
Not quite there yet, lua looks unhappy still...

brauner:
Ah, having fun with `libtool`. Nothing more exciting...

stgraber:
Yeah... that's needed to build on Ubuntu 17.04 unfortunately so I'm converting lxc over to it which is a lot more complicated than it should be (python3 and lua are a bit problematic there...).

stgraber:
Current issue I'm fighting is related to the library version in the elf header, so fixing that with -version-info now, hopefully that'll fix things for good.

PR #1248, opened 2016-10-24
Use external

tych0:
FWIW, the --external flag is new as of 2.6, so I don't think any of this should be SRU'd into xenial until CRIU 2.6 is.

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -443,8 +443,11 @@ static void exec_criu(struct criu_opts *opts)
 				if (strlen(n->name) >= sizeof(eth))
 					goto err;
 				strncpy(eth, n->name, sizeof(eth));
-			} else
-				sprintf(eth, "eth%d", netnr);
+			} else {
+				ret = snprintf(eth, sizeof(eth), "eth%d", netnr);
+				if (ret < 0 || ret >= sizeof(eth))
+					goto err;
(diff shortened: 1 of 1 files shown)


src/lxc/criu.c
```diff
@@ -439,27 +439,53 @@ static void exec_criu(struct criu_opts *opts)
...
+				if (ret < 0 || ret >= sizeof(buf))
+					goto err;
+				break;
+			case LXC_NET_MACVLAN:
+				if (n->name) {
+					if (strlen(n->name) >= sizeof(eth))
+						goto err;
+					strncpy(eth, n->name, sizeof(eth));
+				} else
+					sprintf(eth, "eth%d", netnr);
```
brauner: Should we be using `snprintf()` here as well?

stgraber:
Ignore the jenkins failure, will fix in a minute.

stgraber:
jenkins: test this please

src/lxc/criu.c
tych0: Probably so :)

brauner: :)

tych0: Actually we can just drop this hunk entirely, as it's done above (also without a snprintf... I'll push patches for both).

PR #1255, opened 2016-10-27
remove veth device from host

### LINKED ISSUE #1253: Destroy veth device on stop
This relates to https://github.com/lxc/lxd/issues/2439

To avoid the race condition where the kernel takes too long to expire the network namespace and possibly keeps it alive indefinitely, we should destroy the host side of the veth pair on exit.

brauner:
Closes #1253.
Closes https://github.com/lxc/lxd/issues/2439.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -220,7 +220,6 @@ static void lxc_proc_put_context_info(struct lxc_proc_context_info *ctx)
 
 static int lxc_attach_to_ns(pid_t pid, int which)
 {
-	char path[MAXPATHLEN];
 	/* according to <http://article.gmane.org/gmane.linux.kernel.containers.lxc.devel/1429>,
 	 * the file for user namespaces in /proc/$pid/ns will be called
 	 * 'user' once the kernel supports it
@@ -235,8 +234,7 @@ static int lxc_attach_to_ns(pid_t pid, int which)
 	int i, j, saved_errno;
 
 
-	snprintf(path, MAXPATHLEN, "/proc/%d/ns", pid);
-	if (access(path, X_OK)) {
+	if (access("/proc/self/ns", X_OK)) {
 		ERROR("Does this kernel version support 'attach' ?");
 		return -1;
 	}
@@ -261,7 +259,7 @@ static int lxc_attach_to_ns(pid_t pid, int which)
 				close(fd[j]);
 
 			errno = saved_errno;
-			SYSERROR("failed to open '%s'", path);
+			SYSERROR("failed to open namespace: '%s'.", ns[i]);
 			return -1;
 		}
 	}
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -222,7 +222,7 @@ static int lxc_attach_to_ns(pid_t pid, int which)
 {
 	char path[MAXPATHLEN];
 	/* according to <http://article.gmane.org/gmane.linux.kernel.containers.lxc.devel/1429>,
-	 * the file for user namepsaces in /proc/$pid/ns will be called
+	 * the file for user namespaces in /proc/$pid/ns will be called
 	 * 'user' once the kernel supports it
 	 */
 	static char *ns[] = { "user", "mnt", "pid", "uts", "ipc", "net", "cgroup" };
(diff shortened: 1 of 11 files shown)


brauner:
jenkins: test this please

brauner:
Contains various fixes now:

- add `lxc_preserve_ns()` and replace all manual:

```
snprintf(nspath, MAXPATHLEN, "/proc/%d/ns/net", getpid());
open(nspath, O_RDONLY);
```
idioms with a simple `lxc_preserve_ns()` function call. `lxc_preserve_ns()` also avoids taking the hit of `path[MAXPATHLEN]`.

- Use the correct flag argument in `start.c` when calling `lxc_clone()`. (We somehow missed that bug for a while. I think @hallyn noted it at some point.)

- Improve the log output in `start.c` so the user and we can now properly see which namespaces where cloned at what time.

- Explicitly delete the host part of a veth device pair on container {stop, reboot}. This should prevent issues we've seen with `LXD` and repeated reboots.

- Improve general network deletion on container {stop, reboot}:
  - So far we blindly called `lxc_delete_network()` twice in `start.c` to make sure that we deleted all network devices. This resulted in pointless `netlink` calls, especially when a container had multiple networks defined. Let's be smarter and have `lxc_delete_network()` return a boolean that indicates whether *all* configured networks have been deleted. If so, don't needlessly try to delete them again in `start.c`. This also decreases confusing error messages a user might see.

  - When we receive `-ENODEV` from one of our `lxc_netdev_delete_*()` functions, let's assume that either the network device already got deleted or that it got moved to a different network namespace. Inform the user about this but do not report an error in this case.

  - When we have explicitly deleted the host side of a veth pair let's immediately `free(priv.veth_attr.pair) `and `NULL`, or `memset(priv.veth_attr.pair, ...)` the corresponding member so we don't needlessly try to destroy them again when we have to call `lxc_delete_network()` again in `start.c`.

brauner:
start, utils: improve `preserve_ns()`
- Allocating an error message that the caller must free seems pointless. We can just print the error message in `preserve_ns()` itself. This also allows us to avoid using the `GNU` extension `asprintf()`.
- Improve `lxc_preserve_ns()`: By passing in `NULL` or `""` as the second argument the function can now also be used to check whether namespaces are supported by the kernel.
- Use `lxc_preserve_ns()` in `preserve_ns()`.

start: improve log
Improve log in a bunch of places to make it easier for us on bug reports.

Blub:
If I see this right both `handler->netnsfd` and `handler->nsfd[LXC_NS_NET]` point to the same network namespace (both come from `handler->pid`). Maybe we could just reuse that? We'd have to expclicitly pass `CLONE_NEWNET` to the `preserve_ns(handler->nsfd, ...)` call, and not skip all further files when one fails (currently `preserve_ns()` goes to the `error` label as soon as any `lxc_preserve_ns()` call fails).

I wonder if we should mark the network interfaces `lxc_delete_network()` successfully deleted instead of just returning whether it worked for all of them, as there is a possible (but tiny) race condition where the index could be re-issued by another process between the two tries. (The kernel probably wouldn't since it only counts up, but netlink allows explicit indices at creation AFAIK as well).
Similarly, wouldn't it be better to only delete the host-side veth in the case where the interface has actually been moved (delete via `ifindex` failed)?

brauner:
@Blub, if you want you can checkout the branch, grab my commits and add you own commits on top of it. :)

src/lxc/attach.c
```diff
@@ -222,7 +222,7 @@ static int lxc_attach_to_ns(pid_t pid, int which)
 {
 	char path[MAXPATHLEN];
```
hallyn: In the access(path, X_OK) check below, you could just use "/proc/self/ns" and drop this path variable altogether?

src/lxc/attach.c
```diff
@@ -250,8 +250,7 @@ static int lxc_attach_to_ns(pid_t pid, int which)
 			continue;
 		}
 
-		snprintf(path, MAXPATHLEN, "/proc/%d/ns/%s", pid, ns[i]);
-		fd[i] = open(path, O_RDONLY | O_CLOEXEC);
+		fd[i] = lxc_preserve_ns(pid, ns[i]);
 		if (fd[i] < 0) {
 			saved_errno = errno;
```
hallyn: In my tree, below this there is a SYSERROR("failed to open '%s'", path); below this which you seem to be leaving in place, but path is no longer the right name.

brauner: Adressed. :)

PR #1256, opened 2016-10-28
remove atoi

brauner:
Has been sitting in my tree for a while now.

Replace by two custom functions:

```
lxc_safe_uint()
```

and

```
lxc_safe_int()
```

Should make things safer by switching a few types and adhering more to what we advertise for various keys in `man lxc.container.conf`.

### DIFF
--- a/src/lxc/tools/lxc_autostart.c
+++ b/src/lxc/tools/lxc_autostart.c
@@ -26,6 +26,7 @@
 #include "arguments.h"
 #include "list.h"
 #include "log.h"
+#include "utils.h"
 
 lxc_log_define(lxc_autostart_ui, lxc);
 static struct lxc_list *accumulate_list(char *input, char *delimiter, struct lxc_list *str_list);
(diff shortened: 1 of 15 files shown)


Blub:
In `config_ephemeral()` you're leaving the `ERROR()` message for when it's `> 1`, but are removing the `return -1` in this case whereas other boolean values return -1 but have no `ERROR()` msg (`kmsg`, `lsm_aa_allow_incomplete`, `autodev`).
Similar with `config_no_new_privs()`.

In `lxc_autostart.c` you seem to be removing a `free(value)`.

brauner:
Should be addressed.

Blub:
Looks good.

src/lxc/tools/lxc_autostart.c
```diff
@@ -290,8 +308,8 @@ static int get_config_integer(struct lxc_container *c, char *key) {
 		return 0;
 	}
 
-	ret = atoi(value);
-	free(value);
+	if (lxc_safe_int(value, &ret) < 0)
+		DEBUG("Could not parse config item.");
```
hallyn: You're not freeing value?

src/lxc/tools/lxc_autostart.c
```diff
@@ -311,6 +311,8 @@ static int get_config_integer(struct lxc_container *c, char *key) {
 	if (lxc_safe_int(value, &ret) < 0)
 		DEBUG("Could not parse config item.");
 
+	free(value);
```
hallyn: Hm, I see.  No, I'd really really prefer rebasing to fix the original patches.  Keeping the tree more bisectable.

brauner: Sure, will do that.

PR #1269, opened 2016-11-02
conf: merge network namespace move & rename on shutdown

Blub:
On shutdown we move physical network interfaces back to the
host namespace and rename them afterwards as well as in the
later lxc_network_delete() step. However, if the device had
a name which already exists in the host namespace then the
moving fails and so do the subsequent rename attempts. When
the namespace ceases to exist the devices finally end up
in the host namespace named 'dev<ID>' by the kernel.

In order to avoid this, we do the moving and renaming in a
single step (lxc_netdev_move_by_*()'s move & rename happen
in a single netlink transaction).

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2397,14 +2397,16 @@ static int setup_network(struct lxc_list *network)
 }
 
 /* try to move physical nics to the init netns */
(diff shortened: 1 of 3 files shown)


brauner:
jenkins: test this please

brauner:
Ok, tested it. Seems to work fine. :) However, I need to file a bug. Restoring wireless device names is currently not working.

PR #1276, opened 2016-11-04
Resource Limits

Blub:
Currently containers inherit the resource limits of the process starting them, but it can be useful to configure them explicitly.
If for instance a container can be started via a system service or manually via `lxc-start` or some wrapper, these 2 ways of starting the container can end up giving it different limits. Both have the option to change the limits before starting, but if they do not have `CAP_SYS_RESOURCE` or are unprivileged they won't be able to raise them back up if they lowered them and would have to resort to forking first.

Limits which aren't explicitly specified will be inherited as before, so the default behavior stays the same.
As usual the last specified value for a limit takes precedence, specifying an empty value deletes it (iow. declares it as to be inherited as before).

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -4256,10 +4256,14 @@ int lxc_clear_cgroups(struct lxc_conf *c, const char *key)
 {
 	struct lxc_list *it,*next;
 	bool all = false;
-	const char *k = key + 11;
+	const char *k = NULL;
 
 	if (strcmp(key, "lxc.cgroup") == 0)
 		all = true;
+	else if (strncmp(key, "lxc.cgroup.", sizeof("lxc.cgroup.")-1) == 0)
+		k = key + sizeof("lxc.cgroup.")-1;
+	else
+		return -1;
 
 	lxc_list_for_each_safe(it, &c->cgroup, next) {
 		struct lxc_cgroup *cg = it->elem;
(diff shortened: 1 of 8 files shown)


src/lxc/conf.c
```diff
@@ -4037,6 +4147,27 @@ int lxc_clear_cgroups(struct lxc_conf *c, const char *key)
 	return 0;
 }
 
+int lxc_clear_limits(struct lxc_conf *c, const char *key)
+{
+	struct lxc_list *it, *next;
+	bool all = false;
+	const char *k = key + 10;
```
Blub: Yeah could make it less magical (note the access to `k` later depends on `all` being `false`, which is an easily overlooked dependency, and error prone). I should probably also change `lxc_clear_cgroups` where I took the pattern from ;-).

brauner:
jenkins: test this please

stgraber:
jenkins: test this please

Blub:
Squashed the above mentioned change into it: https://github.com/Blub/lxc/commit/be95f354d4e764d4ebed87240e84e9cadcad4623
And added a 4th commit doing the same to the other functions of this pattern.

stgraber:
Looks like we dropped the ball on this one... Any chance you can rebase on master and then we can get some reviews done and merge this?

I agree having this feature would be pretty useful so it'd be nice to have this included.

brauner:
@Blub, once you've rebased I'm more than happy to merge. (If you happen to have time to add some sort of minimal tests for this than ++ if not I understand. :))

mmxe:
I spent a day and a half fighting this problem. If you can push this to ubuntu updates asap it's greatly appreciated.

Blub:
I can revisit this next week I think.

Blub:
Rebased and did the proposed changes. (Changes can be viewed separately here: https://github.com/Blub/lxc/commits/limits-3 as the `fixup:...` commits. (Not sure if the last one should be a separate commit here as well?))

brauner:
jenkins: test this please

brauner:
@Blub, thanks for implementing this! :)

Issue #1254, opened 2016-10-27
[Question] Reasons to limit copy/snapshot to stopped containers


Anthony25:
Hello,
lxc-copy and lxc-snapshot check that the target is stopped to continue. Is there a reason to do so? 

Backends like btrfs or zfs can do instant snapshots. I understand that it could be an issue if a database is running, but if it is the only reason, you could just warn the user about it in your manpages, no? Plus, tell me if I am wrong, but lxd allows hot snapshot/copy.

I patched lxc because I need to backup running containers, so I would like to know what possible drawbacks it could involve.

Thanks!

brauner:
I couldn't find an explanation why @hallyn used `is_stopped()` in `do_lxcapi_clone()`. Maybe he can enlighten us. :)

hallyn:
If lack of instant fs snapshots was the only problem we could have just frozen the containers around the snapshot action.  The problem (iirc) was inconsistent application state.  Applications may be in the middle of a transaction which cannot be correctly completed or reversed without knowing the state of the application's memory.

hallyn:
Note if you wanted to add a sufficiently scary sounding flag to override that that would be fine.

Anthony25:
Yep, I did a PR: #1277
You can always change the flag if it doesn't look scary enough :p

PR #1277, opened 2016-11-05
Option to force lxc-copy and lxc-snapshot for running domains

Anthony25:
Hello,
This PR is linked with the issue #1254. It adds a flag to force the clone of a container, even if it is running.
The python API has not been touched, though.

### DIFF
--- a/doc/lxc-copy.sgml.in
+++ b/doc/lxc-copy.sgml.in
@@ -57,6 +57,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
       <arg choice="opt">-s, --snapshot</arg>
       <arg choice="opt">-K, --keepname</arg>
       <arg choice="opt">-D, --keepdata</arg>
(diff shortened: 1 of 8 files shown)


Anthony25:
Sorry about the tests failure, the needed changes in the Python API were not added.

brauner:
I'm not convinced of this change. This will likely mean a lot of bug reports of inconsistent container clones and weird failures.

stgraber:
You'd also need to sign-off your commit before we can consider it for inclusion. (See CONTRIBUTING)

Anthony25:
@brauner: I agree, but personally I think that giving responsibility to the user is not a bad thing if they are aware about the possible dangers. Maybe some users might want to lock their database, make a snapshot/copy and then unlock their database, what would be safe but not possible right now.

@stgraber: Done, sorry about that.

brauner:
As I said, I don't think this patch provides enough value. While I understand the problem we shouldn't just allow this. Especially given that this involves some internal change and given from what I've seen also an API break. Which is not a thing we do. :)

PR #1282, opened 2016-11-06
cgroups: remove isolated cpus from cpuset.cpus  …

brauner:
In case the system was booted with

    isolcpus=n_i-n_j,n_k,n_m

we cannot simply copy the cpuset.cpus file from our parent cgroup. For example,
in the root cgroup cpuset.cpus will contain all of the cpus including the
isolated cpus. Copying the values of the root cgroup into a child cgroup will
lead to a wrong view in /proc/self/status: For the root cgroup
/sys/fs/cgroup/cpuset /proc/self/status will correctly show

    Cpus_allowed_list:      0-1,3

even though cpuset.cpus will show

    0-3

However, initializing a subcgroup in the cpuset controller by copying the
cpuset.cpus setting from the root cgroup will cause /proc/self/status to
incorrectly show

    Cpus_allowed_list:      0-3

Hence, we need to make sure to remove the isolated cpus from cpuset.cpus. Seth
has argued that this is not a kernel bug but by design. So let us be the smart
guys and fix this in liblxc.

The solution is straightforward: To avoid having to work with raw cpulist
strings we create cpumasks based on uint32_t bit arrays.

See discussion in https://bugs.launchpad.net/ubuntu/+source/lxc/+bug/1623143, especially [Seth's comment](https://bugs.launchpad.net/ubuntu/+source/lxc/+bug/1623143/comments/8).

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -33,21 +33,25 @@
  * under /sys/fs/cgroup/clist where clist is either the controller, or
  * a comman-separated list of controllers.
  */
+
 #include "config.h"
-#include <stdio.h>
+
+#include <ctype.h>
+#include <dirent.h>
+#include <errno.h>
+#include <grp.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <errno.h>
-#include <sys/types.h>
+#include <string.h>
 #include <unistd.h>
-#include <dirent.h>
-#include <grp.h>
+#include <sys/types.h>
 
 #include "bdev.h"
-#include "log.h"
 #include "cgroup.h"
-#include "utils.h"
 #include "commands.h"
+#include "log.h"
+#include "utils.h"
 
 lxc_log_define(lxc_cgfsng, lxc);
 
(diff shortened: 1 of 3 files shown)


hallyn:
I don't see any obvious bugs here (leaks or anything).  Will take a closer look later, but will be happy if someone else wants to merge it earlier.

src/lxc/utils.c
```diff
@@ -1930,3 +1930,37 @@ bool task_blocking_signal(pid_t pid, int signal)
...
+
+	*list = tmp;
+	(*list)[newentry + 1] = NULL;
+
+	return newentry;
+}
+
+int lxc_append_string(char ***list, char *entry)
+{
+	int newentry = lxc_append_null_to_list((void ***)list);
```
hallyn: Need to check for newentry == -1

src/lxc/cgroups/cgfsng.c
```diff
@@ -329,15 +591,18 @@ bool handle_cpuset_hierarchy(struct hierarchy *h, char *cgname)
 		return false;
 	}
 
+	/* Make sure any isolated cpus are removed from cpuset.cpus. */
+	if (!set_cpuset(cgpath, v == '1'))
```
hallyn: Given the comment, it sounds like set_cpuset() should be renamed to something like filter_cpuset() or remove_isol_cpuset()?

hallyn: Oh, I guess not.  Ignore.

brauner: Can do.

src/lxc/utils.c
brauner: Thanks!

brauner:
@hallyn, updated. :)

PR #1286, opened 2016-11-08
create symlink for /var/run

mgariepy:
this patch create /var/run link to point to /run.

This will fix various issue present when /var/run is persistent.


brauner:
Hi, you need to sign-off you commit (`git commit -s ... `). :)

PR #1290, opened 2016-11-10
cgroups: skip v2 hierarchy entry

brauner:


hallyn:
(Asking here for posterity) - is this the right thing to do, as opposed to actually supporting v2?

brauner:
We will support v2, but then we need separate logic anyway instead of hacking it into this. I'd rather have a dedicated `get_existing_subsystems_v1()` and `get_existing_subsystems_v2()`. Also, we should probably wait on what the `systemd` guys and see what they do before we make any moves(?).

PR #1293, opened 2016-11-12
also stop lxc-net in runlevels 0 and 6

evgeni:
there is no reason to not do this :)


brauner:
jenkins: test this please

PR #1295, opened 2016-11-12
install bash completion where pkg-config tells us to

evgeni:


brauner:
jenkins: test this please

PR #1299, opened 2016-11-14
lxc-checkpoint: enable dirty memory tracking in criu

adrianreber:
CRIU supports dirty memory tracking to take incremental checkpoints.
Incremental checkpoints are one way of reducing downtime during
migration. The first checkpoint dumps all the memory pages and the
second (and third, and fourth, ...) only dumps pages which have changed.

Most of the necessary code has already been implemented. This just adds
the existing functionality to lxc-checkpoint:

  -p, --pre-dump            Only pre-dump the memory of the container.
                            Container keeps on running and following
                            checkpoints will only dump the changes.
  --predump-dir=DIR         path to images from previous dump (relative to -D)

The following is an example from a container running CentOS 7 with psql
and tomcat:

 # lxc-checkpoint -n c7 -D /tmp/cp -p
Container keeps on running
 # du -h /tmp/cp
 229M	/tmp/cp
Sync initial checkpoint to destination
 # rsync -a /tmp/cp host2:/tmp/
Sync file-system
 # rsync -a /var/lib/lxc/c7 host2:/var/lib/lxc/
Final dump; container is stopped
 # lxc-checkpoint -n c7 -D /tmp/cp --predump-dir=../cp -s
 # du -h /tmp/cp2
 90M	/tmp/cp2

After transferring the second (incremental checkpoint) and the changes
to the container's file system the container can be restored on the
second host by pointing lxc-checkpoint to the second checkpoint
directory:

 # lxc-checkpoint -n c7 -D /tmp/cp2 -r

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -398,6 +398,7 @@ static void exec_criu(struct criu_opts *opts)
 		if (opts->user->predump_dir) {
 			DECLARE_ARG("--prev-images-dir");
 			DECLARE_ARG(opts->user->predump_dir);
+			DECLARE_ARG("--track-mem");
 		}
 
 		if (opts->user->pageserver_address && opts->user->pageserver_port) {
--- a/src/lxc/tools/lxc_checkpoint.c
+++ b/src/lxc/tools/lxc_checkpoint.c
@@ -36,6 +36,10 @@ static bool stop = false;
 static bool verbose = false;
 static bool do_restore = false;
 static bool daemonize_set = false;
+static bool pre_dump = false;
+static char *predump_dir = NULL;
+
+#define OPT_PREDUMP_DIR OPT_USAGE+1
 
 static const struct option my_longopts[] = {
 	{"checkpoint-dir", required_argument, 0, 'D'},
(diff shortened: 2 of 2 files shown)


adrianreber:
It seems my commit messages is formatted by github differently than expected (not at all).

@tych0 This is my preparation for pre-dump support in lxd to easier test the code path.

src/lxc/criu.c
```diff
@@ -398,6 +398,10 @@ static void exec_criu(struct criu_opts *opts)
 		if (opts->user->predump_dir) {
 			DECLARE_ARG("--prev-images-dir");
 			DECLARE_ARG(opts->user->predump_dir);
+			/* Using --prev-images-dir works in older CRIU versions (2.3).
+			 * For newer CRIU versions like 2.8 --track-mem needs to
+			 * be added to the command-line. */
+			DECLARE_ARG("--track-mem");
```
tych0: Will the parser fail on older versions?

tych0:
Other than the one comment, LGTM. We don't currently have any users of this code path, so if it's easier to just merge this as-is to experiment, that's fine with me.

adrianreber:
New version pushed without the comment. --track-mem is now always enabled when --prev-images-dir is used.

tych0:
Ok, cool. +1 then, thanks!

brauner:
jenkins: test this please

brauner:
jenkins: test this please

PR #1300, opened 2016-11-14
debian: Don't depend on libui-dialog-perl

stgraber:
This package doesn't exist in stretch anymore, and it's unclear why we
were depending on a library to begin with (as opposed to having it
brought by whatever needs it).


evgeni:
:+1:
I also don't see any reason to have this here. Just for archeological reasons, it was added in c952d1b9156ca3aecd74a3a7bde4b55a81641afe but there is no explanation why.

None of my non-LXC installs have this package installed :)

PR #1302, opened 2016-11-15
separate the limiting from the namespaced cgroup root

Blub:
When cgroup namespaces are enabled a privileged container
with mixed cgroups has full write access to its own root
cgroup effectively allowing it to overwrite values written
from the outside or configured via lxc.cgroup.*.

This patch causes an additional 'inner/' directory to be
created in all cgroups if cgroup namespaces and cgfsng are
being used in order to combat this.

This is a request for comments, it currently has the subdirectory name
hardcoded and is non-optional. All of this could of course be made
configurable if a "feature" like this is acceptable. It seems to only be
necessary for privileged containers which are, well, "privileged" after
all, but cgroup limits do seem pointless if they can't be enforced even
when it's privileged containers we're dealing with.

On the other hand for unprivileged containers maybe this could allow us to
make the entire directory the container sees as root cgroup writable by
the container? (Not sure about this, currently we only chown
`cgroup.procs` and `tasks`, and otherwise the container can only create
subdirectories.)

PS: I'm not sure whether criu.c needs special code there, but from the
host's point of view it's a subdirectory within the container's cgroup,
only the point at which `CLONE_NEWCGROUP` is used is moved, so only the
container's inner view changes.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -72,6 +72,7 @@ struct hierarchy {
 	char *mountpoint;
 	char *base_cgroup;
 	char *fullcgpath;
+	char *innercgpath;
 };
 
 /*
@@ -814,6 +815,7 @@ static void add_controller(char **clist, char *mountpoint, char *base_cgroup)
 	new->mountpoint = mountpoint;
 	new->base_cgroup = base_cgroup;
 	new->fullcgpath = NULL;
+	new->innercgpath = false;
 
 	newentry = append_null_to_list((void ***)&hierarchies);
 	hierarchies[newentry] = new;
@@ -1286,6 +1288,8 @@ static void cgfsng_destroy(void *hdata, struct lxc_conf *conf)
 				free(h->fullcgpath);
 				h->fullcgpath = NULL;
 			}
+			free(h->innercgpath);
+			h->innercgpath = NULL;
 		}
 	}
 
@@ -1299,18 +1303,25 @@ struct cgroup_ops *cgfsng_ops_init(void)
 	return &cgfsng_ops;
 }
 
-static bool create_path_for_hierarchy(struct hierarchy *h, char *cgname)
+static bool create_path_for_hierarchy(struct hierarchy *h, char *cgname, bool inner)
 {
-	h->fullcgpath = must_make_path(h->mountpoint, h->base_cgroup, cgname, NULL);
-	if (dir_exists(h->fullcgpath)) { // it must not already exist
-		ERROR("Path \"%s\" already existed.", h->fullcgpath);
+	char *path;
+	if (inner) {
+		path = must_make_path(h->fullcgpath, CGROUP_NAMESPACE_SUBDIR, NULL);
+		h->innercgpath = path;
+	} else {
+		path = must_make_path(h->mountpoint, h->base_cgroup, cgname, NULL);
+		h->fullcgpath = path;
+	}
+	if (dir_exists(path)) { // it must not already exist
+		ERROR("Path \"%s\" already existed.", path);
 		return false;
 	}
-	if (!handle_cpuset_hierarchy(h, cgname)) {
+	if (!inner && !handle_cpuset_hierarchy(h, cgname)) {
 		ERROR("Failed to handle cgroupfs v1 cpuset controller.");
 		return false;
 	}
-	return mkdir_p(h->fullcgpath, 0755) == 0;
+	return mkdir_p(path, 0755) == 0;
 }
 
 static void remove_path_for_hierarchy(struct hierarchy *h, char *cgname)
@@ -1325,7 +1336,8 @@ static void remove_path_for_hierarchy(struct hierarchy *h, char *cgname)
  * Try to create the same cgroup in all hierarchies.
  * Start with cgroup_pattern; next cgroup_pattern-1, -2, ..., -999
  */
-static inline bool cgfsng_create(void *hdata)
+static inline bool cgfsng_create_inner(struct cgfsng_handler_data*);
+static inline bool cgfsng_create(void *hdata, bool inner)
 {
 	struct cgfsng_handler_data *d = hdata;
 	char *tmp, *cgname, *offset;
@@ -1335,7 +1347,13 @@ static inline bool cgfsng_create(void *hdata)
 	if (!d)
 		return false;
 	if (d->container_cgroup) {
+		if (inner)
+			return cgfsng_create_inner(d);
 		WARN("cgfsng_create called a second time");
+		return false;
+	}
+	if (inner) {
+		ERROR("cgfsng_create called twice for innner cgroup");
 		return false;
 	}
 
@@ -1358,7 +1376,7 @@ again:
 	if (idx)
 		snprintf(offset, 5, "-%d", idx);
 	for (i = 0; hierarchies[i]; i++) {
-		if (!create_path_for_hierarchy(hierarchies[i], cgname)) {
+		if (!create_path_for_hierarchy(hierarchies[i], cgname, false)) {
 			int j;
 			SYSERROR("Failed to create %s: %s", hierarchies[i]->fullcgpath, strerror(errno));
 			free(hierarchies[i]->fullcgpath);
@@ -1378,7 +1396,24 @@ out_free:
 	return false;
 }
 
-static bool cgfsng_enter(void *hdata, pid_t pid)
+static inline bool cgfsng_create_inner(struct cgfsng_handler_data *d)
+{
+	size_t i;
+	bool ret = true;
+	char *cgname = must_make_path(d->container_cgroup, CGROUP_NAMESPACE_SUBDIR, NULL);
+	for (i = 0; hierarchies[i]; i++) {
+		if (!create_path_for_hierarchy(hierarchies[i], cgname, true)) {
+			SYSERROR("Failed to create %s namespace subdirectory: %s", hierarchies[i]->fullcgpath, strerror(errno));
+			ret = false;
+			break;
+		}
+	}
+	free(cgname);
+	return ret;
+}
+
+
+static bool cgfsng_enter(void *hdata, pid_t pid, bool inner)
 {
 	char pidstr[25];
 	int i, len;
@@ -1388,7 +1423,13 @@ static bool cgfsng_enter(void *hdata, pid_t pid)
 		return false;
 
 	for (i = 0; hierarchies[i]; i++) {
-		char *fullpath = must_make_path(hierarchies[i]->fullcgpath,
+		char *fullpath;
+		if (inner)
+			fullpath = must_make_path(hierarchies[i]->fullcgpath,
+						CGROUP_NAMESPACE_SUBDIR,
+						"cgroup.procs", NULL);
+		else
+			fullpath = must_make_path(hierarchies[i]->fullcgpath,
 						"cgroup.procs", NULL);
 		if (lxc_write_to_file(fullpath, pidstr, len, false) != 0) {
 			SYSERROR("Failed to enter %s", fullpath);
@@ -1404,6 +1445,7 @@ static bool cgfsng_enter(void *hdata, pid_t pid)
 struct chown_data {
 	struct cgfsng_handler_data *d;
 	uid_t origuid; // target uid in parent namespace
+	bool inner;
 };
 
 /*
@@ -1432,13 +1474,20 @@ static int chown_cgroup_wrapper(void *data)
 	for (i = 0; hierarchies[i]; i++) {
 		char *fullpath, *path = hierarchies[i]->fullcgpath;
 
+		if (arg->inner)
+			path = must_make_path(path, CGROUP_NAMESPACE_SUBDIR, NULL);
+
 		if (chown(path, destuid, 0) < 0) {
 			SYSERROR("Error chowning %s to %d", path, (int) destuid);
+			if (arg->inner)
+				free(path);
 			return -1;
 		}
 
 		if (chmod(path, 0775) < 0) {
 			SYSERROR("Error chmoding %s", path);
+			if (arg->inner)
+				free(path);
 			return -1;
 		}
 
@@ -1462,12 +1511,14 @@ static int chown_cgroup_wrapper(void *data)
 		if (chmod(fullpath, 0664) < 0)
 			WARN("Error chmoding %s: %m", path);
 		free(fullpath);
+
+		free(path);
 	}
 
 	return 0;
 }
 
-static bool cgfsns_chown(void *hdata, struct lxc_conf *conf)
+static bool cgfsns_chown(void *hdata, struct lxc_conf *conf, bool inner)
 {
 	struct cgfsng_handler_data *d = hdata;
 	struct chown_data wrap;
@@ -1480,6 +1531,7 @@ static bool cgfsns_chown(void *hdata, struct lxc_conf *conf)
 
 	wrap.d = d;
 	wrap.origuid = geteuid();
+	wrap.inner = inner;
 
 	if (userns_exec_1(conf, chown_cgroup_wrapper, &wrap) < 0) {
 		ERROR("Error requesting cgroup chown in new namespace");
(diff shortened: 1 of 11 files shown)


stgraber:
So one concern with doing this for unprivileged containers too is the added overhead of being one level further from the root of the cgroup tree. Some controllers are pretty expensive in that regard (or at least were).

However, I'm certainly fine with the approach for privileged containers since it's effectively the only way to prevent them from escaping their limits.

Blub:
Could be made configurable for both? Perhaps a global-config value like `lxc.cgroup.separate = none | privileged | unprivileged | both`?

brauner:
I guess the cgroup that is most likely to be expensive in that regard is the memory controller. Especially when moving tasks since I imagine the memory pages need to be relabeled.

brauner:
Making it configurable for both seems like a good option. A global value also seems to make the most sense. It should always be activated for privileged containers and deactivated for unprivileged containers.

brauner:
@Blub, up for implementing this? :)

Blub:
Sure, so far cgfsng seems to be working, not sure whether it's needed with cgmanager and the exact use case differences between cgfs and cgfsng.

brauner:
`cgfs` is our fallback backend. If you can hack it in there, go ahead. I'd say it's safe to ignore `cgmanager`. @hallyn, any objections? :)

src/lxc/cgroups/cgfsng.c
```diff
@@ -1303,18 +1303,24 @@ struct cgroup_ops *cgfsng_ops_init(void)
...
 {
-	h->fullcgpath = must_make_path(h->mountpoint, h->base_cgroup, cgname, NULL);
-	if (dir_exists(h->fullcgpath)) { // it must not already exist
-		ERROR("Path \"%s\" already existed.", h->fullcgpath);
-		return false;
-	}
-	if (!handle_cpuset_hierarchy(h, cgname)) {
-		ERROR("Failed to handle cgroupfs v1 cpuset controller.");
-		return false;
+	char *path;
```
hallyn: @path is leaked?

hallyn:
Is there a better name we can use?  lxc.cgroup.nest?  That's not ideal, but 'lxc.cgroup.separate' says something different to me.

src/lxc/initutils.c
```diff
@@ -139,11 +139,11 @@ const char *lxc_global_config_value(const char *option_name)
 	}
 
 	const char * const (*ptr)[2];
-	size_t i;
+	size_t option;
```
hallyn: Kinda bike-shedding, but 'option' next to 'options' (both of which can be safely incremented) also seems scary. 'opt' maybe?  Of course the best solution would be to break this function up, but I'm not asking you to do this as part of this PR.

hallyn:
Yeah if anyone wants this for cgmanager they can do it themselves (and in a cgmanager world the "can't write to /" is a more natural fit anyway).

src/lxc/cgroups/cgfsng.c
Blub: thanks, will fix it

src/lxc/start.c
```diff
@@ -1067,6 +1067,7 @@ static int lxc_spawn(struct lxc_handler *handler)
 	int saved_ns_fd[LXC_NS_MAX];
 	int preserve_mask = 0, i, flags;
 	int netpipepair[2], nveths;
+	bool privileged = !!lxc_list_empty(&handler->conf->id_map);
```
brauner: Double negation?

brauner:
jenkins: test this please

hallyn:
On Tue, Nov 29, 2016 at 08:29:33PM -0800, Christian Brauner wrote:
> 
> 
> 
> > @@ -1067,6 +1067,7 @@ static int lxc_spawn(struct lxc_handler *handler)
>  	int saved_ns_fd[LXC_NS_MAX];
>  	int preserve_mask = 0, i, flags;
>  	int netpipepair[2], nveths;
> +	bool privileged = !!lxc_list_empty(&handler->conf->id_map);
> 
> Double negation?

Yeah it's not unusual, although it's one of those things that naturally
makes the brain pause.  Using == 0 would probably be less jarring, but
this is not wrong.

brauner:
Yep, my point is just about readability.

src/lxc/initutils.c
```diff
@@ -96,6 +96,7 @@ const char *lxc_global_config_value(const char *option_name)
 		{ "lxc.default_config",     NULL            },
 		{ "lxc.cgroup.pattern",     NULL            },
 		{ "lxc.cgroup.use",         NULL            },
+		{ "lxc.cgroup.separate",    DEFAULT_CGSEPARATE },
```
brauner: Maybe `lxc.cgroup.protect_limits`?

Blub:
`lxc.cgroup.protect_limits` could work,
about the double negation, can change to `==0`, (hey at least the variable isn't also called _un_-privileged ;-) )

brauner:
If you have updated this PR leave a ping here because I'm not sure whether I'd be notified if you don't. :)

Blub:
@brauner: Updated with the requested changes.

Also fixed an issue with `lxc-attach`: since cgroup operations use the monitor cgroup which is outside the namespace-directory in order for changes to end up there, `cgroup_attach()` used to use the monitor's cgroup as well and so `lxc-attach` ended up attaching to the wrong namespace.
To more easily review this change I copied out the patch from before squashing into a gist: https://gist.github.com/Blub/9190fc64c01636af7ce40bcb16cb7d72

Blub:
Actually, hold on a bit...

Blub:
Okay since the attaching process doesn't necessarily have all the information (in particular the new global `protect_limits` may have changed in `/etc/` by then) I had to add another `LXC_CMD_` to get the inner cgroup if it exists. (It'll return the regular one if the inner one is not being used.)
(Updated the gist as well)

brauner:
@blub, does the `LXC_CMD_` addendum cause an API break?

brauner:
jenkins: test this please

Blub:
Right, sort of. Currently a new `lxc-attach` will not attach to the cgroups. I can change `lxc_cmd_get_attach_cgroup_path()` to try and fallback to the old `...get_cgroup_path()` on error (`ret==0`). It seems to behave correctly then.

Blub:
Squashed this into it: https://github.com/Blub/lxc/commit/924a6d05de08b3869fc50c0ba28851e7b7a3b806

brauner:
@Blub, cool. Thanks. :)

brauner:
jenkins: test this please

src/lxc/cgroups/cgfsng.c
```diff
@@ -1299,18 +1303,25 @@ struct cgroup_ops *cgfsng_ops_init(void)
...
 
-static bool create_path_for_hierarchy(struct hierarchy *h, char *cgname)
+static bool create_path_for_hierarchy(struct hierarchy *h, char *cgname, bool inner)
 {
-	h->fullcgpath = must_make_path(h->mountpoint, h->base_cgroup, cgname, NULL);
-	if (dir_exists(h->fullcgpath)) { // it must not already exist
-		ERROR("Path \"%s\" already existed.", h->fullcgpath);
+	char *path;
+	if (inner) {
+		path = must_make_path(h->fullcgpath, "ns", NULL);
```
brauner: I take it that the `h->fullcgpath` is already guaranteed to be initialized once this codepath is hit.

src/lxc/cgroups/cgfsng.c
```diff
@@ -1299,18 +1303,25 @@ struct cgroup_ops *cgfsng_ops_init(void)
...
+	} else {
+		path = must_make_path(h->mountpoint, h->base_cgroup, cgname, NULL);
+		h->fullcgpath = path;
+	}
+	if (dir_exists(path)) { // it must not already exist
+		ERROR("Path \"%s\" already existed.", path);
 		return false;
 	}
-	if (!handle_cpuset_hierarchy(h, cgname)) {
+	if (!inner && !handle_cpuset_hierarchy(h, cgname)) {
```
brauner: Just checking: can there be a case where the `cpuset` hierarchy still needs to be handled?

src/lxc/cgroups/cgfsng.c
```diff
@@ -1378,7 +1396,24 @@ static inline bool cgfsng_create(void *hdata)
 	return false;
 }
 
-static bool cgfsng_enter(void *hdata, pid_t pid)
+static inline bool cgfsng_create_inner(struct cgfsng_handler_data *d)
+{
+	size_t i;
+	bool ret = true;
+	char *cgname = must_make_path(d->container_cgroup, "ns", NULL);
```
brauner: Feels like bike-shedding but is the `ns` supposed to meaningful or just random? Maybe there's a smarter name for this, so it's clear in the `/sys/fs/cgroup/*/*/ns` tree why this cgroup exists. :)

brauner: Also, maybe you can make this a `#define` somewhere and add a little comment. :)

src/lxc/commands.c
```diff
@@ -128,15 +128,16 @@ static int fill_sock_name(char *path, int len, const char *name,
...
+		[LXC_CMD_CONSOLE]           = "console",
+		[LXC_CMD_STOP]              = "stop",
+		[LXC_CMD_GET_STATE]         = "get_state",
+		[LXC_CMD_GET_INIT_PID]      = "get_init_pid",
+		[LXC_CMD_GET_CLONE_FLAGS]   = "get_clone_flags",
+		[LXC_CMD_GET_CGROUP]        = "get_cgroup",
+		[LXC_CMD_GET_CONFIG_ITEM]   = "get_config_item",
+		[LXC_CMD_GET_NAME]          = "get_name",
+		[LXC_CMD_GET_LXCPATH]       = "get_lxcpath",
+		[LXC_CMD_GET_ATTACH_CGROUP] = "get_attach_cgroup",
```
brauner: Just a reminder: `get_attach_cgroup()` should be really well thought-through because once this is a thing in the wild you cannot change it since this would be an api break. The only way to change it after that is by adding yet another member to this struct. :)

src/lxc/cgroups/cgfsng.c
Blub: Yes, it comes from  `cgfsng_create()` which errors if `inner` is set on the first call (or not set on the second call). But I can add an additional check there as well if you'd prefer that.

src/lxc/cgroups/cgfsng.c
Blub: Can do. It was supposed to mean "namespace" as it's there for the cgroup-namespace. Could write it out as well?

src/lxc/commands.c
Blub: I just read through the existing `LXC_CMD_GET_CGROUP` again and think it might even be possible to add it as a parameter. It currently passes the cgroup name zero-terminated (as it simply passes it over to `cgroup_get_cgroup()`, so by checking the data against the length it should be possible distinguish the two. New code connecting to old monitors should work because they don't check `strlen()` against `datalen`, old code connecting to new monitors would work because they'd always use the old method and new+new would be able to distinguish the two calls.

src/lxc/cgroups/cgfsng.c
brauner: Nah, it's fine then. :)

src/lxc/cgroups/cgfsng.c
brauner: If you could please simply name it `cgns`, make it a `#define` and give the `#define` a comment that explains what it is supposed to represent. :)

src/lxc/commands.c
brauner: Cool, if you want to code that up I'm very much for it. :)

src/lxc/cgroups/cgfsng.c
Blub: AFAIK no. It sets `cgroup.clone_children`, filters cpus and clones the parent's `mems`. When the inner group is created `clone_children` will already be in effect (and if it had failed to set it it'll have errored out), so this should not be necessary at this point anymore.

src/lxc/commands.c
Blub: Testing it out right now.

Blub:
Updates:
`#define` for `cgns/` subdir: https://github.com/Blub/lxc/commit/f1873b3d599d29174c3e5b7d0db9ed5816135440
Extending existing `LXC_CMD_GET_CGROUP`: https://github.com/Blub/lxc/commit/b08a732a55786e1737fdd2c03a26e26ca00ef8dc

src/lxc/cgroups/cgfsng.c
brauner: Cool, thanks.

brauner:
jenkins: test this please

brauner:
@Blub, thanks! :)

src/lxc/cgroups/cgfsng.c
```diff
@@ -814,6 +815,7 @@ static void add_controller(char **clist, char *mountpoint, char *base_cgroup)
 	new->mountpoint = mountpoint;
 	new->base_cgroup = base_cgroup;
 	new->fullcgpath = NULL;
+	new->innercgpath = false;
```
tych0: This should be NULL.

src/lxc/commands.c
```diff
@@ -128,15 +128,15 @@ static int fill_sock_name(char *path, int len, const char *name,
...
-		[LXC_CMD_GET_LXCPATH]     = "get_lxcpath",
+		[LXC_CMD_CONSOLE]           = "console",
+		[LXC_CMD_STOP]              = "stop",
+		[LXC_CMD_GET_STATE]         = "get_state",
+		[LXC_CMD_GET_INIT_PID]      = "get_init_pid",
+		[LXC_CMD_GET_CLONE_FLAGS]   = "get_clone_flags",
+		[LXC_CMD_GET_CGROUP]        = "get_cgroup",
+		[LXC_CMD_GET_CONFIG_ITEM]   = "get_config_item",
+		[LXC_CMD_GET_NAME]          = "get_name",
+		[LXC_CMD_GET_LXCPATH]       = "get_lxcpath",
```
tych0: What's the point of this hunk?

tych0:
Two questions here:

1. Why does cgns matter here? Doesn't the problem happen even without cgns?
2. What do things like lxc_cmd_get_cgroup_path() return? if it's `/lxc/foo/` I think this should be fine, but `/lxc/foo/cgns` will cause the container's limits not to be checkpointed. This may or may not be a problem, assuming we set things up correctly again, but it's worth noting.

src/lxc/commands.c
Blub: Leftover from when I had an extra entry there, sorry.

src/lxc/cgroups/cgfsng.c
Blub: Ouch, yes.

Blub:
1. non-cgns case: I'll have to check again, haven't used non-cgns containers in a while. The issue comes from the permissions, so they should have the same issue, and considering the AppArmor profile would prevent mounting the outer cgroups the same fix would probably work. I'll have to test this soon.

2. `lxc_cmd_*` return values: The existing function returns the same value as before, `/lxc/foo/`. The new function `lxc_cmd_get_attach_cgroup()` returns the inner one. They both use the same `LXC_CMD_GET_CGROUP`. This previously only took the subsystem name as a string argument, and now checks whether the received data length is greater than that string in which case the byte afterwards can be set to `1` to request the inner cgroup. If no inner cgroup exists the regular one will be returned. `cgroup_attach()` uses the new function as it is used by `lxc-attach` (this is currently the only user of this function). Actually, I could add that as a parameter to `lxc_cmd_get_cgroup()` instead as well. Somehow I wasn't sure whether it would be usable from external tools but it's not in the headers and wrapped in `lxccontainer.c`'s `lxcapi_get_config_item`.

src/lxc/commands.c
brauner: Did you update this?

brauner:
@Blub, do you intend to do update this?

brauner:
I'm leaving this one open as I'd really like to have it and I'm not letting @Blub off the hook. :P

brauner:
Closing this for now.

PR #1312, opened 2016-11-22
tools: replace non-standard namespace identifiers

brauner:
The identifiers for namespaces used with lxc-unshare and lxc-attach as given on
the manpage do not align with the standard identifiers. This affects network,
mount, and uts namespaces. The standard identifiers are: "mnt", "uts", and
"net" whereas lxc-unshare and lxc-attach use "MOUNT", "UTSNAME", and "NETWORK".
I'm weary to hack this into namespace.{c.h} by e.g. adding additional members
to the ns_info struct or to special case this in lxc_fill_namespace_flags().
Internally, we should only accept standard identifiers to ensure that we are
always correctly aligned with the kernel. So let's use some cheap memmove()s to
replace them by their standard identifiers in lxc-unshare and lxc-attach.

### DIFF
--- a/src/lxc/tools/lxc_attach.c
+++ b/src/lxc/tools/lxc_attach.c
@@ -107,6 +107,8 @@ static int add_to_simple_array(char ***array, ssize_t *capacity, char *value)
 
 static int my_parser(struct lxc_arguments* args, int c, char* arg)
 {
+	char **it;
+	char *del;
 	int ret;
 
 	switch (c) {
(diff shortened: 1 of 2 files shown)


Blub:
I think in the two `memmove(del + 3, del + 7, strlen(del) - 4);` you want to subtract `6` instead of `4`. While 4 is small enough to include the terminating nul bytes you also copy some trailing bytes after the 0. (If you execute this on a `strdup()`ed string (to get exact allocation sizes) while using valgrind on a string that doesn't contain `MOUNT` before `NETWORK` or `UTSNAME` it shows out-of-bounds accesses. (If `MOUNT` was part of the string then the removal of `OU` makes sure that the difference of 2 doesn't go out of bounds already).
```
==29997== Invalid read of size 1
==29997==    at 0x4C2D6A0: memcpy@GLIBC_2.2.5 (vg_replace_strmem.c:914)
```

stgraber:
Android regression: https://jenkins.linuxcontainers.org/view/All/job/lxc-build-android/1006/console

brauner:
Yep, but that has something to do with an explicit integer overflow I caused and is unrelated to this patch. I'll take care of that right after this.

brauner:
To be precise, the android regression is related to testing `lxc_safe_{u}int()`.

PR #1314, opened 2016-11-22
Configure a static MAC address on the LXC bridge

### LINKED ISSUE #604: lxcbr0 MAC address is unstable, causing sproradic network outages
In Linux, bridges take on the MAC address of the lowest attached interface by default. Thus, if that interface is removed (by stopping a container, for instance), the MAC changes, temporarily disrupting network connectivity for all other interfaces on the bridge. Please configure a MAC address on the bridge when it is created to prevent it from changing.

ccope:
Requires Linux >= 2.6.26. Prior to that, you could not set a fixed mac address on a bridge.

This is just a random MAC address, the maintainers of LXC should request a MAC Address Block Small assignment from the IEEE here: https://regauth.standards.ieee.org/standards-ra-web/app

Resolves #604

### DIFF
--- a/config/init/common/lxc-net.in
+++ b/config/init/common/lxc-net.in
@@ -9,6 +9,7 @@ varlib="@LOCALSTATEDIR@/lib"
 
 USE_LXC_BRIDGE="true"
 LXC_BRIDGE="lxcbr0"
+LXC_BRIDGE_MAC="00:16:3e:00:00:00"
 LXC_ADDR="10.0.3.1"
 LXC_NETMASK="255.255.255.0"
 LXC_NETWORK="10.0.3.0/24"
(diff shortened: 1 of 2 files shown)


brauner:
You forgot to sign-off your commit: `git commit -s ...`.

ccope:
@brauner Ok, I'll amend this commit once you guys get a real MAC address assignment from the IEEE

stgraber:
Just use a 00:16:3e:xx:xx:xx MAC, that's what LXC uses by default for containers (registered by Xen for use by virtual machines and containers).

Blub:
Isn't a fixed static mac address like this going to hurt people with multiple lxc-using machines in the same LAN? Or are the trailing zeroes randomized somewhere already?

PR #1315, opened 2016-11-23
conf, criu: add make_anonymous_mount_file()

### LINKED ISSUE #1296: [lxc-android] Can't bind mount inside container
Hello,

For lxc on android, when I try to bind mount the directory inside container.

```
lxc.mount.entry = /system system/ none bind,optional,create=dir
```

it fails with cryptic error

```
lxc_conf - conf.c:write_mount_file:1960 - tmpfile error: m                       
lxc_conf - conf.c:lxc_setup:3731 - failed to setup the mount entries for 'system'
```

I am not really sure whatever is "tmpfile error: m"?

brauner:
Before we used tmpfile() to write out mount entries for the container. This
requires a writeable /tmp file system which can be a problem for Android where
this filesystem is not necessarily present. This commit switches from tmpfile()
to using the memfd_create() syscall. It allows us to create an anonymous tmpfs
file (And is somewhat similar to mmap().) which is automatically deleted as
soon as any references to it are dropped.

Closes #1296.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -39,7 +39,6 @@
 #include <unistd.h>
 #include <arpa/inet.h>
 #include <linux/loop.h>
-#include <linux/memfd.h>
 #include <net/if.h>
 #include <netinet/in.h>
 #include <sys/mman.h>
@@ -61,6 +60,10 @@
 #include <pty.h>
 #else
 #include <../include/openpty.h>
+#endif
+
+#ifdef HAVE_LINUX_MEMFD_H
+#include <linux/memfd.h>
 #endif
 
 #include "af_unix.h"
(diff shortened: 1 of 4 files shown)


brauner:
Let's see if all architectures like this.

stgraber:
This effectively means that LXC won't work on anything prior to 3.17 though, right?

If so, that's not acceptable.

brauner:
Yeah, I will add fallback code for that case. I just want to see whether our Android is in general fine with this.

stgraber:
Well, no shipping Android kernels have that syscall, so the direct syscall will "work" with bionic but won't give you much :)

brauner:
*sigh*, Android. Well, `memfd_create()` has its value in any case as it allows us to avoid `/tmp` in most cases.

brauner:
Well, `Marshmallow` and `Nougat` should have `memfd_create()` supporrt. :)

stgraber:
@brauner nope, they ship with a 3.12 kernel.

brauner:
*sigh* again. In any case, I added fallback code in case the `memfd_create()` syscall is not implemented. I still see value in not relying on `/tmp` being present when the kernel allows us to.

src/lxc/conf.c
```diff
@@ -49,37 +62,25 @@
...
-#include <sys/mount.h>
-#include <sys/mman.h>
-#include <sys/prctl.h>
-
-#include <arpa/inet.h>
-#include <fcntl.h>
-#include <netinet/in.h>
-#include <net/if.h>
-#include <libgen.h>
+#ifdef HAVE_LINUX_MEMFD_H
```
stgraber: You don't have a configure.ac entry for this

src/lxc/conf.c
```diff
@@ -167,6 +166,57 @@ return -1;
...
+/* memfd_create() */
+#ifndef MFD_CLOEXEC
+#define MFD_CLOEXEC 0x0001U
+#endif
+
+#ifndef MFD_ALLOW_SEALING
+#define MFD_ALLOW_SEALING 0x0002U
+#endif
+
+#ifndef HAVE_MEMFD_CREATE
```
brauner: Will do.

src/lxc/conf.c
brauner: Will do.

brauner:
@stgraber, this should check for `memfd_create()` in `configure.ac` now.

PR #1318, opened 2016-11-25
log: use thread-safe localtime_r()

brauner:
This fixes a race in liblxc logging which can lead to deadlocks. The reproducer
for this issue before this is to simply compile with --enable-tests and then
run:

    lxc-test-concurrent -j 20 -m create,start,stop,destroy -D

which should deadlock.

### DIFF
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -20,6 +20,8 @@
  * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
+
+#define _GNU_SOURCE
 #include <assert.h>
 #include <stdio.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


brauner:
I'm pretty convinced this will also fix https://github.com/lxc/lxd/issues/2189.

src/lxc/log.c
```diff
@@ -155,8 +155,10 @@ static int log_append_logfile(const struct lxc_log_appender *appender,
 	if (fd_to_use == -1)
 		return 0;
 
-	t = localtime(&event->timestamp.tv_sec);
-	strftime(date, sizeof(date), "%Y%m%d%H%M%S", t);
+	if (!localtime_r(&event->timestamp.tv_sec, &newtime))
```
brauner: For future reference, this fix is not at all complete. The race actually comes from `localtime()`/`localtime_r()` and friends using `tzset()` internally which seems to deadlock in multi-threaded scenarios. My suspicion is that `fork()`ing on daemonized start in multi-threaded environments messes up the mutexes this function uses internally.

PR #1319, opened 2016-11-25
log: fix race

brauner:
localtime_r() can lead to deadlocks because it calls __tzset() and
__tzconvert() internally. The deadlock stems from an interaction between these
functions and the functions in monitor.c and commands.{c,h}. The latter
functions will write to the log independent of the container thread that is
currently running. Since the monitor fork()ed it seems to duplicate the mutex
states of the time functions mentioned above causing the deadlock.
As a short termm fix, I suggest to simply disable receiving the time when
monitor.c or command.{c,h} functions are called. This should be ok, since the
[lxc monitor] will only emit a few messages and thread-safety is currently more
important than beautiful logs. The rest of the log stays the same as it was
before.

Here is an example output from logs where I printed the pid and tid of the
process that is currently writing to the log:

```
            lxc 20161125170200.619 INFO     lxc_start:   18695-18695: - start.c:lxc_check_inherited:243 - Closed inherited fd: 23.
            lxc 20161125170200.640 DEBUG    lxc_start:   18677-18677: - start.c:__lxc_start:1334 - Not dropping CAP_SYS_BOOT or watching utmp.
            lxc 20161125170200.640 INFO     lxc_cgroup:  18677-18677: - cgroups/cgroup.c:cgroup_init:68 - cgroup driver cgroupfs-ng initing for lxc-test-concurrent-0

----------> lxc 20150427012246.000 INFO     lxc_monitor: 13017-18622: - monitor.c:lxc_monitor_sock_name:178 - using monitor sock name lxc/ad055575fe28ddd5//var/lib/lxc

            lxc 20161125170200.662 DEBUG    lxc_cgfsng:  18677-18677: - cgroups/cgfsng.c:filter_and_set_cpus:478 - No isolated cpus detected.
            lxc 20161125170200.662 DEBUG    lxc_cgfsng:  18677-18677: - cgroups/cgfsng.c:handle_cpuset_hierarchy:648 - "cgroup.clone_children" was already set to "1".
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -628,7 +628,7 @@ AC_CHECK_DECLS([PR_GET_NO_NEW_PRIVS], [], [], [#include <sys/prctl.h>])
 AC_CHECK_HEADERS([sys/signalfd.h pty.h ifaddrs.h sys/capability.h sys/personality.h utmpx.h sys/timerfd.h])
 
 # Check for some syscalls functions
-AC_CHECK_FUNCS([setns pivot_root sethostname unshare rand_r confstr faccessat])
+AC_CHECK_FUNCS([setns pivot_root sethostname unshare rand_r confstr faccessat gettid])
 
 # Check for some functions
 AC_CHECK_LIB(pthread, main)
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -155,7 +155,7 @@ static int log_append_logfile(const struct lxc_log_appender *appender,
 	char buffer[LXC_LOG_BUFFER_SIZE];
 	struct tm newtime;
 	int n;
-	int ms;
+	int ms = 0;
 	int fd_to_use = -1;
 
 #ifndef NO_LXC_CONF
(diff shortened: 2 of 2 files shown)


stgraber:
Approach is fine with me, but we should definitely print something other than 20150427012246.000 as the timestamp in such case.

stgraber:
You made Android sad: https://jenkins.linuxcontainers.org/view/All/job/lxc-build-android/1031/console

brauner:
@stgraber, new log output now looks like this:

        lxc 20161125201943.819 INFO     lxc_start - start.c:lxc_check_inherited:243 - Closed inherited fd: 54.
    --> lxc N/A                INFO     lxc_monitor - monitor.c:lxc_monitor_sock_name:178 - using monitor sock name lxc/ad055575fe28ddd5//var/lib/lxc
        lxc 20161125201943.958 DEBUG    lxc_commands - commands.c:lxc_cmd_handler:893 - peer has disconnected
    --> lxc N/A                DEBUG    lxc_commands - commands.c:lxc_cmd_get_state:579 - 'lxc-test-concurrent-0' is in 'RUNNING' state
        lxc 20161125201943.960 DEBUG    lxc_commands - commands.c:lxc_cmd_handler:893 - peer has disconnected
        lxc 20161125201944.009 INFO     lxc_start - start.c:lxc_check_inherited:243 - Closed inherited fd: 3.

brauner:
Also, Android's now happy. :)

brauner:
Should fix https://github.com/lxc/lxd/issues/2189.

PR #1331, opened 2016-11-27
log: drop all timezone conversion functions 

brauner:
Our log functions need to make extra sure that they are thread-safe. We had
some problems with that before. This especially involves time-conversion
functions. I don't want to find any localtime() or gmtime() functions or
relatives in here. Not even localtime_r() or gmtime_r() or relatives. They all
fiddle with global variables and locking in various libcs. They cause deadlocks
when liblxc is used multi-threaded and no matter how smart you think you are,
you __will__ cause trouble using them.
(As a short example how this can cause trouble: LXD uses forkstart to fork off
a new process that runs the container. At the same time the go runtime LXD
relies on does its own multi-threading thing which we can't control. The
fork()ing + threading then seems to mess with the locking states in these time
functions causing deadlocks.)
The current solution is to be good old unix people and use the Epoch as our
reference point and simply use the seconds and nanoseconds that have past since
then. This relies on clock_gettime() which is explicitly marked MT-Safe with no
restrictions! This way, anyone who is really strongly invested in getting the
actual time the log entry was created, can just convert it for themselves. Our
logging is mostly done for debugging purposes so don't try to make it pretty.
Pretty might cost you thread-safety.

### DIFF
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -22,9 +22,12 @@
  */
 
 #define _GNU_SOURCE
+#define __STDC_FORMAT_MACROS /* Required for PRIu64 to work. */
 #include <assert.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <errno.h>
+#include <inttypes.h>
 #include <limits.h>
 #include <unistd.h>
 #include <sys/types.h>
@@ -138,6 +141,44 @@ static int log_append_stderr(const struct lxc_log_appender *appender,
 }
 
 /*---------------------------------------------------------------------------*/
+int lxc_unix_epoch_to_utc(char *buf, size_t bufsize, const struct timespec *time)
+{
+	int64_t days, z, era, doe, yoe, year, doy, mp, day, month, d_in_s,
+	    hours, h_in_s, minutes, seconds;
+	char nanosec[__LXC_NUMSTRLEN];
+	int ret;
+
+	/* See https://howardhinnant.github.io/date_algorithms.html for an
+	 * explanation of the algorithm used here.
+	 */
+	days = time->tv_sec / 86400;
+	z = days + 719468;
+	era = (z >= 0 ? z : z - 146096) / 146097;
+	doe = (z - era * 146097);
+	yoe = (doe - doe / 1460 + doe / 36524 - doe / 146096) / 365;
+	year = (yoe) + era * 400;
+	doy = doe - (365 * yoe + yoe / 4 - yoe / 100);
+	mp = (5 * doy + 2) / 153;
+	day = doy - (153 * mp + 2) / 5 + 1;
+	month = mp + (mp < 10 ? 3 : -9);
+	d_in_s = days * 86400;
+	hours = (time->tv_sec - d_in_s) / 3600;
+	h_in_s = hours * 3600;
+	minutes = (time->tv_sec - d_in_s - h_in_s) / 60;
+	seconds = (time->tv_sec - d_in_s - h_in_s - (minutes * 60));
+
+	ret = snprintf(nanosec, __LXC_NUMSTRLEN, "%ld", time->tv_nsec);
+	if (ret < 0 || ret >= __LXC_NUMSTRLEN)
+		return -1;
+	ret = snprintf(buf, bufsize, "%" PRId64 "%" PRId64 "%" PRId64 "%" PRId64
+				     "%" PRId64 "%" PRId64 ".%.3s",
+		       year, month, day, hours, minutes, seconds, nanosec);
+	if (ret < 0 || (size_t)ret >= bufsize)
+		return -1;
+
+	return 0;
+}
+
 /* This function needs to make extra sure that it is thread-safe. We had some
  * problems with that before. This especially involves time-conversion
  * functions. I don't want to find any localtime() or gmtime() functions or
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -203,10 +203,9 @@ static int log_append_logfile(const struct lxc_log_appender *appender,
 			      struct lxc_log_event *event)
 {
 	char buffer[LXC_LOG_BUFFER_SIZE];
-	struct timespec zero_timespec = {0};
-	int ret, n;
+	char date_time[LXC_LOG_TIME_SIZE];
+	int n;
 	int fd_to_use = -1;
-	char nanosec[__LXC_NUMSTRLEN];
 
 #ifndef NO_LXC_CONF
 	if (!lxc_log_use_global_fd && current_config)
@@ -219,19 +218,15 @@ static int log_append_logfile(const struct lxc_log_appender *appender,
 	if (fd_to_use == -1)
 		return 0;
 
-	ret = snprintf(nanosec, __LXC_NUMSTRLEN, "%ld", event->timestamp.tv_nsec);
-	if (ret < 0 || ret >= LXC_LOG_TIME_SIZE)
+	if (lxc_unix_epoch_to_utc(date_time, LXC_LOG_TIME_SIZE, &event->timestamp) < 0)
 		return 0;
 
 	n = snprintf(buffer, sizeof(buffer),
-			"%15s%s%s [ %.0f.%.6s ] %-8s %s - %s:%s:%d - ",
+			"%15s%s%s %s %-8s %s - %s:%s:%d - ",
 			log_prefix,
 			log_vmname ? " " : "",
 			log_vmname ? log_vmname : "",
-			/* sec_since_epoch: Safely convert time_t type to
-			 * double. */
-			difftime(event->timestamp.tv_sec, zero_timespec.tv_sec),
-			nanosec,
+			date_time,
 			lxc_log_priority_to_string(event->priority),
 			event->category,
 			event->locinfo->file, event->locinfo->func,
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -143,33 +143,87 @@ static int log_append_stderr(const struct lxc_log_appender *appender,
 /*---------------------------------------------------------------------------*/
 int lxc_unix_epoch_to_utc(char *buf, size_t bufsize, const struct timespec *time)
 {
-	int64_t days, z, era, doe, yoe, year, doy, mp, day, month, d_in_s,
-	    hours, h_in_s, minutes, seconds;
+	int64_t epoch_to_days, z, era, doe, yoe, year, doy, mp, day, month,
+	    d_in_s, hours, h_in_s, minutes, seconds;
 	char nanosec[__LXC_NUMSTRLEN];
 	int ret;
 
 	/* See https://howardhinnant.github.io/date_algorithms.html for an
 	 * explanation of the algorithm used here.
 	 */
-	days = time->tv_sec / 86400;
-	z = days + 719468;
+
+	/* Convert Epoch in seconds to number of days. */
+	epoch_to_days = time->tv_sec / 86400;
+
+	/* Shift the Epoch from 1970-01-01 to 0000-03-01. */
+	z = epoch_to_days + 719468;
+
+	/* compute the era from the serial date by simply dividing by the number
+	 * of days in an era (146097).
+	 */
 	era = (z >= 0 ? z : z - 146096) / 146097;
+
+	/* The day-of-era (doe) can then be found by subtracting the era number
+	 * times the number of days per era, from the serial date.
+	 */
 	doe = (z - era * 146097);
+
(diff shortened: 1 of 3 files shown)


brauner:
Sample log output after the change:

```
      lxc-start yak [ 1480287447.412928 ] INFO     lxc_conf - conf.c:run_script_argv:418 - Executing script "/usr/share/lxcfs/lxc.mount.hook" for container "yak", config section "lxc".
      lxc-start yak [ 1480287447.460716 ] INFO     lxc_conf - conf.c:fill_autodev:1193 - Creating initial consoles under container /dev
      lxc-start yak [ 1480287447.460756 ] INFO     lxc_conf - conf.c:fill_autodev:1204 - Populating container /dev
      lxc-start yak [ 1480287447.460797 ] INFO     lxc_conf - conf.c:fill_autodev:1241 - Populated container /dev
      lxc-start yak [ 1480287447.460865 ] INFO     lxc_conf - conf.c:setup_dev_console:1489 - console has been setup
      lxc-start yak [ 1480287447.460910 ] INFO     lxc_utils - utils.c:mount_proc_if_needed:1780 - I am 1, /proc/self points to '1'
      lxc-start yak [ 1480287447.478652 ] DEBUG    lxc_conf - conf.c:setup_rootfs_pivot_root:1102 - pivot_root syscall to '/usr/lib/x86_64-linux-gnu/lxc' successful
      lxc-start yak [ 1480287447.478972 ] DEBUG    lxc_conf - conf.c:lxc_create_tty:3421 - allocated pty '/dev/pts/0' (9/12)
      lxc-start yak [ 1480287447.479041 ] DEBUG    lxc_conf - conf.c:lxc_create_tty:3421 - allocated pty '/dev/pts/1' (14/15)
      lxc-start yak [ 1480287447.479097 ] DEBUG    lxc_conf - conf.c:lxc_create_tty:3421 - allocated pty '/dev/pts/2' (16/17)
      lxc-start yak [ 1480287447.479149 ] DEBUG    lxc_conf - conf.c:lxc_create_tty:3421 - allocated pty '/dev/pts/3' (18/19)
      lxc-start yak [ 1480287447.479163 ] INFO     lxc_conf - conf.c:lxc_create_tty:3432 - tty's configured
      lxc-start yak [ 1480287447.479491 ] INFO     lxc_conf - conf.c:setup_tty:1047 - 4 tty(s) has been setup
      lxc-start yak [ 1480287447.479523 ] INFO     lxc_conf - conf.c:setup_personality:1445 - set personality to '0x0'
      lxc-start yak [ 1480287447.479536 ] DEBUG    lxc_conf - conf.c:setup_caps:2128 - drop capability 'mac_admin' (33)
      lxc-start yak [ 1480287447.479548 ] DEBUG    lxc_conf - conf.c:setup_caps:2128 - drop capability 'mac_override' (32)
      lxc-start yak [ 1480287447.479558 ] DEBUG    lxc_conf - conf.c:setup_caps:2128 - drop capability 'sys_time' (25)
      lxc-start yak [ 1480287447.479568 ] DEBUG    lxc_conf - conf.c:setup_caps:2128 - drop capability 'sys_module' (16)
      lxc-start yak [ 1480287447.479577 ] DEBUG    lxc_conf - conf.c:setup_caps:2128 - drop capability 'sys_rawio' (17)
      lxc-start yak [ 1480287447.479586 ] DEBUG    lxc_conf - conf.c:setup_caps:2137 - capabilities have been setup
      lxc-start yak [ 1480287447.479594 ] NOTICE   lxc_conf - conf.c:lxc_setup:3969 - 'yak' is setup.
      lxc-start yak [ 1480287447.479767 ] INFO     lxc_start - start.c:do_start:865 - Unshared CLONE_NEWCGROUP.
      lxc-start yak [ 1480287447.479882 ] INFO     lxc_apparmor - lsm/apparmor.c:apparmor_process_label_set:238 - changed apparmor profile to lxc-container-default-cgns
      lxc-start yak [ 1480287447.480283 ] NOTICE   lxc_start - start.c:start:1445 - Exec'ing "/sbin/init".
      lxc-start yak [ 1480287447.524969 ] NOTICE   lxc_start - start.c:post_start:1456 - Started "/sbin/init" with pid "29710".
      lxc-start yak [ 1480287447.525074 ] WARN     lxc_start - start.c:signal_handler:336 - Invalid pid for SIGCHLD. Received pid 29702, expected pid 29710.
      lxc-start yak [ 1480287447.525442 ] DEBUG    lxc_commands - commands.c:lxc_cmd_handler:893 - Peer has disconnected.
      lxc-start yak [ 1480287447.525520 ] DEBUG    lxc_commands - commands.c:lxc_cmd_handler:893 - Peer has disconnected.
      lxc-start yak [ 1480287447.525521 ] DEBUG    lxc_commands - commands.c:lxc_cmd_get_state:578 - Container "yak" is in "RUNNING" state.
```

hallyn:
On Sun, Nov 27, 2016 at 02:58:41PM -0800, Christian Brauner wrote:
> Sample log output after the change:
> 
> ```
>       lxc-start yak [ 1480287447.412928 ] INFO     lxc_conf - conf.c:run_script_argv:418 - Executing script "/usr/share/lxcfs/lxc.mount.hook" for container "yak", config section "lxc".

Hi,

What about having lxc_log_init() define starttime, (maybe just
noting the tv_sec), and then showing the difftime?  Maybe we lose
info when comparing two containers, but spitting out a message at
end of the first lxc_log_init() should suffice for that.

src/lxc/log.c
```diff
@@ -148,15 +138,34 @@ static int log_append_stderr(const struct lxc_log_appender *appender,
...
 {
-	char date[LXC_LOG_DATEFOMAT_SIZE] = "N/A";
 	char buffer[LXC_LOG_BUFFER_SIZE];
-	struct tm newtime;
-	int n;
-	int ms = 0;
+	struct timespec zero_timespec = {0};
+	int ret, n;
 	int fd_to_use = -1;
+	char nanosec_since_epoch[__LXC_NUMSTRLEN];
```
hallyn: Is this variable name correct?

hallyn:
Looks good to me.  I understand @stgraber may object, so will wait to let him merge or object.

src/lxc/log.c
brauner: Well strictly speaking, no. It should simply be `nanosec`. It's obvs. not all nanoseconds since `1970`. (Though that would be impressive. :))

stgraber:
So is it thread safe to use ctime()? if so, it'd be more acceptable to land a change to switch from localtime to UTC time, with a clear mention of UTC in the timestamp.

That's something I'd probably be fine backporting to stable, whereas the change above ^ will break anyone who currently attempts at parsing the log.

brauner:
`ctime()` and `ctime_r()` will have the same issue.

brauner:
So all of them:

```
       ┌───────────────┬───────────────┬─────────────────────────────────┐
       │Interface      │ Attribute     │ Value                           │
       ├───────────────┼───────────────┼─────────────────────────────────┤
       │asctime()      │ Thread safety │ MT-Unsafe race:asctime locale   │
       ├───────────────┼───────────────┼─────────────────────────────────┤
       │asctime_r()    │ Thread safety │ MT-Safe locale                  │
       ├───────────────┼───────────────┼─────────────────────────────────┤
       │ctime()        │ Thread safety │ MT-Unsafe race:tmbuf            │
       │               │               │ race:asctime env locale         │
       ├───────────────┼───────────────┼─────────────────────────────────┤
       │ctime_r(),     │ Thread safety │ MT-Safe env locale              │
       │gmtime_r(),    │               │                                 │
       │localtime_r(), │               │                                 │
       │mktime()       │               │                                 │
       ├───────────────┼───────────────┼─────────────────────────────────┤
       │gmtime(),      │ Thread safety │ MT-Unsafe race:tmbuf env locale │
       │localtime()    │               │                                 │
       └───────────────┴───────────────┴─────────────────────────────────┘
```

internally might call `tzset()` or `__tzconvert()` which use global locking on a whole file in `glibc` and this is messing us up afaict.

stgraber:
So there is no way to get something you can feed strftime from inside a thread?

brauner:
@stgraber, if your concern is with the the length of the date field only, we can just log the seconds and give it a maximum field with of `15` chars that should last us for a couple of hundred maybe thousand years before that field overflows. A log entry then would look like this:

```
 lxc-start yak 1480287447      WARN     lxc_start - start.c:signal_handler:336 - Invalid pid for SIGCHLD. Received pid 29702, expected pid 29710.
```

brauner:
@stgraber, well even `strftime()` though noted as `MT-Safe` has restrictions on thread-safety in terms of accessing the `locale` and `env`:

```
       ┌───────────┬───────────────┬────────────────────┐
       │Interface  │ Attribute     │ Value              │
       ├───────────┼───────────────┼────────────────────┤
       │strftime() │ Thread safety │ MT-Safe env locale │
       └───────────┴───────────────┴────────────────────┘
```

And though `POSIX` notes that it should usually be safe, I fear that our specific use-case in `LXD` might still cause weird, though rare scenarios where we might end up deadlocked. Maybe I'm being to bleak here but I much rather go down the safe route.

stgraber:
No, my concern is that we have folks who run log parsers against our log files, which will explode if they get a unix epoch instead of a formated date string.

Having the timestamp change from local time to UTC time isn't a huge deal because those parsers will still work, the sysadmin may just need to change the offset at some point if they care about merging logs, but at least things wouldn't break.

brauner:
So e.g. in `glibc` `ctime_r()` is just:

```
char *
ctime_r (const time_t *t, char *buf)
{
  struct tm tm;
  return __asctime_r (__localtime_r (t, &tm), buf);
}
```

brauner:
We can maybe try to do our own calculations based on the unix Epoch timestamp.

stgraber:
You'd think there should be a function somewhere to convert an epoch to its string representation without poking at locales or timezone...

brauner:
@stgraber, I added our own UTC conversion function. Here's the new log output:

```      lxc-start yak 20161128212530.440 INFO     lxc_cgroup - cgroups/cgroup.c:cgroup_init:68 - cgroup driver cgroupfs-ng initing for yak
      lxc-start yak 20161128212530.440 INFO     lxc_start - start.c:lxc_spawn:1171 - Cloned CLONE_NEWUSER.
      lxc-start yak 20161128212530.440 INFO     lxc_start - start.c:lxc_spawn:1171 - Cloned CLONE_NEWNS.
      lxc-start yak 20161128212530.441 INFO     lxc_start - start.c:lxc_spawn:1171 - Cloned CLONE_NEWPID.
      lxc-start yak 20161128212530.441 INFO     lxc_start - start.c:lxc_spawn:1171 - Cloned CLONE_NEWUTS.
      lxc-start yak 20161128212530.441 INFO     lxc_start - start.c:lxc_spawn:1171 - Cloned CLONE_NEWIPC.
      lxc-start yak 20161128212530.448 INFO     lxc_start - start.c:do_start:759 - Unshared CLONE_NEWNET.
      lxc-start yak 20161128212530.535 NOTICE   lxc_start - start.c:do_start:785 - Switching to uid=0 and gid=0 in new user namespace.
      lxc-start yak 20161128212530.535 DEBUG    lxc_conf - conf.c:setup_rootfs:1267 - mounted '/home/chb/.local/share/lxc/yak/rootfs' on '/usr/lib/x86_64-linux-gnu/lxc'
      lxc-start yak 20161128212530.535 INFO     lxc_conf - conf.c:setup_utsname:895 - 'yak' hostname has been setup
      lxc-start yak 20161128212530.536 DEBUG    lxc_conf - conf.c:setup_hw_addr:2217 - mac address '00:16:3e:c4:9c:6f' on 'eth0' has been setup
      lxc-start yak 20161128212530.536 DEBUG    lxc_conf - conf.c:setup_netdev:2444 - 'eth0' has been setup
      lxc-start yak 20161128212530.536 INFO     lxc_conf - conf.c:setup_network:2465 - network has been setup
      lxc-start yak 20161128212530.536 INFO     lxc_conf - conf.c:mount_autodev:1124 - Mounting container /dev
      lxc-start yak 20161128212530.536 INFO     lxc_conf - conf.c:mount_autodev:1147 - Mounted tmpfs onto /usr/lib/x86_64-linux-gnu/lxc/dev
      lxc-start yak 20161128212530.536 INFO     lxc_conf - conf.c:mount_autodev:1165 - Mounted container /dev
```

stgraber:
Cool. That seems acceptable to me. @hallyn can you take a look at the code?

brauner:
This is again related to fixing https://github.com/lxc/lxd/issues/2189.

PR #1338, opened 2016-12-02
tools: lxc-start: set configfile after load_config

Blub:
Same change as in 6118210e0a which was missing in lxc-start
and back then is_defined() wasn't being called.


brauner:
jenkins: test this please

PR #1340, opened 2016-12-02
conf: clearly report to either use drop or keep

brauner:


brauner:
See https://github.com/lxc/lxd/issues/2673.

PR #1343, opened 2016-12-07
confile: support the network link string pattern matching

lifupan:
Enable lxc network config support the following type and link:

lxc.network.type = phys
lxc.network.link = eth*

Thus, when lxc find any network interfaces name prefixed
with "eth" such as "eth0", "eth1" and so on, it will
try to move them the container's namespace; If it didn't
find any matching, it would do nothing for this configure
line.

Signed-off-by: fli <fupan.li@windriver.com>

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -50,6 +50,12 @@
 #include "network.h"
 #include "lxcseccomp.h"
 
+#if HAVE_IFADDRS_H
+#include <ifaddrs.h>
+#else
+#include <../include/ifaddrs.h>
+#endif
+
 #if HAVE_SYS_PERSONALITY_H
 #include <sys/personality.h>
 #endif
@@ -672,16 +678,85 @@ static int config_network_flags(const char *key, const char *value,
 	return 0;
 }
 
+static int set_network_link(const char *key, const char *value, struct lxc_conf *lxc_conf)
+{
+        struct lxc_netdev *netdev;
+
+        netdev = network_netdev(key, value, &lxc_conf->network);
+        if (!netdev)
+                return -1;
+
+        return network_ifname(&netdev->link, value);
+}
+
+static int create_matched_ifnames(const char *value, struct lxc_conf *lxc_conf)
+{
+	struct ifaddrs *ifaddr, *ifa;
+	const char *type_key = "lxc.network.type";
+	const char *link_key = "lxc.network.link";
+	const char *tmpvalue = "phys";
+	int n, ret = 0;
+
+	if (getifaddrs(&ifaddr) == -1) {
+		SYSERROR("Get network interfaces failed");
+		return -1;
+	}
+
+	for (ifa = ifaddr, n = 0; ifa != NULL; ifa = ifa->ifa_next, n++) {
+		if (!ifa->ifa_addr)
+			continue;
+		if (ifa->ifa_addr->sa_family != AF_PACKET)
+			continue;
+
+		if (!strncmp(value, ifa->ifa_name, strlen(value)-1)) {
+			ret = config_network_type(type_key, tmpvalue, lxc_conf);
+			if (!ret) {
+				ret = set_network_link(link_key, ifa->ifa_name, lxc_conf);
+				if (ret) {
+					ERROR("failed to create matched ifnames");
+					break;
+				}
+			} else {
+				ERROR("failed to create matched ifnames");
+				break;
+			}
+		}
+	}
+
+	freeifaddrs(ifaddr); /* free the dynamic memory */
+	ifaddr = NULL;	    /* prevent use after free */
+	
+	return ret;
+}
+
 static int config_network_link(const char *key, const char *value,
 			       struct lxc_conf *lxc_conf)
 {
 	struct lxc_netdev *netdev;
+	struct lxc_list * it;
+	int ret = 0;
 
 	netdev = network_netdev(key, value, &lxc_conf->network);
 	if (!netdev)
 		return -1;
 
-	return network_ifname(&netdev->link, value);
+	if (value[strlen(value) - 1] == '+' && netdev->type == LXC_NET_PHYS) {
+		//get the last network list and remove it.
+		it = lxc_conf->network.prev;
+		if (((struct lxc_netdev *)(it->elem))->type != LXC_NET_PHYS) {
+			ERROR("lxc config cannot support string pattern matching for this link type");
+			return -1;
+		}
+
+		lxc_list_del(it);
+		free(it);
+		ret = create_matched_ifnames(value, lxc_conf);
+
+	} else {
+		ret = network_ifname(&netdev->link, value);
+	}
+
+	return ret;
 }
 
 static int config_network_name(const char *key, const char *value,


stgraber:
You can't glob network interfaces as you can have a "*" in an actual interface name:

root@castiana:~# ip link add eth* type dummy
root@castiana:~# ifconfig eth*
eth*: flags=130<BROADCAST,NOARP>  mtu 1500
        ether a2:40:8b:92:08:01  txqueuelen 1000  (Ethernet)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 0  bytes 0 (0.0 B)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lifupan:
Hi, Stgraber

For this case, how about the following  solution:

Support the escape characters of "\\" and
"\*"  when they appeared at the end of the value string.

Such as "eth\*" to match the actual interface name "eth*",
and "eth\\\*" to match the actual interface name "eth\*",
and so on.  If they appeared in the middle if the string,
such as "et\*h", it mean the literal meaning of itself.

Thanks!

Fupan

stgraber:
That'd regress any user which uses such interface names today, so no.

lifupan:
Hmm, since the kernel accept any character as an interface name, 
such as "eth\*", so if I change it to "\*" as an string pattern matching trigger,
than it  cannot glob the actual interface name "eth\*", and need to reset the key
value to "eth\\*". 

In one word, it's impossible to not to regress any users with supporting this string pattern matching.

So, this feature wouldn't be accepted, right? 

Fupan

stgraber:
So I think the least bad suffix to use is "+". The kernel does let you have it in an interface name, but it's already used as a wildcard character in iptables so anyone using this will already be running into problems.

lifupan:
That's great.

I'll change it to '+' and send a new pull.

Thanks very much!

Fupan

lifupan:
Hi, stgraber

I had updated the commit, can you review this pull again?

Thanks!

Fupan

stgraber:
jenkins: test this please

src/lxc/confile.c
```diff
@@ -115,6 +121,7 @@ static int config_init_uid(const char *, const char *, struct lxc_conf *);
 static int config_init_gid(const char *, const char *, struct lxc_conf *);
 static int config_ephemeral(const char *, const char *, struct lxc_conf *);
 static int config_no_new_privs(const char *, const char *, struct lxc_conf *);
+static int _config_network_link(const char *key, const char *value,struct lxc_conf *lxc_conf);
```
brauner: I think the underscore is unnecessary. Just name it `config_network_link()`

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+	const char *tmpvalue = "phys";
+	int n, ret = 0;
+
+	if (getifaddrs(&ifaddr) == -1) {
+		SYSERROR("Get network interfaces failed");
+		return -1;
+	}
+
+	for (ifa = ifaddr, n = 0; ifa != NULL; ifa = ifa->ifa_next, n++)
+	{
```
brauner: bracket should go on the same line as the `for`.

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+		if (!ifa->ifa_addr)
+			continue;
+		if(ifa->ifa_addr->sa_family != AF_PACKET)
+			continue;
+
+		if(!strncmp(value, ifa->ifa_name, strlen(value)-1))
+		{
```
brauner: bracket should go on the same line as the `if`.

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+			ret = config_network_type(type_key, tmpvalue, lxc_conf);
+			if (!ret)
+			{
```
brauner: bracket should go on the same line as the `if`.

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+				ret = _config_network_link(link_key, ifa->ifa_name, lxc_conf);
+				if (ret){
```
brauner: Please add a space between closing round bracket and curly bracket.

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+					ERROR("failed to create mached ifnames");
+					break;
+				}
+			}else
```
brauner: please add a space between the curly bracket and `else`

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+			{
```
brauner: bracket should go on the same line as `else`

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+	struct lxc_list * it;
+	int ret = 0;
 
 	netdev = network_netdev(key, value, &lxc_conf->network);
 	if (!netdev)
 		return -1;
 
-	return network_ifname(&netdev->link, value);
+	if (value[strlen(value) - 1] == '+' && netdev->type == LXC_NET_PHYS)
+	{
```
brauner: bracket should go on the same line as `if`

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+		//get the last network list and remove it.
+		it = lxc_conf->network.prev;
+		if (((struct lxc_netdev *)(it->elem))->type != LXC_NET_PHYS)
+		{
```
brauner: bracket should go on the same line as `if`

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+			ERROR("lxc config cannot support string pattern matching for this link type");
+			return -1;
+		}
+		lxc_list_del(it);
+		free(it);
+
+		ret = create_matched_ifnames(value, lxc_conf);
+	}else
```
brauner: please add a space between curly bracket and `else`

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
...
+		ret = network_ifname(&netdev->link, value);
```
brauner: please enclose this statement in curly brackets as well

src/lxc/confile.c
```diff
@@ -672,16 +679,88 @@ static int config_network_flags(const char *key, const char *value,
 	return 0;
 }
 
+static int create_matched_ifnames(const char *value,
```
brauner: leave the function header on one line in this case

lifupan:
Hi, Brauner

I had addressed all of your comments and please have a look again. 

Thanks!

Fupan

src/lxc/confile.c
```diff
@@ -672,16 +678,82 @@ static int config_network_flags(const char *key, const char *value,
...
+	int ret = 0;
 
 	netdev = network_netdev(key, value, &lxc_conf->network);
 	if (!netdev)
 		return -1;
 
-	return network_ifname(&netdev->link, value);
+	if (value[strlen(value) - 1] == '+' && netdev->type == LXC_NET_PHYS) {
+		//get the last network list and remove it.
+		it = lxc_conf->network.prev;
```
brauner: I assume you want to remove the e.g. `eth+` entry here? But why are you using the previous pointer of the list here? Wouldn't you be removing the previous entry and not the `eth+` entry or am I misreading?

src/lxc/confile.c
```diff
@@ -672,16 +678,82 @@ static int config_network_flags(const char *key, const char *value,
...
+
+static int create_matched_ifnames(const char *value, struct lxc_conf *lxc_conf)
+{
+	struct ifaddrs *ifaddr, *ifa;
+	const char *type_key = "lxc.network.type";
+	const char *link_key = "lxc.network.link";
+	const char *tmpvalue = "phys";
+	int n, ret = 0;
+
+	if (getifaddrs(&ifaddr) == -1) {
```
brauner: Is there a reason why you're not calling `freeifaddrs()`?

src/lxc/confile.c
lifupan: In fact I want to remove the network list created by the configure line of "lxc.network.type = phys" which followed by
a configure line of "lxc.network.link = xxx+".  As you know, the configure line  lxc.network.link will not create any netdev instance, in fact, it only set the netdev's "link" field created by the previous line of "lxc.network.type".  

Here  I removed it just in case it didn't find any matched network interface. Once it find one matched interface, it will
recreate one netdev instance by the following call in function create_matched_ifnames

712 +			ret = config_network_type(type_key, tmpvalue, lxc_conf);

and set the corresponding netdev link by the call of:

714  +				ret = set_network_link(link_key, ifa->ifa_name, lxc_conf);


Fupan

hallyn:
How is this intended to work with hotplug?  This might be a corner case we can ignore (well, document), but it looks like if I do

c = lxc_container_new(...)
<wait 10 minutes; during which two nics I wanted are hotplugged>
c.start()

I won't see the new ones?

lifupan:
@hallyn 

Since the call chain lxc_container_new->lxcapi_load_config had already do the network interfaces scan and match, 
and thus the network devices added   after that call will not be detected, that's why you wouldn't see them in the containers.

So, the watershed is the time calling lxcapi_load_config, the networks hotplugged before that time, it would be detected and added to the container, otherwise, it wouldn't be added to the container.

Fupan

hallyn:
On Thu, Dec 15, 2016 at 12:33:46AM -0800, fli wrote:
> @hallyn 
> 
> Since the call chain lxc_container_new->lxcapi_load_config had already do the network interfaces scan and match, 
> and thus the network devices added   after that call will not be detected, that's why you wouldn't see them in the containers.
> 
> So, the watershed is the time calling lxcapi_load_config,

And is that good enough?

>  the networks hotplugged before that time, it would be detected and added to the container, otherwise, it wouldn't be added to the container.

lifupan:
@hallyn

Hmm, I don't think so. 
For hotplug, I think there should be a notification mechanism instead
of the one time static scanning. 

Fupan

hallyn:
Quoting fli (notifications@github.com):
> @hallyn
> 
> Hmm, I don't think so. 
> For hotplug, I think there should be a notification mechanism instead
> of the one time static scanning. 

That's doable for lxd but not lxc.

But I see no good solution, and no problem with this patch, so merging.

Thanks!

PR #1351, opened 2016-12-10
conf, attach: save errno across call to close

Blub:
Save errno across some calls to close() since it can be
interrupted.

Signed-off-by: Wolfgang Bumiller <wry.git@bumiller.com>


brauner:
jenkins: test this please

PR #1360, opened 2016-12-18
Add --enable-gnutls option

ffontaine:
Previously HAVE_LIBGNUTLS was never set in config.h even if gnutls was
detected as AC_CHECK_LIB default action-if-found was overriden by
enable_gnutls=yes
This patch adds an --enable-gnutls option and will call AC_CHECK_LIB
with the default action to write HAVE_LIBGNUTLS in config.h

Signed-off-by: Fabrice Fontaine <fabrice.fontaine@orange.com>


stgraber:
jenkins: test this please

Issue #1356, opened 2016-12-14
Fedora template maintenance


ganto:
I lately put some [effort](https://copr.fedorainfracloud.org/coprs/ganto/lxd/) in creating community packages for LXD on Fedora and therefore I also had a look at the `lxc-fedora` template in LXC. It's a very feature rich template that still works fine also with newer Fedora releases but it hasn't seen much love lately (maybe because of its complexity?).

To get know the template a bit better I ported the bootstrapping part for non-yum distributions from the currently used Fedora 20 squashfs image to the Fedora 24 image and therefore was able to get rid of some complexity (see my [fedora](https://github.com/ganto/lxc/tree/fedora) branch). By dropping `python` (`dnf` now requires `python-3`), `yum`, `rsyslog` and `chkconfig`, I was able to reduce the image size from 230MB to around 160MB, what I think is a big win.

However, the shift to `dnf` made me think of some issues with the current code paths that I would like to address here:

* Is it worth to put more effort in updating the `lxc-fedora` template or is this something that might not be widely used anymore with the availability of the Fedora [CloudImages](http://archives.fedoraproject.org/pub/fedora/linux/releases/24/CloudImages/x86_64/images/)?

 * Are there users which use this template to build containers of unsupported versions (non-`dnf`, non-`systemd`) of Fedora?

* Getting rid of all this compatibility code for old Fedora versions would simplify the template quite a bit. Would you be interested if I put some effort in doing this?

Please let me know what you think.

brauner:
I would vote for getting rid of the complexity. (@stgraber)

stgraber:
1) I don't think too many people use the fedora cloud images since there's no straightforward way to turn those into LXC containers (no template script). I'm not opposed changing lxc-fedora to just repack those images though.
2) It's hard to know what our users do, but I'd say it's certainly not worth keeping a lot of complexity for such corner cases.
3) Assuming that "old" here means no longer supported, I'm absolutely fine with removing that code.

ganto:
Ok, thanks for your reply. I will work on a suggestion for a new template in the next few days and maybe also try out how this would work when converting a CloudImage.

brauner:
The template has been completely reworked and merged. This can be closed.

PR #1364, opened 2016-12-24
[RFC] New fedora-cloud container template

ganto:
This is a major adjustment to the Fedora template to bootstrap from the Fedora Cloud image instead of a `dnf` chroot installation. This was created in response to #1356.

Changes in comparison with the current Fedora template:

**Container:**
* only support Fedora >=24
* use `systemd-networkd`/`systemd-resolved` instead of traditional network init
* don't set hostname to FQDN by default (adjustment to other templates)
* start `cloud-init` by default

**Template Options:**
* drop support for `--arch`. So far the Fedora Cloud Image is only available on x86_64
* preserve CLI compatibility with `lxc-fedora` template
* add `--mirror` option for setting custom download mirror
* add `--debug` for verbose script output

**Dependencies:**
* additionally use `mount` and `unxz` helper applications during the bootstrap
* drop `rsync` dependency

For simplicity I didn't add the checks for the required host binaries yet and ignored the btrfs handling found in other templates as I'm not using btrfs. Also some other smaller enhancements might still be required. 

As you can see, it's pretty simple but lost a lot of flexibility compared with the `lxc-fedora` template. Before I put more effort into it, what do you think? I guess the work should rather go into cleaning up `lxc-fedora`?

### DIFF
--- a/templates/lxc-fedora-cloud.in
+++ b/templates/lxc-fedora-cloud.in
@@ -1,0 +1,513 @@
+#!/bin/bash
+
+#
+# template script for generating Fedora Cloud container for LXC
+#
+
+#
+# lxc: linux Container library
+
+# Authors:
+# Daniel Lezcano <daniel.lezcano@free.fr>
+# Ramez Hanna <rhanna@informatiq.org>
+# Michael H. Warfield <mhw@WittsEnd.com>
+# Reto Gantenbein <reto.gantenbein@linuxmonk.ch>
+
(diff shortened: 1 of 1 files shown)


brauner:
Thanks for the new template. :) You still need to sign-off your commit though. We follow-this kernel-style tradition. :)

brauner:
Closing in favor of https://github.com/lxc/lxc/pull/1371.

PR #1365, opened 2016-12-25
tests: skip unpriv tests on broken overlay module

brauner:
This mainly affects Trusty. The 3.13 kernel has a broken overlay module which
does not handle symlinks correctly. This is a problem for containers that use
an overlay based rootfs since `safe_mount()` uses `/proc/<pid>/fd/<fd-number>` in
its calls to `mount()`.

### DIFF
--- a/src/tests/lxc-test-unpriv
+++ b/src/tests/lxc-test-unpriv
@@ -27,6 +27,41 @@ if [ $(id -u) -ne 0 ]; then
 	echo "ERROR: Must run as root."
 	exit 1
 fi
+
+# Test if we're using an overlayfs module that handles symlinks correctly. If
+# not, we skip these tests since overlay clones will not work correctly.
+if modprobe -q overlayfs; then
+        TMPDIR=$(mktemp -d)
+
+        MOUNTDIR="${TMPDIR}/ovl_symlink_test"
+
+        mkdir ${MOUNTDIR}
+
+        mount -t tmpfs none ${MOUNTDIR}
+
+        mkdir "${MOUNTDIR}/lowerdir" "${MOUNTDIR}/upperdir" "${MOUNTDIR}/overlayfs"
+        mount -t overlayfs -o lowerdir="${MOUNTDIR}/lowerdir",upperdir="${MOUNTDIR}/upperdir" none "${MOUNTDIR}/overlayfs"
+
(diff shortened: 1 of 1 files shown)


src/tests/lxc-test-unpriv
```diff
@@ -27,6 +27,43 @@ if [ $(id -u) -ne 0 ]; then
 	echo "ERROR: Must run as root."
```
stgraber: Since the script must run as root, there's no need to use sudo below.

brauner: Addressed.

src/tests/lxc-test-unpriv
```diff
@@ -27,6 +27,42 @@ if [ $(id -u) -ne 0 ]; then
...
 	exit 1
 fi
+
+# Test if we're using an overlayfs module that handles symlinks correctly. If
+# not, we skip these tests since overlay clones will not work correctly.
+OVERLAYFS=$(modprobe -q overlayfs)
```
stgraber: You don't actually use ${OVERLAYFS} anywhere below, so you probably should just make this:

    if modprobe -q overlayfs; then

stgraber:
One minor nit, the rest looks good.

src/tests/lxc-test-unpriv
brauner: Thanks, missed that.

src/tests/lxc-test-unpriv
```diff
@@ -27,6 +27,42 @@ if [ $(id -u) -ne 0 ]; then
 	echo "ERROR: Must run as root."
 	exit 1
 fi
+
+# Test if we're using an overlayfs module that handles symlinks correctly. If
+# not, we skip these tests since overlay clones will not work correctly.
+modprobe -q overlayfs
+if [ "${?}" -eq 0 ]; then
```
stgraber: Why do this instead of just my suggested one line:

    if modprobe -q overlayfs; then

brauner: It just irritated me because in my mind, this line should only trigger when then command returns a non-zero exit status and not, as it actually does, when it returns a zero exit status. In my mind writing `if modprobe -q overlayfs; then` implies `if some command != 0; then` so that the `if` block is hit when `modprobe -q overlayfs` fails while we actually want to hit it when it succeeds. Calling before and explicitly comparing exit status seemed clearer to me. But I'm happy to simplify as well.

Issue #1176, opened 2016-09-08
lxc- start is not working with specific UID and GID map for privilege container


ragavaluvijay:
Hi Everyone ,

I would like to create & start LXC container for specific UID & GID for below purpose.

Login to root user(sudo user) , Do lxc-create of container, then do lxc-start of the container with same root user id. while staring container i need to start all the process inside the container with low privilege(un-priv) user id in which access to the user is restricted to only container and not to host.basically when i say ps -axu in host , i should be able to see the container process with low privilege user id(basically mapped UID/GID).

PS : my host which is VM with ubuntu 16.04 LTS is systemd init system and my container(type busybox) will be started with systemd services . for that i have modified busybox template as below
Code:
`lxc.init_cmd=/lib/systemd/systemd`
and also removed all rcS & inittab related scripts in the busybox default template
and for user id map , i have changed as below in my template
Code:

```
   lxc.id_map=u 0 165536 65536
   lxc.id_map=g 0 165536 65536
```

after that i just followed below commends

Container created normally with below commend (and i cross checked config file at /var/lib/lxc/testecho_uid/config to make sure that all my lxc template changes are reflected or not, found ok)
Command:
`sudo lxc-create -n testecho_uid -t busybox_systemd_uid`

Entered below command to start the container with foreground
Command:
`sudo lxc-start -n testecho_uid -F`

Error is :

>  vijay@vijay-VirtualBox:~$ sudo lxc-start -n testecho_uid -F
>     lxc-start: cgfsng.c: cgfsng_create: 1072 No such file or directory - Failed to create /sys/fs/cgroup/systemd//lxc/testecho_uid: No such file or directory
>     lxc-start: cgfsng.c: cgfsng_create: 1072 No such file or directory - Failed to create /sys/fs/cgroup/systemd//lxc/testecho_uid-1: No such file or directory
>     newuidmap: uid range [0-65536) -> [165536-231072) not allowed
>     lxc-start: start.c: lxc_spawn: 1161 failed to set up id mapping
>     lxc-start: start.c: __lxc_start: 1353 failed to spawn 'testecho_uid'
>     newuidmap: uid range [0-65536) -> [165536-231072) not allowed
>     lxc-start: conf.c: userns_exec_1: 4315 Error setting up child mappings
>     lxc-start: cgfsng.c: recursive_destroy: 983 Error destroying /sys/fs/cgroup/systemd//lxc/testecho_uid-2
[...]

PR #1373, opened 2017-01-02
start: fix execute and improve setgroups() calls

brauner:
lxc_execute() and lxc-execute where broken when a user tried to switch to a
non-root uid/gid. This prevented necessary setup operations like mounting the
rootfs which require root in the user namespace. This commit separates
switching to root in the user namespace from switching to the requested uid/gid
by lxc_execute().
This should be safe: Once we switched to root in the user namespace via
setuid() and then switch to a non-root uid/gid in the user namespace for
lxc_execute() via setuid() we cannot regain root privileges again. So we can
only make us safer (Unless I forget about some very intricate user namespace
nonsense; which is not as unlikely as I try to make it sound.).

Closes https://github.com/lxc/lxc/issues/1176.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -769,32 +769,17 @@ static int do_start(void *data)
 		goto out_warn_father;
 
 	/* If we are in a new user namespace, become root there to have
-	 * privilege over our namespace. When using lxc-execute we default to
-	 * root, but this can be overriden using the lxc.init_uid and
-	 * lxc.init_gid configuration options.
+	 * privilege over our namespace.
 	 */
 	if (!lxc_list_empty(&handler->conf->id_map)) {
-		gid_t new_gid = 0;
-		if (handler->conf->is_execute && handler->conf->init_gid)
-			new_gid = handler->conf->init_gid;
+		if (lxc_switch_uid_gid(0, 0) < 0)
+			goto out_warn_father;
 
-		uid_t new_uid = 0;
-		if (handler->conf->is_execute && handler->conf->init_uid)
-			new_uid = handler->conf->init_uid;
-
-		NOTICE("Switching to uid=%d and gid=%d in new user namespace.", new_uid, new_gid);
-		if (setgid(new_gid)) {
-			SYSERROR("Failed to setgid().");
+		/* Drop groups only after we switched to a valid gid in the new
+		 * user namespace.
+		 */
+		if (lxc_setgroups(0, NULL) < 0)
 			goto out_warn_father;
-		}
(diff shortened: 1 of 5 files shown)


brauner:
@hallyn, if you have a minute. :)

src/lxc/utils.c
```diff
@@ -2048,3 +2049,32 @@ int lxc_safe_long(const char *numstr, long int *converted)
...
+	}
+	NOTICE("Switched to uid %d.", uid);
+
+	return 0;
+}
+
+/* Simple covenience function which enables uniform logging. */
+int lxc_setgroups(int size, gid_t list[])
+{
+	if (setgroups(0, NULL) < 0) {
```
hallyn: While this currently isn't a regression since this is the only way this has been used so far, your function is promising to take a size and list[], but it doesn't use them.

src/lxc/start.c
```diff
@@ -900,6 +885,34 @@ static int do_start(void *data)
...
 	}
 
+	/* The container has been setup. We can now switch to an unprivileged
+	 * uid/gid.
+	 */
+	if (handler->conf->is_execute) {
+		uid_t new_uid = 0;
+		gid_t new_gid = 0;
+
+		if (handler->conf->init_uid > 0)
```
hallyn: uid_t is (according to the headers I'm looking at) unsigned int, so I think you can just do

uid_t new_uid = handler->conf->init_uid;
gid_t new_gid = handler->conf->init_gd;

saving a few lines.

src/lxc/start.c
```diff
@@ -900,6 +885,34 @@ static int do_start(void *data)
...
+
+		if (handler->conf->init_gid > 0)
+			new_gid = handler->conf->init_gid;
+
+		/* If we are in a new user namespace we already dropped all
+		 * groups when we switched to root in the new user namespace
+		 * further above. Only drop groups if we can, so ensure that we
+		 * have necessary privilege.
+		 */
+		bool can_setgroups = ((getuid() == 0) && (getgid() == 0));
```
hallyn: I could be wrong but AFAIK getgid() == 0 is not required.  Only CAP_GID is.

brauner: Addressed.

src/lxc/start.c
brauner: Addressed.

src/lxc/utils.c
brauner: Addressed.

hallyn:
Thanks!

PR #1379, opened 2017-01-04
tools: only check for O_RDONLY

brauner:
On some Android systems the lxc folders where containers are stored might be
read-only and so checking for O_RDWR, will effectively make the tools useless
on these systems, so let's dumb the check down to O_RDONLY.


brauner:
See https://bugs.launchpad.net/ubuntu/+source/lxc-android-config/+bug/1653725.

stgraber:
jenkins: test this please

PR #1381, opened 2017-01-11
tools/lxc-start: remove c->is_defined(c) check

brauner:
We do not check here whether the container is defined, because we support
volatile containers. Which means the container does not need to be created for
it to be started. You can just pass a configuration file as argument and start
the container right away.


brauner:
See the discussion around: https://lists.linuxcontainers.org/pipermail/lxc-users/2017-January/012720.html.

PR #1382, opened 2017-01-12
add cgroupfs-mount to Should-Start/Stop sysvinit LSB headers

evgeni:
otherwise init might try to start the containers before cgroupfs was mounted.

Debian-Bug: https://bugs.debian.org/850212


brauner:
jenkins: test this please

evgeni:
FYI: should-* instead of required-* as $user might have a different way to mount cgroupfs than cgroupfs-mount

PR #1389, opened 2017-01-23
Use external API of bash completion

traylenator:
The _have method of bash completion is internal
and have should not  be used.

In particular when using old versions of bash
completion the _have method does not even
exist.

### DIFF
--- a/config/bash/lxc.in
+++ b/config/bash/lxc.in
@@ -1,4 +1,4 @@
-_have lxc-start && {
+have lxc-start && {
     _lxc_names() {
         COMPREPLY=( $( compgen -W "$( lxc-ls )" "$cur" ) )
     }


thmo:
3db8dd39a797f87f8b348f1b6b44953a25f3f170 claims that `have` had been deprcated in favor of `_have` in newer bash_completion versions.

traylenator:
I agree with your analysis.

thmo:
Maybe this was a misunderstanding - there still is a problem: For bash_completion before version 2, there is no `_have`, so this beats us at least on rhel6. And I think the `_have` function of newer bash_completion versions is also not meant to be generally used. `have` still exists, but is deprecated, because it is "generally not needed with dynamically loaded completions". See e.g. http://stackoverflow.com/a/12874971.

PR #1392, opened 2017-01-26
Make lxc-start-ephemeral Python 3.2-compatible

cjwatson:
On Ubuntu 12.04 LTS with Python 3.2, `lxc-start-ephemeral` breaks as
follows:

    Traceback (most recent call last):
      File "/usr/bin/lxc-start-ephemeral", line 371, in attach_as_user
      File "/usr/lib/python3.2/subprocess.py", line 515, in check_output
      File "/usr/lib/python3.2/subprocess.py", line 732, in __init__
    LookupError: unknown encoding: ANSI_X3.4-1968

This is because `universal_newlines=True` causes `subprocess` to use
`io.TextIOWrapper`, and in versions of Python earlier than 3.3 that
fetched the preferred encoding using `locale.getpreferredencoding()`
rather than `locale.getpreferredencoding(False)`, thereby changing the
locale and causing codecs to be reloaded.  However, `attach_as_user`
runs inside the container and thus can't rely on having access to the
same Python standard library on disk.

The workaround is to decode by hand instead, avoiding the temporary
change of locale.

Signed-off-by: Colin Watson <cjwatson@ubuntu.com>

### DIFF
--- a/src/lxc/tools/lxc-start-ephemeral.in
+++ b/src/lxc/tools/lxc-start-ephemeral.in
@@ -28,6 +28,7 @@
 import argparse
 import gettext
 import lxc
+import locale
 import os
 import sys
 import subprocess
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

PR #1398, opened 2017-01-28
Add LXC template script of Sabayon OS

geaaru:
Hi,
please add lxc-sabayon script for add support to Sabayon containers.
Thanks.
geaaru

### DIFF
--- a/config/templates/Makefile.am
+++ b/config/templates/Makefile.am
@@ -40,4 +40,6 @@ templatesconfig_DATA = \
 	sparclinux.userns.conf \
 	voidlinux.common.conf \
 	voidlinux.userns.conf \
+	sabayon.common.conf \
+	sabayon.userns.conf \
 	userns.conf
--- a/config/templates/sabayon.common.conf.in
+++ b/config/templates/sabayon.common.conf.in
@@ -1,0 +1,77 @@
+# Default configuration for Sabayon containers
+
+# Setup the default mounts
+lxc.mount.auto = cgroup:mixed proc:mixed sys:mixed
+
+# Allow for 1024 pseudo terminals
+lxc.pts = 1024
+
+# Setup 1 tty devices for lxc-console command
+lxc.tty = 1
+
+# Needed for systemd distro
+lxc.autodev = 1
+
+# Doesn't support consoles in /dev/lxc/
+lxc.devttydir =
+
+# CGroup whitelist
+lxc.cgroup.devices.deny = a
+
+## Allow any mknod (but not reading/writing the node)
+#lxc.cgroup.devices.allow = c *:* m
+#lxc.cgroup.devices.allow = b *:* m
+
+## Allow specific devices
+### /dev/null
+lxc.cgroup.devices.allow = c 1:3 rwm
+### /dev/zero
+lxc.cgroup.devices.allow = c 1:5 rwm
+### /dev/full
+lxc.cgroup.devices.allow = c 1:7 rwm
+### /dev/random
+lxc.cgroup.devices.allow = c 1:8 rwm
+### /dev/urandom
+lxc.cgroup.devices.allow = c 1:9 rwm
+### /dev/pts/*
+#lxc.cgroup.devices.allow = c 136:* rwm
+### /dev/tty
+#lxc.cgroup.devices.allow = c 5:0 rwm
+### /dev/console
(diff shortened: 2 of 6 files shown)


stgraber:
This needs to be signed-off before we can review.

geaaru:
rebase and signed-off done. ;)

stgraber:
@geaaru yes, your template must be using the separate template config mechanism (see config/templates) for it to be suitable for integration with images.linuxcontainers.org and the lxc-download template.

Then if your template works on Ubuntu 16.04, I'll be able to easily add daily image generation to our Jenkins so that they publish on images.linuxcontainers.org. If your template instead only works on its own distro (sabayon), then we'll have to jump through a few more hoops to get bootstrap containers in place and sabayon support in lxc_ci.

stgraber:
So if your goal with this pull request was inclusion on images.linuxcontainers.org and support for unprivileged containers through lxc-download (as well as availability for LXD users), then you have a bit more work to do on the template to have it be in line with those other templates that use config/templates/.

geaaru:
Ok, thank you very much for support. I will try to add this others files.

geaaru:
But, I have some questions:
1) on my script, unlike others script I support unprivileged container (without download) through a different configuration of lxc.conf file and of container's rootfs (see unprivileged_rootfs function). While, on current scripts I see always that unprivileged is supported only through download image. Why doesn't support unprivileged container directly through template script ?
2) For common template files is it needed have two configuration for both privileged and unprivileged container ? 
3) on common.conf.in I see that as default fuse is mounted, but I think that this is not mandatory, so is correct create a sabayon.common.conf that doesn't include comons.conf.in?
Thanks in advance.

stgraber:
1) That's fine. I believe we have one other such one, the ubuntu-cloud template which works fine unprivileged. It's usually not an easy option for other templates because of some script that they run from their distribution which attempts to run mknod and so fails. But if yours doesn't hit such problems, then that's great.

2) Both files must exist for things to work, but you can perfectly have the .userns.conf one only include the common userns.conf and not do anything else (archlinux does that for example).

3) Why is LXC mounting /sys/fs/fuse/connections a problem for you? Note that this mount entry is marked optional, so if fuse isn't supported on the system, the mount will no occur.

geaaru:
Hi, thank you very very much for your informations.

About point 1... perfect, it is needed only that on target host is available fuidshift tool.

About point 3, less features are available on unprivileged containers less potentials exploits or holes could be happens. So, I think that fuse stack could be add manually from user if it is needed not as default. Indeed I disabled mount of fuse sysfs on unprivileged container at bootstrap.

Let me understand last thing, I haven't found documentation about this. What are targets of both common.conf and userns.conf ? Is userns.conf for unprivileged container ?

Thank you.

stgraber:
@geaaru the preferred way for users to disable fuse if they so wish should be by modifying common.conf rather than individual templates causing differing behavior from the others.

common.conf is applied to all containers, userns.conf is applied on top to add the bits needed for unprivileged containers.

geaaru:
@stgraber  I see only now your reply. sorry. For now I create a sabayon.common.conf script that doesn't include common.conf but I tested script on:
- amd64 gentoo env (without apparmor) kernel 4.5
- armv7 sabayon env (without apparmor) kernel 4.9
and both privileged and unprivileged containers works fine.
I think that works fine also under ubuntu.
Let me known if these scripts could be fine for merge.

Thanks

stgraber:
jenkins: test this please

geaaru:
Hi, I see test suite process. But sabayon doesn't support i386 architecture... current script support only x86_64 and armv7. I don't known if is required a change of test suite for this.
Let me know.
Thanks

stgraber:
Nah, the testsuite won't actually run your template, but it does a good job at validating that the autoconf stuff wasn't broken.

geaaru:
Hi, is there any other stuff to do for daily generation images in images.linuxcontainers.org of sabayon template ?
Thank in advance

stgraber:
I'll need to update github.com/lxc/lxc-ci and then configure Jenkins for it. I'll hopefully have some time to do so this week.

That'll get us images that work with LXD. For the lxc-download template, it'll take until the next major release as we don't include new templates into our bugfix releases.

geaaru:
Ok, thank you very much for your support.

PR #1399, opened 2017-01-29
Allow build without sys/capability.h

bneumeier:
There is no guard clause around a reference to CAP_EFFECTIVE and
CAP_SETGID, causing compilation to fail if sys/capability.h is not
available.


brauner:
Hi, thanks for the patch. Unfortunately you forgot to sign-off your commit. Could you please resend with a "Signed-off-by" line added.

PR #1402, opened 2017-01-30
lxc-opensuse: fix default value for release code

mabes:
The template scripts assumes that DISTRO should contain leap/4* when dealing with Opensuse Leap releases, but the default value did not contain the leap/ prefix


brauner:
@mabes, thanks for the patch. Unfortunately, you forgot to sign-off your commit. Please add a Signed-off-by line at the end of your commit message.

mabes:
Done, sorry about that

Issue #1158, opened 2016-08-27
templates should not set insecure default passwords


evgeni:
Hi,

promted by https://bugs.debian.org/831850, I had a look at the various templates and how they handle passwords.
There is an (outdated) overview done by the Fedora people at https://fedoraproject.org/wiki/LXC_Template_Security_Improvements

Let's update it and discuss :)

| template | default root password | root password changeable | notes |
| --- | --- | --- | --- |
| alpine | N/A | No |  |
| altlinux | rooter | No |  |
| archlinix | N/A | Yes |  |
| busybox | root | No |  |
| centos | random | Yes | IMHO not really random |
| cirros | N/A | No | password not touched, probably cirros default |
| debian | N/A | No |  |
| fedora | random | Yes | IMHO not really random |
| gentoo | toor | Yes | can be disabled |
| openmandriva | root | No |  |
| opensuse | root | No |  |
| oracle | root | No | additional oracle user with oracle password |
| plamo | root | No |  |
| pld | root | No |  |
| slackware | root | No |  |
| sparclinux | root | No |  |
| sshd | N/A | No |  |
| ubuntu | N/A | No | ubuntu user with ubuntu password, overridable |

IMHO containers should ship with no default passwords, and have a way to set them at creation time (either via `--pasword` or from the ENV).

brauner:
I think we decided that we want no root password to be set per default. Setting a password at creation time is probably not necessary since you can always just `lxc-attach` and set a password. We've discussed both of these issues before and I think this sums up the outcome. (/cc @stgraber). Patches to remove the passwords from templates are very much welcome. :)

stgraber:
Right and that's what I've been doing for the download template.

We were discussing whether we wanted to use the fancy random password generation code that fedora/centos have in their templates but now that we can basically assume all users have a kernel capable of setns. So I'd be fine with just removing all of them and having the templates tell the user how to set a password with lxc-attach.

evgeni:
Updated version for master (as of b49bb7dce83d6ea55785b7c7ccacca2370a3ba63)

| template | default root password | root password changeable | notes |
| --- | --- | --- | --- |
| alpine | N/A | No |  |
| altlinux | rooter | No | still unfixed |
| archlinix | N/A | Yes |  |
| busybox | N/A | No |  fixed |
| centos | random | Yes | IMHO not really random |
| cirros | N/A | No | password not touched, probably cirros default |
| debian | N/A | No |  |
[...]

Issue #302, opened 2014-08-19
 lxc-create creates vulnerable system in default debian template


ghost:
This is a forward of https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=758643:

---snip---
From: Ondřej Surý <ondrej@debian.org>
To: Debian Bug Tracking System <submit@bugs.debian.org>
Subject: lxc-create creates vulnerable system in default debian template
Date: Tue, 19 Aug 2014 17:10:33 +0200

Package: lxc
Version: 1:1.0.5-1
Severity: grave
Tags: security upstream patch
Justification: user security hole

lxc-create will by default set root password to 'root'.

This is a horrible practice and together with default installation of
openssh-server with PermitRootLogin yes create a security hole in each
and every container created and booted with IP address.

Attached is a simple fix that generates random password using pwgen
package.

All other templates needs to be checked as well.

Cheers,
Ondrej

diff --git a/debian/control b/debian/control
index 9a890a1..a5fa12c 100644
--- a/debian/control
+++ b/debian/control
@@ -17,6 +17,7 @@ Build-Depends:
  linux-libc-dev,
  pkg-config,
  python3-dev,
+ pwgen
 Standards-Version: 3.9.5
 Homepage: http://linuxcontainers.org/
 Vcs-Browser: http://daniel-baumann.ch/gitweb/?p=debian/packages/lxc.git

$ cat debian/patches/0010-Generate-a-random-password-in-default-debian-templat.patch 
From: =?utf-8?q?Ond=C5=99ej_Sur=C3=BD?= <ondrej@sury.org>
Date: Tue, 19 Aug 2014 17:08:16 +0200
Subject: Generate a random password in default debian template

---
 templates/lxc-debian.in | 6 ++++--
 1 file changed, 4 insertions(+), 2 deletions(-)

diff --git a/templates/lxc-debian.in b/templates/lxc-debian.in
index c2076dc..5cf928d 100644
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -148,8 +148,10 @@ EOF
         echo "Timezone in container is not configured. Adjust it manually."
     fi
 
-    echo "root:root" | chroot $rootfs chpasswd
-    echo "Root password is 'root', please change !"
+    password=$(pwgen 10 1)
+
+    echo "root:$password" | chroot $rootfs chpasswd
+    echo "Root password is '$password', please change !"
 
     return 0
 }

-- System Information:
Debian Release: jessie/sid
[...]

PR #1418, opened 2017-02-05
[stable-2.0] debian template: don't set a root password

zorun:
This is a backport of 515fb8d20c ("do not set the root password in the
debian template") from master.

See #302 and #1158 for details.

### DIFF
--- a/templates/lxc-debian.in
+++ b/templates/lxc-debian.in
@@ -158,9 +158,6 @@ EOF
         echo "Timezone in container is not configured. Adjust it manually."
     fi
 
-    echo "root:root" | chroot "$rootfs" chpasswd
-    echo "Root password is 'root', please change !"
-
     return 0
 }
 


stgraber:
I'm going to reject this for stable-2.0 at this time.

To be clear, I'd like the default behavior of our templates to be better and I do agree that not setting root passwords is an important part of that.

I however don't like the idea of "sneaking" this one change to one template into a bugfix release. I'd feel much better if we changed ALL templates in lxc to behave the same way and then backported that as one big behavior change (for security reasons) to the 2.0 branch.

That would make it much easier to highlight in the release announcement and make it very clear to everyone that there is a small behavior change in LXC 2.0.x which they may need to work around and that we've agreed to push anyway because of the current behavior being insecure.

evgeni:
@stgraber then how about combining this pr (which is a backport of https://github.com/lxc/lxc/pull/973) with a backport of https://github.com/lxc/lxc/pull/1358?

stgraber:
@evgeni yeah, that sounds good. Does the combination of the two actually cover all templates in stable-2.0?

evgeni:
@stgraber not completelly. see #1158. I will work on the others and the submit a backport of all changes for 2.0.x inclusion

evgeni:
@stgraber more template fixes in https://github.com/lxc/lxc/pull/1474

PR #1428, opened 2017-02-12
fix seccomp blocking x32 guests on amd64 (userspace) hosts

kilobyte:
While lxc has x32 support, it currently works only "natively" when the host runs amd64 kernel x32 userspace.  This patch fixes the more common case of regular amd64 on the host.

Tested only atop of current Debian package (2.0.7).


brauner:
jenkins: test this please.

brauner:
lgtm and I don't see any issues with allowing that. @hallyn, any objections?

hallyn:
No objections.

Thanks.

PR #1430, opened 2017-02-12
Add HAVE_LIBCAP

### LINKED ISSUE #1361: Impossible to build with --disable-capabilities if libcap sys/capability.h is present
Currently, it is impossible to build lxc with --disable-capabilities if the user has libcap-dev installed on his system as:
- calls to cap_xxx functions in caps.c are not protected by HAVE_LIBCAP defines. The whole file is only protected by HAVE_SYS_CAPABILITY_H.
- AC_CHECK_LIB default action-if-found is overriden by [true] so HAVE_LIBCAP is never written to config.in

As a result, if the user specifies --disable-capabilities, lxc's configure will not include -lcap which will result in a link error as HAVE_SYS_CAPABILITY_H will nevertheless be true (as sys/capability.h will be found).

So I would suggest to at least protect all calls to cap_xxx functions by HAVE_LIBCAP. An other option would be to replace all HAVE_SYS_CAPABILITY_H defines by HAVE_LIBCAP defines.

What's your opinion on this?

ffontaine:
Currently it is impossible to build lxc with --disable-capabilities if
the user has libcap-dev installed on his system as:
 - calls to cap_xxx functions are not protected by HAVE_LIBCAP defines.
 The whole file is only protected by HAVE_SYS_CAPABILITY_H.
 - AC_CHECK_LIB default action-if-found is overriden by [true] so
 HAVE_LIBCAP is never written to config.h

This patch replaces all HAVE_SYS_CAPABILITY_H checks by HAVE_LIBCAP
checks (fix #1361)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -370,7 +370,8 @@ fi
 AM_CONDITIONAL([ENABLE_CAP], [test "x$enable_capabilities" = "xyes"])
 
 AM_COND_IF([ENABLE_CAP],
-	[AC_CHECK_LIB(cap,cap_set_proc,[true],[AC_MSG_ERROR([You are missing libcap support.])])
+	[AC_CHECK_HEADER([sys/capability.h],[],[AC_MSG_ERROR([You must install the libcap development package in order to compile lxc])])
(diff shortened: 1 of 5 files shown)


brauner:
jenkins: test this please.

PR #1442, opened 2017-02-23
Fix issue with the clonehostname hook not working for overlayfs snapshot clones

mkeeler:
Previously this hook did not work when cloning containers using an overlayfs snapshot as the LXC_ROOTFS_PATH didn't point to the actual filesystem that the container would see. LXC_ROOTFS_MOUNT should be used instead and in fact lxc.container.conf man page says that you usually would want to use the _MOUNT variant.

### DIFF
--- a/hooks/clonehostname
+++ b/hooks/clonehostname
@@ -19,9 +19,9 @@
 
 # Note that /etc/hostname is updated by lxc itself
 for file in \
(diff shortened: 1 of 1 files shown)


brauner:
You need to sign-off your commit kernel-style. :)

Issue #1446, opened 2017-02-27
Weird event timestamp in log since lxc-2.0.7


ganto:
When digging around the log files I found an interesting change of behaviour in the lxc logging facility. Since the update to `lxc-2.0.7` the timestamps in the logs are shifted by one year. E.g. from `/var/lib/lxd/containers/lxc-monitord.log`:
```
   lxc-monitord 20170204193605.228 NOTICE   lxc_monitord - lxc_monitord.c:main:417 - pid:4059 monitoring lxcpath /var/lib/lxd/containers
   lxc-monitord 20170204193636.031 NOTICE   lxc_monitord - lxc_monitord.c:main:422 - no remaining clients, exiting
   lxc-monitord 20170204193636.031 NOTICE   lxc_monitord - lxc_monitord.c:main:430 - monitor exiting
   lxc-monitord 20160214053329.179 INFO     lxc_monitor - monitor.c:lxc_monitor_sock_name:185 - Using monitor socket name "lxc/d78a9d7e97b4b375//var/lib/lxd/containers".
   lxc-monitord 20160214053329.179 NOTICE   lxc_monitord - lxc_monitord.c:main:422 - lxc-monitord with pid 13391 is now monitoring lxcpath /var/lib/lxd/containers.
   lxc-monitord 20160214053329.182 DEBUG    lxc_monitord - lxc_monitord.c:lxc_monitord_fifo_create:108 - lxc-monitord already running on lxcpath /var/lib/lxd/containers.
   lxc-monitord 20160214053329.183 INFO     lxc_monitord - lxc_monitord.c:lxc_monitord_sock_accept:222 - Accepted client file descriptor 5. Number of accepted file descriptors is now 1.
```

The update to `lxc-2.0.7` was done on Feb 10, 2017.

I can see the same behaviour when starting a container via the regular `lxc-*` tools. E.g.:
```
# lxc-info -n debops-jessie64-mailer-lxc_mail_1487803394149_34284 -l debug -o debug.log ; cat debug.log ; date
Name:           debops-jessie64-mailer-lxc_mail_1487803394149_34284
State:          RUNNING
PID:            26280
IP:             192.168.122.110
CPU use:        353.06 seconds
Memory use:     84.61 MiB
KMem use:       19.66 MiB
Link:           vethYFP3EB
 TX bytes:      8.21 MiB
 RX bytes:      145.04 MiB
 Total bytes:   153.25 MiB
[...]

PR #1447, opened 2017-02-27
log: fix lxc_unix_epoch_to_utc()

brauner:
The conversion algorithm used uses a clever trick by letting a year start at 1
March. So we need to add 1 for January and February.


brauner:
closes #1446

Issue #80, opened 2013-10-27
Kernel module fuse can not be used in container.


absolute8511:
The fuse is already loaded in host 

modprobe -l | grep fuse  shows
kernel/fs/fuse/fuse.ko
kernel/fs/fuse/cuse.ko

But in container while using it, something error:

fuse: device not found, try 'modprobe fuse' first

I also tried to copy the /lib/modules from host to container and the same error.
Another module ip_queue has the same problem. 
Is it supported to use the kernel module in container by now?  Or is any plan to support this feature?

Thank you!

PR #1453, opened 2017-03-06
seccomp: set SCMP_FLTATR_ATL_TSKIP if available

hallyn:
Newer libseccomp has a flag called SCMP_FLTATR_ATL_TSKIP which
allows syscall '-1' (nop) to be executed.  Without that flag,
debuggers cannot skip system calls inside containers.  For reference,
see the seccomp(2) manpage, which says:

	The tracer can skip the system call by changing the system call  number  to  -1.

and see the seccomp issue #80


brauner:
@hallyn, did you forgot to sign-off your commit?

hallyn:
no way!  that's supposed to be auto.  pushed with sig, thx

PR #1458, opened 2017-03-11
lxc-user-nic: improvements

brauner:


brauner:
@stgraber, also adds the code to cleanup the link when the rename fails.

stgraber:
Looks good to me, though given this is a setuid tool, I'd feel better if @hallyn could review it too.

hallyn:
In the future, I'd really really appreciate if any patch which has a long sequence of "should have zero functionality changes" not be mixed with functional changes.

Note also that alloca() per the manpage does not return NULL on error - just gives SEGV on access :)  There may be some libc where that's not the case though so let's leave your fix as is.

hallyn:
(yeah, thanks stgraber, I'd actually downloaded the patch yesterday morning but have been largely sans laptop since)

hallyn:
In patch 2 can you please --amend with a description of the bugfix?

brauner:
Hm, actually the only patch that changes functionality should be the third one. I intended to keep them separate. Are there functionality changes in the others?

hallyn:
Oh, I was going by the commit message.  I hadn't yet seen a functional change which was why i was asking :)

brauner:
@hallyn, yeah, the mention of "bugfix" in the second commit referred to a couple of unchecked `strdup()`s. :)

brauner:
Which I didn't consider functional changes in any relevant sense. :)

PR #1463, opened 2017-03-15
[templates] archlinux: resolve conflicting files

otofune:
- already found /etc/localtime
- duplicate creation /etc/resolv.conf

### DIFF
--- a/templates/lxc-archlinux.in
+++ b/templates/lxc-archlinux.in
@@ -87,9 +87,6 @@ configure_arch() {
 
     # hostname and nameservers
     echo "${name}" > "${rootfs_path}/etc/hostname"
-    while read r; do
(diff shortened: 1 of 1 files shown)


stgraber:
I can see why the resolv.conf bits aren't needed because of the networkd code, but why isn't the timezone code needed?

otofune:
First of all, there is the `/etc/localtime` before generating. it's soft link of `../usr/share/zoneinfo/UTC`.
I think pacman doing to generate it. I'll check it.

stgraber:
Ok, since default_timezone can't be overriden by the user through environment, argument or based on the host timezone, I guess it's fine to just drop that part of the code.

Though you should also drop the line defining default_timezone then.

PR #1474, opened 2017-03-19
don't set a default password for altlinux, gentoo, openmandriva and pld

evgeni:
Refs: #1158


brauner:
+1

PR #1476, opened 2017-03-20
tree-wide: include <sys/sysmacros.h> directly

brauner:


brauner:
Let's see if this is enough to fix it.

PR #1504, opened 2017-04-11
start: fix error handling when limits fail to apply

Blub:
(The code was moved here from the child side of the startup
without adapting the error case.)


brauner:
jenkins: test this please

PR #1509, opened 2017-04-15
idmap improvements

brauner:

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3321,11 +3321,13 @@ static int write_id_mapping(enum idtype idtype, pid_t pid, const char *buf,
 
 int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
 {
-	struct lxc_list *iterator;
 	struct id_map *map;
-	int ret = 0, use_shadow = 0;
+	struct lxc_list *iterator;
 	enum idtype type;
-	char *buf = NULL, *pos, *cmdpath = NULL;
+	char *pos;
+	char *buf = NULL, *cmdpath = NULL;
+	bool use_shadow = false;
+	int ret = 0;
 
 	/*
 	 * If newuidmap exists, that is, if shadow is handing out subuid
@@ -3335,7 +3337,7 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
 	 */
 	cmdpath = on_path("newuidmap", NULL);
 	if (cmdpath) {
-		use_shadow = 1;
+		use_shadow = true;
 		free(cmdpath);
 	}
 
@@ -3344,50 +3346,51 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
 		return -1;
 	}
 
-	for(type = ID_TYPE_UID; type <= ID_TYPE_GID; type++) {
+	for (type = ID_TYPE_UID; type <= ID_TYPE_GID; type++) {
 		int left, fill;
-		int had_entry = 0;
+		bool had_entry = false;
 		if (!buf) {
-			buf = pos = malloc(4096);
+			buf = pos = malloc(LXC_IDMAPLEN);
 			if (!buf)
 				return -ENOMEM;
 		}
 		pos = buf;
 		if (use_shadow)
-			pos += sprintf(buf, "new%cidmap %d",
-				type == ID_TYPE_UID ? 'u' : 'g',
-				pid);
+			pos += sprintf(buf, "new%cidmap %d", type == ID_TYPE_UID ? 'u' : 'g', pid);
 
 		lxc_list_for_each(iterator, idmap) {
-			/* The kernel only takes <= 4k for writes to /proc/<nr>/[ug]id_map */
+			/* The kernel only takes <= 4k for writes to
+			 * /proc/<nr>/[ug]id_map
+			 */
 			map = iterator->elem;
 			if (map->idtype != type)
 				continue;
 
-			had_entry = 1;
-			left = 4096 - (pos - buf);
+			had_entry = true;
+
+			left = LXC_IDMAPLEN - (pos - buf);
 			fill = snprintf(pos, left, "%s%lu %lu %lu%s",
-					use_shadow ? " " : "",
-					map->nsid, map->hostid, map->range,
+					use_shadow ? " " : "", map->nsid,
+					map->hostid, map->range,
 					use_shadow ? "" : "\n");
 			if (fill <= 0 || fill >= left)
-				SYSERROR("snprintf failed, too many mappings");
+				SYSERROR("Too many {g,u}id mappings defined.");
+
 			pos += fill;
 		}
 		if (!had_entry)
 			continue;
 
 		if (!use_shadow) {
-			ret = write_id_mapping(type, pid, buf, pos-buf);
+			ret = write_id_mapping(type, pid, buf, pos - buf);
 		} else {
-			left = 4096 - (pos - buf);
+			left = LXC_IDMAPLEN - (pos - buf);
 			fill = snprintf(pos, left, "\n");
 			if (fill <= 0 || fill >= left)
-				SYSERROR("snprintf failed, too many mappings");
+				SYSERROR("Too many {g,u}id mappings defined.");
 			pos += fill;
 			ret = system(buf);
 		}
-
 		if (ret)
 			break;
 	}
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3319,30 +3319,89 @@ static int write_id_mapping(enum idtype idtype, pid_t pid, const char *buf,
 	return ret < 0 ? ret : closeret;
 }
 
+/* Check whether a binary exist and has either CAP_SETUID, CAP_SETGID or both. */
+static int idmaptool_on_path_and_privileged(const char *binary, cap_value_t cap)
+{
+	char *path;
+	int ret;
+	struct stat st;
+	int fret = 0;
+
+	path = on_path(binary, NULL);
+	if (!path)
+		return -ENOENT;
+
+	ret = stat(path, &st);
+	if (ret < 0) {
+		fret = -errno;
+		goto cleanup;
+	}
+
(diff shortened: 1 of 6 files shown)


src/lxc/conf.c
```diff
@@ -3323,28 +3323,95 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
...
+		/* Check if the newuidmap binary is setuid. */
+		if ((sb.st_mode & S_ISUID) && sb.st_uid == 0)
+			has_sufficient_privilege = true;
+
+		#if HAVE_LIBCAP
+		/* Check if it has file capabilities. */
+		if (!has_sufficient_privilege &&
+		    lxc_file_cap_is_set(newuidmap_path, CAP_SETUID, CAP_EFFECTIVE) &&
+		    lxc_file_cap_is_set(newuidmap_path, CAP_SETUID, CAP_PERMITTED) &&
+		    sb.st_uid == 0)
```
hallyn: File does not need to be owned by root for filecaps to take effect.  The sb.st_uid check can be removed afaics.

src/lxc/conf.c
```diff
@@ -3323,28 +3323,95 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
...
 	 * will protected it by preventing another user from being handed the
 	 * range by shadow.
 	 */
-	cmdpath = on_path("newuidmap", NULL);
-	if (cmdpath) {
-		use_shadow = true;
-		free(cmdpath);
+	newuidmap_path = on_path("newuidmap", NULL);
+	if (newuidmap_path) {
+		bool has_sufficient_privilege = false;
```
hallyn: This block sure seems like a good candidate for a helper function :)  Not requisite for this patch, just a comment.

src/lxc/conf.c
```diff
@@ -3382,21 +3449,21 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
...
 		} else {
 			left = LXC_IDMAPLEN - (pos - buf);
 			fill = snprintf(pos, left, "\n");
 			if (fill <= 0 || fill >= left)
 				SYSERROR("Too many {g,u}id mappings defined.");
 			pos += fill;
-			ret = system(buf);
+			fret = system(buf);
 		}
 		if (ret)
```
hallyn: Should this check still be on ret, or on fret?  What was the reason for switching from ret to fret?

src/lxc/conf.c
brauner: You would still be able to write to `new_{g,u}idmap` even if `euid > 0`?

src/lxc/conf.c
brauner: Yeah, thought about it as well. Might just write it now.

src/lxc/conf.c
brauner: `ret` is used in the `new{g,u}idmap` block before and might get initialized to a non-zero value but the `for`-loop, as unlikely as this is, might not be run and then you'd exit with error.

src/lxc/conf.c
hallyn: Write to it?  Aren't you going to execute it?

src/lxc/conf.c
hallyn: So if you turn that block into a helper then it goes away? :)

In either case, the check here to break out of the loop needs to be for fret right?  or not?

hallyn: Oops, that last comment was supposed to go with the block below.

src/lxc/conf.c
brauner: I mean will `new{g,u}idmap` be able to write to `/proc/<pid>/uid_map` when it is not uid `0`?

brauner: Unless of course, `uid_map` is owned by another the unpriv user.

src/lxc/conf.c
```diff
@@ -3321,73 +3320,107 @@ static int write_id_mapping(enum idtype idtype, pid_t pid, const char *buf,
...
+			return -1;
+		}
+		free(newgidmap_path);
+		use_shadow++;
+	} else if (euid > 0) {
+		ERROR("The newgidmap binary is missing and euid is %d.", euid);
+		return -1;
+	}
+
+	if (has_privileged_newuidmap && !has_privileged_newgidmap) {
```
brauner: I think it makes sense to error out when we find only one of the two binaries.

src/lxc/utils.c
```diff
@@ -2070,3 +2071,36 @@ int lxc_setgroups(int size, gid_t list[])
...
+	if ((st.st_mode & S_ISUID) && st.st_uid == uid) {
+		DEBUG("The binary \"%s\" has the setuid bit set.", path);
+		return true;
+	}
+
+	#if HAVE_LIBCAP
+	/* Check if it has file capabilities. */
+	if (lxc_file_cap_is_set(path, CAP_SETUID, CAP_EFFECTIVE) &&
+	    lxc_file_cap_is_set(path, CAP_SETUID, CAP_PERMITTED) &&
+	    st.st_uid == uid) {
```
brauner: I left this in for now. But if we're sure I'll remove it before the merge.

src/lxc/conf.c
```diff
@@ -3321,73 +3320,107 @@ static int write_id_mapping(enum idtype idtype, pid_t pid, const char *buf,
...
+	/* If new{g,u}idmap exists, that is, if shadow is handing out subuid
+	 * ranges, then insist that root also reserve ranges in subuid. This
 	 * will protected it by preventing another user from being handed the
 	 * range by shadow.
 	 */
-	cmdpath = on_path("newuidmap", NULL);
-	if (cmdpath) {
-		use_shadow = 1;
-		free(cmdpath);
+	newuidmap_path = on_path("newuidmap", NULL);
```
brauner: I didn't move the `on_path()` part into the setuid checks. But I probably could. Should we do that, what do you think?

hallyn:
> 
> 
> 
> > +	newgidmap_path = on_path("newgidmap", NULL);
> +	if (newgidmap_path) {
> +		has_privileged_newgidmap = is_setuid_binary(newgidmap_path, 0);
> +		if (!has_privileged_newgidmap) {
> +			ERROR("The newgidmap binary lacks sufficient privilege.");
> +			free(newgidmap_path);
> +			return -1;
> +		}
> +		free(newgidmap_path);
> +		use_shadow++;
> +	} else if (euid > 0) {
> +		ERROR("The newgidmap binary is missing and euid is %d.", euid);
> +		return -1;
> +	}
> +
> +	if (has_privileged_newuidmap && !has_privileged_newgidmap) {
> 
> I think it makes sense to error out when we find only one of the two binaries.

Yes. I'm only suggesting using a helper to check if the binary is there and
sufficiently privileged, since you have around 30 lines of code duplicated.

hallyn:
> 
> 
> 
> > +		if (ret < 0) {
> +			SYSERROR("Failed to stat \"%s\".", newuidmap_path);
> +			free(newuidmap_path);
> +			return -1;
> +		}
> +
> +		/* Check if the newuidmap binary is setuid. */
> +		if ((sb.st_mode & S_ISUID) && sb.st_uid == 0)
> +			has_sufficient_privilege = true;
> +
> +		#if HAVE_LIBCAP
> +		/* Check if it has file capabilities. */
> +		if (!has_sufficient_privilege &&
> +		    lxc_file_cap_is_set(newuidmap_path, CAP_SETUID, CAP_EFFECTIVE) &&
> +		    lxc_file_cap_is_set(newuidmap_path, CAP_SETUID, CAP_PERMITTED) &&
> +		    sb.st_uid == 0)
> 
> I mean will `new{g,u}idmap` be able to write to `/proc/<pid>/uid_map` when it is not uid `0`?

Yes, you only need CAP_SETUID for uid_map and CAP_SETGID for gid_map.

hallyn:
> 
> 
> 
> >  	 * will protected it by preventing another user from being handed the
>  	 * range by shadow.
>  	 */
> -	cmdpath = on_path("newuidmap", NULL);
> -	if (cmdpath) {
> -		use_shadow = 1;
> -		free(cmdpath);
> +	newuidmap_path = on_path("newuidmap", NULL);
> 
> I didn't move the `on_path()` part into the setuid checks. But I probably could. Should we do that, what do you think?

I don't know, depends on whether the privilege-only check is more reusable.
I suspect just a on_path_and_privileged("newuidmap", CAP_SETUID) and
on_path_privileged("newgidmap", CAP_SETGID | CAP_SETUID) works out nicely.

brauner:
@hallyn, updated.

hallyn:
Hi,

looks good.  Just two more comments for now:

1. unfortunately (if i read this right) writing to map_file also requires cap_sys_admin against the user_ns, in addition to cap_setuid/cap_setgid.
2. newgidmap is actually setuid-root, not setgid-root.  So the filecaps check should check for CAP_SETGID and CAP_SYS_ADMIN, but the 'check if the binary is setuid' check should check for setuid-root.

brauner:
> Hi,

> looks good.  Just two more comments for now:

> 1. unfortunately (if i read this right) writing to map_file also requires cap_sys_admin against the user_ns, in addition to cap_setuid/cap_setgid.
> 2. newgidmap is actually setuid-root, not setgid-root.  So the filecaps check should check for CAP_SETGID and CAP_SYS_ADMIN, but the 'check if the binary is setuid' check should check for setuid-root.

As discussed on IRC `CAP_SYS_ADMIN` is actually not required. What we need to check is:
- 1. `newuidmap` has the setuid bit set --> good to go
- 2. `newuidmap` has `CAP_SETUID` file capability in `CAP_EFFECTIVE` and `CAP_PERMITTED` --> good to go
- 3. `newgidmap` has the setuid bit set --> good to go
- 4. `newgidmap` has `CAP_SETGID` file capability in `CAP_EFFECTIVE` and `CAP_PERMITTED` --> good to go.
Anything else we don't consider for now.

brauner:
@hallyn, should be ready to merge.

hallyn:
Awesome, thanks.

PR #1510, opened 2017-04-16
ls: simplify the judgment condition when list active containers

0x0916:
In the following code 
```
if ( args->ls_active || (args->ls_active && args->ls_frozen))
```
`(args->ls_active && args->ls_frozen)` always false if `args->ls_active` is false.
so we can delete the second  judgment condition and only keep the first one.

### DIFF
--- a/src/lxc/tools/lxc_ls.c
+++ b/src/lxc/tools/lxc_ls.c
@@ -356,7 +356,7 @@ static int ls_get(struct ls **m, size_t *size, const struct lxc_arguments *args,
 	}
 
 	/* Do not do more work than is necessary right from the start. */
-	if (args->ls_active || (args->ls_active && args->ls_frozen))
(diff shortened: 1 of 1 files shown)


src/lxc/tools/lxc_ls.c
```diff
@@ -356,7 +356,7 @@ static int ls_get(struct ls **m, size_t *size, const struct lxc_arguments *args,
 	}
 
 	/* Do not do more work than is necessary right from the start. */
-	if (args->ls_active || (args->ls_active && args->ls_frozen))
+	if (args->ls_active)
```
brauner: Actually that should be `args->ls_active || args->ls_frozen`.

0x0916: ok,  I will update the patch. thanks

brauner:
jenkins: test this please

PR #1513, opened 2017-04-17
caps: skip file capability checks on android

brauner:
Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

PR #1514, opened 2017-04-18
autotools: check for cap_get_file

brauner:

cf. the discussion on https://github.com/lxc/lxc/pull/1513

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -372,7 +372,9 @@ AM_CONDITIONAL([ENABLE_CAP], [test "x$enable_capabilities" = "xyes"])
 AM_COND_IF([ENABLE_CAP],
 	[AC_CHECK_HEADER([sys/capability.h],[],[AC_MSG_ERROR([You must install the libcap development package in order to compile lxc])])
 	AC_CHECK_LIB(cap,cap_set_proc,[],[AC_MSG_ERROR([You must install the libcap development package in order to compile lxc])])
-	AC_SUBST([CAP_LIBS], [-lcap])])
+        # Test whether we support getting file capabilities via cap_get_file().
+        AC_CHECK_LIB(cap,cap_get_file, AC_DEFINE(LIBCAP_SUPPORTS_FILE_CAPABILITIES,1,[Have cap_get_file]),[],[])
+        AC_SUBST([CAP_LIBS], [-lcap])])
 
 # HAVE_SCMP_FILTER_CTX=1 will tell us we have libseccomp api >= 1.0.0
 OLD_CFLAGS="$CFLAGS"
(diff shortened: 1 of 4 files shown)


src/lxc/caps.c
```diff
@@ -247,6 +244,8 @@ bool lxc_file_cap_is_set(const char *path, cap_value_t cap, cap_flag_t flag)
 	cap_is_set = lxc_cap_is_set(caps, cap, flag);
 	cap_free(caps);
 	return cap_is_set;
+	#else
+	return true;
```
hallyn: Should this return false instead?

configure.ac
```diff
@@ -372,7 +372,9 @@ AM_CONDITIONAL([ENABLE_CAP], [test "x$enable_capabilities" = "xyes"])
 AM_COND_IF([ENABLE_CAP],
 	[AC_CHECK_HEADER([sys/capability.h],[],[AC_MSG_ERROR([You must install the libcap development package in order to compile lxc])])
 	AC_CHECK_LIB(cap,cap_set_proc,[],[AC_MSG_ERROR([You must install the libcap development package in order to compile lxc])])
-	AC_SUBST([CAP_LIBS], [-lcap])])
+        # Test whether we support getting file capabilities via cap_get_file().
+        AC_CHECK_LIB(cap,cap_get_file, AC_DEFINE(HAVE_FILE_CAPABILITIES,1,[Have cap_get_file]),[],[])
```
hallyn: Would LIBCAP_HAS_FILE_CAPABILITIES be a better name?  Otherwise others may see this and think it means fscaps are or are not support on the system, rather than only in libcap.

src/lxc/caps.c
brauner: Currently all of the caps functions return `true` when they are not defined. Not sure if reversing it is worth it.

brauner: That being said not a big deal if we do it.

configure.ac
brauner: Yeah, best not confuse users.

src/lxc/caps.c
hallyn: But if cap_get_file() call itself fails, you return false :)

i think returning false, and demanding different privilege, is probably best.  Or will that break android?  Is there a platform out there that relies on filecaps but then doesn't offer a libcap that can read them?

hallyn: In fact another way to implement this would be to do

#ifndef LIBCAP_HAS_FILE_CAPS
#define cap_get_flags(a, b, c, d) { errno = ENODATA; return NULL; }
#endif

 and just treat it as a failed helper call.

brauner: What advantage would this provide over a simple `return false`?

brauner: If we change the `return true` we should switch over all to `return false`.

brauner: Setting `errno = ENODATA` in `lxc_file_cap_is_set()` on `#ifnded LIBCAP_SUPPORTS_FILE_CAPABILTIES` is still worth it.

hallyn:
Thanks.

PR #1519, opened 2017-04-21
conf: use bind-mount for /dev/ptmx

brauner:

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1456,8 +1456,7 @@ static int setup_pivot_root(const struct lxc_rootfs *rootfs)
 static int lxc_setup_devpts(int num_pts)
 {
 	int ret;
-	char target[PATH_MAX];
-	char *devpts_mntopts = "newinstance,ptmxmode=0666,mode=0620,gid=5";
+	const char *devpts_mntopts = "newinstance,ptmxmode=0666,mode=0620,gid=5";
 
 	if (!num_pts) {
 		DEBUG("no new devpts instance will be mounted since no pts "
@@ -1465,9 +1464,9 @@ static int lxc_setup_devpts(int num_pts)
 		return 0;
 	}
 
+	/* Unmount old devpts instance. */
 	ret = access("/dev/pts/ptmx", F_OK);
 	if (!ret) {
-		/* Unmount old devpts instance. */
 		ret = umount("/dev/pts");
 		if (ret < 0) {
 			SYSERROR("failed to unmount old devpts instance");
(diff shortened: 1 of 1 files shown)


src/lxc/conf.c
```diff
@@ -1455,48 +1455,92 @@ static int setup_pivot_root(const struct lxc_rootfs *rootfs)
...
+		}
+		created_ptmx = true;
+		DEBUG("Created dummy \"/dev/ptmx\" file as bind mount target.");
+	}
 
-	/* fallback here, /dev/pts/ptmx exists just mount bind */
-	if (mount("/dev/pts/ptmx", "/dev/ptmx", "none", MS_BIND, 0)) {
-		SYSERROR("mount failed '/dev/pts/ptmx'->'/dev/ptmx'");
+	/* Bind mount /dev/pts/ptmx to /dev/ptmx. */
+	ret = mount("/dev/pts/ptmx", "/dev/ptmx", "none", MS_BIND, 0);
```
stgraber: So what if /dev/ptmx exists as a symlink to /dev/pts/ptmx already?

The mount() call on it will deref the symlink and cause over-mounting of /dev/pts/ptmx won't it?

src/lxc/conf.c
```diff
@@ -1455,48 +1455,92 @@ static int setup_pivot_root(const struct lxc_rootfs *rootfs)
...
+			goto success;
+		}
+		SYSERROR("Failed to create symlink \"/dev/ptmx\" -> \"/dev/pts/ptmx\"");
 		return -1;
 	}
 
-	INFO("created new pts instance");
+	/* Check if the symlink is valid. */
+	if (realpath("/dev/ptmx", target) && !strcmp(target, "/dev/pts/ptmx"))
+		goto success;
```
stgraber: That's not really doing anything is it? Whether it succeeds or fails we hit success :)

brauner: Nah, that's why I'm re-doing the commits and also to split out *all* non-functional changes. :) Coming up in a bit. :)

src/lxc/conf.c
brauner: Right, so we should probably remove the symlink. That should be safe.

src/lxc/conf.c
```diff
@@ -1453,50 +1453,86 @@ static int setup_pivot_root(const struct lxc_rootfs *rootfs)
...
-	INFO("created new pts instance");
+	/* Fallback option: create symlink /dev/ptmx -> /dev/pts/ptmx  */
+	ret = mount("/dev/pts/ptmx", "/dev/ptmx", "none", MS_BIND, 0);
+	if (!ret) {
+		DEBUG("bind mounted \"/dev/pts/ptmx\" to \"/dev/ptmx\"");
+		return 0;
+	} else {
+		/* Fallthrough and try to create a symlink. */
+		ERROR("failed to bind mount \"/dev/pts/ptmx\" to \"/dev/ptmx\"");
+	}
```
setharnold: I think the symlink case has been overlooked; or should the comment simply be modified? Thanks

brauner: My intention was to fallthrough and retry with a symlink if the bind-mount fails. Is that your question?

setharnold: Ah. I failed to read. I expected to see the symlink() right here in the `else { }` block. THanks.

PR #1531, opened 2017-05-02
Change sshd template to work with Ubuntu 17.04

mar-kolya:
A few things have changed and this patch makes container generated for sshd work in Ubuntu

Signed-off-by: Nikolay Martynov <mar.kolya@gmail.com>

### DIFF
--- a/templates/lxc-sshd.in
+++ b/templates/lxc-sshd.in
@@ -38,7 +38,6 @@ install_sshd()
     rootfs=$1
 
     tree="\
-$rootfs/var/run/sshd \
 $rootfs/var/empty/sshd \
 $rootfs/var/lib/empty/sshd \
 $rootfs/etc/init.d \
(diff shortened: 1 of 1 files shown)


stgraber:
But doing that will presumably break the template on older Ubuntu releases and other distros.

mar-kolya:
I do not think that this is really the case, unless I'm really missing something:
There is still symlink in old `/var/run` location so that should take care of `sshd` that uses that
And sshd config options that were deleted seem to be quite old.

stgraber:
I seem to remember that /run isn't a thing on some of the distros we support, but I can't remember specifics right now.

mar-kolya:
It should not matter: this template will create `/run` and link `/var/run` to it - so container will have both. And it becomes to `sshd` to choose which one to use.

stgraber:
jenkins: test this please

PR #1535, opened 2017-05-03
[RFC] api extension: lxc_config_item_is_supported()

brauner:
This adds lxc_config_item_is_supported() as API extension. It allows to check
whether a given config item (e.g. lxc.autodev) is supported by this LXC
instance. The function is useful in the following scenarios:
1. Users have compiled liblxc from source and have removed a config items from
   the corresponding struct in confile.c. (For example, embedded users might
   decide to gut a bunch of options that they cannot use.)
2. Callers that want to check for a specific configuration item independent of
   the version numbers exposed in our version.h header.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -4522,3 +4522,8 @@ free_ct_name:
 	free(ct_name);
 	return ret;
 }
+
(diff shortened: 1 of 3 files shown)


hallyn:
Thanks, Lgtm. Sadly I cannot merge from this interface.

PR #1536, opened 2017-05-04
Fix abstract socket address length issue

0x0916:
commit: https://github.com/lxc/lxc/commit/073135baa78511c26e502362840f2c950cfddfe2
change `sockname[sizeof(addr->sun_path)-2] = '\0';` to
`sockname[sizeof(addr->sun_path)-3] = '\0';` and can not explain the reason.

because we use `memset` to initialize `addr` to zero, we need not the code to make
sure that the last byte of addr is zero.

when delete this code, it doesn't seem to work, because the wrong check for abstract
socker addr's length in function `lxc_mointor_open`. The max length of the address
can be 107 bytes.

also fix this in function `lxc_abstract_unix_connect` and `lxc_abstract_unix_open`.

### DIFF
--- a/src/lxc/af_unix.c
+++ b/src/lxc/af_unix.c
@@ -56,7 +56,7 @@ int lxc_abstract_unix_open(const char *path, int type, int flags)
 	addr.sun_family = AF_UNIX;
 
 	len = strlen(&path[1]) + 1;
-	if (len >= sizeof(addr.sun_path) - 1) {
+	if (len >= sizeof(addr.sun_path)) {
 		close(fd);
 		errno = ENAMETOOLONG;
 		return -1;
@@ -110,7 +110,7 @@ int lxc_abstract_unix_connect(const char *path)
 	addr.sun_family = AF_UNIX;
 
 	len = strlen(&path[1]) + 1;
-	if (len >= sizeof(addr.sun_path) - 1) {
+	if (len >= sizeof(addr.sun_path)) {
 		close(fd);
 		errno = ENAMETOOLONG;
 		return -1;
--- a/src/lxc/monitor.c
+++ b/src/lxc/monitor.c
@@ -181,7 +181,7 @@ int lxc_monitor_sock_name(const char *lxcpath, struct sockaddr_un *addr) {
 	if (ret < 0)
 		return -1;
 
-	sockname[sizeof(addr->sun_path)-3] = '\0';
+	addr->sun_path[sizeof(addr->sun_path)-1] = '\0';
 	INFO("Using monitor socket name \"%s\".", sockname);
 
 	return 0;
@@ -205,7 +205,7 @@ int lxc_monitor_open(const char *lxcpath)
 	}
 
 	len = strlen(&addr.sun_path[1]) + 1;
-	if (len >= sizeof(addr.sun_path) - 1) {
+	if (len >= sizeof(addr.sun_path)) {
 		ret = -1;
 		errno = ENAMETOOLONG;
 		goto on_error;


src/lxc/af_unix.c
```diff
@@ -56,7 +56,7 @@ int lxc_abstract_unix_open(const char *path, int type, int flags)
 	addr.sun_family = AF_UNIX;
 
 	len = strlen(&path[1]) + 1;
-	if (len >= sizeof(addr.sun_path) - 1) {
```
brauner: Abstract unix sockets are `\0` terminated at the first byte and the last byte. So unless I'm missing something this code is correct and should stay the way it is.

brauner:
What issue is this commit supposed to fix? Did you observe a segfault or something?

src/lxc/af_unix.c
0x0916: if we use a long enough lxcpath, in the code `len = strlen(&path[1]) + 1` , len = 107. and `sizeof(addr.sun_path)-1` = 107.  at this situation, the abstract unix socket length is legal. we can fix it to `len > sizeof(addr.sun_path) - 1` or `len >= sizeof(addr.sun_path) `

0x0916:
@brauner,  I did not observer a segfault or something else wrong, This pr only makes the code more correctly. 

if the abstract unix socket(`addr.sun_path`) is `\0abstract...socket\0`, with this patch, the max length of  `abstract...socket` can be `106` , and without this patch it only can be `105`.

src/lxc/af_unix.c
Blub: They're not actually null-terminated, only `\0` initiated and use all bytes up to the given length of the address structure. From `unix(7)`: `The socket's address in this namespace is given by the additional bytes in sun_path that are covered by the specified length of the address structure. (Null bytes in the name have no special significance.)`.
If I'm reading this right the current code enforces space for a `\0` while the new code doesn't. (Having a `\0` makes sense if they're checked with `strlen` and seems like a useful convention in general.)

brauner: Yeah, I know you can give them names including \0 bytes but I'm always weary to mention this fact since I really don't want to deal with tracking down bugs that stem from a missing \0 byte at the end. :)

src/lxc/monitor.c
```diff
@@ -181,7 +181,6 @@ int lxc_monitor_sock_name(const char *lxcpath, struct sockaddr_un *addr) {
 	if (ret < 0)
 		return -1;
 
-	sockname[sizeof(addr->sun_path)-3] = '\0';
```
Blub: The `-3` looks suspicious btw. It used to be `-2` which makes more sense, should check why that errored. Taking this away however is definitely a problem due to the address being checked with `strlen()` and `snprintf` potentially truncating the data (`ret == len`).

src/lxc/monitor.c
```diff
@@ -205,7 +204,7 @@ int lxc_monitor_open(const char *lxcpath)
 	}
 
 	len = strlen(&addr.sun_path[1]) + 1;
```
Blub: With the removal of the 0-termination in `lxc_monitor_sock_name` this `strlen` potentially reads past the `addr` on the stack giving weird results.
(Btw. - I think the length check could be moved above the `socket()` call?)

src/lxc/monitor.c
brauner: -2 for the first and last \0 byte and to actually write a \0 byte you need to do -3.

0x0916: `sockname` is pointer to `&addr->sun_path[1]`, so i think here should be `-2`.

Blub: `(sockname = sun_path+1) + sizeof(sun_path) - 3 = sun_path + sizeof(sun_path) - 2`
So this would set `sun_path[sizeof(sun_path)-2] = 0` which is the second to last byte of `sun_path` (while IMO only the last one should need to be `\0`? (and first one sure, but this is about the size of `sun_path`, not the length of its contents so the initial byte's content doesn't matter for this offset)).

src/lxc/monitor.c
0x0916:
```c
int snprintf(char *str, size_t size, const char *format, ...);
```
`snprintf` will write at most **size** bytes (including the terminating null byte ('\0')) to **str**, so the  `addr.sun_path` awlays have a `null byte` and the end.

src/lxc/monitor.c
0x0916: I think that only the last one should be set to `\0`. so we change it to  `sockname[sizeof(addr->sun_path)-2] = '\0';`  or `addr->sun_path[sizeof(addr->sun_path)-1] = '\0';`

stgraber:
Wouldn't fixing this issue actually break backward compatibility (new monitor client, old monitor process)?

brauner:
@stgraber, it's something i considered as well but haven't tested since I'm not yet convinced that the change itself is actually correct.

stgraber:
jenkins: test this please

brauner:
So while there is indeed something to fix here this will break us all over the place since connecting to older sockets would become a problem. Also, the fix as it is is not complete. I'm looking into this and if I come up with a solution I will pick the compatible parts of your commit and add it with your signed-of-by line to my branch if that's alright.

0x0916:
brauner opens another pr(https://github.com/lxc/lxc/pull/1539) to fix the issue, so close this one.

brauner:
@0x0916, I included a `Reported-by`-line for you into one of my patches in the other branch. If you're unhappy with that, yell. :)

PR #1539, opened 2017-05-06
bugfixes: {segfaults, hashes, abstract unix sockets}

brauner:

### DIFF
--- a/src/lxc/af_unix.c
+++ b/src/lxc/af_unix.c
@@ -55,8 +55,9 @@ int lxc_abstract_unix_open(const char *path, int type, int flags)
 
 	addr.sun_family = AF_UNIX;
 
-	len = strlen(&path[1]) + 1;
-	if (len >= sizeof(addr.sun_path) - 1) {
+	len = strlen(&path[1]);
+	/* do not enforce \0-termination */
+	if (len >= sizeof(addr.sun_path)) {
 		close(fd);
 		errno = ENAMETOOLONG;
 		return -1;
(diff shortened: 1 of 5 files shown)


src/lxc/monitor.c
```diff
@@ -153,36 +153,52 @@ int lxc_monitor_close(int fd)
...
+		ERROR("failed to create name for monitor socket");
 		return -1;
 	}
 
+	/* Note: snprintf() will \0-terminate addr->sun_path on the 106th byte
+	 * and so the abstract socket name has 105 "meaningful" characters. This
+	 * is absolutely intentional. For further info read the comment for this
+	 * function above!
+	 */
 	len = sizeof(addr->sun_path) - 1;
```
brauner: Note that we could easily guarantee that monitors created by 105-char liblxc's are attachable for 106-char liblxc's. It's the reverse case that is not easily dealt with. It's better to stick with the 105-char version.

brauner: This is obviously only a problem for lxcpaths >= sizeof(sun_path) (or sizeof(sun_path) - 3 to be pedantic. -1 for the starting \0-byte and another -1 for the terminating \0-byte that we use to be able to print out the abstract socket name in the logs, and another -1 for the \0-byte caused by incorrect checks in af_unix.c).

brauner:
jenkins: test this please

brauner:
jenkins: test this please

src/lxc/af_unix.c
```diff
@@ -55,7 +55,7 @@ int lxc_abstract_unix_open(const char *path, int type, int flags)
 
 	addr.sun_family = AF_UNIX;
 
-	len = strlen(&path[1]) + 1;
+	len = strlen(&path[1]);
 	if (len >= sizeof(addr.sun_path) - 1) {
```
0x0916: In these low-level functions, we allow the socket to be 107 length, so the check should be `if (len >= sizeof(addr.sun_path)) {`

brauner: `sizeof(addr.sun_path)` will give you `108` so `len(bla) >= 108 - 1` is correct.

0x0916: yes, I know, but `len(bla)` can be 107, I think it is valid socket name at this situation. The current code will return `ENAMETOOLONG` if `len = 107`.

brauner: Yeah, at least if we don't want to enforce \0-termination. Which we probably shouldn't here.

brauner: Thanks!

PR #1547, opened 2017-05-10
start: remove umount2() 

brauner:
I really fail to see the point of this and git {blame, log -S} don't really
enlighten me on the reason for this as well. But I might be dense. The way I
see it the only thing this line achieves is causing trouble when the container
is started as root because the umount2() call will umount e.g.
/usr/lib/x86_64-linux-gnu/lxc in case it is a mountpoint on the host. Note,
this is because lxc_spawn() is still called in the hosts namespaces.

Closes https://github.com/lxc/lxd/#3255.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1287,23 +1287,23 @@ int prepare_ramfs_root(char *root)
 	char *p2;
 
 	if (realpath(root, nroot) == NULL)
-		return -1;
+		return -errno;
 
 	if (chdir("/") == -1)
-		return -1;
+		return -errno;
 
 	/*
 	 * We could use here MS_MOVE, but in userns this mount is
 	 * locked and can't be moved.
 	 */
-	if (mount(root, "/", NULL, MS_REC | MS_BIND, NULL)) {
+	if (mount(root, "/", NULL, MS_REC | MS_BIND, NULL) < 0) {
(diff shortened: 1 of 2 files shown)


hallyn:
Can you verify that lxc-start on fedora or arch still works with this?

hallyn:
Background: the kernel enforces that the parent of fs must not be MS_SHARED else pivot_root fails. In the original commit, 

commit cc28d0b0a66bd956645dc7b8fc85b917711f2472
Author: Serge Hallyn <serge.hallyn@canonical.com>
Date:   Wed Dec 19 23:58:44 2012 -0600

    Support MS_SHARED /

I mentioned that arch and fedora had this problem at the time, but I would think that Ubuntu 16.10 onward would also satisfy that.

commit 91e93c71c7487bce07eada582397af1104d64a8e
Author: Andrey Vagin <avagin@gmail.com>
Date:   Sun Oct 5 01:49:16 2014 +0400

    lxc: don't call pivot_root if / is on a ramfs

Made changes to that, but for the unrelated case of root being on ramfs.

In any case, the umount2 by itself seems useless, so this seems ok.

brauner:
Right, the kernel part was obvious what I didn't know was why `umount2()` was called in the hosts mount namespace. This must have been changed sometime after I assume. Thanks for digging this up.

PR #1551, opened 2017-05-11
utils: fix num parsing functions

brauner:
Suggested-by: Benedikt Rosenkranz beluro@web.de

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -23,6 +23,7 @@
 
 #include "config.h"
 
+#include <ctype.h>
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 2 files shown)


brauner:
The reasoning behind this is:

- The POSIX standard mandates that `errno` cannot be unconditionally inspected if the function does not return an error code (e.g. `-1` or something). So only when specific values are returned can we rely on errno otherwise the implementation is free to change it.
- `strtol()` and friends are bath-sh*t crazy implementations which you can gather from the fact that the whole world writes wrappers around them.

src/lxc/utils.c
```diff
@@ -2040,14 +2050,14 @@ int lxc_safe_long(const char *numstr, long int *converted)
 
 	errno = 0;
 	sli = strtol(numstr, &err, 0);
-	if (errno > 0)
+	if (errno == ERANGE && (sli == LONG_MAX || sli == LONG_MIN))
 		return -errno;
 
-	if (!err || err == numstr || *err != '\0')
-		return -EINVAL;
+	if (errno != 0 && sli == 0)
```
hallyn: I see the manpage alluding to == 0 if there wer eno digits at all, but it only says that nptr original value will be returned in endptr, nothing about return value.  Do you have a pointer to better documentation about that?

PR #1562, opened 2017-05-15
monitor: delete unneccessory include file

0x0916:
we want to export `monitor.h`, `#include "conf.h"` will cause error
and it is unneccessory so just delete it.


brauner:
jenkins: test this please

PR #1571, opened 2017-05-17
Fix build of conf.c when capabilities is disabled

ffontaine:
Do not call idmaptool_on_path_and_privileged with CAP_SETUID and
CAP_SETGID if HAVE_LIBCAP is not defined otherwise compilation fails

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3469,8 +3469,10 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
 	 * will protected it by preventing another user from being handed the
 	 * range by shadow.
 	 */
+	#if HAVE_LIBCAP
 	uidmap = idmaptool_on_path_and_privileged("newuidmap", CAP_SETUID);
 	gidmap = idmaptool_on_path_and_privileged("newgidmap", CAP_SETGID);
+	#endif
 	if (uidmap > 0 && gidmap > 0) {
 		DEBUG("Functional newuidmap and newgidmap binary found.");
 		use_shadow = true;


brauner:
jenkins: test this please

src/lxc/conf.c
```diff
@@ -3469,8 +3469,10 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
 	 * will protected it by preventing another user from being handed the
 	 * range by shadow.
 	 */
+	#if HAVE_LIBCAP
 	uidmap = idmaptool_on_path_and_privileged("newuidmap", CAP_SETUID);
 	gidmap = idmaptool_on_path_and_privileged("newgidmap", CAP_SETGID);
+	#endif
```
brauner: If we do not have caps we should probably still give the benefit of the doubt and proceed but give a warning.

brauner: I actually wonder why this would need fixing in the first place? `CAP_SET{G,U}ID` are now `#ifndef`ined when not available and `idmaptool_on_path_and_privileged()` does:

```	
        #if HAVE_LIBCAP && LIBCAP_SUPPORTS_FILE_CAPABILITIES
	/* Check if it has the CAP_SETUID capability. */
	if ((cap & CAP_SETUID) &&
	    lxc_file_cap_is_set(path, CAP_SETUID, CAP_EFFECTIVE) &&
	    lxc_file_cap_is_set(path, CAP_SETUID, CAP_PERMITTED)) {
		DEBUG("The binary \"%s\" has CAP_SETUID in its CAP_EFFECTIVE "
		      "and CAP_PERMITTED sets.", path);
		fret = 1;
		goto cleanup;
	}

	/* Check if it has the CAP_SETGID capability. */
	if ((cap & CAP_SETGID) &&
	    lxc_file_cap_is_set(path, CAP_SETGID, CAP_EFFECTIVE) &&
	    lxc_file_cap_is_set(path, CAP_SETGID, CAP_PERMITTED)) {
		DEBUG("The binary \"%s\" has CAP_SETGID in its CAP_EFFECTIVE "
		      "and CAP_PERMITTED sets.", path);
		fret = 1;
		goto cleanup;
	}
	#else
	/* If we cannot check for file capabilities we need to give the benefit
	 * of the doubt. Otherwise we might fail even though all the necessary
	 * file capabilities are set.
	 */
	DEBUG("Cannot check for file capabilites as full capability support is "
	      "missing. Manual intervention needed.");
	fret = 1;
	#endif
```

ffontaine: Sorry, you're right, I will close this PR. I found the issue on lxc-2.0.8 and I didn't see that the bug was fixed by bc5b27d6f6d166d2a6df47982cbe36041ce6b735.

brauner: np :)

PR #1573, opened 2017-05-18
utils: fix ppc64le builds

brauner:
I suspect that there's a glibc bug on ppc64le. Both clang and gcc a very
unhappy when you return -errno from these functions. Instead, let's return
concrete errno numbers, e.g. -EINVAL.


brauner:
jenkins: test this please

PR #1576, opened 2017-05-19
DO NOT add the handles of adjust winsize when the 'stdin' is not a tty

lifeng68:
I think it is not necessary that  the 'stdin'  must be a tty device.  For example, we can pass a FIFO fd as the ‘stdin’ parameter, then we can do more flexible work on the container console.

### DIFF
--- a/src/lxc/console.c
+++ b/src/lxc/console.c
@@ -121,6 +121,11 @@ struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
 	ts->masterfd = dstfd;
 	ts->sigfd = -1;
 
+	if (!isatty(srcfd)) {
+		INFO("fd %d does not refer to a tty device", srcfd);
+		return ts;
+	}
+
 	/* add tty to list to be scanned at SIGWINCH time */
 	lxc_list_add_elem(&ts->node, ts);
 	lxc_list_add_tail(&lxc_ttys, &ts->node);
@@ -128,20 +133,20 @@ struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
 	sigemptyset(&mask);
 	sigaddset(&mask, SIGWINCH);
 	if (sigprocmask(SIG_BLOCK, &mask, &ts->oldmask)) {
-		SYSERROR("failed to block SIGWINCH.");
+		SYSERROR("failed to block SIGWINCH");
 		ts->sigfd = -1;
 		return ts;
 	}
 
 	ts->sigfd = signalfd(-1, &mask, 0);
 	if (ts->sigfd < 0) {
-		SYSERROR("failed to get signalfd.");
+		SYSERROR("failed to create signal fd");
 		sigprocmask(SIG_SETMASK, &ts->oldmask, NULL);
 		ts->sigfd = -1;
 		return ts;
 	}
 
-	DEBUG("%d got SIGWINCH fd %d", getpid(), ts->sigfd);
+	DEBUG("process %d created signal fd %d to handle SIGWINCH events", getpid(), ts->sigfd);
 	return ts;
 }
 
@@ -647,16 +652,17 @@ int lxc_console(struct lxc_container *c, int ttynum,
 	struct lxc_epoll_descr descr;
 	struct termios oldtios;
 	struct lxc_tty_state *ts;
+	int istty = 0;
 
-	if (!isatty(stdinfd)) {
-		ERROR("stdin is not a tty");
-		return -1;
-	}
-
-	ret = lxc_setup_tios(stdinfd, &oldtios);
-	if (ret) {
-		ERROR("failed to setup tios");
-		return -1;
+	istty = isatty(stdinfd);
+	if (istty) {
+		ret = lxc_setup_tios(stdinfd, &oldtios);
+		if (ret) {
+			ERROR("failed to setup terminal properties");
+			return -1;
+		}
+	} else {
+		INFO("fd %d does not refer to a tty device", stdinfd);
 	}
 
 	ttyfd = lxc_cmd_console(c->name, &ttynum, &masterfd, c->config_path);
@@ -684,8 +690,10 @@ int lxc_console(struct lxc_container *c, int ttynum,
 	ts->winch_proxy = c->name;
 	ts->winch_proxy_lxcpath = c->config_path;
 
-	lxc_console_winsz(stdinfd, masterfd);
-	lxc_cmd_console_winch(ts->winch_proxy, ts->winch_proxy_lxcpath);
+	if (istty) {
+		lxc_console_winsz(stdinfd, masterfd);
+		lxc_cmd_console_winch(ts->winch_proxy, ts->winch_proxy_lxcpath);
+	}
 
 	ret = lxc_mainloop_open(&descr);
 	if (ret) {
@@ -733,8 +741,10 @@ err2:
 	close(masterfd);
 	close(ttyfd);
 err1:
-	tcsetattr(stdinfd, TCSAFLUSH, &oldtios);
+	if (istty) {
+		if (tcsetattr(stdinfd, TCSAFLUSH, &oldtios) < 0)
+			WARN("failed to reset terminal properties: %s.", strerror(errno));
+	}
 
 	return ret;
 }
-


lifeng68:
Jenkins: test this please.

brauner:
jenkins: test this please

src/lxc/console.c
```diff
@@ -121,27 +121,29 @@ struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
...
+			return ts;
+		}
 
-	ts->sigfd = signalfd(-1, &mask, 0);
-	if (ts->sigfd < 0) {
-		SYSERROR("failed to get signalfd.");
-		sigprocmask(SIG_SETMASK, &ts->oldmask, NULL);
-		ts->sigfd = -1;
-		return ts;
+		DEBUG("%d got SIGWINCH fd %d", getpid(), ts->sigfd);
```
brauner: While you're at it could you change this to:

    DEBUG("process %d created signal fd %d to handle SIGWINCH events", getpid(), ts->sigfd)

src/lxc/console.c
```diff
@@ -121,27 +121,29 @@ struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
 	ts->masterfd = dstfd;
 	ts->sigfd = -1;
 
-	/* add tty to list to be scanned at SIGWINCH time */
-	lxc_list_add_elem(&ts->node, ts);
-	lxc_list_add_tail(&lxc_ttys, &ts->node);
+	if (isatty(srcfd)) {
```
brauner: From a coding style point of view it would be nicer to just do:

```
if (!isatty(srcfd)) {
        INFO("fd %d does not refer to a tty device", srcfd);
        return ts;
}
```
This lets you avoid the additional indentation level afterwards.

src/lxc/console.c
```diff
@@ -647,16 +649,15 @@ int lxc_console(struct lxc_container *c, int ttynum,
...
-		ERROR("stdin is not a tty");
-		return -1;
-	}
-
-	ret = lxc_setup_tios(stdinfd, &oldtios);
-	if (ret) {
-		ERROR("failed to setup tios");
-		return -1;
+	istty = isatty(stdinfd);
+	if (istty) {
```
brauner: If you want to remove that test you need to at least add an `INFO()` to inform users that they are not attaching to a tty device.

lifeng68:
@brauner ，the use-case :
For example, we can pass FIFO fds as the ‘stdin’ and 'stdout', we can control the container by the FIFOs.
It can be used in conjunction with other software developping by our company. Through the software, you can control the container from a different machine.

brauner:
@lifeng68, sounds reasonable. I'll leave a few more comments to be addressed.

src/lxc/console.c
```diff
@@ -121,27 +121,29 @@ struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
...
-	lxc_list_add_tail(&lxc_ttys, &ts->node);
+	if (isatty(srcfd)) {
+		/* add tty to list to be scanned at SIGWINCH time */
+		lxc_list_add_elem(&ts->node, ts);
+		lxc_list_add_tail(&lxc_ttys, &ts->node);
+
+		sigemptyset(&mask);
+		sigaddset(&mask, SIGWINCH);
+		if (sigprocmask(SIG_BLOCK, &mask, &ts->oldmask)) {
+			SYSERROR("failed to block SIGWINCH.");
```
brauner: Can you please remove the `.` at the end of the `SYSERROR()` statement?

src/lxc/console.c
```diff
@@ -121,27 +121,29 @@ struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
...
-	sigemptyset(&mask);
-	sigaddset(&mask, SIGWINCH);
-	if (sigprocmask(SIG_BLOCK, &mask, &ts->oldmask)) {
-		SYSERROR("failed to block SIGWINCH.");
-		ts->sigfd = -1;
-		return ts;
-	}
+		ts->sigfd = signalfd(-1, &mask, 0);
+		if (ts->sigfd < 0) {
+			SYSERROR("failed to get signalfd.");
```
brauner: Remove the `.` at the end please and add reword to `failed to create signal fd`.

src/lxc/console.c
```diff
@@ -647,16 +649,15 @@ int lxc_console(struct lxc_container *c, int ttynum,
...
-
-	ret = lxc_setup_tios(stdinfd, &oldtios);
-	if (ret) {
-		ERROR("failed to setup tios");
-		return -1;
+	istty = isatty(stdinfd);
+	if (istty) {
+		ret = lxc_setup_tios(stdinfd, &oldtios);
+		if (ret) {
+			ERROR("failed to setup tios");
```
brauner: Please reword to "failed to setup terminal properties"

src/lxc/console.c
```diff
@@ -733,8 +736,8 @@ int lxc_console(struct lxc_container *c, int ttynum,
 	close(masterfd);
 	close(ttyfd);
 err1:
-	tcsetattr(stdinfd, TCSAFLUSH, &oldtios);
+	if (istty)
+		tcsetattr(stdinfd, TCSAFLUSH, &oldtios);
```
brauner: Can you also issue an `INFO()` or `WARN()` including the output from `strerror(errno)` right here in case `tcsetattr()` failed. Might be useful in debugging.

brauner:
So it definitely looks like something we could support but we need to consider two things:
1. possible security implications of allowing non-tty-referring fds to be used
2. while technically not an api break it is definitely an api extension so we need to make sure that this doesn't mess with current use-cases

src/lxc/console.c
lifeng68: Thanks for your suggestions, I had fixed it

src/lxc/console.c
lifeng68: Thanks, i added an "INFO("fd %d does not refer to a tty device", stdinfd);" for inform the users

src/lxc/console.c
lifeng68: fixed

src/lxc/console.c
lifeng68: fixed

src/lxc/console.c
lifeng68: fixed

src/lxc/console.c
lifeng68: fixed

lifeng68:
@brauner , I do not quite understand what is 'security implications'  about,  check the permissions of the non-tty-referring fds? In the  use-case i mentioned above, the caller of the method `lxc_console` should check whether there is permission to open the FIFOs. So, i think the lxc  just need provide the method   `lxc_console ` as a library function， the check of the fds should be the caller's job.
I had tested the current use-cases already. The 'lxc-console' and 'lxc-start -F ...' ， the modification had no effect on them, and from the analysis of the code logic, it would not mess with current use-cases.

brauner:
So existing users might be confused by the fact that the console API function now accepts non-tty-referring arguments for std{I,out,err}. But the public API never advertises the fact that it only allows tty-referring fds so I think we're good on that front.

My security concerns are about allowing arbitrary file descriptors to be attached to the containers tty fd. Which this commit essentially would allow you to do by special casing tty-referring fds internally. I'm cautious as we've seen privilege escalation caused in his area.

brauner:
So I convinced myself that this patch should not increase the  console attack vector anymore as it is now. Here's my reasoning for future reference so that people can tell at me later on. :)

In contrast to attaching to a containers namespaces without allocating a fresh tty where you're susceptible to tty pushback attacks the situation here is different. The tty file descriptors which are used by console have been setup and opened before the container init has started and so should be pristine. Also, there's no immediate danger of leaking host fds into the container by allowing this.

brauner:
jenkins: test this please

lifeng68:
jenkins: test this please

lifeng68:
@brauner  ,The Testsuite failed， the erro info as below, is anything wrong with the` GPG keyserver`?
```
FAIL: lxc-tests: /usr/bin/lxc-test-apparmor-mount
---
/usr/sbin/deluser: The user `lxcunpriv' does not exist.
Setting up the GPG keyring
ERROR: Unable to fetch GPG key from keyserver.
lxc-create: tmp.kOeCIWBTOn: lxccontainer.c: create_run_template: 1295 container creation template for tmp.kOeCIWBTOn failed
lxc-create: tmp.kOeCIWBTOn: tools/lxc_create.c: main: 318 Error creating container tmp.kOeCIWBTOn
Container is not defined
umount: /sys/kernel/security/apparmor/features/mount: not mounted
---
```

brauner:
@lifeng68, yeah, this has nothing to do with your branch. But let's retry.

brauner:
jenkins: test this please

brauner:
I'll perform some local test and then will merge.

PR #1577, opened 2017-05-19
Fix the bug of 'ts->stdoutfd' did not fill with parameters 'stdoutfd'

lifeng68:
The 'ts->stdoutfd' did not fill with parameters 'stdoutfd', it used the default value "0".
I think we should fix that mistake.

### DIFF
--- a/src/lxc/console.c
+++ b/src/lxc/console.c
@@ -683,6 +683,7 @@ int lxc_console(struct lxc_container *c, int ttynum,
 	ts->escape = escape;
 	ts->winch_proxy = c->name;
 	ts->winch_proxy_lxcpath = c->config_path;
+	ts->stdoutfd = stdoutfd;
 
 	lxc_console_winsz(stdinfd, masterfd);
 	lxc_cmd_console_winch(ts->winch_proxy, ts->winch_proxy_lxcpath);


lifeng68:
Jenkins: test this please.

brauner:
jenkins: test this please

src/lxc/console.c
```diff
@@ -683,6 +683,7 @@ int lxc_console(struct lxc_container *c, int ttynum,
 	ts->escape = escape;
 	ts->winch_proxy = c->name;
 	ts->winch_proxy_lxcpath = c->config_path;
+	ts->stdoutfd = stdoutfd;
```
brauner: What about

    ts->stdinfd
    ts->stderrfd

?

lifeng68: ts->stdinfd had been filled in the method `lxc_console_sigwinch_init`
```
struct lxc_tty_state *lxc_console_sigwinch_init(int srcfd, int dstfd)
{
	sigset_t mask;
	struct lxc_tty_state *ts;

	ts = malloc(sizeof(*ts));
	if (!ts)
		return NULL;

	memset(ts, 0, sizeof(*ts));
	ts->stdinfd = srcfd;
	ts->masterfd = dstfd;
	ts->sigfd = -1;
        .......

```
The ts->stderrfd did not use at all. becase the stdin\stdout\stderr of container had been duplicated to the same slave of pty . So the stdout and stderr of container are combined. 
```
{
    if (fd < 0)
        return 0;

    if (isatty(STDIN_FILENO))
            return -1;
        }

    if (isatty(STDOUT_FILENO))
            return -1;
        }

    if (isatty(STDERR_FILENO))
            return -1;
        }

    return 0;
} 
```
So, we can get the stdout\stderr info of container by reading the master of pty (ts->masterfd) . And then we just need to write the info to the ts->stdout, the info include the 'stdout' and 'stderr'.
```
int lxc_console_cb_tty_master(int fd, uint32_t events, void *cbdata,
		struct lxc_epoll_descr *descr)
{
	struct lxc_tty_state *ts = cbdata;
	char buf[1024];
	int r, w;

	if (fd != ts->masterfd)
		return 1;

	r = lxc_read_nointr(fd, buf, sizeof(buf));
	if (r <= 0)
		return 1;

	w = lxc_write_nointr(ts->stdoutfd, buf, r);
	if (w <= 0) {
		return 1;
	} else if (w != r) {
		SYSERROR("failed to write");
		return 1;
	}

	return 0;
}
```

brauner: Ah right. I should remember that. I wrote this code. :)

PR #1584, opened 2017-05-23
[RFC]: pre-allocate and retrieve pre-allocated pts devices from container

brauner:
This implements `lxc.pts.allocate` which when set to a non-zero value will pre-allocate a number of pty devices from the container's private devpts mount inside of the container's namespaces **before** the init binary is spawned. `LXC` caches the master and slave side file descriptors for later retrieval. This guarantees that the `fd`s have not been tampered with by the program running inside of the container.

In order to retrieve a set of cached pty {master,slave} fd pair from the container this PR adds

    int lxc_cmd_get_pty(const char *name, int *ttynum, int *masterfd, int *slavefd, const char *lxcpath);

a to the lxc commands api. The function will retrieve a pty {master,slave} pair and hand it back to the caller. The fds can then be used to e.g. run an interactive command on a pts device that actually exists inside the container's namespaces at `/dev/pts/<n>`. This is different from the current `lxc.tty` fds which refer to prepared `/dev/tty<n>` devices which likely are running a login session from the init binary.

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -142,6 +142,7 @@ static const char *lxc_cmd_str(lxc_cmd_t cmd)
 		[LXC_CMD_GET_CONFIG_ITEM] = "get_config_item",
 		[LXC_CMD_GET_NAME]        = "get_name",
 		[LXC_CMD_GET_LXCPATH]     = "get_lxcpath",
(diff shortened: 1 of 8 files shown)


brauner:
This will hopefully soon be done way smarter.

brauner:
Closing in favor of a different implementation.

Issue #140, opened 2014-01-17
Can we get a template for Kali Linux ?


lethargic87:
hi there
I was wondering if there is a template for Kali (new BackTrack) in the way?!
As you know this distro is mostly used for penetration and security reasons and it is not suitable for everyday use so we use it mostly in VMs! I think a template for this specific distro can help a lot of people as full Virtualization uses a lot of resources and is cumbersome! 
thanks

ak2766:
I'm late to the party.

Even if Kali can run inside LXC, I doubt it will be possible to do any wifi hacking - educate me please!

ak.

postfix:
https://testpurposes.net/2015/05/05/using-lxc-containers-to-partially-replace-virtual-machines/

Kentoseth:
Going back to this old issue (I am working on sweeping up LXC issues), I don't think using Kali within LXC is as good an idea as using a full VM layer (these are my own security considerations - but everyone is entitled to their own opinions).

With that being said, @lethargic87 have you tested out @vince06fr's suggestion?

This issue is quite old, so it will be a good idea to resolve it.

brauner:
As it seems that kali can simply be created from debian I think we can close this issue. Feel free to reopen if I'm wrong.

PR #1587, opened 2017-05-26
Kali Linux template for LXC (modified debian)

binkybear:
This is a modification of lxc-debian.in and debian template to support Kali Linux.  

Tested on Ubuntu 16.

Supports following issue: https://github.com/lxc/lxc/issues/140

Signed-off-by: binkybear <binkybear@nethunter.com>

### DIFF
--- a/config/templates/Makefile.am
+++ b/config/templates/Makefile.am
@@ -20,6 +20,8 @@ templatesconfig_DATA = \
 	gentoo.common.conf \
 	gentoo.moresecure.conf \
 	gentoo.userns.conf \
+	kali.common.conf \
+	kali.userns.conf \
 	nesting.conf \
 	opensuse.common.conf \
 	opensuse.userns.conf \
--- a/config/templates/kali.common.conf.in
+++ b/config/templates/kali.common.conf.in
@@ -1,0 +1,28 @@
+# This derives from the global common config
+lxc.include = @LXCTEMPLATECONFIG@/common.conf
+
+# Doesn't support consoles in /dev/lxc/
+lxc.devttydir =
+
+# When using LXC with apparmor, the container will be confined by default.
+# If you wish for it to instead run unconfined, copy the following line
+# (uncommented) to the container's configuration file.
+#lxc.aa_profile = unconfined
+
+# If you wish to allow mounting block filesystems, then use the following
+# line instead, and make sure to grant access to the block device and/or loop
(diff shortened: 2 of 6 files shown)


hallyn:
First, it seems you're re-enabling root password logins and re-setting the default root password of 'toor'.  I'd certainly prefer you drop that.

Also, looking at the diff between lxc-kali.in and lxc-debian.in, I can't help but feel we'd be better served with a more generalized lxc-debian-base.in which takes a few variables, and by default installs a debian system;  debian derivatives could then use it and make any other tweaks they need.

Maybe that actually doesn't work out long term as debian changes and derivatives do not...  But this sure seems like a lot of copied boilerplate which, if there is a security issue inherent in all of them, will need to be reproduced, probably be forgotten, hence not fixed.

So I guess this (commonality) is something for later discussion.  But please reply about the root password.

binkybear:
@hallyn - By default in all Kali distributions the password is "toor" and it has always been that way (since Backtrack/before Kali).  Kali is not meant to be a "daily driver" and so by default it usually has SSH and default passwords set.

I see your point though in having to manage two different templates that are pretty much based off the same template.  Changes to Debian template may not necessarily make it into Kali template.

hallyn:
Please make the default no password, with a --set-root-pwd option. I.e.

Lxc-create -t kali -n c1 -- --set-root-pwd

Will make toor the pwd

hallyn:
Do you have any objections to my doing these changes for you?

binkybear:
@hallyn - Sorry been a bit busy.  I have no objections if you have time to make changes.  Thank you!

hallyn:
I've updated the template to follow the same env variables as lxc-fedora and lxc-centos (which I don't much like, but is better than having even more incompatible password schemes).  The result is in branch 'kali2' in github.com/hallyn/lxc.  Now when I try to use it, I get:

Installing packages:
chroot: failed to run command ‘apt-key’: No such file or directory
chroot: failed to run command ‘apt-get’: No such file or directory
chroot: failed to run command ‘apt-get’: No such file or directory

hallyn:
So, to be clear, I'm leaving it at that, you can build on that or not, as you choose.

brauner:
We've been keeping this PR around for a couple of months. I'm closing now but if you add another PR with the changes requested we won't object to merging.

Issue #984, opened 2016-04-21
(2.0.0 regression) container name not found by lxc-ls and lxc-attach (>= kernel 4.5) (lxc-info and lxc-console seem to work)


Auch999:
If a(ny) (unprivileged) container is started like this (without having been 'created' beforehand): 

$ lxc-start -n containerName -F/-d -f /path/to/conf -s 'lxc.id_map = u 0 100000 65536' -s 'lxc.id_map = g 0 100000 65536' -s 'lxc.rootfs = /path/to/rootfs' -s 'lxc.init_cmd = /path/to/initcmd' ,

the container name ('containerName' here) is not recognized by the lxc tools. 'lxc-ls' shows nothing at all and 'lxc-attach -n containerName' prints 'Error: container containerName is not defined'

Note: The container starts fine and is well usable in foreground mode. 
(Linux 4.5 and up (same behavior with 4.6.0-rc4) (CHECKPOINT_RESTORE enabled))

Is there a workaround to start the container in background mode and attach to it later on?

How do the lxc tools determine which containers are currently running btw.?

EDIT: lxc-info seems to work: (container name m3all1)

[The cgroups have been created like this by root:
 cgm create all username
 cgm chown all username uid gid

 The user moved the pid of the the shell into the cgroups before launching lxc-start:
 cgm movepid all username $$ && lxc-start -n m3all1 -F -f /path/to/conf -s 'lxc.id_map = u 0 100000 65536' -s 'lxc.id_map = g 0 100000 65536' -s 'lxc.rootfs = /path/to/rootfs' -s 'lxc.init_cmd = /path/to/initcmd' ]

$ cgm movepid all username $$
$ lxc-ls
$ lxc-attach -n m3all1
Error: container m3all1 is not defined
$ lxc-info -n m3all1
Name:           m3all1
State:          RUNNING
PID:            4664
IP:             10.0.3.13
CPU use:        0.01 seconds
BlkIO use:      0 bytes
Memory use:     1.17 MiB
KMem use:       804.00 KiB

How can I attach to it?

After having downgraded to lxc 1.1.5, I can confirm that the above works and it is definitely a bug in lxc 2.0.0 .

panic-amt:
I'm also seeing this after upgrading to Linux 4.5.0.

tobiaspal:
This bug is still present in the latest LXC 2.0.4.

jordemort:
[...]

PR #1591, opened 2017-05-27
lxc-ls: return all containers by default, new filter - list only defi…

grzgrzgrz3:
…ned containers.

1.x.x lxc-ls list all containers, new lxc-ls (2.x.x) implementation is
ignoring not defined containers.

related with issue: #984

Signed-off-by: Grzegorz Grzywacz <grzgrzgrz3@gmail.com>

### DIFF
--- a/doc/lxc-ls.sgml.in
+++ b/doc/lxc-ls.sgml.in
@@ -55,6 +55,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
       <arg choice="opt">--frozen</arg>
       <arg choice="opt">--running</arg>
       <arg choice="opt">--stopped</arg>
+      <arg choice="opt">--defined</arg>
       <arg choice="opt">-f</arg>
       <arg choice="opt">-F <replaceable>format</replaceable></arg>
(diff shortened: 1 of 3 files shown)


hallyn:
jenkins: test this please

brauner:
@hallyn, @stgraber, can you make a call on this. I don't have strong opinions. :)

hallyn:
I'm in favor of restoring the old behavior.

hallyn:
IMO we should revert the previous behavior, and if we consider
the new behavior desirable (which I don't), we should announce
it with the other changes for 3.0.

Otherwise I'd worry large sites will be caught off guard.

stgraber:
agree with @hallyn

brauner:
@grzgrzgrz3, could you restore the old behavior, that lists all containers, even undefined ones?

grzgrzgrz3:
What you mean restore the old behavior, this commit actually restore it, list all containers by default.

If i understood correctly:
old behavior - list all containers 
new behavior - list only defined containers

lxc-ls 1.x.x:
old behavior
lxc-ls   master branch:
new behavior
lxc-ls this merge:
old behavior

brauner:
Ah, I misunderstood. I thought added a flag for the old behavior. :) Right on!

PR #1592, opened 2017-05-28
idmap improvements

brauner:

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3473,16 +3473,35 @@ cleanup:
 	return fret;
 }
 
+int lxc_map_ids_exec_wrapper(void *args)
+{
+	execl("/bin/sh", "sh", "-c", (char *)args, (char *)NULL);
+	return -1;
+}
+
 int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
 {
 	struct id_map *map;
 	struct lxc_list *iterator;
 	enum idtype type;
+	char u_or_g;
 	char *pos;
-	int euid;
-	int ret = 0, use_shadow = 0;
-	int uidmap = 0, gidmap = 0;
-	char *buf = NULL;
+	int euid, fill, left;
+	char cmd_output[MAXPATHLEN];
+	/* strlen("new@idmap") = 9
+	 * +
+	 * strlen(" ") = 1
+	 * +
+	 * LXC_NUMSTRLEN64
+	 * +
+	 * strlen(" ") = 1
+	 *
+	 * We add some additional space to make sure that we really have
+	 * LXC_IDMAPLEN bytes available for our the {g,u]id mapping.
+	 */
+	char mapbuf[9 + 1 + LXC_NUMSTRLEN64 + 1 + LXC_IDMAPLEN] = {0};
+	int ret = 0, uidmap = 0, gidmap = 0;
+	bool use_shadow = false, had_entry = false;
 
 	euid = geteuid();
 
@@ -3507,17 +3526,12 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
 		return -1;
 	}
 
-	for (type = ID_TYPE_UID; type <= ID_TYPE_GID; type++) {
-		int left, fill;
-		bool had_entry = false;
-		if (!buf) {
-			buf = pos = malloc(LXC_IDMAPLEN);
-			if (!buf)
-				return -ENOMEM;
-		}
-		pos = buf;
+	for (type = ID_TYPE_UID, u_or_g = 'u'; type <= ID_TYPE_GID;
+	     type++, u_or_g = 'g') {
+		pos = mapbuf;
+
 		if (use_shadow)
-			pos += sprintf(buf, "new%cidmap %d", type == ID_TYPE_UID ? 'u' : 'g', pid);
+			pos += sprintf(mapbuf, "new%cidmap %d", u_or_g, pid);
 
 		lxc_list_for_each(iterator, idmap) {
 			/* The kernel only takes <= 4k for writes to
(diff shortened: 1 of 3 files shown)


src/lxc/conf.c
```diff
@@ -3384,27 +3384,32 @@ int lxc_assign_network(const char *lxcpath, char *lxcname,
...
-	if (ret < 0)
-		SYSERROR("writing id mapping");
-	closeret = fclose(f);
-	if (closeret)
-		SYSERROR("writing id mapping");
-	return ret < 0 ? ret : closeret;
+
+	ret = write(fd, buf, buf_size);
+	if (ret != buf_size)
+		SYSERROR("failed to write %cid mapping to \"%s\"",
```
hallyn: SYSERROR could be misleading here under weird circumstances.  I wouldn't expect it to happen here, but in theory write could return 0 < ret < buf_size, with errno not specified.  SYSERROR in that case would report an errno from a past operation.

src/lxc/conf.c
```diff
@@ -3384,27 +3384,32 @@ int lxc_assign_network(const char *lxcpath, char *lxcname,
...
+			 idtype == ID_TYPE_UID ? 'u' : 'g', path);
+	else
+		ret = 0;
+
+	close(fd);
+
+	return ret;
```
hallyn: If write returned 0, you'll claim success here.

src/lxc/conf.c
```diff
@@ -3506,17 +3522,12 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
...
-		int left, fill;
-		bool had_entry = false;
-		if (!buf) {
-			buf = pos = malloc(LXC_IDMAPLEN);
-			if (!buf)
-				return -ENOMEM;
-		}
-		pos = buf;
+	for (type = ID_TYPE_UID, u_or_g = 'u'; type <= ID_TYPE_GID;
+	     type++, type == ID_TYPE_GID ? (u_or_g = 'g') : (u_or_g = '@')) {
```
hallyn: Any reason not to just make this

for (type = ID_TYPE_UID, u_or_g = 'u'; type <= ID_TYPE_GID;
      type++, u_or_g = 'g') {

src/lxc/conf.c
```diff
@@ -3541,22 +3552,83 @@ int lxc_map_ids(struct lxc_list *idmap, pid_t pid)
 		if (!had_entry)
 			continue;
 
-		if (!use_shadow) {
-			ret = write_id_mapping(type, pid, buf, pos - buf);
+		/* Try to catch the ouput of new{g,u}idmap to make debugging
+		 * easier.
+		 */
+		if (use_shadow) {
```
hallyn: The rest of this looks good, thanks.

src/lxc/conf.c
brauner: If somehow the `enum` is altered I didn't want to just rewrite the gid mapping. But I think I just directly check that size(id types) == 2.

brauner:
> Can I ask about the motivation here? Admittedly there appears to be no good reason to use FILE* instead of int fd here, but this seems like begging for an accidental regression. It still may be worth it, but I'm wondering whether there is something specific you're fixing/improving here (increasing its urgency).

I'm not sure if you're asking for the motivation of this whole patch or only the first part that rewrites `write_id_mapping()`. I'll just answer to both. :)
- `write_id_mapping()`: So the old code contained

```
        ret = fwrite(buf, buf_size, 1, f);
	if (ret < 0)
		SYSERROR("writing id mapping");
	closeret = fclose(f);
	if (closeret)
		SYSERROR("writing id mapping");
```
which to me was a little confusing. First of all, it checked for `ret < 0` from `fwrite()` whereas `fwrite()` returns either a short-write or `0` on error. The manpage does not mention it returning `< 0`. This also is a problem in the return statement `ret < 0 ? ret : closeret`. The next thing I found confusing is why the return value of `fclose()` matters here and why it is returned. So I wanted to get rid of this confusion and give `write_id_mapping()` a cleaner and lower-level interface by using `read()`/`write()` directly. Presumably it would also be worth it to use `lxc_write_nointr()` when writing the mapping to exclude `EINTR`.
- gathering `new{g,u}idmap` output: Without gathering new{g,u}idmap`'s `std{err,out}` output we never know what went wrong which makes debugging very unpleasant. Did we get `EINVAL`, did we get that this mapping is not allowed? Since using `new{g,u}idmap` is the usual case it would be nice to be properly informed about what actually went wrong. (This technically also affects all cases where we call out to `lxc-usernsexec` but that's future work.)

src/lxc/conf.c
brauner: I'll reset `errno = 0` before the call.

src/lxc/conf.c
brauner: fixed

src/lxc/conf.c
hallyn: Ok that should be fine.  It could still become -EINTR during normal run but that's ok.

src/lxc/utils.c
```diff
@@ -2246,3 +2246,61 @@ int lxc_unstack_mountpoint(const char *path, bool lazy)
 
 	return umounts;
 }
+
+int run_command(char *buf, size_t buf_size, int (*child_fn)(void *), void *args)
+{
+	pid_t child;
+	int ret, fret, pipefd[2];
+	ssize_t bytes;
+
```
hallyn: Best to do a 
 if (buf_size > 0 && buf) buf[0] = '\0';

because note that right now if pipe() fails here, lxc_map_ids() will print the un-initialized buffer out in its error message.

brauner: done

brauner:
@hallyn, updated

hallyn:
Thanks.

PR #1593, opened 2017-05-28
Add a new hook named privileged-start

superboum:
Hi everyone,

**I'm proposing a new hook named "privileged-start" that is run inside the namespaces but before the capability drop. I use it to start systemd without `CAP_SYS_ADMIN` on a kernel with cgroup namespaces.**

It's quite difficult to run systemd without `CAP_SYS_ADMIN` as you have to pre-mount everything. I found a very detailed article named [LXC containers without CAP_SYS_ADMIN under Debian Jessie ](https://blog.iwakd.de/lxc-cap_sys_admin-jessie) that explain the steps. Unfortunately, the way cgroup are managed in this article doesn't work if your kernel support cgroup namespaces.
In other words, `lxc.mount.auto = cgroup` doesn't work as it does nothing if it detects cgroup ns support.

When you have cgroup namespaces, you can't use `lxc.mount.entry` as you'll mount the wrong cgroup hierarchy. It appears that you have to mount cgroup AFTER unsharing the cgroup namespace but BEFORE the capability drop, as `CAP_SYS_ADMIN` is required to mount them. So my PR is about the hook that I added just before the capability drop to solve this problem.

So, now, I'm able to start a container with the following configuration:

```ini
# path of the file on the host: /var/lib/lxc/test/config
# ... (the default configuration)
lxc.mount.auto = proc:rw sys:rw
lxc.mount.entry = tmpfs dev/shm tmpfs rw,nosuid,nodev,create=dir 0 0
lxc.mount.entry = tmpfs run tmpfs rw,nosuid,nodev,mode=755,create=dir 0 0
lxc.mount.entry = tmpfs run/lock tmpfs rw,nosuid,nodev,noexec,relatime,size=5120k,create=dir 0 0
lxc.mount.entry = tmpfs run/user tmpfs rw,nosuid,nodev,mode=755,size=50m,create=dir 0 0
lxc.hook.priv-start = /bin/mountall
lxc.cap.drop = sys_admin
```

And with this hook:

```bash
#!/bin/bash
# path of the file on the host: /var/lib/lxc/test/rootfs/bin/mountall
mount -t tmpfs tmpfs /sys/fs/cgroup
mkdir -p /sys/fs/cgroup/systemd
mount -t cgroup -o rw,nosuid,nodev,noexec,relatime,xattr,name=systemd cgroup /sys/fs/cgroup/systemd
```

*I don't want to keep `CAP_SYS_ADMIN` as this capability is often considered as [the new root](https://lwn.net/Articles/486306/) by authorizing [too many things](http://man7.org/linux/man-pages/man7/capabilities.7.html).*


brauner:
Wait, why can't this be done with `lxc.hook.mount`? This runs explicitly before any caps are dropped.

superboum:
I thought that `lxc.mount.hook` was run before `CLONE_NEW*` but reality is more complex. 

Actually, `lxc.mount.hook` is run in the `ipc`, `mnt`, `net`, `pid`, `user` and `uts` namespaces but not in the `cgroup` namespace that is unshared later (cf `src/lxc/start.c:833`). Or I want to mount my cgroup hierarchy in the `cgroup` namespace.

So, my hook might no be the best solution to fix this problem. I'm open to suggestions :)

superboum:
Again, I was wrong :(
So in the right order :

   1. Clone `mnt`, `pid`, `uts`, `ipc`, `net`
   2. Cgroup setup limits
   3. Mount autodev
   4. Mount entries
   5. pivot_root
   6. drop capabilities
   7. Cgroup setup limits (again ?)
   8. Unshare `cgroup`

As `cap_sys_admin` is dropped before `cgroup` unsharing, this solution can't work. During my tests, I was able to mount the cgroup hierarchy but after a check it was the hierarchy of the host... I was trapped by the fact that I got a permission denied during `lxc.mount.hook` but everything went well with my new (useless) hook. So I'll close this issue and think to another way to solve the problem. LXC might drop the capabilities later ?

brauner:
Well, it's intricate but it should be doable by shuffling around when what is run. Basically one would have to carefully split `lxc_setup()`. It's crucial that the cgroups have been created before unsharing the cgroup namespace.

PR #1596, opened 2017-05-29
confile: fix parsing

### LINKED ISSUE #1595: lxc-start-ephemeral in 2.0.8 creates configuration it can't parse
# Required information

 * Distribution: Ubuntu 14.04
 * Distribution version: Ubuntu 14.04.5 LTS
 * The output of
   * `lxc-start --version` : 2.0.8
   * `uname -a` : Linux vagrant-ubuntu-trusty-64 3.13.0-119-generic #166-Ubuntu SMP Wed May 3 12:18:55 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

# Issue description

Given I have a container and I want to create a new ephemeral container from it, a configuration is being created, that leads to lxc erroring out on parsing it and thus lxc tooling not being able to interact with the new container. This wasn't the case in 2.0.7 but is now the case in 2.0.8.

# Steps to reproduce

`lxc-create -n foobar -t ubuntu`
`lxc-start-ephemeral -n foobar-ephemeral -o foobar -d`
`lxc-ls -f` leads to:
`lxc-ls: parse.c: lxc_file_for_each_line: 57 Failed to parse config: lxc.ephemeral =`

# Information to attach

- the configuration file looks as follows
```
cat /var/lib/lxc/foobar-ephemeral/config
# Template used to create this container: /usr/share/lxc/templates/lxc-ubuntu
# Parameters passed to the template:
# Template script checksum (SHA-1): 4d7c613c3c0a0efef4b23917f44888df507e662b
# For additional config options, please look at lxc.container.conf(5)
# Uncomment the following line to support nesting containers:
#lxc.include = /usr/share/lxc/config/nesting.conf
# (Be aware this has security implications)
# Common configuration
lxc.include = /usr/share/lxc/config/ubuntu.common.conf
# Container specific configuration
lxc.rootfs.backend = dir
lxc.arch = amd64
# Network configuration
lxc.network.type = veth
lxc.network.link = lxcbr0
lxc.network.flags = up
lxc.network.hwaddr = 00:16:3e:9b:4e:ab
lxc.utsname =
lxc.utsname = foobar-ephemeral
lxc.rootfs =
lxc.rootfs = /var/lib/lxc/foobar-ephemeral/rootfs
lxc.network.0.hwaddr = 00:16:3e:6f:fd:69
lxc.hook.pre-mount =
lxc.hook.pre-mount = /var/lib/lxc/foobar-ephemeral/pre-mount
lxc.ephemeral =
lxc.ephemeral = 1
```

I believe the line `lxc.ephemeral =` is the one triggering the configuration parse issue.

brauner:
So this has grown into a major rework of the configuration system. My reasons for this are plenty most of all:
- there were/are a bunch of configuration items that had a set but no get method
- there were a bunch of configuration items that could not be cleared via the api
- adding new configuration items was error prone since the set method was enforced by the callback system but neither the get nor clear methods were

This changes to a new callback-based configuration system that is enforced via the lxc_config_t struct. If someone wants to implement a new configuration item things should now be more straightforward. To add a new configuration items one needs to:
- create a new member in the lxc_config_t struct and implement
  1. a set method
  2. a get method
  3. a clear method
simple as that.

The rework is also necessary because we recently fixed a bug whereby lots of config items could be set but couldn't be unset via clear_config_item(). This was especially problematic when copying containers because the config items where stacked onto each other if they were changed. The same problem for api calls to clear_config_item() + set_config_item(). The fix however did not take into account that this means that whitespace character for all of these config items need to be accounted for. This commit does.

Closes #1595.

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -1358,7 +1358,7 @@ static int set_config_hooks(const char *key, const char *value,
 	if (config_value_empty(value))
 		return lxc_clear_hooks(lxc_conf, key);
 
-	if (strcmp(key, "lxc.hook") == 0) {
+	if (strcmp(key + 4, "hook") == 0) {
 		ERROR("lxc.hook cannot take a value");
 		return -1;
 	}
@@ -1368,23 +1368,23 @@ static int set_config_hooks(const char *key, const char *value,
 		return -1;
 	}
 
-	if (strcmp(key, "lxc.hook.pre-start") == 0)
+	if (strcmp(key + 9, "pre-start") == 0)
 		return add_hook(lxc_conf, LXCHOOK_PRESTART, copy);
-	else if (strcmp(key, "lxc.hook.pre-mount") == 0)
+	else if (strcmp(key + 9, "pre-mount") == 0)
 		return add_hook(lxc_conf, LXCHOOK_PREMOUNT, copy);
-	else if (strcmp(key, "lxc.hook.autodev") == 0)
+	else if (strcmp(key + 9, "autodev") == 0)
 		return add_hook(lxc_conf, LXCHOOK_AUTODEV, copy);
-	else if (strcmp(key, "lxc.hook.mount") == 0)
+	else if (strcmp(key + 9, "mount") == 0)
 		return add_hook(lxc_conf, LXCHOOK_MOUNT, copy);
-	else if (strcmp(key, "lxc.hook.start") == 0)
+	else if (strcmp(key + 9, "start") == 0)
 		return add_hook(lxc_conf, LXCHOOK_START, copy);
-	else if (strcmp(key, "lxc.hook.stop") == 0)
+	else if (strcmp(key + 9, "stop") == 0)
 		return add_hook(lxc_conf, LXCHOOK_STOP, copy);
-	else if (strcmp(key, "lxc.hook.post-stop") == 0)
+	else if (strcmp(key + 9, "post-stop") == 0)
 		return add_hook(lxc_conf, LXCHOOK_POSTSTOP, copy);
-	else if (strcmp(key, "lxc.hook.clone") == 0)
+	else if (strcmp(key + 9, "clone") == 0)
 		return add_hook(lxc_conf, LXCHOOK_CLONE, copy);
-	else if (strcmp(key, "lxc.hook.destroy") == 0)
+	else if (strcmp(key + 9, "destroy") == 0)
 		return add_hook(lxc_conf, LXCHOOK_DESTROY, copy);
 
 	SYSERROR("Unknown key: %s", key);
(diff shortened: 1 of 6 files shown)


brauner:
Actually I'm not sure whether we should consider whitespace a problem or not. I mean one could also make the reasonable argument that it must always be `lxc.some_int_value =` and not `lxc.some_int_value=\n\t`. The latter allows for more flexibility. @stgraber, @hallyn

brauner:
We should only consier `!value || !strlen(value)` valid.

stgraber:
Looks good to me but will give some time for @hallyn to take a look

brauner:
All of the changes are very very incremental which is intentional so that we reduce regression potential as much as possible.

hallyn:
Dude.

hallyn:
Are there any libc's out there which will crash on free(NULL)?

brauner:
@hallyn, never heard of such a thing but then we would crash in a bunch of places. :)

Also, what? Too much little commits. I actually think this provides some value given that the whole config handling has gotten really mucky. When people implemented something new, they often forgot either the getter, the clearer or both. I think this way of using the jump table for everything makes it very straightforward to grasp what is needed and how this works. But feel free to disagree and argue against it. The amount of commits should play no role in the decision. :)

hallyn:
No no i like the bunch of little commits.  It's looking for the more complicated/longer ones in the set of little commits that's stressful :)

Ok - I thought I had seen us crash on free(NULL) before, but can't reproduce it with my libc's here, so let's assume i'm wrong.  Crazy assumption.

brauner:
So one of the bigger commits is that I implemented a new parser for lxc.id_map and a getter for lxc.id_map which we didn't have before.

src/lxc/confile.c
```diff
@@ -2967,6 +2967,54 @@ static int lxc_get_item_network(struct lxc_conf *c, char *retv, int inlen)
 	return fulllen;
 }
 
+static int lxc_get_idmaps(struct lxc_conf *c, char *retv, int inlen)
+{
+	struct lxc_list *it;
+	int len, listlen, ret;
+	int fulllen = 0;
```
hallyn: What's up with fulllen?

brauner: What do you mean? The `strprint()` function expects `len`, `fulllen` and a bunch of others to be defined. I don't find it the best way to do it either.

hallyn:
> 
> 
> 
> > @@ -2967,6 +2967,54 @@ static int lxc_get_item_network(struct lxc_conf *c, char *retv, int inlen)
>  	return fulllen;
>  }
>  
> +static int lxc_get_idmaps(struct lxc_conf *c, char *retv, int inlen)
> +{
> +	struct lxc_list *it;
> +	int len, listlen, ret;
> +	int fulllen = 0;
> 
> What do you mean? The `strprint()` function expects `len`, `fulllen` and a bunch of others to be defined. I don't find it the best way to do it either.

D'oh, I (a) misread the s*print and (b) forgot we had that
abomination :)

thanks.

PR #1599, opened 2017-05-30
conf: use minimal {g,u}id map

### LINKED ISSUE #1598: only map minimally required set of {g,u}ids when executing function in new user namespace
https://bugs.launchpad.net/ubuntu/+source/lxd/+bug/1692111

brauner:
Afaict, userns_exec_1() is only used to operate based on privileges for the
user's own {g,u}id on the host and for the container root's unmapped {g,u}id.
This means we require only to establish a mapping from:
- the container root {g,u}id as seen from the host -> user's host {g,u}id
- the container root -> some sub{g,u}id

The former we add, if the user did not specifiy a mapping. The latter we
retrieve from the ontainer's configured {g,u}id mappings.

Closes #1598.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -4628,108 +4628,158 @@ static int run_userns_fn(void *data)
 {
 	struct userns_fn_data *d = data;
 	char c;
-	// we're not sharing with the parent any more, if it was a thread
 
+	/* Close write end of the pipe. */
 	close(d->p[1]);
+
+	/* Wait for parent to finish establishing a new mapping in the user
+	 * namespace we are executing in.
+	 */
 	if (read(d->p[0], &c, 1) != 1)
 		return -1;
+
+	/* Close read end of the pipe. */
 	close(d->p[0]);
+
+	/* Call function to run. */
 	return d->fn(d->arg);
 }
 
-/*
- * Add ID_TYPE_UID/ID_TYPE_GID entries to an existing lxc_conf,
- * if they are not already there.
- */
-static struct lxc_list *idmap_add_id(struct lxc_conf *conf,
-		uid_t uid, gid_t gid)
+static struct id_map *mapped_hostid_entry(unsigned id, struct lxc_conf *conf,
+					  enum idtype idtype)
 {
-	int hostuid_mapped = mapped_hostid(uid, conf, ID_TYPE_UID);
-	int hostgid_mapped = mapped_hostid(gid, conf, ID_TYPE_GID);
-	struct lxc_list *new = NULL, *tmp, *it, *next;
-	struct id_map *entry;
+	struct lxc_list *it;
+	struct id_map *map;
+	struct id_map *retmap = NULL;
 
-	new = malloc(sizeof(*new));
-	if (!new) {
-		ERROR("Out of memory building id map");
+	lxc_list_for_each(it, &conf->id_map) {
+		map = it->elem;
+		if (map->idtype != idtype)
+			continue;
+
+		if (id >= map->hostid && id < map->hostid + map->range) {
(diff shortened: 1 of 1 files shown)


stgraber:
What do we use to invoke the templates? If that's userns_exec_1 then that'd be a problem, but since Jenkins passed, I guess it's not :)

brauner:
Hm, why would that be a problem? In simple terms, you want to be able to chown stuff to the root {g,u}id of the container as seen from the host. Meaning, if uid 0 in the container is represented via 65536 outside, you establish a new mapping of 65536 -> $(id -u) and also map in the container root, meaning 0 -> 165536.

brauner:
So I added this to the `TRACE()` log and with this patch what happens is the following. :)

```
      lxc-start alp2 20170530160908.766 DEBUG    lxc_conf - conf.c:idmap_add_id:4716 - adding uid mapping: nsid 65536 hostid 1000 range 1
      lxc-start alp2 20170530160908.767 DEBUG    lxc_conf - conf.c:idmap_add_id:4741 - adding gid mapping: nsid 65536 hostid 1000 range 1
      lxc-start alp2 20170530160908.767 TRACE    lxc_conf - conf.c:userns_exec_1:4871 - establishing uid mapping for "17213" in new user namespace: nsuid 65536 - hostid 1000 - range 1
      lxc-start alp2 20170530160908.767 TRACE    lxc_conf - conf.c:userns_exec_1:4871 - establishing gid mapping for "17213" in new user namespace: nsuid 65536 - hostid 1000 - range 1
      lxc-start alp2 20170530160908.767 TRACE    lxc_conf - conf.c:userns_exec_1:4871 - establishing uid mapping for "17213" in new user namespace: nsuid 0 - hostid 165536 - range 65536
      lxc-start alp2 20170530160908.767 TRACE    lxc_conf - conf.c:userns_exec_1:4871 - establishing gid mapping for "17213" in new user namespace: nsuid 0 - hostid 165536 - range 65536
```

brauner:
Ooh, not LXD templates, LXC templates... No, `userns_exec_1()` doesn't execute templates at least that's what I gather from the codebase. :)

src/lxc/conf.c
```diff
@@ -4625,108 +4625,158 @@ static int run_userns_fn(void *data)
...
+	}
+
+	if (!retmap)
+		return NULL;
+
+	retmap = malloc(sizeof(*retmap));
+	if (!retmap)
+		return NULL;
+
+	memset(retmap, 0, sizeof(*retmap));
```
hallyn: Any advantage to doing this memset before the memcpy?

src/lxc/conf.c
```diff
@@ -4625,108 +4625,158 @@ static int run_userns_fn(void *data)
...
+		goto on_error;
+	entry = hostuid_idmap;
+	if (!hostuid_idmap) {
 		hostuid_mapped = find_unmapped_nsuid(conf, ID_TYPE_UID);
 		if (hostuid_mapped < 0)
-			goto err;
-		tmp = malloc(sizeof(*tmp));
-		if (!tmp)
-			goto err;
+			goto on_error;
```
hallyn: Does this leak @tmp?

brauner: Yeah, tmp should show in under `on_error`.

src/lxc/conf.c
brauner: removed.

src/lxc/conf.c
brauner: added under `on_error()` and also removed a duplicate call to adding `tmp` to the list I just detected

brauner:
@hallyn, updated.

PR #1606, opened 2017-06-01
call lxc_setup() after unshare(CLONE_NEWCGROUP)

brauner:
When the running kernel supports cgroup namespaces and users want to manually
set up cgroups via lxc.hook.mount before the init binary starts the cgroup
namespace needs to be already unshared. Otherwise the view on the cgroup mounts
is wrong. This commit places the call to lxc_setup() after the
LXC_SYNC_POST_CGROUP barrier.

Before this commit, the tty fds we allocate from a fresh devpts instance in the
container's namespaces before the init binary starts were referring to the
host's cgroup namespace since lxc_setup() was called before
unshare(CLONE_NEWCGROUP). Although not a security risk at this point since
setns() restricts its calls to /proc/<self>/ns files it's still better to do it
*after* the cgroup namespace has been unshared.

Adding a Suggested-by line for the lxc.mount.hook fix for Quentin.

Closes #1597.

Suggested-by: Quentin Dufour <quentin@dufour.tk>


brauner:
@hallyn, this is the last one and should hopefully be fairly uncontroversial. :)

superboum:
Thanks for your patch, without no doubt, it's a cleaner solution than mine. 

After some tests, I might have found a problem with the cgroup `device` subsystem. By running `lxc_setup()` after the cgroup logic, you need to whitelist the devices you want to mount in your container. That was not needed before.

brauner:
Does this affect devices that are currently setup by default or only new devices that you want to add? Also, I assume you're running privileges containers, right?

superboum:
Let's assume we have the following configuration :

```ini
# ... default configuration
lxc.rootfs = /dev/vg_root/lv_lxc_[container].cdg
lxc.cgroup.devices.deny = a
## /dev/null and zero
lxc.cgroup.devices.allow = c 1:3 rwm lxc.cgroup.devices.allow = c 1:5 rwm 
## full 
lxc.cgroup.devices.allow = c 1:7 rwm 
## consoles 
lxc.cgroup.devices.allow = c 5:0 rwm 
## /dev/{,u}random 
lxc.cgroup.devices.allow = c 1:8 rwm lxc.cgroup.devices.allow = c 1:9 rwm 
## /dev/pts/* 
lxc.cgroup.devices.allow = c 5:2 rwm lxc.cgroup.devices.allow = c 136:* rwm 
## rtc 
lxc.cgroup.devices.allow = c 254:0 rm 
## hpet
lxc.cgroup.devices.allow = c 10:228 rm 
```

 * Without the patch, I can boot the container
 * With the patch I can't boot the container as I get an `Operation not permitted` due to the fact that my device is not in the cgroup whitelist. I must add a `lxc.cgroup.devices.allow = xxx` line to allow my device in order to boot my container.

brauner:
Oh, I know why. Thanks for spotting this!

superboum:
Is there any way I can help to merge this pull request ? Could it be a solution to artificially add the rootfs to the cgroup whitelist to keep backward compatibility ?

brauner:
@superboum, I'm looking at this again now, sorry.

brauner:
@superboum, I updated the PR. Care to test the new approach? :)

brauner:
@hallyn, so the idea is to basically be more fine-grained wrt to setting up cgroups and the container itself:
1. create new cgroups and move cloned child into new cgroups
2. unshare the cgroup namespace in the child
3. apply device cgroup limits in the parent
4. exec the init binary

superboum:
A little bit late but I've tested LXC repository's HEAD (ebc5243 - after the merge) on my test environment and it works fine ! I removed the line `lxc.cgroup.devices.allow` and my container boots without any problem. Thanks !

Issue #1589, opened 2017-05-27
need to expose `lxc_log_init`


0x0916:
When we use only `liblxc`, we have no way to get DEBUG/TRACE
information. So we need to export function `lxc_log_init`.

More discuss: https://github.com/lxc/lxc/pull/1586

PR #1586, opened 2017-05-25
log: add trace feature for liblxc

0x0916:
When we use only `liblxc`, we have no way to get DEBUG/TRACE
information.

With this patch, we can set `LXC_TRACE=xxx`(xxx may be `trace`,`debug`,
`ifno`, `notice`, `warn` and `error` etc.). and then we can see more info
from stdout.

This is very useful for developers when they debug programs which only
use `liblxc`.

For example:

run lxc tests without `LXC_TRACE=trace`

```bash
$ ./src/tests/lxc-test-startone
all lxc_container tests passed for lxctest1
Destroyed container lxctest1
```

and run lxc tests with `LXC_TRACE=trace`

```bash
$ LXC_TRACE=trace ./src/tests/lxc-test-startone
[INFO] lxc: lsm/lsm.c: lsm_init: 62 Initialized LSM security driver AppArmor
[INFO] lxc: lsm/lsm.c: lsm_init: 62 Initialized LSM security driver AppArmor
[DEBUG] lxc: monitor.c: lxc_monitord_spawn: 330 Going to wait for pid 16607.
[DEBUG] lxc: monitor.c: lxc_monitord_spawn: 349 Trying to sync with child process.
[INFO] lxc: start.c: lxc_check_inherited: 243 Closed inherited fd: 3.
[INFO] lxc: start.c: lxc_check_inherited: 243 Closed inherited fd: 4.
[DEBUG] lxc: monitor.c: lxc_monitord_spawn: 364 Successfully synced with child process.
[DEBUG] lxc: monitor.c: lxc_monitord_spawn: 333 Finished waiting on pid 16607.
[INFO] lxc: lxccontainer.c: do_lxcapi_start: 802 Attempting to set proc title to [lxc monitor] /usr/local/var/lib/lxc lxctest1
[INFO] lxc: start.c: lxc_check_inherited: 243 Closed inherited fd: 3.
[DEBUG] lxc: monitor.c: lxc_monitord_spawn: 330 Going to wait for pid 16611.
[DEBUG] lxc: monitor.c: lxc_monitord_spawn: 349 Trying to sync with child process.
[INFO] lxc: start.c: lxc_check_inherited: 243 Closed inherited fd: 3.
[INFO] lxc: start.c: lxc_check_inherited: 243 Closed inherited fd: 4.
[DEBUG] lxc: monitor.c: lxc_monitord_spawn: 364 Successfully synced with child process.
[...]

PR #1615, opened 2017-06-05
API: expose function lxc_log_init

0x0916:
Fix: https://github.com/lxc/lxc/issues/1589
More discuss: https://github.com/lxc/lxc/pull/1586

### DIFF
--- a/src/lxc/lxccontainer.h
+++ b/src/lxc/lxccontainer.h
@@ -1017,6 +1017,22 @@ int list_active_containers(const char *lxcpath, char ***names, struct lxc_contai
  */
 int list_all_containers(const char *lxcpath, char ***names, struct lxc_container ***cret);
 
+struct lxc_log {
+	const char *name;
+	const char *lxcpath;
+	const char *file;
+	const char *priority;
+	const char *prefix;
+	bool quiet;
+};
+
+/*!
+ *\brief Initialize the log
+ *
+ *\param log lxc log configuration.
+ */
+int lxc_log_init(struct lxc_log *log);
+
 /*!
  * \brief Close log file.
  */
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -44,6 +44,7 @@
 #include "log.h"
 #include "caps.h"
 #include "utils.h"
+#include "lxccontainer.h"
 
 /* We're logging in seconds and nanoseconds. Assuming that the underlying
  * datatype is currently at maximum a 64bit integer, we have a date string that
@@ -543,9 +544,7 @@ extern void lxc_log_enable_syslog(void)
  * Called from lxc front-end programs (like lxc-create, lxc-start) to
  * initalize the log defaults.
  */
-extern int lxc_log_init(const char *name, const char *file,
-			const char *priority, const char *prefix, int quiet,
-			const char *lxcpath)
+extern int lxc_log_init(struct lxc_log *log)
 {
 	int lxc_priority = LXC_LOG_PRIORITY_ERROR;
 	int ret;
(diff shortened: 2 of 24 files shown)


src/lxc/lxccontainer.h
```diff
@@ -1017,6 +1017,22 @@ int list_active_containers(const char *lxcpath, char ***names, struct lxc_contai
...
+	int quiet;
+	const char *lxcpath;
+};
+
+/*!
+ *\brief Initialize the log
+ *
+ *\param log lxc log configuration.
+ */
+int lxc_log_init(struct lxc_log_conf log);
```
brauner: Don't you want this to be a pointer?

0x0916: We will not change anything that `log` holds. If you prefer pointer, I will change it later today, thanks

brauner: It's just good practice to not copy the whole struct into the function. By using a pointer the function becomes a lot more lightweight. If you really think that we'll never change anything that is inside the struct then make it `const struct *log`. But the thing about a struct is that you can extend it later on.

src/lxc/lxccontainer.h
```diff
@@ -1017,6 +1017,22 @@ int list_active_containers(const char *lxcpath, char ***names, struct lxc_contai
  */
 int list_all_containers(const char *lxcpath, char ***names, struct lxc_container ***cret);
 
+struct lxc_log_conf {
```
brauner: What do you think about simply `lxc_log`?

src/lxc/lxccontainer.h
```diff
@@ -1017,6 +1017,22 @@ int list_active_containers(const char *lxcpath, char ***names, struct lxc_contai
...
+	const char *name;
+	char *file;
+	char *priority;
+	const char *prefix;
+	int quiet;
```
brauner: This should be a `bool`.

src/lxc/lxccontainer.h
```diff
@@ -1017,6 +1017,22 @@ int list_active_containers(const char *lxcpath, char ***names, struct lxc_contai
...
+	const char *lxcpath;
```
brauner: Sot the struct should be:

    struct lxc_log {
            const char *name;
            const char *lxcpath;
            char *logfile;
            char *loglevel;
            char *logprefix;
            bool quiet;
    };

Why aren't the other not marked as `const`? Any reason for that?

brauner:
@hallyn, @stgraber, since this is a an API addition I would like a review and ack from all maintainers before merging. :)

brauner:
@Blub, feel free to chime in as well. :)

src/lxc/lxccontainer.h
brauner: Note that the order of the struct is important and should not be changed later as otherwise callers of the shared library might have the wrong offset if they were compiled against a different version.

src/lxc/lxccontainer.h
0x0916: thank you. change it to pointer.

src/lxc/lxccontainer.h
0x0916: ok changed.

src/lxc/lxccontainer.h
0x0916: updated

src/lxc/lxccontainer.h
0x0916: thanks, all of the struct member marked as `const`, except the `quiet`.

hallyn:
Thanks, looks good to me.

I'm going to wait for @stgraber to merge it as it's late and I'm failing to think clearly on whether we need to do anything API-wise.  We're just adding another symbol, so we just should bump minor version to make sure newer userspace knows whether it can use the fn?

brauner:
jenkins: test this please

stgraber:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
Ok, as I see it all maintainers agree that this is a reasonable suggestion. I tested it locally and it all seems to work fine. Logging from the API now works nicely. Thanks!

PR #1618, opened 2017-06-08
start: kill lxc-monitord

brauner:
#### Introduction

A LXC container's lifecycle is regulated by the states STARTING, RUNNING, STOPPING, STOPPED, ABORTING. These states are tracked in the LXC handler and can be checked via approriate functions in the command socket callback system. (The freezer stages are not part of a container's lifecycle since they are not recorded in the LXC handler. This might change in the future but given that the freezer controller will be removed from future cgroup implementations it is unlikely.) So far, LXC was using an external helper to track the states of a container (lxc-monitord). This solution was error prone. For example, the external state server would hang in various scenarios that seemed to be caused by either very subtle internal races or irritation of the external state server by signals.

#### Proposal
LXC will switch from an external state monitor (lxc-monitord) which serves as a state server for state clients to a native implementation using the individual container's command socket. This solution was discussed and outlined by @stgraber and @brauner during the Denver, CO LX{C,D} sprint.

#### Outline
The LXC handler will gain an additional field to track state clients. In order for a state client to receive state notifications from the command server he will need to register himself via the lxc_cmd_state_server() function in the state client list. The state client list will be served by lxc_set_state() during the container's lifecycle. lxc_set_state() will also take care of removing any clients from the state list in the LXC handler once the requested state has been reached and sent to the client.
In order to prevent races between adding and serving new state clients the state client list and the state field in the LXC handler will be protected by a lock.

#### Corner Cases
- Adding new clients when the container is in STARTING state:
  Assume that lxc_cmd_state_server() acquires the lock on the states client list and detects that the current state is STARTING. In this case lxc_cmd_state_server() will only add clients that want to wait on the RUNNING, STOPPING, or STOPPED states. (Clients that want to wait on STARTING state will be immediately served by lxc_cmd_state_server() without adding them to the state client list.)
- Registering clients that want to wait on RUNNING state when the container already is in RUNNING state: This shouldn't be a problem. If the container is already in RUNNING state and lxc_cmd_state_server() holds the lock on the state clients lists it will detect that the container is already running and immediately report back to the caller without adding a new client to the list. If lxc_set_state() succeeds in taking the lock and setting a new state the only possible states that can follow RUNNING are STOPPING, STOPPED, or ABORTING. In all cases the container will eventually be shut down, the command socket closed, and the client list free()ed. So there's no danger of accumulating zombie clients. Furthermore, the implementation updates the state field in the lxc handler with the process lock held which should prevent moving on to another state while lxc_cmd_state_server() acquires the lock.
- Adding new clients when the container is in STOPPING state: Assume that lxc_cmd_state_server() acquires the lock on the states client list and detects that the current state is STOPPING. In this case
  lxc_cmd_state_server() will only add clients that want to wait on the STOPPED state.
  (Clients that want to wait on STOPPING will be immediately served by lxc_cmd_state_server() without adding them to the state client list.)
- Adding new clients when the container is in STOPPED state: Assume that lxc_cmd_state_server() acquires the lock on the states client list and detects that the current state is STOPPED. In this case  lxc_cmd_state_server() will not add new clients. (Clients that want to wait on STOPPED will be immediately served by lxc_cmd_state_server() without adding them to the state client list.)
- Adding new clients when the container is in STOPPED state: Assume that lxc_cmd_state_server() acquires the lock on the states client list and detects that the current state is ABORTING. In this case
  lxc_cmd_state_server() will not add new clients. (Clients that want to wait on ABORTING will be immediately served by lxc_cmd_state_server() without adding them to the state client list.)

#### Implementation Details
- The lock is always taken on the whole list and the state field in the lxc handler struct while lxc_cmd_state_server() or lxc_set_state() are adding or serving clients. This prevents a bunch of weird races.

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -297,12 +297,20 @@ static int lxc_cmd(const char *name, struct lxc_cmd_rr *cmd, int *stopped,
 		return -1;
 
 	sock = lxc_abstract_unix_connect(path);
+	TRACE("command %s tries to connect to \"@%s\"",
+	      lxc_cmd_str(cmd->req.cmd), offset);
 	if (sock < 0) {
-		if (errno == ECONNREFUSED)
+		if (errno == ECONNREFUSED) {
+			TRACE("command %s failed to connect to \"@%s\": %s",
+			      lxc_cmd_str(cmd->req.cmd), offset,
+			      strerror(errno));
 			*stopped = 1;
-		else
-			SYSERROR("Command %s failed to connect to \"@%s\".",
-				 lxc_cmd_str(cmd->req.cmd), offset);
+		} else {
+			SYSERROR("command %s failed to connect to \"@%s\": %s",
+				 lxc_cmd_str(cmd->req.cmd), offset,
+				 strerror(errno));
+		}
+
 		return -1;
 	}
 
@@ -462,19 +470,25 @@ char *lxc_cmd_get_cgroup_path(const char *name, const char *lxcpath,
 	};
 
 	ret = lxc_cmd(name, &cmd, &stopped, lxcpath, NULL);
-	if (ret < 0)
+	if (ret < 0) {
+		TRACE("command %s failed for container \"%s\": %s.",
+		      lxc_cmd_str(cmd.req.cmd), name, strerror(errno));
 		return NULL;
+	}
 
 	if (!ret) {
-		WARN("Container \"%s\" has stopped before sending its state.", name);
+		WARN("container \"%s\" has stopped before sending its state", name);
 		return NULL;
 	}
 
 	if (cmd.rsp.ret < 0 || cmd.rsp.datalen < 0) {
-		ERROR("Command %s failed for container \"%s\": %s.",
+		ERROR("command %s failed for container \"%s\": %s",
 		      lxc_cmd_str(cmd.req.cmd), name, strerror(-cmd.rsp.ret));
 		return NULL;
 	}
+
+	TRACE("command %s successful for container \"%s\"",
+	      lxc_cmd_str(cmd.req.cmd), name);
 
 	return cmd.rsp.data;
 }
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -43,7 +43,9 @@
 #include "commands.h"
 #include "console.h"
 #include "confile.h"
+#include "lxclock.h"
 #include "mainloop.h"
+#include "monitor.h"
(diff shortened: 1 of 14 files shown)


brauner:
jenkins: test this please

stgraber:
jenkins: test this please

brauner:
jenkins: test this please

src/lxc/execute.c
```diff
@@ -111,16 +111,15 @@ static struct lxc_operations execute_start_ops = {
 };
 
 int lxc_execute(const char *name, char *const argv[], int quiet,
```
brauner: Note, this change prevents old `lxc-execute` binaries to be useable with `liblxc`'s that have a new version of `lxc_execute()`. This is fine since `lxc-execute` used internal symbols which is something we won't support in the future anymore. Also, we never intended to support old binaries + new `liblxc`.

brauner: Also, I've switched `lxc-execute` to rely on public API. A full switch to API-only will happen when we exposed a log function in our API. A PR for this already exists.

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
@stgraber, Jenkins failures are unrelated to the branch. It seems sad overall. :)

stgraber:
jenkins: test this please

brauner:
Fwiw, the failure on the AppArmor tests is not really a failure. The container is supposed to fail to start which it does but the way `lxc_wait()` reports back to the caller has changed to before. I'll tweak this a little now.

brauner:
This should be ready for review. @stgraber  @hallyn

src/lxc/commands.c
```diff
@@ -861,6 +868,143 @@ static int lxc_cmd_get_lxcpath_callback(int fd, struct lxc_cmd_req *req,
...
+	newclient = malloc(sizeof(*newclient));
+	if (!newclient)
+		return -1;
+
+	/* copy requested states */
+	memcpy(newclient->states, req->data, sizeof(newclient->states));
+	newclient->clientfd = fd;
+
+	tmplist = malloc(sizeof(*tmplist));
+	if (!tmplist)
```
hallyn: leaking newclient

brauner: Thanks, updated.

brauner:
@stgraber, @hallyn updated :)

hallyn:
Thanks.

brauner:
@hallyn, thanks for reviewing. :)

PR #1621, opened 2017-06-12
lxc-alpine: Add support for ppc64le

leitao:
Starting at version 3.6, Alpine Linux has support for ppc64le
architecture. Adding the new keys also.

Signed-off-by: Breno Leitao <breno.leitao@gmail.com>


leitao:
@jirutka FYI

PR #1626, opened 2017-06-14
new network config parser

brauner:
This implements a new network configuration parser.

Previous versions of `liblxc` had suffered from several legacy design decisions such that network configuration was kind of unintuitive. For example, to specify a network the order of the keys mattered:

```
lxc.network.type = veth
lxc.network.flags = up
```
would work fine whereas

```
lxc.network.flags = up
lxc.network.type = veth
```
would not. This obviously was caused by the fact that multiple networks were supposed to be specified by using keys without any index. Meaning that each new `lxc.network.type` key marked the configuration of a new network. Later, network configuration based on indexes was supported meaning that:

```
lxc.network.0.type = veth
lxc.network.0.flags = up

lxc.network.1.type = veth
lxc.network.1.flags = up
```
could be used to specify multiple networks. While an improvement, the ordering of the keys still mattered. 

The new network configuration parser doesn't care about ordering. It will internally simply keep the index of the network and maintain an ordered list configured networks. For example:

```
lxc.network.2.flags = up
lxc.network.2.type = veth
lxc.network.2.name=eth5

lxc.network.0.type=veth
lxc.network.0.link=lxcbr0
lxc.network.0.name=eth1
lxc.network.0.flags=up

lxc.network.3.type=veth
lxc.network.3.link=lxcbr0
lxc.network.3.name=eth3
lxc.network.3.flags=up
```

would internally be tracked as:
```
0
2
3
```


brauner:
- [x] set all individual configuration items
  This specifically means that you can set each configuration key independent of any other network configuration key. That implies that a new network configuration can be started with any key. So the dependency on key ordering is completely removed.
- [x] clear all individual configuration items
  Prior to this work one could e.g. remove a whole network definition by setting either `lxc.network` or `lxc.network.type` to an empty value. The latter option is removed since it is wildly inconsistent with user expectation. To remove a network set `lxc.network` to an empty value. For example, to clear networks `1` and `5` set `lxc.network.1` and `lxc.network.5` to an empty value.

brauner:
jenkins: test this please

brauner:
- [x] preserve backwards compatibility until LXC 3.0
When this is merged we consider specifying multiple networks without indices deprecated and print an appropriate warning. Until LXC 3.0 it is still possible to specify multiple networks via `lxc.network.type` such that each new entry of `lxc.network.type` starts a new network configuration. However, users should update their container configurations to use index based network configuration via `lxc.network.[i].*` which is order independent and has a much clearer internal interface that allows sanely setting, getting, and clearing all configuration keys.

brauner:
jenkins: test this please

brauner:
jenkins: test this please

hallyn:
Looks very good overall.  However, I do disagree with the deprecation warnings.  Supporting lxc.network.[i].X out of order is a huge improvement.  But requiring [i] is in my opinion a step backwards.  I may well have a wrapper program which creates containers which doesn't care to track [i], just knows to print the entries in order beginning with the type.

hallyn:
To be clear, the intuitive, unambiguous thing to do is to always support 'lxc.network.X' as referring to the highest defined [i].

src/lxc/conf.c
```diff
@@ -2342,8 +2343,21 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
 				return -1;
 			}
 		}
-		if (netdev->type != LXC_NET_VETH)
+
+		if (netdev->type != LXC_NET_EMPTY)
```
hallyn: Did you mean == here?

src/lxc/conf.c
```diff
@@ -2342,8 +2343,21 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
...
+			return 0;
+
+		if (netdev->type != LXC_NET_NONE)
 			return 0;
+
+		if (netdev->type != LXC_NET_VETH) {
```
hallyn: You'll never get to the code in this block since you've already ensured that netdev->type == LXC_NET_EMPTY :)  if i'm reading the patch right.

src/lxc/confile.c
```diff
@@ -695,11 +695,29 @@ static int set_config_network_type(const char *key, const char *value,
...
 		return clr_config_network_type(key, lxc_conf, data);
 
-	/* lxc.network.* without an index */
 	if (!data) {
+		/* REMOVE IN LXC 3.0:
+		 * lxc.network.* without an index
+		 */
+
+		/* We maintain a negative count for legacy network devices. */
+		ssize_t negidx = -1;
```
brauner: @hallyn, there it is. :)

brauner:
I should probably give a short summary what we've discussed wrt to deprecating `lxc.network.*` without an index on IRC yesterday. As the community might have noticed @hallyn was against deprecating the non-index and order-dependent way of specifying multiple networks where each new occurrence of `lxc.network.type` starts the definition of a new network. I am very much in favor of killing this when `LXC 3.0` will be released. But, as a sensible upstream which tries to focus on our end-users needs, @hallyn has a valid point. So instead of simply calling it right now we want the community to have a say in whether we should go on supporting the `lxc.network.*` syntax without an index or if we should kill it. So we've decided to print a deprecation warning and requiring users to report to us whether they rely on the `lxc.network.*` syntax. So if you do **need** this functionality please report back in the next couple of months. If we don't hear a bunch of people yelling at us we will very likely kill this feature starting with `LXC 3.0`.

PR #1627, opened 2017-06-15
tree-wide: priority -> level

brauner:
When we merged the new logging function for the api we exposed the log level
argument in the struct as "priority" which we actually requested to be changed
to "level" which somehow didn't happen and we missed it. Given the fact there
has been no new liblxc release let's fix it right now before it hits users.
Also, take the chance to change the terminology in the log from "priority" to
"level" globally. This is to prevent confusion with syslog's "priority"
argument which we also support.


hallyn:
Thanks, looks good

Issue #1625, opened 2017-06-14
is this a problem to send a signal when the signal is blocked in function do_lxcapi_shutdown


tanyifeng:
In function "do_lxcapi_shutdown" which is in file src/lxc/lxccontainer.c . lxc detects whether a signal is blocked:
```
 /* Detect whether we should send SIGRTMIN + 3 (e.g. systemd). */
	if (task_blocking_signal(pid, (SIGRTMIN + 3)))
		haltsignal = (SIGRTMIN + 3);
```
I wonder why still set haltsignal to SIGRTMIN + 3 when the signal is blocked ? should it be:
```
if (!task_blocking_signal(pid, (SIGRTMIN + 3)))
```

Another question is :
In function "task_blocking_signal", I think the way to detect whether the signal is blocked is wrong.
```
if (sigblk & signal)
	bret = true;
```
Cause signal is type of int, so we should detect like this:
```
if (sigblk & (1 << (signal - 1)))
	bret = true;
```

brauner:
Hey there,
When the signal is blocked it means that there's likely a signal handler installed to be able to run e.g. cleanup code before shutting down so the current version is correct.
About testing whether the bit is set. I'm not sure how testing the way you suggested would be an improvement to what we do right now.

tanyifeng:
Thanks for comment.
Actually, sigblk is a bit set,  assume that signal is 37, then we should detect whether the 37th bit of sigblk is 1 or 0, rather than using (sigblk & signal) which may have wrong result. And using function sigismember is a good way, thus:
```
if (sigismember((sigset_t)sigblk, signal))
        bret = true;
```

tanyifeng:
Sorry, using sigismember is wrong, we should use bit operation due to what i comment above.
```
if (sigblk & (1 << (signal - 1)))
	bret = true;
```

brauner:
Can you explain how `(a & b)` may have the wrong result?

tanyifeng:
when the 7th bit of sigblk is 0, and we want to detect it, so:
```
sigblk = 0x FFFF FFFF FFFF FFBF;
signal = 0x 0000 0000 0000 0007;
```
using (sigblk & signal) will get true, and (sigblk & (signal - 1)) get false. is this right?

tanyifeng:
reference to this :
[bitmask](https://unix.stackexchange.com/questions/85364/how-can-i-check-what-signals-a-process-is-listening-to)

brauner:
[...]

PR #1628, opened 2017-06-19
utils: fix the way to detect blocking signal

tanyifeng:
see issue #1625 

"sigblk" is a bitmask, so we use bit shift operators to detect whether the "signal" is blocked.

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -1928,7 +1928,7 @@ bool task_blocking_signal(pid_t pid, int signal)
 				goto out;
 	}
 
-	if (sigblk & signal)
+	if (sigblk & (1 << (signal - 1)))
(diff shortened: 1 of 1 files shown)


stgraber:
jenkins: test this please

brauner:
@tanyifeng, thanks for spotting this. I pushed another small fix on top of it that takes care we're not overflowing. :)

tanyifeng:
I'm so glad for receiving your mention, cause I'm a new employee just graduate from college, and it's my 1st commit in open source software. Thank you for your fix.

brauner:
@tanyifeng, it's all good. :) We all started at some point. As long as you send decent code our way and are willing to argue a little about it now and then it'll all be fine. :)

PR #1633, opened 2017-06-21
Fix memory leak of 'lxc_tty_state'

lifeng68:
This patch fix the memory leak of 'lxc_tty_state' while `lxc_console_sigwinch_init` returned` lxc_tty_state*` with the sigfd is -1. We did not call the `lxc_console_sigwinch_fini` to release the ` lxc_tty_state*` memory, so the memory had been lost.


brauner:
jenkins: test this please

brauner:
Cool, thanks! Need to take a closer look whether this is correct but I'm sure it is.

PR #1634, opened 2017-06-22
lxc: let it compile ok for gcc7

DengkeDu:
After gcc upgrade to 7.1, gcc7 contains a number of enhancements that help
detect buffer overflow and other forms of invalid memory accesses.

When compiling lxc with gcc7, the system outputs errors:

   ../../../lxc-2.0.4/src/lxc/bdev/lxcloop.c:297:30: error: '%s' directive
   output may be truncated writing up to 255 bytes into a region of size 95
   [-Werror=format-truncation=]

This is because in configure.ac, the flag "-Werror" let the compiler treat
the warnings as errors, in order to compile successfully, we should add the
following to the configure.ac:

   -Wno-error=format-truncation

Signed-off-by: Dengke Du <dengke.du@windriver.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -686,7 +686,7 @@ LXC_CHECK_TLS
 if test "x$GCC" = "xyes"; then
 	CFLAGS="$CFLAGS -Wall"
 	if test "x$enable_werror" = "xyes"; then
-		CFLAGS="$CFLAGS -Werror"
+		CFLAGS="$CFLAGS -Werror -Wno-error=format-truncation"
 	fi
 fi
 


DengkeDu:
jenkins: test this please

brauner:
Have you checked first that the warning isn't legitimate?

DengkeDu:
The new version of gcc can detect the buffer overflow in advance, especially for the snprintf family of formatted output functions, so I think the warning maybe isn't legitimate, the right fix is adjust the string size to fit for the max size as much as possible?

stgraber:
Isn't that just https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78969 which already mentions LXC as being affected and is treated as a compiler regression upstream?

If so, I don't think we should start papering over compiler bugs as making this warning category non-fatal may lead to legitimate issues being hidden in the future.

brauner:
Yeah, we were not going to merge this. But this is another code-path it seems to me. The last time the codepath was in the `cgfsng.c`. So I was wondering whether it might be legit this time.

stgraber:
Error shows LXC 2.0.4, so maybe it's something that got moved or changed since?

brauner:
So that codepath doesn't exist anymore since I've switched LXC to use the same low-level helpers that I implemented for LXD. They perform more rigorous checks.

PR #1639, opened 2017-06-23
confile: prepare for 2.1.

brauner:


brauner:
@0x0916, it would probably a good idea to wait for @hallyn to merge this PR otherwise you very likely have to resolve a bunch of merge conflicts.

0x0916:
@brauner ok

hallyn:
Perhaps we should (1) add a 'lxc.version' container config tag, and then ship tiny programs or sed scripts that can update container config versions.

brauner:
@hallyn, yes that sounds like a good idea. Well, I'd like to discuss this in a little more depth and this pr is not the best location I'll make a Github issue and drag you into the discussion. :D

Issue #1641, opened 2017-06-23
[TAKEN]: Community/Up-For-Grabs: tree-wide: wipe all references to lxc.network.{[i]}.* and replace with lxc.net.[i].*


brauner:
This relates to templates, documentation and configuration files. The code itself is already ready. :)

0x0916:
Assign to me.

brauner:
### Note, that this issue has been assigned to @0x0916 . In loser terms, @0x0916 called dibs please respect that. :)

PR #1649, opened 2017-06-27
wipe all references lxc.network.{[i]}.* and replace with lxc.net.[i].*

0x0916:
Fix: https://github.com/lxc/lxc/issues/1641

This pr also add the new `lxc.net` key support for function `network_new_hwaddrs`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,8 @@ LXC is configured via a simple set of keys. For example,
 - `lxc.mount.entry`
 
 LXC namespaces configuration keys by using single dots. This means complex
-configuration keys such as `lxc.network` expose various subkeys such as
-`lxc.network.type`, `lxc.network.link`, `lxc.network.ipv6`, and others for even
+configuration keys such as `lxc.net.0` expose various subkeys such as
+`lxc.net.0.type`, `lxc.net.0.link`, `lxc.net.0.ipv6`, and others for even
 more fine-grained configuration.
 
 LXC is used as the default runtime for [LXD](https://github.com/lxc/lxd),
--- a/src/lxc/confile_legacy.c
+++ b/src/lxc/confile_legacy.c
@@ -945,34 +945,33 @@ int get_config_network_legacy_item(const char *key, char *retv, int inlen,
 static int lxc_clear_nic(struct lxc_conf *c, const char *key)
 {
 	char *p1;
-	int ret, idx, i;
-	struct lxc_list *it;
-	struct lxc_netdev *netdev;
+	int idx;
+	struct lxc_list *it = NULL;
+	struct lxc_netdev *netdev = NULL;
+
+	if (lxc_list_empty(&c->network)) {
+		ERROR("network is not created for %s", key);
+		return -1;
+	}
+
+	if ((idx = get_network_netdev_idx(key)) == EINVAL)
+		netdev = lxc_list_last_elem(&c->network);
+	else {
+		lxc_list_for_each(it, &c->network) {
+			netdev = it->elem;
+			if (idx == netdev->idx)
+				break;
+			netdev = NULL;
+		}
+	}
+	if (!netdev)
+		return -1;
 
 	p1 = strchr(key, '.');
 	if (!p1 || *(p1+1) == '\0')
 		p1 = NULL;
 
-	ret = sscanf(key, "%d", &idx);
-	if (ret != 1) return -1;
-	if (idx < 0)
-		return -1;
-
-	i = 0;
-	lxc_list_for_each(it, &c->network) {
-		if (i == idx)
-			break;
-		i++;
-	}
-	if (i < idx)  // we don't have that many nics defined
-		return -1;
-
-	if (!it || !it->elem)
-		return -1;
-
-	netdev = it->elem;
-
-	if (!p1) {
+	if (!p1 && it) {
 		lxc_remove_nic(it);
 	} else if (strcmp(p1, ".ipv4") == 0) {
 		struct lxc_list *it2,*next;
@@ -989,7 +988,7 @@ static int lxc_clear_nic(struct lxc_conf *c, const char *key)
 			free(it2);
 		}
 	}
-		else return -1;
+	else return -1;
 
 	return 0;
 }
--- a/src/lxc/confile_utils.c
+++ b/src/lxc/confile_utils.c
@@ -593,9 +593,9 @@ bool lxc_config_net_hwaddr(const char *line)
 }
 
 /*
- * If we find a lxc.net.hwaddr in the original config file, we expand it in
- * the unexpanded_config, so that after a save_config we store the hwaddr for
- * re-use.
+ * If we find a lxc.net.[i].hwaddr or lxc.network.hwaddr in the original config
+ * file, we expand it in the unexpanded_config, so that after a save_config we
+ * store the hwaddr for re-use.
  * This is only called when reading the config file, not when executing a
  * lxc.include.
  * 'x' and 'X' are substituted in-place.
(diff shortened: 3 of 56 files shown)


0x0916:
jenkins: test this please

stgraber:
jenkins: test this please

brauner:
Note that the new network parse only supports indexed keys. That means you e.g. cannot replace `lxc.network.type` with `lxc.net.type`. You rather need to replace it with `lxc.net.0.type`.

0x0916:
jenkins: test this please

brauner:
jenkins: test this please

0x0916:
@brauner Can you tell me that how to run testsuite on my laptop, or update the instructions to README?

brauner:
The easiest way, assuming you have all build dependencies, is to grab the compile instructions from the jenkins output, and compiling locally. Then in a shell set `LD_LIBRARY_PATH` to the appropriate directory. This part (minus the compile flags) can be found in `README.md`. The you can simply run the tests from that shell (as root).

0x0916:
jenkins: test this please

brauner:
jenkins: test this please

0x0916:
@brauner failed on i386 clang, do you know reason?

brauner:
@0x0916 that looks like a false negative. Will restart jenkins to confirm.

brauner:
Oh wait, it stopped before running the utility tests. Hm, let's see what the next jenkins run will do.

README.md
```diff
@@ -83,8 +83,8 @@ LXC is configured via a simple set of keys. For example,
 - `lxc.mount.entry`
 
 LXC namespaces configuration keys by using single dots. This means complex
-configuration keys such as `lxc.network` expose various subkeys such as
-`lxc.network.type`, `lxc.network.link`, `lxc.network.ipv6`, and others for even
+configuration keys such as `lxc.net.0` expose various subkeys such as
+`lxc.net.0.type`, `lxc.net.0.link`, `lxc.net.0.ipv6`, and others for even
```
brauner: Excellent that you also thought of that. :)

src/tests/get_item.c
```diff
@@ -77,12 +77,12 @@ int main(int argc, char *argv[])
 	}
```
brauner: Actually, until `LXC 3.0` we should still run tests for the old configuration keys. So this file should only test for `lxc.network.{[i]}.*` related keys. The new network configuration keys are already tested in the `parse_config_file.c` test I added a while back.

src/tests/getkeys.c
```diff
@@ -41,7 +41,7 @@ int main(int argc, char *argv[])
 		goto out;
```
brauner: Similar comment: until LXC 3.0 we should still run tests for the old configuration keys. So this file should only test for both `lxc.network.{[i]}.*` and `lxc.net.[i].*` keys.

src/lxc/confile_utils.c
```diff
@@ -532,10 +532,44 @@ int rand_complete_hwaddr(char *hwaddr)
...
+		SYSERROR("failed to allocate memory");
+		return false;
+	}
+	if (*(copy + 8) >= '0' && *(copy + 8) <= '9') {
+		p = strchr(copy + 8, '.');
+		if (!p) {
+			free(copy);
+			return false;
+		}
+		strcpy(copy + 8, p + 1);
```
brauner: Hm, not a fan of `strcpy()` without bounds checking tbh.

src/lxc/confile_utils.c
```diff
@@ -532,10 +532,44 @@ int rand_complete_hwaddr(char *hwaddr)
...
+	}
+	if (strncmp(copy, "lxc.net.hwaddr", 14) == 0) {
```
brauner: Isn't this supposed to handle bot `lxc.net.[i].hwaddr` and `lxc.network.{[i]}.hwaddr.*`? If so you should add:

```
if (strncmp(copy, "lxc.network.hwaddr", appropriate-length)  == 0) {
        /* same error handling */
}
```
You can probably move this into a single line by using `&&`.

src/lxc/confile_utils.c
brauner: I'd prefer a

    strncpy(copy + 8, p + 1, length-you-want);
    copy[length-you-want - 1] = '\0';

or you use the same pattern but use `memmove()`. Sorry for the bikeshedding but I really don't approve of `strcpy()` That just calls for trouble later on.

0x0916: ok, just copy `hwaddr`(length = 6) is enough.

src/lxc/confile_utils.c
0x0916: yes, in order to handle `lxc.network.{[i]}`, we need to dup another line and process it with the same logic.

src/tests/get_item.c
0x0916: updated

brauner:
jenkins: test this please

brauner:
I think there's still an issue with retrieving:

    308: lxc.network.0.ipv4 returned -1

brauner:
You could probably just copy and paste the old `get_item.c` file and see if it works correctly locally. I guess it doesn't need to be changed.

0x0916:
@brauner it not the problem of `get_item.c`. just revert the following change

```
diff --git a/config/etc/default.conf.lxcbr b/config/etc/default.conf.lxcbr
index e53047d..661718b 100644
--- a/config/etc/default.conf.lxcbr
+++ b/config/etc/default.conf.lxcbr
@@ -1,4 +1,4 @@
-lxc.network.type = veth
-lxc.network.link = lxcbr0
-lxc.network.flags = up
-lxc.network..hwaddr = 00:16:3e:xx:xx:xx
+lxc.net.0.type = veth
+lxc.net.0.link = lxcbr0
+lxc.net.0.flags = up
+lxc.net.0.hwaddr = 00:16:3e:xx:xx:xx
```

the testcase pass.

0x0916:
@brauner Maybe we can merge this pr, and solve `config/etc/default.conf.lxcbr` issue in another pr.

brauner:
Oh ok, that makes perfect sense and isn't your fault. LXC will read in the `/etc/default.conf` if run as root and it won't find a defined network for `lxc.network.0.ipv4`. So what you can do is use `set_config_item()` and set

```
lxc.network.type = veth
lxc.network.link = lxcbr0
lxc.network.flags = up
lxc.network..hwaddr = 00:16:3e:xx:xx:xx
```

before performing the checks.

0x0916:
@brauner when I define a network before checks in `get_item.c`, There is still some error when clear the ipv4 config item. `c->clear_config_item(c, "lxc.network.0.ipv4")`.  I have fixed that. see https://github.com/lxc/lxc/pull/1649/commits/cf4b7999cd90c2f9c58dd1bc2217ad7cc254673a

0x0916:
jenkins: test this please

src/lxc/confile_legacy.c
```diff
@@ -944,32 +944,31 @@ int get_config_network_legacy_item(const char *key, char *retv, int inlen,
...
-	lxc_list_for_each(it, &c->network) {
-		if (i == idx)
-			break;
-		i++;
 	}
-	if (i < idx)  // we don't have that many nics defined
-		return -1;
 
-	if (!it || !it->elem)
+	if ((idx = get_network_netdev_idx(key)) == EINVAL)
```
brauner: Good catch.

0x0916:
jenkins: test this please

brauner:
jenkins: test this please

PR #1657, opened 2017-06-28
Fix possible buffer overflow / compile failure with gcc 7.1.0

ljward10:
Fix compile failure with gcc 7.1.0 by expanding buffer size by one byte to prevent possible overflow in file src/lxc/cgroups/cgfsng.c.

Fixes gcc 7.1.0 error:

cgroups/cgfsng.c:1381:23: error: ‘snprintf’ output may be truncated before the last format character [-Werror=format-truncation=]
   snprintf(offset, 5, "-%d", idx);
                       ^~~~~
cgroups/cgfsng.c:1381:3: note: ‘snprintf’ output between 3 and 6 bytes into a destination of size 5
   snprintf(offset, 5, "-%d", idx);

Signed-off-by: Lance Ward <ljward10@gmail.com>

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1370,11 +1370,11 @@ static inline bool cgfsng_create(void *hdata)
 		ERROR("Failed expanding cgroup name pattern");
 		return false;
 	}
-	len = strlen(tmp) + 5; // leave room for -NNN\0
+	len = strlen(tmp) + 6; // leave room for -NNN\0
 	cgname = must_alloc(len);
 	strcpy(cgname, tmp);
 	free(tmp);
-	offset = cgname + len - 5;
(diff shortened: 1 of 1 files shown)


brauner:
That's a gcc bug which is already reported upstream. :)

brauner:
@ljward10, sorry for closing this directly. :) We've had similar PRs in the last couple of days. The gist is that there's a bug report against gcc upstream ( https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78969 ) and the marked it as a proper bug. So until that is fixed there's nothing for `LXC` to do. :)

ljward10:
No problem.  Personally I would be willing to give up one byte to not have to worry about when gcc will get around to fixing it.  Still it will likely only effect a few crazy people like me compiling bleeding edge versions of Linux from scratch.

PR #1659, opened 2017-06-29
start: use separate socket on daemonized start

brauner:
Since we killed lxc-monitord we rely on the container's command socket to wait
for the container. This doesn't work nicely on daemonized startup since a
container's init process might be something that is so short-lived that we
won't even be able to add a state client before the mainloop closes. But the
container might still have been RUNNING and executed the init binary correctly.
In this case we would erroneously report that the container failed to start
when it actually started just fine.
This commit ensures that we really all cases where the container successfully
ran by switching to a short-lived per-container anonymous unix socket pair that
uses credentials to pass container states around. It is immediately closed once
the container has started successfully.
This should also make daemonized container start way more robust since we don't
rely on the command socket handler to be running.

For the experienced developer: Yes, I did think about utilizing the command
socket directly for this. The problem is that when the mainloop starts it may
end up end accept()ing the connection that we want
do_wait_on_daemonized_start() to accept() so this won't work and might cause us
to hang indefinitely. The same problem arises when the container fails to start
before the mainloop is created. In this case we would hang indefinitely as
well.


src/lxc/start.c
```diff
@@ -344,12 +344,36 @@ static int signal_handler(int fd, uint32_t events, void *data,
 int lxc_set_state(const char *name, struct lxc_handler *handler,
 		  lxc_state_t state)
 {
+	int state_socket;
 	ssize_t ret;
 	struct lxc_list *cur, *next;
 	struct state_client *client;
 	struct lxc_msg msg = {.type = lxc_msg_state, .value = state};
 
+	if (state != STARTING) {
```
brauner: We need to ingore the `STARTING` state here. We're only interested in `RUNNING` but also want to catch cases like `STOPPING`, `STOPPED`, and `ABORTING` in case the init binary fails to exec or some other error occurs.

src/lxc/start.c
```diff
@@ -344,12 +344,36 @@ static int signal_handler(int fd, uint32_t events, void *data,
...
+		state_socket = lxc_abstract_unix_connect(handler->state_socket_name);
+		if (state_socket < 0) {
+			if (errno == ECONNREFUSED)
```
brauner: This means `wait_on_daemonized_start()` has closed the connection so we won't bother trying to send. Instead, we go right to serving any registered state clients.

src/lxc/start.c
```diff
@@ -481,6 +481,7 @@ struct lxc_handler *lxc_init_handler(const char *name, struct lxc_conf *conf,
 	handler->conf = conf;
 	handler->lxcpath = lxcpath;
 	handler->pinfd = -1;
+	handler->state_socket = -1;
 	lxc_list_init(&handler->state_clients);
```
0x0916: It is better to move the changes in `start.c/start.h` to commit `start: use separate socket on daemonized start`

src/tests/shortlived.c
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	b = c->load_config(c, NULL);
+	if (!b) {
+		fprintf(stderr, "%d: %s failed to read its config\n", __LINE__, c->name);
+		goto out;
+	}
+
+	if (!c->set_config_item(c, "lxc.init_cmd", "echo hello")) {
+		fprintf(stderr, "%d: failed setting lxc.utsname\n", __LINE__);
```
0x0916: s/lxc.utname/lxc.init_cmd

brauner: Wrong `sed` and your PR hasn't been merged so this would fail on jenkins currently.

src/lxc/lxccontainer.c
```diff
@@ -924,9 +1027,15 @@ static bool do_lxcapi_start(struct lxc_container *c, int useinit, char * const a
...
-	if (daemonize)
-		exit(ret == 0 ? true : false);
-	return (ret == 0 ? true : false);
+	if (daemonize && ret < 0)
+		exit(EXIT_FAILURE);
+	else if (daemonize)
+		exit(EXIT_SUCCESS);
+
+	if (ret < 0)
+		return false;
```
0x0916: It is not correct here. when got err, we assign `ret = 1`. it should be `if (ret != 0)` or we just delete `ret = 1` when got error.

src/lxc/lxccontainer.c
```diff
@@ -878,13 +976,16 @@ static bool do_lxcapi_start(struct lxc_container *c, int useinit, char * const a
 
 	/* Unshare the mount namespace if requested */
 	if (conf->monitor_unshare) {
-		if (unshare(CLONE_NEWNS)) {
+		ret = unshare(CLONE_NEWNS);
+		if (ret < 0) {
 			SYSERROR("failed to unshare mount namespace");
 			free_init_cmd(init_cmd);
 			lxc_free_handler(handler);
 			return false;
```
0x0916: If `daemonize` is true, we need  `exit(EXIT_FAILURE)` here, not the `return false`.

src/lxc/lxccontainer.c
brauner: Good catch, thanks!

src/lxc/lxccontainer.c
brauner: Yeah, we should simply `goto out`. Also, that label should be more descriptive.

src/lxc/start.c
brauner: Thanks!

src/tests/shortlived.c
brauner: Sorry, was misreading you're comment. That's correct.

src/lxc/commands_utils.c
```diff
@@ -0,0 +1,92 @@
...
+	struct lxc_msg msg;
+	struct timeval out;
+
+	memset(&out, 0, sizeof(out));
+	out.tv_sec = timeout;
+	ret = setsockopt(state_client_fd, SOL_SOCKET, SO_RCVTIMEO,
+			 (const void *)&out, sizeof(out));
+	if (ret < 0) {
+		SYSERROR("Failed to set %ds timeout on containter state socket", timeout);
+		close(state_client_fd);
```
brauner: Needs to be removed since it's the callers responsibility to close the fd.

src/lxc/start.c
```diff
@@ -195,10 +196,12 @@ static int match_fd(int fd)
  *   which case we also close all open fds.
  * - A daemonized container will always pass closeall=true.
  */
-int lxc_check_inherited(struct lxc_conf *conf, bool closeall, int fd_to_ignore)
+int lxc_check_inherited(struct lxc_conf *conf, bool closeall,
+			int *fds_to_ignore, size_t len_fds)
 {
```
0x0916: It is better to update the description for `lxc_check_inherited`.

src/lxc/start.c
```diff
@@ -228,7 +231,11 @@ int lxc_check_inherited(struct lxc_conf *conf, bool closeall, int fd_to_ignore)
 			continue;
 		}
 
-		if (fd == fddir || fd == lxc_log_fd || fd == fd_to_ignore)
+		for (i = 0; i < len_fds; i++)
+			if (fds_to_ignore[i] == fd)
+				break;
+
+		if (fd == fddir || fd == lxc_log_fd || fd == fds_to_ignore[i])
```
0x0916: If the program does not hit the line `236`,  `i` will be equal to `len_fds`, and `fd == fds_to_ignore[i]`  in line `238` will access out of bounds. 

```
if (fd == fddir || fd == lxc_log_fd || ( i < len_fds && fd == fds_to_ignore[i]))
```

src/tests/shortlived.c
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	b = c->load_config(c, NULL);
+	if (!b) {
+		fprintf(stderr, "%d: %s failed to read its config\n", __LINE__, c->name);
+		goto out;
+	}
+
+	if (!c->set_config_item(c, "lxc.init_cmd", "echo hello")) {
+		fprintf(stderr, "%d: failed setting lxc.init_cmd\n", __LINE__);
```
0x0916: now , we can `s/lxc.init_cmd/lxc.init.cmd`

src/tests/shortlived.c
```diff
@@ -0,0 +1,188 @@
...
+	for (i = 0; i < 10; i++) {
+		if (!c->startl(c, 0, NULL)) {
+			fprintf(stderr, "%d: %s failed to start\n", __LINE__, c->name);
+			goto out;
+		}
+		sleep(1);
+	}
+
+	if (!c->set_config_item(c, "lxc.init_cmd", "you-shall-fail")) {
+		fprintf(stderr, "%d: failed setting lxc.init_cmd\n", __LINE__);
```
0x0916: same as above.

src/lxc/start.c
brauner: good catch

hallyn:
@brauner - are you ready for this to be merged?

hallyn:
Hi,

I'm concerned about lxc_cmd_add_state_client().  It seems to frequently return @state, which is just an enumentry, without setting state_client_fd.  Meanwhile, do_lxcapi_shutdown(), which calls that, seems to assume that if its return value >= 0 then the state_client_fd will be set.

Should do_lxcapi_shutdown() initialize state_client_fd to -1?  Am I misreading the values of the states?

brauner:
@hallyn, nice. Yes, `do_lxcapi_shutdown()` should unconditionally initialize `state_client_fd` to `-1`.

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
Jenkins error is unrelated to the branch.

PR #1661, opened 2017-07-01
start: send state to legacy lxc-monitord state server even if no state clients registered

0x0916:
This pr https://github.com/lxc/lxc/pull/1618 kill lxc-monitord, for backwards compatibility,
we also send state to legacy lxc-monitord state server in function `lxc_set_state`.

we should also send state if there is no state clients registered, otherwise `lxc-monitor` client will
not get state change event if container changed state to `STARTING` or `RUNNING`.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -359,6 +359,7 @@ int lxc_set_state(const char *name, struct lxc_handler *handler,
 	if (lxc_list_empty(&handler->state_clients)) {
 		TRACE("no state clients registered");
 		process_unlock();
+		lxc_monitor_send_state(name, state, handler->lxcpath);
 		return 0;
 	}
 


brauner:
Why wouldn't it get them?

brauner:
jenkins: test this please

brauner:
Ah, I see the issue.

src/lxc/start.c
```diff
@@ -349,6 +349,11 @@ int lxc_set_state(const char *name, struct lxc_handler *handler,
 	struct state_client *client;
 	struct lxc_msg msg = {.type = lxc_msg_state, .value = state};
 
+	/* This function will try to connect to the legacy lxc-monitord state
+	 * server and only exists for backwards compatibility.
+	 */
+	lxc_monitor_send_state(name, state, handler->lxcpath);
```
brauner: The state clients are the one that should always be notified first since monitord will die a horrible horrible death soon. So if you could just add:

```
if (lxc_list_empty(&handler->state_clients)) {
		TRACE("no state clients registered");
		process_unlock();
                lxc_monitor_send_state(...);
		return 0;
}
```

that would be +1.

0x0916: update done. thanks for review.

brauner:
jenkins: Test this please

PR #1663, opened 2017-07-01
Fix lxc-ubuntu template usage without qemu-debootstrap

d4s:
In case if 'qemu-debootstrap' binary is not available then container
creation with 'lxc-ubuntu' template fail.

Signed-off-by: Denis Pynkin <denis.pynkin@collabora.com>

### DIFF
--- a/templates/lxc-ubuntu.in
+++ b/templates/lxc-ubuntu.in
@@ -375,7 +375,7 @@ download_ubuntu()
     choose_container_proxy $cache/partial-$arch/ $arch
     # download a mini ubuntu into a cache
     echo "Downloading ubuntu $release minimal ..."
-    if [ -n "$(which qemu-debootstrap)" ]; then
+    if [ -n "$(which qemu-debootstrap 2>/dev/null || :)" ]; then
         qemu-debootstrap --verbose $debootstrap_parameters --components=main,universe --arch=$arch --include=${packages_template} $release $cache/partial-$arch $MIRROR
     else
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

stgraber:
I don't understand what this is trying to solve.

"which" doesn't return anything when a command is missing, it simply exits with exit code 1.

So:
```
stgraber@castiana:~/data/code/lxc/lxd (stgraber/master)$ which missing
stgraber@castiana:~/data/code/lxc/lxd (stgraber/master)$ echo $?
1
```

There is also no reason for the "|| true" as subshells within a condition will not cause the parent to error out (with -e).

stgraber:
So I guess your change could make sense if some distros have a version of "which" which is noisy (on stderr) when a binary doesn't exist. In such cases, the "2> /dev/null" is probably fine.

I still can't see a reason for the "|| :" part though.

d4s:
This is because of 'set -e' on
https://github.com/lxc/lxc/blob/master/templates/lxc-ubuntu.in#L40
As soon as `which qemu-debootstrap` fails, script also stopped with error.

stgraber:
What shell are you using that'd give you this behavior?

stgraber:
```
#!/bin/sh

if [ -n "$(exit 1)" ]; then
    echo not here
fi

echo here
```

```
stgraber@castiana:~$ sh -ex test.sh
+ exit 1
+ [ -n  ]
+ echo here
here
```

d4s:
You are correct.
hm... interesting. I definitely seen such behavior some months ago and it has been fixed by this commit.
Now I can't reproduce wrong behavior anymore.

PR #1668, opened 2017-07-03
lxclock: bugfix and update some comment

0x0916:
This pr mainly updates comments based on the current code and also fix a small bug(the errno is the wrong value when open lock file failed).


brauner:
jenkins: test this please

PR #1674, opened 2017-07-04
lxc-execute: print error message when failed

0x0916:

### DIFF
--- a/src/lxc/tools/lxc_execute.c
+++ b/src/lxc/tools/lxc_execute.c
@@ -160,7 +160,9 @@ int main(int argc, char *argv[])
 	bret = c->start(c, 1, my_args.argv);
 	ret = c->error_num;
 	lxc_container_put(c);
-	if (!bret)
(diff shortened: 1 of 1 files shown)


src/lxc/tools/lxc_execute.c
```diff
@@ -160,7 +160,9 @@ int main(int argc, char *argv[])
 	bret = c->start(c, 1, my_args.argv);
 	ret = c->error_num;
 	lxc_container_put(c);
-	if (!bret)
+	if (!bret) {
```
brauner: `if (!bret && !my_args.quiet)` so that it can be used in scripts.

0x0916: if `my_args.quiet` is true, the `ERROR` will output nothing to `stderr`.  so there is no need to add judgment

brauner: Oh, because you initalize the log above. Cool.

PR #1680, opened 2017-07-05
Change _have => have into bash/lxc.in

bayaro:
`/etc/bash_completion` contains only `have`


stgraber:
"have" is the deprecated function, the currently supported one is "_have".

PR #1683, opened 2017-07-08
af_unix: remove unlink operation

0x0916:
It is not necessary to unlink the abstract socket pathname when
we have finished using the socket. The abstract name is automatically
removed when the socket is closed.

### DIFF
--- a/src/lxc/af_unix.c
+++ b/src/lxc/af_unix.c
@@ -42,9 +42,6 @@ int lxc_abstract_unix_open(const char *path, int type, int flags)
 	size_t len;
 	struct sockaddr_un addr;
 
-	if (flags & O_TRUNC)
-		unlink(path);
-
 	fd = socket(PF_UNIX, type, 0);
 	if (fd < 0)
 		return -1;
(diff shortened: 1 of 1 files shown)


brauner:
I'd rather leave the `unlink()`s. They exist to remove any non-abstract sockets of the same name that might have been created.

0x0916:
@brauner ,yes, `lxc_abstract_unix_open`  just open an abstract socket, remove any non-abstract sockets of the same name may be cause issue if the non-abstract sockets used by another program. but the probability is very small.

hallyn:
@brauner - but you cannot create a non-abstract socket by the same name (beginning with '\0' :)  And unlink willl always return -ENOENT (tested here), even if the socket is bound.

I'd be half tempted to say let's turn the unlinks into loud warning messages, but I just don't think it's worth it.

So I think we should merge it. Since you objected I don't want to just ignore that so I'll leave it to you :)

Thanks, both!

brauner:
@hallyn, the unlink should start at `path[1]`, granted.

PR #1685, opened 2017-07-08
lxc_abstract_unix_connect: remove the workaround-code

0x0916:
commit bdb3f44147bc1a55a97131b4b39d42844ae4fb9e says that we may undo
the change in august 2014.

I think that it is time to do that.


brauner:
jenkins: test this please

PR #1688, opened 2017-07-09
utils: use access instead of stat

0x0916:
we shoud test whether the file exists and grants
execute permissions.


hallyn:
jenkins: test this please

hallyn:
Thanks, looks good to me.

Issue #1646, opened 2017-06-26
Two questions about the daemonize containers stdfds


lifeng68:
There are a few questions about the daemonize containers stdfds.
Q1: When we start a daemonize container, in the function `do_lxcapi_start`, redirected 0,1,2 to /dev/null
```
		/* like daemon(), chdir to / and redirect 0,1,2 to /dev/null */
		if (chdir("/")) {
			SYSERROR("Error chdir()ing to /.");
			exit(1);
		}
		lxc_check_inherited(conf, true, -1);
		if (null_stdfds() < 0) {
			ERROR("failed to close fds");
			exit(1);
		}
```
In the function `lxc_spawn` ,clone out the container process.  In the process,  `do_start` call `lxc_console_set_stdfds` with the purpose to redirect 0,1,2 to the `handler->conf->console.slave`, but the STDIN_FILENO、STDOUT_FILENO、STDERR_FILENO had been  redirected to /dev/null, they would not be a tty, so lxc_console_set_stdfds will not redirect 0,1,2 to the `handler->conf->console.slave`.
Is this a bug?
```
int lxc_console_set_stdfds(int fd)
{
	if (fd < 0)
		return 0;

	if (isatty(STDIN_FILENO))
		if (dup2(fd, STDIN_FILENO) < 0) {
			SYSERROR("failed to duplicate stdin.");
			return -1;
		}

	if (isatty(STDOUT_FILENO))
		if (dup2(fd, STDOUT_FILENO) < 0) {
			SYSERROR("failed to duplicate stdout.");
			return -1;
		}

	if (isatty(STDERR_FILENO))
		if (dup2(fd, STDERR_FILENO) < 0) {
			SYSERROR("failed to duplicate stderr.");
			return -1;
		}

	return 0;
}
```
Q2: In the function do_start, why we should redirect 0,1,2 to /dev/null when start container backgrounded? I think we should redirect 0,1,2 to the `handler->conf->console.slave` always, and when we start a daemon container,  do not read the master of the pty (`console.master`)
```
	if (devnull_fd < 0) {
		devnull_fd = open_devnull();

		if (devnull_fd < 0)
			goto out_warn_father;
	}

	if (handler->backgrounded && set_stdfds(devnull_fd))
		goto out_warn_father;
```

brauner:
Q1:
[...]

PR #1689, opened 2017-07-10
Fix issue #1646, dup the STD{in,out,err} of the container to the slave of the PTY

lifeng68:
1. In the case of the container had console(handler->conf->console.slave >=0),  we should make STD{in,out,err} of the container dup  to the slave of the PTY allocated in lxc_console_create()

2. In the case of the container did not have console(handler->conf->console.slave < 0), and the container run as background, we should make STD{in,out,err} of the container dup  to the /dev/null


brauner:
jenkins: test this please

brauner:
I've made some small changes to the PR you sent but retained your authorship of course. So I'm closing this one and merging the other one in a bit if Jenkins is happy (https://github.com/lxc/lxc/pull/1690).

PR #1690, opened 2017-07-10
start: dup std{in,out,err} to pty slave

### LINKED ISSUE #1646: Two questions about the daemonize containers stdfds

brauner:
In the case the container has a console with a valid slave pty file descriptor
we duplicate std{in,out,err} to the slave file descriptor so console logging
works correctly. When the container does not have a valid slave pty file
descriptor for its console and is started daemonized we should dup to
/dev/null.

Closes #1646.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -1063,8 +1063,13 @@ static int do_start(void *data)
 	 * setup on its console ie. the pty allocated in lxc_console_create() so
 	 * make sure that that pty is stdin,stdout,stderr.
 	 */
-	if (lxc_console_set_stdfds(handler->conf->console.slave) < 0)
-		goto out_warn_father;
+	 if (handler->conf->console.slave >= 0)
+		 if (set_stdfds(handler->conf->console.slave) < 0) {
+			ERROR("Failed to redirect std{in,out,err} to pty file "
+			      "descriptor %d",
+			      handler->conf->console.slave);
+			goto out_warn_father;
+		 }
 
 	/* If we mounted a temporary proc, then unmount it now. */
 	tmp_proc_unmount(handler->conf);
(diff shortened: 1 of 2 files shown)


brauner:
@lifeng68, just some small formatting changes and a non-functional change added. Your authorship is retained. Thanks!

PR #1692, opened 2017-07-11
cgfsng: only output debug info when we set cgroup data

0x0916:
Only output debug info `cgroup 'xxxx' set to 'yyyy'` when we set
cgroup data.


brauner:
jenkins: test this please

Issue #1697, opened 2017-07-12
Output confusion when  starting container with  LXC_DEBUG_CGFSNG=1


0x0916:
# Issue description

Output confusion when  starting container with  LXC_DEBUG_CGFSNG=1

# Steps to reproduce

 1. Step one
     create an container with name  `bb` with fedora template.

```
# lxc-create -n bb -t download
```
 2. Step two

```
# LXC_DEBUG_CGFSNG=1 lxc-start -n bb -F
```

# Output confusion as following:

```bash
# LXC_DEBUG_CGFSNG=1 lxc-start -n bb -F
basecginfo is:
11:memory:/init.scope
10:blkio:/init.scope
9:pids:/init.scope
8:cpu,cpuacct:/init.scope
7:perf_event:/
6:devices:/init.scope
5:net_cls,net_prio:/
4:hugetlb:/
3:cpuset:/
2:freezer:/
1:name=systemd:/init.scope

kernel subsystem 0: memory
kernel subsystem 1: blkio
kernel subsystem 2: pids
kernel subsystem 3: cpu
kernel subsystem 4: cpuacct
kernel subsystem 5: perf_event
kernel subsystem 6: devices
kernel subsystem 7: net_cls
kernel subsystem 8: net_prio
kernel subsystem 9: hugetlb
kernel subsystem 10: cpuset
kernel subsystem 11: freezer
named subsystem 0: name=systemd
writeable subsystems:
  Hierarchies:
  0: base_cgroup /
      mountpoint /sys/fs/cgroup/systemd
      controllers:
      0: name=systemd
  1: base_cgroup /
      mountpoint /sys/fs/cgroup/freezer
      controllers:
      0: freezer
  2: base_cgroup /
      mountpoint /sys/fs/cgroup/cpuset
      controllers:
      0: cpuset
  3: base_cgroup /
      mountpoint /sys/fs/cgroup/hugetlb
      controllers:
      0: hugetlb
  4: base_cgroup /
      mountpoint /sys/fs/cgroup/net_cls
      controllers:
      0: net_cls
      1: net_prio
  5: base_cgroup /
      mountpoint /sys/fs/cgroup/devices
      controllers:
      0: devices
  6: base_cgroup /
      mountpoint /sys/fs/cgroup/perf_event
      controllers:
      0: perf_event
  7: base_cgroup /
      mountpoint /sys/fs/cgroup/cpu
      controllers:
      0: cpu
      1: cpuacct
  8: base_cgroup /
[...]

PR #1699, opened 2017-07-13
cgfsng: use `\r\n` instead of `\n` in cgfsng debug code

0x0916:
Fix: https://github.com/lxc/lxc/issues/1697

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1000,11 +1000,11 @@ static void trim(char *s)
 
 static void lxc_cgfsng_print_handler_data(const struct cgfsng_handler_data *d)
 {
-	printf("Cgroup information:\n");
-	printf("  container name: %s\n", d->name ? d->name : "(null)");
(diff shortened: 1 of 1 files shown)


hallyn:
Why?

0x0916:
@hallyn During the container start, after call `lxc_set_tios` which change console's terminal settings. 

`\n` will only print on the new line and not at the start of the line.

hallyn:
I don't see that function anywhere - lxc_set_tios

@brauner, I'll leave this to you - I prefer not to splatter \n's about if
we can help it, but if it's needed then it's needed.

IMO in general it's better to encourage users to redirect log to a file
than ever try to read it on a console, especially if non-backgrounded
container.

0x0916:
@hallyn sorry, the function name is `lxc_setup_tios`.

brauner:
Yeah, I'm not sure that this is actually worth it given that we'd introduce some parsing weirdness if we pring `\r\n`. If this doesn't p*ss you off too much @0x0916 I'd close this (or you do). :)

Issue #1702, opened 2017-07-14
How to mount a readonly rootfs?


lifeng68:
The following config leads to a container with a r/w rootfs while it should be r/o instead, assuming  /home/lifeng/myrootfs is a directory containing a container rootfs:

[ lxc.conf ]
```
lxc.rootfs = /home/lifeng/myrootfs
lxc.rootfs.options = ro
lxc.rootfs.backend = dir
```

```
1:25_[/home/lifeng/test]# lxc-execute -n lifeng -f /var/lib/lcrd/lcr/test_run1/config /bin/bash
root@mrsdalloway:/# touch /foo 
root@mrsdalloway:/# ls foo
foo
root@mrsdalloway:/# 
```
The `touch /foo` should be failed, but it successed.

The mount info as follow
```
root@mrsdalloway:/# mount
/dev/sda1 on / type ext4 (rw,relatime,errors=remount-ro,data=ordered)
none on /dev type tmpfs (rw,relatime,size=492k,mode=755)
udev on /dev/sdb type devtmpfs (rw,nosuid,relatime,size=32768948k,nr_inodes=8192237,mode=755)
udev on /dev/sda type devtmpfs (rw,nosuid,relatime,size=32768948k,nr_inodes=8192237,mode=755)
/dev/sda1 on /init.lxc.static type ext4 (rw,relatime,errors=remount-ro,data=ordered)
devpts on /dev/pts type devpts (rw,relatime,gid=5,mode=620,ptmxmode=666)
devpts on /dev/ptmx type devpts (rw,relatime,gid=5,mode=620,ptmxmode=666)
proc on /proc type proc (rw,relatime)
shmfs on /dev/shm type tmpfs (rw,relatime)
mqueue on /dev/mqueue type mqueue (rw,relatime)
```
What actions do I need to do to mount a readonly rootfs?

lifeng68:
I think we should do re-mount with MS_REMOUNT | MS_BIND | MS_RDONLY in the function 
`int dir_mount(struct bdev *bdev)` if mntflags has option` ro`, reference to #570

brauner:
@lifeng68, sure send a patch we can argue over. :)

brauner:
I assigned you but that of course doesn't mean you have to do it. :)

lifeng68:
OK, I will do it , I will submit a patch next week.

brauner:
Cool, thanks @lifeng68.

Issue #570, opened 2015-06-16
Bind mounted rootfs is writable when it should be r/o instead


minipli:
It looks like LXC is prune to ["kernel bug #24912"](https://bugzilla.kernel.org/show_bug.cgi?id=24912) in regard to that it doesn't handle the `ro` mount option correctly for a bind mounted rootfs.

The following config leads to a container with a r/w rootfs while it should be r/o instead, assuming `/chroot/wheezy` is a directory containing a wheezy chroot:

> ,-[ lxc.conf ]-
> | lxc.rootfs = /chroot/wheezy
> | lxc.rootfs.options = ro
> `---

`$ lxc-execute -n wheezy -f lxc.conf /bin/bash`
`(wheezy)root@alpha:/# touch /foo`
`(wheezy)root@alpha:/# ls foo`
`foo`

The `touch /foo` should have failed.

The issue can be confirmed by stracing lxc-execute. It contains the follwing line:

> mount("/chroot/wheezy", "/chroot/jail-apps/lib/lxc/rootfs", 0x7faa280a0ca3, MS_RDONLY|MS_BIND|MS_REC, NULL) = 0

So, LXC calls `mount()` only once, to do the bind mount and apply the `ro` mount flag. But because of the behavior of the Linux kernel regarding bind mounts it ignores the `MS_RDONLY` flag and the bind mount stays writable.

Changing mount options requires a second `mount()` with the `MS_REMOUNT` flag. LXC should therefore handle this case be doing the rootfs mount in two steps:
1. bind mounting the rootfs
2. re-mount with `MS_REMOUNT | MS_BIND | MS_RDONLY`

This way the rootfs will be r/o.

evgeni:
This should be fixed in https://github.com/lxc/lxc/commit/7c5b6e7c3524031c3e9e203a9ab3beea7efa7a41

@brauner @stgraber @hallyn please review and close if correct :)

hallyn:
I'll buy that.  Please re-open if in fact the problem was not solved.

PR #1706, opened 2017-07-17
Fix issue #1702, do remount with the MS_REMOUNT flag when mounts with MS_RDONLY

lifeng68:
If we want to mount a readonly rootfs. LXC calls mount() only once in the function `dir_mount `, to do the bind mount with the ro mount flag. But because of the behavior of the Linux kernel regarding bind mounts it ignores the MS_RDONLY flag and the bind mount stays writable. ["kernel bug #24912" ](https://bugzilla.kernel.org/show_bug.cgi?id=24912)

Changing mount options requires a second mount() with the MS_REMOUNT flag. LXC should therefore handle this case be doing the rootfs mount in two steps:

1. bind mounting the rootfs
2. re-mount with MS_REMOUNT | MS_BIND | MS_RDONLY

This patch will resolve issue #1702 and #570.

### DIFF
--- a/src/lxc/bdev/lxcdir.c
+++ b/src/lxc/bdev/lxcdir.c
@@ -136,6 +136,7 @@ int dir_mount(struct bdev *bdev)
 	unsigned long mntflags;
 	char *src, *mntdata;
 	int ret;
+	unsigned long mflags;
 
 	if (strcmp(bdev->type, "dir"))
 		return -22;
@@ -154,7 +155,8 @@ int dir_mount(struct bdev *bdev)
 	if ((0 == ret) && (mntflags & MS_RDONLY)) {
 		DEBUG("remounting %s on %s with readonly options",
 			src ? src : "(none)", bdev->dest ? bdev->dest : "(none)");
-		ret = mount(src, bdev->dest, "bind", MS_BIND | MS_REC | mntflags | MS_REMOUNT, mntdata);
+		mflags = add_required_remount_flags(src, bdev->dest, MS_BIND | MS_REC | mntflags | MS_REMOUNT);
+		ret = mount(src, bdev->dest, "bind", mflags, mntdata);
 	}
 
 	free(mntdata);
--- a/src/lxc/bdev/lxcdir.c
+++ b/src/lxc/bdev/lxcdir.c
@@ -151,6 +151,12 @@ int dir_mount(struct bdev *bdev)
 	src = lxc_storage_get_path(bdev->src, bdev->type);
 
 	ret = mount(src, bdev->dest, "bind", MS_BIND | MS_REC | mntflags, mntdata);
+	if ((0 == ret) && (mntflags & MS_RDONLY)) {
+		DEBUG("remounting %s on %s with readonly options",
+			src ? src : "(none)", bdev->dest ? bdev->dest : "(none)");
+		ret = mount(src, bdev->dest, "bind", MS_BIND | MS_REC | mntflags | MS_REMOUNT, mntdata);
+	}
+
 	free(mntdata);
 	return ret;
 }
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -628,7 +628,7 @@ int pin_rootfs(const char *rootfs)
  * If we are asking to remount something, make sure that any
  * NOEXEC etc are honored.
  */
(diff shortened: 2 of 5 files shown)


brauner:
@lifeng68, thanks for the patch. Can you add a test for this someplace suitable in https://github.com/lxc/lxc/tree/master/src/tests ?

lifeng68:
jenkins: test this please

lifeng68:
I added a test script` lxc-test-rootfs` to test the rootfs ro options.

brauner:
jenkins: test this please

brauner:
@hallyn, lgtm. Any objections?

hallyn:
Is this a regression? I'm convinced we were doing that... (presumably wrongly convinced) 

I'll look tonight thanks.

lifeng68:
In the commit [7c5b6e7](https://github.com/lxc/lxc/commit/7c5b6e7c3524031c3e9e203a9ab3beea7efa7a41),  it resolved the readonly mount problems in the` lxc_conf->fstab `and` lxc_conf->mount_list`, but the problem of rootfs had not been resolved. 
I tested in my computer without remount_flags, it worked.  But I  think we should  require using add-required_remount_flags. Because there is no guarantee that there will be no problems on any other kernel version without the MS_REMOUNT flags.
```
static int lxc_mount_auto_mounts(struct lxc_conf *conf, int flags, struct lxc_handler *handler)
{
	int r;
	int i;
	static struct {
		int match_mask;
		int match_flag;
		const char *source;
		const char *destination;
		const char *fstype;
		unsigned long flags;
		const char *options;
	} default_mounts[] = {
		/* Read-only bind-mounting... In older kernels, doing that required
		 * to do one MS_BIND mount and then MS_REMOUNT|MS_RDONLY the same
		 * one. According to mount(2) manpage, MS_BIND honors MS_RDONLY from
		 * kernel 2.6.26 onwards. However, this apparently does not work on
		 * kernel 3.8. Unfortunately, on that very same kernel, doing the
		 * same trick as above doesn't seem to work either, there one needs
		 * to ALSO specify MS_BIND for the remount, otherwise the entire
		 * fs is remounted read-only or the mount fails because it's busy...
		 * MS_REMOUNT|MS_BIND|MS_RDONLY seems to work for kernels as low as
		 * 2.6.32...
		 */
```

hallyn:
I'm not sure what you are referring to with "without remount_flags".  I was talking about src/lxc/conf.c:add_required_remount_flags(), which ensures that if the parent mount was noexec, nodev, or nosuid, even though the config did not request it, we still add those to the mount flags, else the remount may fail.  See commit e2a7e8dc .

lifeng68:
Sorry, I misunderstood what you meant. 
Do you mean that if the lxc.rootfs (src) path has a nodev mount flags, and we should honor the honor the existing nodev in the function 'dir_mount'?
If so, I tested that when the src rootfs had nodev flags, it did not honor the existing nodev flags. So I should modify the code as follow？
```
--- a/src/lxc/bdev/lxcdir.c
+++ b/src/lxc/bdev/lxcdir.c
@@ -136,6 +136,7 @@ int dir_mount(struct bdev *bdev)
         unsigned long mntflags;
         char *src, *mntdata;
         int ret;
+        unsigned long mflags;
 
         if (strcmp(bdev->type, "dir"))
                 return -22;
@@ -154,7 +155,8 @@ int dir_mount(struct bdev *bdev)
         if ((0 == ret) && (mntflags & MS_RDONLY)) {
                 DEBUG("remounting %s on %s with readonly options",
                         src ? src : "(none)", bdev->dest ? bdev->dest : "(none)");
-                ret = mount(src, bdev->dest, "bind", MS_BIND | MS_REC | mntflags | MS_REMOUNT, mntdata);
+                mflags = add_required_remount_flags(src, bdev->dest, MS_BIND | MS_REC | mntflags | MS_REMOUNT);
+                ret = mount(src, bdev->dest, "bind", mflags, mntdata);
         }
```

hallyn:
Yup, that looks good, thanks.

brauner:
@lifeng68, just ping me when you've updated the pr.

lifeng68:
@brauner , i had updated the code.

brauner:
@lifeng68, perfect, thanks!

brauner:
jenkins: test this please

PR #1708, opened 2017-07-19
Allow full path in lxc.devices.allow

aeris:
Some devices like LVM or cryptsetup entries have no stable major/minor, changing between host reboots.
In this case, hardcoded numbers are not usable in config file and there is currently no way to use hook with lxc-device to do the link at guest startup :

    * `pre-start`/`autodev` hook runs in host context but has the guest in stopped state and so lxc-device not usable
    * `start` hook is in running state but runs in guest context and so lxc-device not available

This patch converts fullpath in lxc.devices.allow to current major/minor numbers to address those changing numbers.

Signed-off-by: aeris <aeris@imirhil.fr>

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -47,6 +47,9 @@
 #include <unistd.h>
 #include <sys/types.h>
 
+#include <linux/types.h>
+#include <linux/kdev_t.h>
+
 #include "bdev.h"
 #include "cgroup.h"
 #include "commands.h"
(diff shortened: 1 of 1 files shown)


src/lxc/cgroups/cgfsng.c
```diff
@@ -1927,12 +1930,62 @@ static int lxc_cgroup_set_data(const char *filename, const char *value, struct c
...
+					MAJOR(dev), MINOR(dev), mode)) {
+			}
+
+			for (i = 0; i < n_parts; ++i) {
+				free(parts[i]);
+			}
+			free(parts);
+		}
+	} else {
+		strcpy(converted_value, value);
```
hallyn: You never use converted_value anywhere.

src/lxc/cgroups/cgfsng.c
```diff
@@ -1927,12 +1930,62 @@ static int lxc_cgroup_set_data(const char *filename, const char *value, struct c
...
+			case S_IFBLK:
+				type = 'b';
+				break;
+			case S_IFCHR:
+				type = 'c';
+				break;
+			}
+
+			if (!snprintf(converted_value, 50, "%c %lu:%lu %s", type,
+					MAJOR(dev), MINOR(dev), mode)) {
```
hallyn: Did you intend to do something if this snprintf fails?

aeris: No, just a trailing test. Removed.

src/lxc/cgroups/cgfsng.c
aeris: Missed a changed chunk during commit. Fixed.

aeris:
The goal of this patch it to transform `lxc.devices.allow /dev/sda rwm` to `lxc.devices.allow b 8:0 rwm` on the fly, to allow full path in `lxc.devices.allow`. With `lxc.mount.entry`, it allows to mount devices with dynamic major/minor, like LVM or cryptsetup mapper, not stable between reboot.

For example, currently this is not possible:

    lxc.devices.allow b 254 3 rwm
    lxc.mount.entry = /dev/system/data dev/system/data none bind,create=file

Because `/dev/system/data` is a LV and its minor sometime changes between reboot.
With this patch, I can do

    lxc.devices.allow /dev/system/data rwm
    lxc.mount.entry = /dev/system/data dev/system/data none bind,create=file

hallyn:
This patch cannot be doing what you want.  You write the result into a new variable 'converted_value', which you do not later use.

aeris:
See the missing commit chunk added after the initial commit https://github.com/lxc/lxc/pull/1708/commits/8d4f36eedfa4dc94422df46876028ce4dcaac377

hallyn:
Ah yes, I missed that.

In the case where 

	if (n_parts == 2) {

is not true, you will be leaving converted_value uninitialized, and
write the uninitialized value.

You should also probably not write out 'type' as a %c if it is still
0.

aeris:
I add more tests on e7a51c5eb2b2e87f592e088057ceb9ca24d4260b

aeris:
Here we go ! feed9c74f4c209bf36924a751e8429ef7aa1bf76

hallyn:
No, it should be (ret < 0 || ret >= 50)

hallyn:
And then please squash them all into one commit, and I'll apply.  Thanks!

aeris:
I missred the snprintf doc…
Fixed and squashed into 064cc58c01e597d65c59d3511d7b0e3308a8bf64

hallyn:
I'm sorry, I noticed two more things.

First, you are mallocing to_split twice.

Secondly, you are not always checking that malloc and strdup succeeded.

Please fix those, or feel free at this point to tell me you prefer I just fix those and I merge.

aeris:
`to_split` must be duplicated, `strtok_r` modifies its first argument, so reusing the first value leads to segfault.
For `malloc/strdup`, I add more tests.

PR #1714, opened 2017-07-27
cgroups: workaround gcc-7 bug

brauner:
Also: LOL


kolyshkin:
Wonder why 'FILE* f' is declared inside the block, but 'int ret' is at the beginning of the function.

brauner:
Yeah, this could probably be scoped tighter faire coding-style point!

PR #1715, opened 2017-07-27
lvm: check whether lxc.bdev.lvm.vg is empty

brauner:

### DIFF
--- a/src/lxc/bdev/lxclvm.c
+++ b/src/lxc/bdev/lxclvm.c
@@ -307,6 +307,11 @@ int lvm_clonepaths(struct bdev *orig, struct bdev *new, const char *oldname,
 			return -1;
 		}
 		vg = lxc_global_config_value("lxc.bdev.lvm.vg");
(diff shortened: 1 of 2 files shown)


stgraber:
what's with the cgroup change?

brauner:
@stgraber, it's just scoping the `int ret` declaration tighter which we do for other variables in this function as well.

brauner:
I've introduced the wider-scoping variable in my other change for the gcc fix before.

stgraber:
@brauner can you move it to a separate commit since it doesn't appear to have anything to do with lvm? :)

brauner:
@stgraber, updated. :)

stgraber:
If the intent was to fix the build failure on archlinux, this didn't work ;)

https://jenkins.linuxcontainers.org/view/All/job/lxc-template-archlinux/524/arch=amd64,release=current,restrict=lxc-priv,variant=default/console

PR #1719, opened 2017-07-29
Fix up c8bf519d to fit our coding style

hallyn:

PR #1720, opened 2017-07-29
Fix some bugs in #1719

hallyn:
In particular, any text (like a #comment) following the mode
would end up in our mode line.


brauner:
jenkins: test this please

PR #1721, opened 2017-07-30
network parser fixes

brauner:
When we are passed a network key like "lxc.net.[i].ipv4.address" we need to
make sure that we pass the deindexed key "lxc.net.ipv4.address" to the
{get,clr,set} methods otherwise we'll end up in an endless loop.

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -3680,72 +3680,88 @@ static int get_config_includefiles(const char *key, char *retv, int inlen,
 	return -ENOSYS;
 }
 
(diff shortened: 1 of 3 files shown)


brauner:
The rationale is: Imagine you're passed a non-valid key like `lxc.net.1.asd` then you will end up in an endless loop since you're de-indexing the key to retrieve the correct jump table entry but then you call the jump table entry function {clr,get,set} with the non-indexed key which means you'll end up calling the de-indexing function again. :)

PR #1722, opened 2017-07-31
devpts: use max=<count> option on mount

brauner:
This will only work with kernels >= 3.4


hallyn:
The commit message should probably point out that this means only kernels >= 3.4 are supported.  (Which is fine, but let's document it so everyone doesn't have to keep searching for it)

brauner:
@hallyn, updated.

hallyn:
Hm, lxc-test-console failed?

brauner:
@hallyn, if the `max=<idx` value is too low the allocation would fail. I'll bump it for the test.

brauner:
@hallyn, should be good to go.

PR #1728, opened 2017-08-03
[RFC] commands: add LXC_CMD_OPENPTY

### LINKED ISSUE #1620: spec: pty allocation
#### Introduction
For security reasons, LXC allocates fresh pty devices before attaching to container namespaces. Current LXC will perform this allocation on the host. A downside with host-allocated pty devices was that the corresponding slave pty file descriptor is either not present under the devpts mount in the container's mount namespace or is accidently present. This would cause functions like ttyname{_r}() to report NULL while isatty() would report 1 (true) when asked whether the current process is attached to a pty. This non-standard behavior has caused a number of programs (e.g. sudo, tmux, screen, etc.) to not correctly
work in a container. A better solution would thus be to allocate the pty file descriptors from the container's devpts mount in the container's mount namespace. So far, this was not possible since there was no safe way to allocate the corresponding slave file descriptor of a pty master file descriptor inside
the container since the allocation was path based. A container could easily trick the caller by e.g. mounting a malicious fuse filesystem as its devpts mount. This way the container could have tricked the caller into opening a /dev/pts/<n> path that would not allocate from the original devpts mount instantiated during container setup but rather from the fuse filesystem. Furthermore, slave pty allocation was inherently racy. After a pty file descriptor for the master side of a pty device was opened there was no direct way to allocate a slave side pty file descriptor based solely on the master file descriptor without resorting to - as mentioned before - path based resolution based on an index provided by the kernel.
Based on an outline and discussion of this problem Aleksa Sarai and I came up with a viable in-kernel solution. Essentially, a new ioctl() flag was added which allows to allocate a slave side file descriptor solely based on the provided master file descriptor without having to resort to path resolution. This means pty file descriptors can now be safely allocated from inside the container's mount namespace.
 
#### Outline
For safety reasons it is still not possible to simply rely on the callback system to get a new master pty file descriptor from the container's devpts mount. When the init binary of the container has started LXC will not be able to guarantee that the devpts mount in the container's mount namespace is pristine.
The init binary or any other suitably privileged process inside the container could still mount a custom fuse filesystem instead of the pristine devpts mount. The only safe way is to pre-allocate a file descriptor for the /dev/pts/ptmx mount inside the container's mount namespaces on container setup before the init
binary is spawned, i.e. before control is handed over to the container itself. The problem is that opening the /dev/pts/ptmx device will not return a file descriptor referring to the /dev/pts/ptmx device but rather to a new master pty file descriptor. Thus, we need to resort to a trick. The /dev/pts/ptmx device will be opened with the O_PATH flag. We then cache the O_PATH /dev/pts/ptmx device file descriptor and cache it in the LXC handler. Furthermore, a new command lxc_cmd_pty_allocate() will be added. The callback for this command will open the corresponding /proc/self/<ptmx-fd-nr> file and allocate a new master file descriptor. After it has succeeded in doing so it will also allocate the corresponding slave pty file descriptor. Both file descriptors will then be sent back to the caller via the container's command socket. The caller is will be responsible for closing the corresponding file descriptor.

brauner:
Closes #1620.

#### Introduction
For security reasons, LXC allocates fresh pty devices before attaching to container namespaces. Current LXC will perform this allocation on the host. A downside with host-allocated pty devices was that the corresponding slave pty file descriptor is either not present under the devpts mount in the container's mount namespace or is accidently present. This would cause functions like ttyname{_r}() to report NULL while isatty() would report 1 (true) when asked whether the current process is attached to a pty. This non-standard behavior has caused a number of programs (e.g. sudo, tmux, screen, etc.) to not correctly
work in a container. A better solution would thus be to allocate the pty file descriptors from the container's devpts mount in the container's mount namespace. So far, this was not possible since there was no safe way to allocate the corresponding slave file descriptor of a pty master file descriptor inside
the container since the allocation was path based. A container could easily trick the caller by e.g. mounting a malicious fuse filesystem as its devpts mount. This way the container could have tricked the caller into opening a /dev/pts/<n> path that would not allocate from the original devpts mount instantiated during container setup but rather from the fuse filesystem. Furthermore, slave pty allocation was inherently racy. After a pty file descriptor for the master side of a pty device was opened there was no direct way to allocate a slave side pty file descriptor based solely on the master file descriptor without resorting to - as mentioned before - path based resolution based on an index provided by the kernel.
Based on an outline and discussion of this problem Aleksa Sarai and I came up with a viable in-kernel solution. Essentially, a new ioctl() flag was added which allows to allocate a slave side file descriptor solely based on the provided master file descriptor without having to resort to path resolution. This means pty file descriptors can now be safely allocated from inside the container's mount namespace.
 
#### Outline
For safety reasons it is still not possible to simply rely on the callback system to get a new master pty file descriptor from the container's devpts mount. When the init binary of the container has started LXC will not be able to guarantee that the devpts mount in the container's mount namespace is pristine.
The init binary or any other suitably privileged process inside the container could still mount a custom fuse filesystem instead of the pristine devpts mount. The only safe way is to pre-allocate a file descriptor for the /dev/pts/ptmx mount inside the container's mount namespaces on container setup before the init
binary is spawned, i.e. before control is handed over to the container itself. The problem is that opening the /dev/pts/ptmx device will not return a file descriptor referring to the /dev/pts/ptmx device but rather to a new master pty file descriptor. Thus, we need to resort to a trick. The /dev/pts/ptmx device will be opened with the O_PATH flag. We then cache the O_PATH /dev/pts/ptmx device file descriptor and cache it in the LXC handler. Furthermore, a new command lxc_cmd_pty_allocate() will be added. The callback for this command will open the corresponding /proc/self/<ptmx-fd-nr> file and allocate a new master file descriptor. After it has succeeded in doing so it will also allocate the corresponding slave pty file descriptor. Both file descriptors will then be sent back to the caller via the container's command socket. The caller is will be responsible for closing the corresponding file descriptor.

Marked `BLOCKED` because of the following bug:
Recently the kernel has implemented the TIOCGPTPEER ioctl() call which allows
users to retrieve an fd for the slave side of a pty solely based on the
corresponding fd for the master side. The ioctl()-based fd retrieval however
causes the "/proc/<pid>/fd/<pty-slave-fd>" symlink to point to the wrong dentry.
Currently, it will always point to "/". This needs to be fixed before this can be merged.

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -140,6 +140,7 @@ static int lxc_cmd_rsp_recv(int sock, struct lxc_cmd_rr *cmd)
 		if (!rspdata) {
 			ERROR("Command %s couldn't allocate response buffer.",
 			      lxc_cmd_str(cmd->req.cmd));
+			close(rspfd);
 			return -1;
 		}
 		rspdata->masterfd = rspfd;
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -21,33 +21,35 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
-#include <stdio.h>
+#include "config.h"
+
 #include <errno.h>
-#include <unistd.h>
-#include <signal.h>
 #include <fcntl.h>
+#include <malloc.h>
 #include <poll.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/param.h>
 #include <sys/socket.h>
 #include <sys/un.h>
-#include <sys/param.h>
-#include <malloc.h>
-#include <stdlib.h>
 
-#include "log.h"
-#include "lxc.h"
-#include "conf.h"
-#include "start.h"	/* for struct lxc_handler */
-#include "utils.h"
+#include "af_unix.h"
 #include "cgroup.h"
 #include "commands.h"
 #include "commands_utils.h"
-#include "console.h"
+#include "conf.h"
+#include "config.h"
 #include "confile.h"
+#include "console.h"
+#include "log.h"
+#include "lxc.h"
 #include "lxclock.h"
 #include "mainloop.h"
 #include "monitor.h"
-#include "af_unix.h"
-#include "config.h"
+#include "start.h" /* for struct lxc_handler */
+#include "utils.h"
 
 /*
  * This file provides the different functions for clients to
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -92,6 +92,7 @@ static const char *lxc_cmd_str(lxc_cmd_t cmd)
 		[LXC_CMD_GET_NAME]         = "get_name",
 		[LXC_CMD_GET_LXCPATH]      = "get_lxcpath",
 		[LXC_CMD_ADD_STATE_CLIENT] = "add_state_client",
+		[LXC_CMD_OPENPTY]          = "openpty",
 	};
 
 	if (cmd >= LXC_CMD_MAX)
@@ -118,10 +119,15 @@ static const char *lxc_cmd_str(lxc_cmd_t cmd)
  */
 static int lxc_cmd_rsp_recv(int sock, struct lxc_cmd_rr *cmd)
 {
-	int ret, rspfd;
+	int ret;
+	int rspfd = -1;
+	int pty_fds[2] = {-1, -1};
 	struct lxc_cmd_rsp *rsp = &cmd->rsp;
 
-	ret = lxc_abstract_unix_recv_fds(sock, &rspfd, 1, rsp, sizeof(*rsp));
+	if (cmd->req.cmd == LXC_CMD_OPENPTY)
+		ret = lxc_abstract_unix_recv_fds(sock, pty_fds, 2, rsp, sizeof(*rsp));
+	else
+		ret = lxc_abstract_unix_recv_fds(sock, &rspfd, 1, rsp, sizeof(*rsp));
 	if (ret < 0) {
 		WARN("Command %s failed to receive response: %s.",
 		     lxc_cmd_str(cmd->req.cmd), strerror(errno));
@@ -148,6 +154,20 @@ static int lxc_cmd_rsp_recv(int sock, struct lxc_cmd_rr *cmd)
 		rspdata->masterfd = rspfd;
 		rspdata->ttynum = PTR_TO_INT(rsp->data);
 		rsp->data = rspdata;
+	} else if (cmd->req.cmd == LXC_CMD_OPENPTY) {
+		struct lxc_cmd_openpty_rsp_data *openpty_rsp = NULL;
+
+		openpty_rsp = malloc(sizeof(*openpty_rsp));
+		if (!openpty_rsp) {
+			ERROR("Command %s couldn't allocate response buffer.",
+					lxc_cmd_str(cmd->req.cmd));
+			close(pty_fds[0]);
+			close(pty_fds[1]);
+			return -1;
+		}
+		openpty_rsp->master = pty_fds[0];
+		openpty_rsp->slave = pty_fds[1];
+		rsp->data = openpty_rsp;
 	}
 
 	if (rsp->datalen == 0) {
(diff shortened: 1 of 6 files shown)


src/lxc/commands.c
```diff
@@ -932,6 +952,90 @@ static int lxc_cmd_add_state_client_callback(int fd, struct lxc_cmd_req *req,
...
+
+	ret = unlockpt(pty_fds[0]);
+	if (ret < 0) {
+		ERROR("%s - Failed to unlock slave pseudoterminal",
+		      strerror(errno));
+		close(pty_fds[0]);
+		return -1;
+	}
+
+	pty_fds[1] = ioctl(pty_fds[0], TIOCGPTPEER, O_RDWR | O_NOCTTY);
```
brauner: Likely needs a kernel or `ENOSYS` check.

brauner: done

src/lxc/conf.c
```diff
@@ -4309,6 +4322,21 @@ int lxc_setup(struct lxc_handler *handler)
 		return -1;
 	}
 
+	/* Allocate "/dev/pts/ptmx" O_PATH file descriptor. */
+	ret = lxc_allocate_dev_ptmx_opath_fd(lxc_conf);
+	if (ret < 0) {
+		ERROR("Failed to allocate \"/dev/pts/ptmx\" file descriptor");
```
hallyn: worth doing a SYSERROR here?

src/lxc/start.c
```diff
@@ -1419,6 +1419,14 @@ static int lxc_spawn(struct lxc_handler *handler)
 	if (lxc_sync_barrier_child(handler, LXC_SYNC_POST_CGROUP))
 		return -1;
 
+	/* Receive "/dev/pts/ptmx" O_PATH file descriptor from child. */
+	if (lxc_abstract_unix_recv_fds(handler->ttysock[1],
+				       &handler->conf->dev_ptmx.opath_fd, 1, NULL,
+				       0) < 0) {
+		ERROR("Failed to send \"/dev/pts/ptmx\" file descriptor");
+		return -1;
```
hallyn: don't you need to goto out_delete_net here?

src/lxc/conf.c
brauner: agreed

src/lxc/start.c
brauner: argh, yeah i do

src/lxc/commands.c
```diff
@@ -935,6 +955,103 @@ static int lxc_cmd_add_state_client_callback(int fd, struct lxc_cmd_req *req,
...
+
+		ERROR("%s - Failed to unlock slave pseudoterminal file "
+		      "descriptor",
+		      strerror(errno));
+		return -1;
+	}
+
+	ret = lxc_abstract_unix_send_fds(fd, pty_fds, 2, NULL, 0);
+	if (ret < 0) {
+		ERROR("Failed to send pseudoterminal file descriptors");
```
brauner: fd leak

src/lxc/commands.c
```diff
@@ -932,6 +955,105 @@ static int lxc_cmd_add_state_client_callback(int fd, struct lxc_cmd_req *req,
...
+
+	/* open the pseudoterminal master */
+	pty_fds[0] = open(proc_self_path, O_RDWR | O_NOCTTY);
+	if (pty_fds[0] < 0) {
+		ERROR("%s - Failed to open \"%s\"", proc_self_path,
+		      strerror(errno));
+		return -1;
+	}
+
+	ret = grantpt(pty_fds[0]);
```
brauner: This needs to be replaced by our own `grantpt()` implementation since libcs don't guarantee that this is not a path-based operation. Instead we should operate directly on the fd itself which shouldn't be too difficult to do.

src/lxc/commands.c
```diff
@@ -932,6 +955,105 @@ static int lxc_cmd_add_state_client_callback(int fd, struct lxc_cmd_req *req,
...
+	if (ret < 0) {
+		ERROR("%s - Failed to grant access to slave pseudoterminal",
+		      strerror(errno));
+		close(pty_fds[0]);
+		return -1;
+	}
+
+	ret = unlockpt(pty_fds[0]);
```
brauner: All libcs I know of use the appropriate `ioctl()` call operating directly on the fd so using `unlockpt()` should be safe.

brauner:
We can now proceed with the patch. See the associated thread that seems to fix the bug I mentioned in the outline for the PR: https://lkml.org/lkml/2017/8/16/752

brauner:
Closing this until we have a viable in-kernel solution. (It's not yet complete.)

PR #1736, opened 2017-08-08
cgfsng: do not use uninitialized variable

brauner:


hallyn:
...  i guess pls ping me when you drop [DO NOT MERGE] :)

Note patch 1/1 seems to change the code more than it needs to just to initialize mode to NULL.

PR #1739, opened 2017-08-11
Fix mem leak with realpath

keloyang:
Signed-off-by: Shukui Yang <yangshukui@huawei.com>


brauner:
jenkins: test this please

hallyn:
Indeed, real_path() already malloc()s the path, so we don't need to dup it.  Thanks.

PR #1761, opened 2017-08-25
further lxc 2.1 preparations

brauner:


brauner:
@hallyn, @stgraber this is quite important so please try to merge relatively soon. :)

hallyn:
This set will require a version bump, right?  lxc-user-nic and liblxc library need to match up, and possibly some other binaries as well.

brauner:
We're already introducing two new API functions already (log and config item checker). So yes, the tools and liblxc need to match so do lxc-user-nic and liblxc. At least lxc-user-nic and library are bundled so we should be good on that front. :)

src/lxc/lxc_user_nic.c
```diff
@@ -1007,6 +1007,25 @@ int main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 
+	if (!strcmp(args.cmd, "delete")) {
+		close(fd);
+
+		if (strcmp(args.type, "ovs")) {
```
hallyn: (I should probably look closer, but)  what's to keep the unprivileged user from deleting any nics he doesn't own?

stgraber:
We don't need a major ABI bump on the library so long as it's backward compatible with older client binaries which it remains. If you build code which uses newer symbols then you obviously won't be able to run that with an older lib, but that doesn't require a major ABI bump on the library side.

src/lxc/lxc_user_nic.c
brauner: That's introduced later by looking at the DB.

PR #1775, opened 2017-09-01
lxc: Do not use strcmp to avoid incorrectly outputs that Criu 2.12.1 …

federicobriata:
…is less than 2.8


brauner:
Hi, you're missing a `Signed-off-by` line which we require in order to include or even look at your code. :)

brauner:
Fwiw, `git commit -s`

evgeni:
Can't we just use [strverscmp](https://linux.die.net/man/3/strverscmp) (which is a GNU extension)?

PR #1777, opened 2017-09-01
use strverscmp instead of strcmp when comparing CRIU versions

evgeni:
Signed-off-by: Evgeni Golov <evgeni@golov.de>


evgeni:
musl has strverscmp, but bionic does not.

well, then you'll have to go with https://github.com/lxc/lxc/pull/1775 instead

federicobriata:
Hi all,
#1775 was closed because not signed and not well formatted, please go with #1780 instead

PR #1781, opened 2017-09-04
stable 2.0: cherry-picks + delta reduction between master and stable 2.0

brauner:


stgraber:
LXD 2.0 still supports the 2.6.32 kernel (since we forgot to announce a bump in base kernel version at release time), so we can't drop the utmp handler in 2.0.x.

PR #1788, opened 2017-09-05
conf: bugfixes

brauner:


hallyn:
I realize this isn't strictly related to your patch, but it brought it to mind - do you think the setup_tty()should be called for lxc_execute?  Is ther a good reason not to?

brauner:
@hallyn, no actually we can set them up. But there won't be an init system to run a getty on them I guess. So sure, send a patch to remove this restriction. :)

hallyn:
Yeah I just looked over the git history.  I'm undecided.

PR #1789, opened 2017-09-05
doc: adapt + update

brauner:


tenforward:
I will translate parts that we need translation later.

PR #1803, opened 2017-09-11
start: switch ids at last possible instance

brauner:

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -962,33 +962,6 @@ static int do_start(void *data)
 		goto out_warn_father;
 	}
 
-	/* The container has been setup. We can now switch to an unprivileged
-	 * uid/gid.
-	 */
-	if (handler->conf->is_execute) {
-		bool have_cap_setgid;
-		uid_t new_uid = handler->conf->init_uid;
-		gid_t new_gid = handler->conf->init_gid;
-
-		/* If we are in a new user namespace we already dropped all
-		 * groups when we switched to root in the new user namespace
-		 * further above. Only drop groups if we can, so ensure that we
-		 * have necessary privilege.
-		 */
-		#if HAVE_LIBCAP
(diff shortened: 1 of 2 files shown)


src/lxc/start.c
```diff
@@ -962,33 +962,6 @@ static int do_start(void *data)
 		goto out_warn_father;
 	}
 
-	/* The container has been setup. We can now switch to an unprivileged
```
hallyn: I see no reason not to do this, but the changelog doesn't tell me why you *want* to do this.  Is there sometimes a problem opening devnull?  Is it just for the principle of the thing?  On failure this task will just go away, right, so there's no cleanup reason to worry about?

src/lxc/storage/storage.c
```diff
@@ -337,6 +337,11 @@ struct lxc_storage *storage_copy(struct lxc_container *c, const char *cname,
 	struct rsync_data data = {0};
 	char cmd_output[MAXPATHLEN] = {0};
 
+	if (!src) {
+		ERROR("No rootfs specified");
```
hallyn: Under what conditions do we get to this case?

I'm wondering whether the ERROR message is warranted.  Is it indicative of a particular misconfiguration on the user's part, or a bug in our code, or is it just a case that sometimes happens, and we should just return NULL quietly?

brauner: If you remove the `lxc.rootfs.path` entry from the containers config file you'd default on copy but you'd work correctly on e.g. execute. That's why I want to check.

src/lxc/start.c
brauner: This is just on principle. In case someone wants to add an operation later on that requires privilege it should not be blocked by switching IDs.

src/lxc/storage/storage.c
hallyn: Sorry - what does "default on copy" mean?  is that autocorrect for segfault?

brauner: Ups, yes "segfault".

PR #1806, opened 2017-09-12
[TRIVIAL] fix regex-typo in documentation example

roques:
To match names beginning with the letters `f` or `b` one can use `[fb].*` or `(f|b).*`, but not `[f|b].*`, which would match strings beginning with `f`, `|`, or `b`.


brauner:
Hi, you need to sign-offs your commit for us to be able to include it.

roques:
@brauner, signed-off.  I'm sorry for the hassle, I'm causing for such a trivial fix.

brauner:
@roques, all good, no problem at all. :)

PR #1811, opened 2017-09-13
Change alpine template file check to also check file size (`-f` => `-s`)

runejuhl:
Because the `fetch` wget wrapper outputs files to stdout we may end up in a
situation where wget fails but the files are still created. This can happen
e.g. when the host date is out of sync leading to a failed certificate
check, resulting in the creation of empty key files.

Once the empty files have been created the template will try to use them which
causes the certificate check to fail.

By using `-s` instead of `-f` the template will re-fetch the files unless they
exist AND have a size greater than zero.

### DIFF
--- a/templates/lxc-alpine.in
+++ b/templates/lxc-alpine.in
@@ -185,7 +185,7 @@ fetch_apk_keys() {
 
 	echo "$APK_KEYS_SHA256" | while read -r line; do
 		keyname="${line##* }"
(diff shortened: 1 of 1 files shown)


stgraber:
Can you update your commit message to include the Signed-off-by line (see CONTRIBUTING).

runejuhl:
@stgraber absolutely -- sorry for missing it in the first place

PR #1824, opened 2017-09-20
zfs: return error directly when zfs creation fails

brauner:


brauner:
@stgraber, fixes `-B best`.

PR #1826, opened 2017-09-20
add a pre-start-host hook

hallyn:
This should satisfy several use cases.  The one I tested for was CNI.
I replaced the network configuration in a root owned container with:

lxc.net.0.type = empty
lxc.hook.pre-start-host = /bin/lxc-prestart-netns

where /bin/lxc-prestart-netns contained:

=================================

echo "starting" > /tmp/debug
ip link add host1 type veth peer name peer1
ip link set host1 master lxcbr0
ip link set host1 up
ip link set peer1 netns "${LXC_PID}"
=================================

The nic 'peer1' was placed into the container as expected.

FOr this to work, we pass the container init's pid as LXC_PID in
an environment variable, since lxc-info cannot work at that point.


brauner:
Looks good to me. @stgraber, any issues with the naming? Suggestions?

hallyn:
Yeah I'm not sold on the name.  It'll also need manpage update of course.

hallyn:
Oh, right - I'm pretty sure that for this to be useful for CNI we will in fact have to get rid of the default arguments we pass to the hooks.  Or I suppose we can require that the config call a script which in turn calls the hook...

Dropping those arguments was discussed recently - are we ready to drop those?  In 3.0?  Earlier?

hallyn:
No wait.  IIUC cni will take config on stdin and environment variables, no arguments.  So maybe no rush on dropping the args.  If that hasn't changed.

brauner:
Hm, how about `start-host`.

hallyn:
yeah start-host makes more sense.  Thanks!

hallyn:
All right I will rename it to start-host and add a manpage entry.

Issue #538, opened 2015-05-26
lxc-start cannot share namespaces in daemon mode or with --close-all-fds


stewartbrodie:
Version: lxc-1.1.2

None of the namespace sharing options for lxc-start (--share-net --share-ipc --share-uts) work if the container is being started with the --close-all-fds option.  Since this option is implied by --daemon, then the problem occurs in daemon mode as well.

Namespaces can be shared succcessfully if the --foreground option is specified and --close-all-fds is not specified.

e.g. lxc-start -n net --share-ipc $$

The initial lxc-start process successfully opens /proc/$$/ns/ipc but after the double-fork, the grandchild process then very early on closes the file descriptor before it's called setns().  Subsequently, the setns() call fails with EINVAL because it doesn't like the file descriptor.  In the cases I've seen, the file descriptor is open as it happens to have been re-used.  If it hadn't been, I expect I'd have seen EBADF instead.

I did a quick experiment whereby I made lxc_check_inherited() ignore any fds it finds in conf->inherit_ns_fd[] and whilst that seems to achieve the effect I want (the container starts and appears to be using the inherited IPC namespace), it makes other things unhappy.  (Is it because I'm interfering with the IPC namespace?)

<log>
/ # lxc-start -n net --logfile /tmp/net.log --logpriority TRACE --share-ipc 1
lxc-start: monitor.c: lxc_monitor_open: 208 connect : backing off 10
lxc-start: monitor.c: lxc_monitor_open: 208 connect : backing off 50
lxc-start: monitor.c: lxc_monitor_open: 208 connect : backing off 100
lxc-start: monitor.c: lxc_monitor_open: 213 connect : Connection refused
lxc-start: lxc_start.c: main: 344 The container failed to start.
lxc-start: lxc_start.c: main: 346 To get more details, run the container in foreground mode.
lxc-start: lxc_start.c: main: 348 Additional information can be obtained by setting the --logfile and --logpriority options.
</log>
<log>
INFO     lxc_monitor - monitor.c:lxc_monitor_sock_name:177 - using monitor sock name lxc/526ce194f37e8eec//containers
ERROR    lxc_monitor - monitor.c:lxc_monitor_open:208 - connect : backing off 10
ERROR    lxc_monitor - monitor.c:lxc_monitor_open:208 - connect : backing off 50
ERROR    lxc_monitor - monitor.c:lxc_monitor_open:208 - connect : backing off 100
[...]

PR #1834, opened 2017-09-28
Fix issue #538

lifeng68:
Fix issue #538, do not close the inherit namespace fds.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -236,6 +236,15 @@ restart:
 		    (i < len_fds && fd == fds_to_ignore[i]))
 			continue;
 
+		if (conf) {
+			for (i = 0; i < LXC_NS_MAX; i++)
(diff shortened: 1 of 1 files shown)


brauner:
Looks good to me. Note however, that `--share-net` will only work with privileged containers now since there's currently no code to handle attaching to the user namespace as well which you need to do because kernel. But we can merge this now and figure out the other part later. I'll grab your commit and give it a more reasonable commit message. :)

brauner:
Grabbed your commit and put it into a branch of mine with a more clear commit message. So closing this in favor of https://github.com/lxc/lxc/pull/1835 .

hallyn:
fwiw I do have a branch where I was working on the share-net thing.  I will hopefully push on that once I rework the CNI branch.

Thanks guys!

PR #1836, opened 2017-09-29
Add support share pid namespace

lifeng68:
use-case for share pid namespace:
If we share the PID namespace, then we can start a container to debug another running container with debug tools (gdb).


hallyn:
jenkins: test this please

PR #1838, opened 2017-10-01
 network: clear ifindeces

brauner:
We need to clear any ifindeces we recorded so liblxc won't have cached stale
data which would cause it to fail on reboot we're we don't re-read the on-disk
config file.


hallyn:
I see no place where you check the return value of the new fn. Is that going to be needed, or can you make it return void?

brauner:
@hallyn, I made the function void. If we need it later on to return something the patch is trivial.

hallyn:
Thanks.

PR #1839, opened 2017-10-02
Add OCI container creation template

hallyn:
This adds preliminary (but working) support for creating application
containers from OCI formats.  Examples:

sudo lxc-create -t oci -n a1 -- -u oci:../oci:alpine -t alpine

create a container from a local OCI layout in ../oci.

sudo lxc-create -t oci -n u1 -- -u docker://ubuntu -t ubuntu

create a container pulling from the docker hub.

I'm pretty sure we'll wan to fine-tune the way "urls' are
specified.  The separate specifying of a 'tag' is unfortunate.
Maybe we can be smart enough to know to pull the name for
dockerhub and the tag for oci:.  But for now it simplified things
to the point of making it easy enough to implement and use.

Comments appreciated.

TODO:
 - implement lxc.execute.cmd configuration key
 - pull more of the OCI runtime information like memory/cpu limits

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -898,6 +898,7 @@ AC_CONFIG_FILES([
 	templates/lxc-fedora
 	templates/lxc-fedora-legacy
 	templates/lxc-gentoo
+	templates/lxc-oci
 	templates/lxc-openmandriva
 	templates/lxc-opensuse
 	templates/lxc-oracle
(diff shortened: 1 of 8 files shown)


hallyn:
Hm, please hold off on this.  Got a little bug in the entry point parsing.

hallyn:
Ok, should be good now.

PR #1840, opened 2017-10-03
drop useless apparmor denies

tych0:
mem and kmem are really in /dev, so this does us no good.

### DIFF
--- a/config/apparmor/abstractions/container-base
+++ b/config/apparmor/abstractions/container-base
@@ -72,8 +72,6 @@
 
   # block some other dangerous paths
   deny @{PROC}/kcore rwklx,
-  deny @{PROC}/kmem rwklx,
-  deny @{PROC}/mem rwklx,
   deny @{PROC}/sysrq-trigger rwklx,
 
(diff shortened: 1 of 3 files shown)


stgraber:
Looks good but some more files need updating, can you run:
 - ./lxc-generate-aa-rules.py container-rules.base > container-rules
 - cat abstractions/container-base.in container-rules > abstractions/container-base

That should take care of the rest.

GamerSource:
Yout talk about removing denies for _/proc/kmem_ and _/proc/mem_ but in config/apparmor/abstractions/container-base.in _kcore_ and _kmem_ gets removed from the deny list, _mem_ even stayed, but kcore is in proc!

Is this really correct?

stgraber:
Oh, that's a good point, the container-base.in diff doesn't match the commit description :)

tych0:
Yep, I just updated. Thanks!

hallyn:
Sorry...  what's the downside to having these in?

I agree they're probably useless, but there *may* be sites out there that have custom templates causing containers to be created with those files.  I'm not nacking, but not quite sold on the tradeoff.

PR #1844, opened 2017-10-04
implement lxc_string_split_quoted

hallyn:
lxc_string_split_quoted() splits a string on spaces, but keeps
groups in single or double qoutes together.  In other words,
generally what we'd want for argv behavior.

Switch lxc-execute to use this for lxc.execute.cmd.

Switch lxc-oci template to put the lxc.execute.cmd inside single
quotes, because parse_line() will eat those.  If we don't do that,
then if we have lxc.execute.cmd = /bin/echo "hello, world", then the
last double quote will disappear.

### DIFF
--- a/src/lxc/tools/lxc_execute.c
+++ b/src/lxc/tools/lxc_execute.c
@@ -99,10 +99,7 @@ static bool set_argv(struct lxc_conf *conf, struct lxc_arguments *args)
 	if (!conf->execute_cmd)
 		return false;
 
-	/* TODO -
-	   we should honor '"' etc; This seems worth a new helper in utils.c.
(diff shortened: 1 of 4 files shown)


stgraber:
missing sign-off?

brauner:
I commented on IRC already, but I'm doing it here again for future reference. Starting with liblxc 3.0 we should maybe consider reworking the `parse_line()` part such that it doesn't strip `""` by default or only where it makes sense. I'm not completely sure right now whether we can do this without regressing the whole world but it is at least worth thinking about it.

hallyn:
We'd have to define "where it makes sense" :)

PR #1851, opened 2017-10-10
init: rework dumb init

brauner:

@hallyn:
[asciicast](https://asciinema.org/a/141675)


hallyn:
Thanks.  lxc-execute is starting to look good.

brauner:
Note that for `lxc-execute` to behave correctly and for `bash` to be able to initialize job control correctly I need to merge https://sourceware.org/bugzilla/show_bug.cgi?id=22145 into `glibc` upstream. The author still needs to send it to `libc-alpha` so that I can review it and get consensus. I expect this to happen rather soon though. Since this is a regression in `glibc` I can backport this back into `glibc 2.26` as well.

PR #1852, opened 2017-10-11
POC: container live patching

brauner:
This adds set_running_config_item() which is the analogue of
get_running_config_item(). In essence it allows a caller to livepatch the
container's in-memory configuration. This POC is severly limited. Here are the
most obvious ones:
- Only the container's in-memory config can be updated but no further actions
  (e.g. on-disk actions) are made.
- Only keys in the "lxc.net." namespace can be changed. This POC also allows
  updating an existing network. For example it allows to change the network
  type of an existing network. This is obviously nonsense and in a non-POC
  implementation this should be blocked.

Use Case:
Callers can hotplug a new network for the container. For example, LXD can
create a pair of veth devices in the host and in the container and add it to
the container's in-memory config. This means, the container can later be
queried for the name of the device later on etc. Note that liblxc will
currently not delete hotplugged network devices on container shutdown since it
won't have the ifindex of the container.

Relates to lxc/lxd#3920 .

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -2838,7 +2838,7 @@ static bool do_lxcapi_set_running_config_item(struct lxc_container *c, const cha
 
 	ret = lxc_cmd_set_config_item(c->name, key, v, do_lxcapi_get_config_path(c));
 	if (ret < 0)
-		ERROR("Failed to live patch container");
+		SYSERROR("%s - Failed to live patch container", strerror(-ret));
 
 	container_mem_unlock(c);
 	return ret == 0;
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -2765,9 +2765,35 @@ static bool do_lxcapi_destroy_with_snapshots(struct lxc_container *c)
 
 WRAP_API(bool, lxcapi_destroy_with_snapshots)
 
-static bool set_config_item_locked(struct lxc_container *c, const char *key, const char *v)
+int lxc_set_config_item_locked(struct lxc_conf *conf, const char *key,
+			       const char *v)
 {
+	int ret;
 	struct lxc_config_t *config;
+	bool bret = true;
+
+	config = lxc_get_config(key);
+	if (!config)
+		return -EINVAL;
+
+	ret = config->set(key, v, conf, NULL);
+	if (ret < 0)
+		return -EINVAL;
+
+	if (lxc_config_value_empty(v))
+		do_clear_unexp_config_line(conf, key);
+	else
+		bret = do_append_unexp_config_line(conf, key, v);
+	if (!bret)
+		return -ENOMEM;
+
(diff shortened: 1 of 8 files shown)


src/lxc/lxccontainer.c
```diff
@@ -2765,29 +2765,47 @@ static bool do_lxcapi_destroy_with_snapshots(struct lxc_container *c)
...
+		do_clear_unexp_config_line(conf, key);
+	else
+		bret = do_append_unexp_config_line(conf, key, v);
+	if (!bret)
+		return -ENOMEM;
+
+	return 0;
+}
+
+static bool do_set_config_item_locked(struct lxc_container *c, const char *key,
```
hallyn: Generally fn() would call do_fn(), right?  This do_fn() and lxc_fn() is confusing and might become problematic...  Unfortunately I don't have any better ideas for naming.  Do you?  (If not, then I guess we leave it as is)

src/tests/livepatch.c
```diff
@@ -0,0 +1,176 @@
...
+	int ret = EXIT_FAILURE;
+
+	c = lxc_container_new("livepatch", NULL);
+	if (!c) {
+		lxc_error("%s", "Failed to create container \"livepatch\"");
+		exit(ret);
+	}
+
+	if (c->is_defined(c)) {
+		lxc_error("%s\n", "Container \"livepatch\" is not defined");
```
hallyn: is *not* defined?

src/tests/livepatch.c
```diff
@@ -0,0 +1,176 @@
...
+		free(value);
+		goto on_error_destroy;
+	}
+	free(value);
+
+	/* Remove in-memory value. */
+	if (!c->set_running_config_item(c, "lxc.net.1.type", "")) {
+		lxc_error("%s\n", "Failed to clear running config item "
+				  "\"lxc.net.1.type\"");
+		goto on_error_destroy;
```
hallyn: Can you add a test having the container reboot, and make sure it comes up right?

And maybe also, separately, stop+start it?

src/tests/livepatch.c
brauner: Yeah, this should be `s/is not defined/is defined/`. :)

src/tests/livepatch.c
brauner: Yeah, can do.

src/lxc/lxccontainer.c
brauner: Yeah, I mean we can reverse the naming for the two functions and export the `do_fn()` method in `lxc.h`. Wouldn't matter much I'd say.

brauner:
@hallyn, updated. Tests for reboot and stop + start added.

src/lxc/lxccontainer.c
hallyn: Yeah, let's leave it.  Thanks.

PR #1855, opened 2017-10-12
arguments: print "-devel" when LXC_DEVEL is true

brauner:
liblxc should inform users that they are using a devel version. This will have
liblxc print

    MAJOR.MINOR.PATCH-devel

if LXC_DEVEL is true and

    MAJOR.MINOR.PATCH

otherwise.

### DIFF
--- a/src/lxc/arguments.c
+++ b/src/lxc/arguments.c
@@ -131,7 +131,7 @@ static void print_usage(const struct option longopts[],
 
 static void print_version()
 {
-	printf("%s\n", LXC_VERSION);
+	printf("%s%s\n", LXC_VERSION, LXC_DEVEL ? "-devel" : "");
 	exit(0);
 }
 


stgraber:
Is that going to affect the version checking code in go-lxc for example?

brauner:
@stgraber, I have a patch for go-lxc.

stgraber:
Good. What happens with the current go-lxc when it gets the new answer from liblxc? does it crash or just move on?

brauner:
@stgraber, it'll move on. The worst case scenario is that the output of go-lxc would be:

    liblxc returns: 2.1.0-devel
    go-lxc returns: 2.1.0-devel (devel)

brauner:
https://github.com/lxc/go-lxc/pull/89 is the corresponding go-lxc PR.

PR #1860, opened 2017-10-15
fix build failures on x32

kilobyte:
There's two new build failures on x32, both due to assumption that some type is identical to ｢long｣.

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -1453,7 +1453,7 @@ static int set_config_prlimit(const char *key, const char *value,
 {
 	struct lxc_list *iter;
 	struct rlimit limit;
-	unsigned long limit_value;
+	rlim_t limit_value;
 	struct lxc_list *limlist = NULL;
 	struct lxc_limit *limelem = NULL;
 
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -222,7 +222,7 @@ int lxc_unix_epoch_to_utc(char *buf, size_t bufsize, const struct timespec *time
 	seconds = (time->tv_sec - d_in_s - h_in_s - (minutes * 60));
 
 	/* Make string from nanoseconds. */
-	ret = snprintf(nanosec, LXC_NUMSTRLEN64, "%ld", time->tv_nsec);
+	ret = snprintf(nanosec, LXC_NUMSTRLEN64, "%"PRId64, (int64_t)time->tv_nsec);
 	if (ret < 0 || ret >= LXC_NUMSTRLEN64)
 		return -1;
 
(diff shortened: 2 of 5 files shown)


src/lxc/log.c
```diff
@@ -222,7 +222,7 @@ int lxc_unix_epoch_to_utc(char *buf, size_t bufsize, const struct timespec *time
 	seconds = (time->tv_sec - d_in_s - h_in_s - (minutes * 60));
 
 	/* Make string from nanoseconds. */
-	ret = snprintf(nanosec, LXC_NUMSTRLEN64, "%ld", time->tv_nsec);
+	ret = snprintf(nanosec, LXC_NUMSTRLEN64, "%ld", (long)time->tv_nsec);
```
brauner: Hm. I'd rather prefer if we'd use `uin64_t` here so:

```
ret = snprintf(nanosec, LXC_NUMSTRLEN64, "%"PRId64, (int64_t)time->tv_nsec);
```

src/lxc/confile.c
```diff
@@ -1453,7 +1453,7 @@ static int set_config_prlimit(const char *key, const char *value,
 {
 	struct lxc_list *iter;
 	struct rlimit limit;
-	unsigned long limit_value;
+	rlim_t limit_value;
```
brauner: @Blub, care to give this a quick look...

Blub:
There's also a check for`sys/resource.h`, we'd have to either put `#if`s around all the code or have conf.h provide `rlim_t` when the header isn't there. Also what I forogot initially is that `setup_resource_limits()` in conf.c needs to deal with a missing resources.h/`prlimit()` by erroring if the limit list is not empty.

src/lxc/log.c
kilobyte: Unlike tv->sec (Y2038), tv->nsec is guaranteed to never exceed 10⁹: it'd roll over to a whole second, which is explicitly forbidden — the kernel checks for and rejects such values on input, and won't produce them itself.  Thus, the ABI using a 64-bit field on some archs here is a bit pointless (I guess, done for alignment reasons?).

But on the other hand, using a too wide field is harmless, and makes it easier to avoid mistakes with time_t next to it.

kilobyte:
As for `sys/resource.h`, I don't quite understand why it's checked for — it's required by POSIX ([ref](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/resource.h.html)).

src/lxc/log.c
brauner: I make it a policy to cast to stdint types and usually the maximum one when it can be safely done. This should put us on the safe side on most platforms.

Blub:
Indeed, I suppose the `prlimit()` check for bionic covers all the special cases, so the resource.h check can probably go.

brauner:
jenkins: test this please

PR #1864, opened 2017-10-18
ringbuffer: implement efficient and performant ringbuffer

### LINKED ISSUE #1857: Implement a console ringbuffer
In LXD (and I suspect other environments like Proxmox), there's the need to provide console access to users, including re-playing the last few messages that were written to the console while nothing was looking.

This addresses both getting an interactive console in the container and early boot debugging where /dev/console tends to be the only way to figure this out.

LXC currently supports lxc.console.logfile where the console output is duplicated to a text file. The problem is that this file will keep growing based on whatever hits /dev/console in the container, making it a resource exhaustion DoS attack point.

To address this, I believe we want:

 - A new option like "lxc.console.buffersize" which one could set to 100kB or some other amount of space.
 - New API functions and monitor commands to retrieve the buffer as well as to flush it. This would allow LXD and other callers to retrieve what output happened while nothing was connected, then flush the buffer when disconnecting.
 - If the buffersize is set, then lxc.console.logfile should become an on-disk copy of that ringbuffer. This will be useful as a way to get console output when the container dies immediately at startup.

brauner:
liblxc will use a ringbuffer implementation that employs mmap()ed memory.
Specifically, the ringbuffer will create an anonymous memory mapping twice the
requested size for the ringbuffer. Afterwards, an in-memory file the requested
size for the ringbuffer will be created. This in-memory file will then be
memory mapped twice into the previously established anonymous memory mapping
thereby effectively splitting the anoymous memory mapping in two halves of
equal size.  This will allow the ringbuffer to get rid of any complex boundary
and wrap-around calculation logic. Since the underlying physical memory is the
same in both halves of the memory mapping only a single memcpy() call for both
reads and writes from and to the ringbuffer is needed.

Design Notes:
- Since we're using MAP_FIXED memory mappings to map the same in-memory file
  twice into the anonymous memory mapping the kernel requires us to always
  operate on properly aligned pages. To guarantee proper page aligment the size
  of the ringbuffer must always be a muliple of the kernel's page size. This
  also implies that the minimum size of the ringbuffer must be at least equal to
  one page size. This additional requirement is reasonably unproblematic.
  First, any ringbuffer smaller than the size of a single page is very likely
  useless since the standard page size on linux is 4096 bytes.
- Because liblxc is not able to predict the output a user is going to produce
  (e.g. users could cat binary files onto the console) and because the
  ringbuffer is located in a hotpath and needs to be as performant as possible
  liblxc will not parse the buffer.

Use Case:
The ringbuffer is needed by liblxc in order to safely log the output of write
intensive callers that produce unpredictable output or unpredictable amounts of
output. The console output created by a booting system and the user is one of
those cases. Allowing a container to log the console's output to a file it
would be possible for a malicious user to fill up the host filesystem by
producing random ouput on the container's console if quota support is either
not enabled or not available for the underlying filesystem. Using a ringbuffer
is a reliable and secure way to ensure a fixed-size log.

Closes #1857.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -80,6 +80,7 @@
 #include "namespace.h"
 #include "network.h"
 #include "parse.h"
+#include "ringbuf.h"
 #include "storage.h"
 #include "storage/aufs.h"
 #include "storage/overlay.h"
@@ -2428,6 +2429,7 @@ struct lxc_conf *lxc_conf_init(void)
 	new->autodev = 1;
 	new->console.log_path = NULL;
 	new->console.log_fd = -1;
+	new->console.log_size = 0;
 	new->console.path = NULL;
 	new->console.peer = -1;
 	new->console.peerpty.busy = -1;
@@ -2436,6 +2438,7 @@ struct lxc_conf *lxc_conf_init(void)
 	new->console.master = -1;
 	new->console.slave = -1;
 	new->console.name[0] = '\0';
+	memset(&new->console.ringbuf, 0, sizeof(struct lxc_ringbuf));
 	new->maincmd_fd = -1;
 	new->nbd_idx = -1;
 	new->rootfs.mount = strdup(default_rootfs_mount);
@@ -3079,6 +3082,64 @@ static bool verify_start_hooks(struct lxc_conf *conf)
 	return true;
 }
 
+/**
+ * Note that this function needs to run before the mainloop starts. Since we
+ * register a handler for the console's masterfd when we create the mainloop
+ * the console handler needs to see an allocated ringbuffer.
+ */
+static int lxc_setup_console_ringbuf(struct lxc_console *console)
+{
+	int ret;
+	struct lxc_ringbuf *buf = &console->ringbuf;
+	uint64_t size = console->log_size;
+
+	/* no ringbuffer previously allocated and no ringbuffer requested */
+	if (!buf->addr && size <= 0)
+		return 0;
+
+	/* ringbuffer allocated but no new ringbuffer requested */
+	if (buf->addr && size <= 0) {
+		lxc_ringbuf_release(buf);
+		buf->addr = NULL;
+		buf->r_off = 0;
+		buf->w_off = 0;
+		buf->size = 0;
+		TRACE("Deallocated console ringbuffer");
+		return 0;
+	}
+
+	if (size <= 0)
+		return 0;
+
+	/* check wether the requested size for the ringbuffer has changed */
+	if (buf->addr && buf->size != size) {
+		TRACE("Console ringbuffer size changed from %" PRIu64
+		      " to %" PRIu64 " bytes. Deallocating console ringbuffer",
+		      buf->size, size);
+		lxc_ringbuf_release(buf);
+	}
+
+	ret = lxc_ringbuf_create(buf, size);
+	if (ret < 0) {
+		ERROR("Failed to setup %" PRIu64 " byte console ringbuffer", size);
+		return -1;
+	}
+
+	TRACE("Allocated %" PRIu64 " byte console ringbuffer", size);
+	return 0;
+}
+
+int lxc_setup_parent(struct lxc_handler *handler)
+{
+	int ret;
+
+	ret = lxc_setup_console_ringbuf(&handler->conf->console);
+	if (ret < 0)
+		return -1;
+
+	return 0;
+}
+
 int lxc_setup_child(struct lxc_handler *handler)
 {
 	int ret;
@@ -3459,6 +3520,8 @@ void lxc_conf_free(struct lxc_conf *conf)
 		current_config = NULL;
 	free(conf->console.log_path);
 	free(conf->console.path);
+	if (conf->console.log_size > 0 && conf->console.ringbuf.addr)
+		lxc_ringbuf_release(&conf->console.ringbuf);
 	free(conf->rootfs.mount);
 	free(conf->rootfs.bdev_type);
 	free(conf->rootfs.options);
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -68,10 +68,6 @@
 #include <../include/openpty.h>
 #endif
 
-#ifdef HAVE_LINUX_MEMFD_H
-#include <linux/memfd.h>
-#endif
-
 #include "af_unix.h"
 #include "caps.h"       /* for lxc_caps_last_cap() */
 #include "cgroup.h"
@@ -179,59 +175,6 @@ static int sethostname(const char * name, size_t len)
 
 #ifndef MS_LAZYTIME
 #define MS_LAZYTIME (1<<25)
-#endif
-
-/* memfd_create() */
-#ifndef MFD_CLOEXEC
-#define MFD_CLOEXEC 0x0001U
-#endif
-
-#ifndef MFD_ALLOW_SEALING
-#define MFD_ALLOW_SEALING 0x0002U
-#endif
-
-#ifndef HAVE_MEMFD_CREATE
-static int memfd_create(const char *name, unsigned int flags) {
-	#ifndef __NR_memfd_create
-		#if defined __i386__
-			#define __NR_memfd_create 356
-		#elif defined __x86_64__
-			#define __NR_memfd_create 319
-		#elif defined __arm__
-			#define __NR_memfd_create 385
-		#elif defined __aarch64__
-			#define __NR_memfd_create 279
-		#elif defined __s390__
-			#define __NR_memfd_create 350
-		#elif defined __powerpc__
-			#define __NR_memfd_create 360
-		#elif defined __sparc__
-			#define __NR_memfd_create 348
-		#elif defined __blackfin__
-			#define __NR_memfd_create 390
-		#elif defined __ia64__
-			#define __NR_memfd_create 1340
-		#elif defined _MIPS_SIM
-			#if _MIPS_SIM == _MIPS_SIM_ABI32
-				#define __NR_memfd_create 4354
-			#endif
-			#if _MIPS_SIM == _MIPS_SIM_NABI32
-				#define __NR_memfd_create 6318
-			#endif
-			#if _MIPS_SIM == _MIPS_SIM_ABI64
-				#define __NR_memfd_create 5314
-			#endif
-		#endif
-	#endif
-	#ifdef __NR_memfd_create
-	return syscall(__NR_memfd_create, name, flags);
-	#else
-	errno = ENOSYS;
-	return -1;
-	#endif
-}
-#else
-extern int memfd_create(const char *name, unsigned int flags);
 #endif
 
 char *lxchook_names[NUM_LXC_HOOKS] = {"pre-start", "pre-mount", "mount",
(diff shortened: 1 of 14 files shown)


brauner:
@hallyn, this only adds the internal plumbing for the ringbuffer but doesn't implement the API. This is on purpose. We should come up with two nice + extensible API functions to retrieve the current console ringbuffer and clear the current console ringbuffer.

brauner:
Also the name of the config key is currently `lxc.console.logsize'. Better naming suggestions very welcome!

src/lxc/ringbuf.c
```diff
@@ -0,0 +1,150 @@
...
+	if (buf->size % getpagesize())
+		return -EINVAL;
+
+	buf->addr = mmap(NULL, buf->size * 2, PROT_NONE,
+			 MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+	if (buf->addr == MAP_FAILED)
+		return -EINVAL;
+
+	/* For kernels < 3.17 we can fallback to mkstemp() + unlink(). Not
+	 * required for the RFC.
```
brauner: Old comment.-

src/lxc/utils.c
```diff
@@ -2339,3 +2360,132 @@ bool lxc_nic_exists(char *nic)
...
+		oflags &= ~FD_CLOEXEC;
+
+	return fcntl(fd, F_SETFD, oflags);
+}
+
+int lxc_make_tmpfile(char *template, int suffixlen, bool cloexec, bool rm)
+{
+	int fd, ret;
+
+	fd = mkstemps(template, suffixlen);
```
stgraber: jenkins doesn't like that on Android

brauner: /me rolls eyes at Android.

src/lxc/utils.c
```diff
@@ -2339,3 +2339,19 @@ bool lxc_nic_exists(char *nic)
 
 	return true;
 }
+
+int lxc_cloexec(int fd, bool set)
+{
+	int oflags;
+
+	oflags = fcntl(fd, F_GETFD, 0);
```
hallyn: I'm probably overthinking this but...   since liblxc is sometimes used in very highly threaded contexts - is this too racy?  (By the time we get around to our F_SETFD, someone else may have changed the base flags)

I'm probably overthinking this.

src/lxc/utils.c
```diff
@@ -2355,3 +2355,31 @@ int lxc_cloexec(int fd, bool set)
...
+	if (rm) {
+		ret = unlink(template);
+		if (ret < 0)
+			goto on_error;
+	}
+
+	return fd;
+
+on_error:
+	close(fd);
```
hallyn: Do you want to rm it here too?

src/lxc/utils.c
brauner: That's actually a good point. We might simply set the flag and call it a day...

src/lxc/utils.c
brauner: Yup.

src/lxc/confile.c
```diff
@@ -1790,6 +1792,45 @@ static int set_config_console_logfile(const char *key, const char *value,
...
+	}
+
+	ret = parse_byte_size_string(value, &size);
+	if (ret < 0)
+		return -1;
+
+	/* Must be at least a page size */
+	pgsz = getpagesize();
+	if (size < pgsz) {
+		ERROR("%s - The size of the console log must be at least %d "
```
hallyn: Would it be better to just automatically set it to the min?

Otherwise users could get annoying start failures when moving a container to a different system.

brauner: But warn users in the log?

src/lxc/start.c
```diff
@@ -937,7 +937,7 @@ static int do_start(void *data)
 	}
 
 	/* Setup the container, ip, names, utsname, ... */
-	ret = lxc_setup(handler);
+	ret = lxc_setup_child(handler);
```
hallyn: This belongs in the previous patch, right?

src/lxc/console.c
```diff
@@ -196,6 +196,15 @@ static int lxc_console_cb_con(int fd, uint32_t events, void *data,
 
 		if (console->peer >= 0)
 			w = lxc_write_nointr(console->peer, buf, r);
+
+		if (console->log_size > 0) {
```
hallyn: I *think* this is right, but just to make sure - is there any case where a console reader or writer will be in a different thread?

src/lxc/conf.c
```diff
@@ -3079,6 +3082,35 @@ static bool verify_start_hooks(struct lxc_conf *conf)
...
+static int lxc_setup_console_ringbuf(struct lxc_console *console)
+{
+	int ret;
+	struct lxc_ringbuf *buf = &console->ringbuf;
+	uint64_t size = console->log_size;
+
+	if (size <= 0)
+		return 0;
+
+	ret = lxc_ringbuf_create(buf, size);
```
hallyn: Doesn't lxc_console_free() need to be updated?

hallyn: I guess it's just an mmap so no big deal

src/lxc/ringbuf.c
```diff
@@ -0,0 +1,146 @@
...
+	return -1;
+}
+
+void lxc_ringbuf_move_read_addr(struct lxc_ringbuf *buf, size_t len)
+{
+	buf->r_off += len;
+
+	if (buf->r_off < buf->size)
+		return;
+	fprintf(stderr, "BLA\n");
```
hallyn: ?

hallyn:
Thanks, looks good overall.  I think the name for lxc.console.logsize is good.

Please ping me when you feel feedback is addressed.

src/lxc/ringbuf.c
brauner: lol

src/lxc/utils.c
brauner: addressed

src/lxc/utils.c
brauner: addressed

src/lxc/start.c
brauner: addressed

src/lxc/ringbuf.c
brauner: addressed

src/lxc/conf.c
brauner: That is an interesting question. I think not because `lxc_console_free()` is only ever called when you allocate the console or ttys from an already running container. In this case the console ringbuffer will have already been setup by the parent so when you detach from the console again but you'd free the ringbuffer you'd make the parent crash since the console epoll handler will still try to write to it.

src/lxc/confile.c
brauner: addressed

src/lxc/console.c
brauner: @hallyn, I don't think so. We need to go through the container's command socket to request an operation on the console ringbuffer to be performed. The command socket handler and console handler will always be in the same thread since they are setup together on container in start in the epoll loop.

brauner: My point being that the lock needed for an operation on the ringbuffer will always be taken by the monitor.

brauner:
@hallyn, updated. The changes we discussed on irc but that haven't been discussed in a comment here are:

- `parse_byte_size_string()` will consider `""` and `NULL`  invalid input and return `-EINVAL`. The redundant check for `end == dup` has been removed.
- `lxc_find_next_power2()` will consider `0` invalid input. Since it returns a `uint64_t` I have it return `0` in response to `0` as input since `0` is not a power of `2`.

brauner:
jenkins: test this please

brauner:
@hallyn, what I wonder about is where the deallocation for the console ringbuffer will need to happen. Should it happen in `lxc_conf_free()` - which means that the deallocation happens on `lxc_container_put()`- or on container stop. I'm tempted to say the deallocation needs to happen on container stop and not on `lxc_container_put()`. Because it's perfectly valid to not call `lxc_container_put()` and simply start the container again which would mean that you'd create a new ringbuffer and loose track of the old mapping without deallocation it first. Dellocating it on stop should also be fine from a users perspective because when the container has stopped the command socket handler will be gone which means - in the implementation I'm thinking about - that users cannot send any requests to get or clear the console ringbuffer anymore. They are basically blocked from accessing the console ringbuffer.
Wait, there might be a smarter option... Currently the console ringbuffer is initialized on start when the check `if (console->log_size > 0) allocate_console_ringbuffer()` is true. We should simply check whether the ringbuffer already has a valid address aka `if (console->ringbuffer.addr) resuse_buffer()`. Yeah, that's the right thing to do. But then the question becomes do we clear the buffer on startup or do we keep the former read and write offset...
I'll wait for you to comment before I make any changes. :) But I think the second option (reusing the buffer) is correct. So deallocation can happen on `lxc_container_put()`...

brauner:
Hm, you need to account for the following API sequences:

- `c = lxc_container_new()`
- `c->start()` /* logsize == 4096 */
- `c->stop()`
- `c->start()` /* logsize == 4096 --> reuse ringbuffer */
- `c->stop()`
- `c->clear_config()`
- `c->load_config()`
- `c->start()` /* logsize == 8096 --> deallocate  old ringbuffer --> allocate newringbuffer */


`clear_config()` and `load_config()` where the logsize has changed. But that's fine in this case we can just deallocate the current buffer and initialize a new one...

hallyn:
> 
> 
> 
> > @@ -196,6 +196,15 @@ static int lxc_console_cb_con(int fd, uint32_t events, void *data,
>  
>  		if (console->peer >= 0)
>  			w = lxc_write_nointr(console->peer, buf, r);
> +
> +		if (console->log_size > 0) {
> 
> My point being that the lock needed for an operation on the ringbuffer will always be taken by the monitor.

so, is the process_lock() needed at all then?  The monitor will
always be the same thread?

hallyn:
... well, if we are re-using the buffer, the user *may* want to still see output
from the last (perhaps failed) boot?  

dealloc on put sounds good.

brauner:
> so, is the process_lock() needed at all then?  The monitor will
> always be the same thread?

I think you're right that we don't need the lock. But let's see if my reasoning is sounds. Both the command socket handler and the console ringbuffer handler run in the same epoll loop in the same thread. Now we send a request `console_get_log()` which will cause a handler to be register via the command socket in the container's epoll loop. Now either the console ringbuffer handler runs or the `console_get_log()` handler runs but never both so they can't race. Does that make sense?

brauner:
@hallyn, added the realloc-dealloc logic to `lxc_setup_console_ringbuf()` called in `lxc_setup_parent()` and removed the `process_{un}lock()` calls. If you have no further objections this should be good to go.

src/lxc/ringbuf.c
```diff
@@ -0,0 +1,145 @@
...
+	char *tmp;
+	int ret;
+	int memfd = -1;
+
+	buf->size = size;
+	buf->r_off = 0;
+	buf->w_off = 0;
+
+	/* verify that we are at least given the multiple of a page size */
+	if (buf->size % getpagesize())
```
brauner: function `getpagesize()` is deprecated

brauner:
jenkins: test this please

PR #1871, opened 2017-10-22
API: add console ringbuffer extension

### LINKED ISSUE #1870: API: add console ringbuffer API extensions
Since we merged the console ringbuffer implementation we now need at least two new API calls. For example:

- `console_get_ringbuffer()`
- `console_clear_ringbuffer()`

Or maybe just one `console_ringbuffer()`?

brauner:
Closes #1870.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,7 @@ src/tests/lxc-test-cgpath
 src/tests/lxc-test-clonetest
 src/tests/lxc-test-concurrent
 src/tests/lxc-test-console
+src/tests/lxc-test-console-log
 src/tests/lxc-test-containertests
 src/tests/lxc-test-createtest
 src/tests/lxc-test-destroytest
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -92,6 +92,7 @@ static const char *lxc_cmd_str(lxc_cmd_t cmd)
 		[LXC_CMD_GET_LXCPATH]      = "get_lxcpath",
 		[LXC_CMD_ADD_STATE_CLIENT] = "add_state_client",
 		[LXC_CMD_SET_CONFIG_ITEM]  = "set_config_item",
+		[LXC_CMD_CONSOLE_LOG]      = "console_log",
 	};
 
 	if (cmd >= LXC_CMD_MAX)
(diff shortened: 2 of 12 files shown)


brauner:
I think we call the API function `console_log()` and give it a `struct console_log` as second argument. I'll push an RFC on top of this patchset soon.

brauner:
Ok, so I've changed the API extension quite a lot in a commit I pushed on top of this patchset. Please don't merge this yet until I a) removed the `lxc-console` patches and b) I squashed the commits once we all agree that the layout for the new API extension is fine. Here's the TL;DR:

- new function is:
`int (*console_log)(struct lxc_container *c, struct console_log *log)`
- new struct is added
```
struct console_log {
	/* Clear the console log. */
	bool clear;

	/* Retrieve the console log. */
	bool read;

	/* This specifies the maximum size to read from the ringbuffer. Setting
	 * it to 0 means that the a read can be as big as the whole ringbuffer.
	 * On return callers can check how many bytes were actually read.
	 * If "read" and "clear" are set to false and a non-zero value is
	 * specified then up to "read_max" bytes of data will be discarded from
	 * the ringbuffer.
	 */
	uint64_t *read_max;

	/* Data that was read from the ringbuffer. If "read_max" is 0 on return
	 * "data" will be NULL.
	 */
	char *data;
};
```

I think the struct makes sense since it will allows us to easily expand or deprecate operations on the ringbuffer.
Having the ringbuffer contents be returned in the struct enables us to return meaningful errors via the return value of `console_log()`. For example, callers can check for `-ENODATA` which I find quite useful. It also allows the API to do sensible logging outside of the actual command infrastructure. I prefer this very much since it keeps it apap (as performant as possible :)).

brauner:
See here for an illustration:

[asciicast](https://asciinema.org/a/143528)

hallyn:
(Hi - I'm assuming this has been on hold for Prague (given the last commit title) - please ping me when ready)

brauner:
@hallyn, actually it is ready for review. :)

stgraber:
@hallyn

hallyn:
Present.

(looking)

hallyn:
One thing about this API is that - if I don't miss something - it is easy to have a situation where

buffer has 200 bytes
user asks to read 100 bytes and clear the buffer
second 100 bytes are lost.

I'm probably misreading.

brauner:
If the user sets `clear = true` the ringbuffer gets wiped completely even if the user just requests to read 100 bytes. If the user only wants to discard 100 bytes after the read then `clear` should simply be left unset. What would you prefer?

hallyn:
Sorry, I'm not thinking about this quite right.  I need to look through and understand what happens with multiple concurrent readers.  Will comment after I grok that again.

hallyn:
...  so is there any point to ever calling ringbuf_clear()?  since doing a read advances the read pointer.

brauner:
If you want to clear the complete buffer without reading from it.

hallyn:
Ok.  I'm good with it - thanks.

Issue #1878, opened 2017-10-27
lxc-execute in 2.0.9 doesn't honor rootfs passed via command-line


jordemort:
# Required information

 * Distribution: Debian
 * Distribution version: jessie & stretch (backport of package from sid)
 * The output of
   * `lxc-start --version`: 2.0.9
   * `lxc-checkconfig`: all green and 👍 
   * `uname -a`:
      `Linux redacted 4.9.0-0.bpo.3-amd64 #1 SMP Debian 4.9.30-2+deb9u5~bpo8+1 (2017-09-28) x86_64 GNU/Linux`
   * `cat /proc/self/cgroup`
     ```
     10:pids:/
     9:perf_event:/
     8:net_cls,net_prio:/
     7:freezer:/
     6:devices:/
     5:memory:/
     4:blkio:/
     3:cpu,cpuacct:/
     2:cpuset:/
     1:name=systemd:/system.slice/ssh.service
     ```
   * `cat /proc/1/mounts`
      ```
      sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
      proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
      udev /dev devtmpfs rw,relatime,size=10240k,nr_inodes=15725876,mode=755 0 0
      devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000 0 0
      tmpfs /run tmpfs rw,nosuid,relatime,size=25166172k,mode=755 0 0
      /dev/dm-0 / ext4 rw,relatime,errors=remount-ro,data=ordered 0 0
      securityfs /sys/kernel/security securityfs rw,nosuid,nodev,noexec,relatime 0 0
      tmpfs /dev/shm tmpfs rw,nosuid,nodev 0 0
      tmpfs /run/lock tmpfs rw,nosuid,nodev,noexec,relatime,size=5120k 0 0
      tmpfs /sys/fs/cgroup tmpfs ro,nosuid,nodev,noexec,mode=755 0 0
      cgroup /sys/fs/cgroup/systemd cgroup rw,nosuid,nodev,noexec,relatime,xattr,release_agent=/lib/systemd/systemd-cgroups-agent,name=systemd 0 0
      pstore /sys/fs/pstore pstore rw,nosuid,nodev,noexec,relatime 0 0
[...]

PR #1879, opened 2017-10-27
Call lxc_config_define_load from lxc_execute again

jordemort:
At some point, `lxc-execute` lost its call to `lxc_config_define_load` - this means it accepted configuration parameters on the command-line via `-s` but didn't actually do anything with them. This puts the call back more-or-less where it was before.

Fixes https://github.com/lxc/lxc/issues/1878

### DIFF
--- a/src/lxc/tools/lxc_execute.c
+++ b/src/lxc/tools/lxc_execute.c
@@ -168,8 +168,11 @@ int main(int argc, char *argv[])
 		}
 	}
 
-	if (lxc_config_define_load(&defines, c->lxc_conf))
+	ret = lxc_config_define_load(&defines, c->lxc_conf);
+	if (ret) {
+		lxc_container_put(c);
 		exit(EXIT_FAILURE);
+	}
 
 	if (my_args.uid)
 		c->lxc_conf->init_uid = my_args.uid;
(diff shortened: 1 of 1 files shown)


jordemort:
Added `Signed-off-by: Jordan Webb <jordemort@github.com>` per CONTRIBUTING

src/lxc/tools/lxc_execute.c
```diff
@@ -168,6 +168,9 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	if (lxc_config_define_load(&defines, c->lxc_conf))
+		exit(EXIT_FAILURE);
```
brauner: This misses a `lxc_container_put(c);` :) So this should be:

```
ret = lxc_config_define_load(&defines, c->lxc_conf);
if (ret) {
        lxc_container_put(c);
        exit(EXIT_FAILURE);
}
```

jordemort:
@brauner How's that look?

brauner:
jenkins: test this please

brauner:
@jordemort, looks good, thanks. :)

jordemort:
Thanks for merging! I think this probably ought to be backported to stable-2.0 and stable-2.1 - we're applying it against 2.0.9. Would you like me to create PRs for that?

brauner:
@jordemort, yeah, you can do that. :)

PR #1883, opened 2017-10-29
attach: correctly handle namespace inheritance

brauner:
When attaching to a container's namespaces we did not handle the case where we
inherited namespaces correctly. In essence, liblxc on start records the
namespaces the container was created with in the handler. But it only records
the clone flags that were passed to clone() and doesn't record the namespaces
we e.g. inherited from other containers. This means that attach only ever
attached to the clone flags. But this is only correct if all other namespaces
not recorded in the handler refer to the namespaces of the caller. However,
this need not be the case if the container has inherited namespaces from
another container. To handle this case we need to check whether caller and
container are in the same namespace. If they are, we know that things are all
good. If they aren't then we need to attach to these namespaces as well.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -249,24 +249,89 @@ static void lxc_proc_put_context_info(struct lxc_proc_context_info *ctx)
 	free(ctx);
 }
 
+/**
+ * in_same_namespace - Check whether two processes are in the same namespace.
+ * @pid1 - PID of the first process.
+ * @pid2 - PID of the second process.
+ * @ns   - Name of the namespace to check. Must correspond to one of the names
+ *         for the namespaces as shown in /proc/<pid/ns/
+ *
+ * If the two processes are not in the same namespace returns an fd to the
+ * namespace of the second process identified by @pid2. If the two processes are
+ * in the same namespace returns -EINVAL, -1 if an error occurred.
+ */
+static int in_same_namespace(pid_t pid1, pid_t pid2, const char *ns)
+{
+	int ns_fd1 = -1, ns_fd2 = -1, ret = -1;
+	struct stat ns_st1, ns_st2;
+
+	ns_fd1 = lxc_preserve_ns(pid1, ns);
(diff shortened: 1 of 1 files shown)


src/lxc/attach.c
```diff
@@ -249,24 +249,89 @@ static void lxc_proc_put_context_info(struct lxc_proc_context_info *ctx)
...
+			if (fd[i] == -EINVAL) {
+				DEBUG("Inheriting %s namespace from %d",
+				      ns_info[i].proc_name, pid);
+				continue;
+			}
 		}
 
-		fd[i] = lxc_preserve_ns(pid, ns_info[i].proc_name);
+		if (fd[i] == -EINVAL)
+			fd[i] = lxc_preserve_ns(pid, ns_info[i].proc_name);
```
hallyn: This could be a pain to rework, but in the inheriting ns case you are opening this fd twice.  Not a huge deal in most cases, but when starting up thousands of containers quickly it could make a difference I suppose...

Not doing it would require something like passing back a pool for in_same_namespace return value, and leaving the open fd in an int pointer argument to the fn.

I don't know whether it's worth it.

hallyn: Well, attaching, not creating - which during initial container could happen many times I suppose.

brauner: `is_same_namespace()` returns an fd to `pid2`'s namespace in which case `fd[i]` will be `>= 0` so `lxc_preserve_ns()` doesn't actually open this twice. Or am I missing your point?

hallyn: In the case where in_same_namespace() returns -EINVAL.

brauner: But doesn't the `continue` in line `329` prevent that?

brauner: I'm probably being dense.

hallyn: d'oh.  I see.  EINVAL at that point means something different.  Thanks.

Issue #1881, opened 2017-10-28
tools: move tools to only use API functions


brauner:
This will be a lot of work actually. Let's use this issue to track some of that work and discuss a few things.

I've been looking into `lxc-start` today because it seems one of the most important candidates and immediately made me aware of one issue. No tool can dereference `struct lxc_conf` because it is an internal symbol. This is fine and I'd like to keep it that way! But this will mean we need to do some thinking. For `lxc-start` the main problem is namespace inheritance by specifying a PID or container name. To achieve this `lxc-start` dereferences `lxc_conf->inherit_ns_fd[<idx>]` which is - as said before - internal. To avoid this but also not introduce any additional API call or a new public struct member. We could simply introduce a new generically namespaced configuration key similar to `lxc.cgroup.*` and `lxc.prlimit.*`. The new key for namespace inheritance would be:

    lxc.inherit.namespace.<namespace-name>

It would currently recognize the following namespaces:

    lxc.inherit.namespace.cgroup
    lxc.inherit.namespace.ipc
    lxc.inherit.namespace.net
    lxc.inherit.namespace.mnt
    lxc.inherit.namespace.user
    lxc.inherit.namespace.uts
    lxc.inherit.namespace.pid

and can be set to either a PID or the name of a container. Additionally, I suggest we enable the key to take the path of a namespace fd as argument too (e.g. `/proc/<pid>/ns/<namespace>`. The namespace names for the new config key itself simply correspond to `/proc/<pid>/ns/<namespace>`. The double namespacing of the key `lxc.inherit.namespace` would be done to allow for the possibility of more inheritance rules. I think this might add some value for API users like `LXD` as well since `LXD` could now specify namespace inheritance between containers. This e.g. could be very useful when e.g. having multiple containers use the same `VF` of an `SR-IOV` enabled network device. @stgraber, @hallyn, opinions?

brauner:
Actually, dropping the `inherit` namespace might make more sense:

    lxc.namespace.<namespace-name>

and would be more in line with `lxc.{cgroup,prlimig}.*`.

brauner:
Hm, actually I'm seeing a problem with this approach. If you can specify this in the on-disk configuration file you'd run into problems everytime you try an operation on a container sharing a namespace with a container that has been stopped. If it is stopped then the parsing of the config file on `lxc_container_new()` would fail. I don't want to special case this in the API so I might go with an `inherit_ns_fd` struct member after all if I can't think of a nicer way.

[...]

PR #1884, opened 2017-10-29
confile: add lxc.namespace.<namespace-key> + add user namespace sharing + rework start logic

brauner:
Relates to #1881.

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -835,7 +835,7 @@ static int lxc_cmd_get_name_callback(int fd, struct lxc_cmd_req *req,
 
 	memset(&rsp, 0, sizeof(rsp));
 
-	rsp.data = handler->name;
+	rsp.data = (char *)handler->name;
 	rsp.datalen = strlen(handler->name) + 1;
 	rsp.ret = 0;
 
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -159,6 +159,8 @@ static bool preserve_ns(int ns_fd[LXC_NS_MAX], int clone_flags, pid_t pid)
 		ns_fd[i] = lxc_preserve_ns(pid, ns_info[i].proc_name);
 		if (ns_fd[i] < 0)
 			goto error;
+
+		DEBUG("Preserved %s namespace via fd %d", ns_info[i].proc_name, ns_fd[i]);
 	}
 
 	return true;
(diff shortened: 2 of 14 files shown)


src/lxc/start.c
```diff
@@ -729,11 +722,18 @@ void lxc_fini(const char *name, struct lxc_handler *handler)
...
 	while (namespace_count--)
 		free(namespaces[namespace_count]);
+
 	for (i = 0; i < LXC_NS_MAX; i++) {
 		if (handler->nsfd[i] != -1) {
 			close(handler->nsfd[i]);
 			handler->nsfd[i] = -1;
 		}
+
+		if (handler->conf->inherit_ns[i].fd < 0) {
```
brauner: wrong check

src/lxc/commands.c
```diff
@@ -835,7 +835,7 @@ static int lxc_cmd_get_name_callback(int fd, struct lxc_cmd_req *req,
 
 	memset(&rsp, 0, sizeof(rsp));
 
-	rsp.data = handler->name;
+	rsp.data = (char *)handler->name;
```
hallyn: How about making rsp.data const void * (as send(2) shows for its data type)?

brauner: Yeah, I guess that makes sense.

brauner: Actually no. That won't work since `rsp.data` is also passed to `recv()`.

src/lxc/start.c
```diff
@@ -561,11 +558,7 @@ struct lxc_handler *lxc_init_handler(const char *name, struct lxc_conf *conf,
 	for (i = 0; i < LXC_NS_MAX; i++)
 		handler->nsfd[i] = -1;
 
-	handler->name = strdup(name);
-	if (!handler->name) {
-		ERROR("failed to allocate memory");
-		goto on_error;
-	}
+	handler->name = name;
```
hallyn: It makes me uneasy, since the container* is shared among threads in a way the handler isn't, and the c->name was previously dereferenced under container_mem_lock().  I *think* it's fine - or rather I think if it isn't we have other problems - but I'm not convinced.  We can try it and just keep this in mind if weird failures end up happening.

hallyn:
Is there a testcase for sharing userns?

stgraber:
@brauner where is the doc for this?

PR #1891, opened 2017-10-31
cgroups/cgfsng: fixes, features, and improved cgroup2 handling

brauner:

### DIFF
--- a/config/templates/archlinux.common.conf.in
+++ b/config/templates/archlinux.common.conf.in
@@ -6,7 +6,6 @@ lxc.tty.max = 6
 
 # Set the halt/stop signals
 lxc.signal.halt=SIGRTMIN+4
-lxc.signal.stop=SIGRTMIN+14
 
 # Uncomment to disable creating tty devices subdirectory in /dev
 # lxc.tty.dir =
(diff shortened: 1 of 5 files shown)


brauner:
Fwiw, the failure is unrelated to any interesting changes. The current failure is caused by the new logging functions I implemented for the constructor. They print to stderr so when `liblxc` `exec()`s `lxc-user-nic` and then reads back from `std{out,err}` to capture the output that is sent to it it'll read back the contents of the log as well which is why `lxc-user-nic` fails which is is why the container fails to start. This only affects unprivileged containers started by unprivileged users they are the only containers that need to use `lxc-user-nic`. Note, the two tests that fail need to run as root but they drop to an unprivileged users before they do anything interesting with containers.

brauner:
Note that for unprivileged containers that - for whatever strange reason - still drop `CAP_SYS_ADMIN` in their own user namespace we only ever mount hierarchies that are writable by she container. I think this is more secure and correct behavior because it means that any hierarchy in which the container is not placed into its own writable cgroup will not be visible to it and the container also won't be able to mount it by itself. This prevents quite some information leaks.

brauner:
jenkins: test this please

hallyn:
Looks good, though the last patch seems misplaced.

Issue #1623, opened 2017-06-13
d5cb35d breaks systemd's PrivateDevices on CentOS 7


major:
# Required information

 * Distribution: CentOS
 * Distribution version: 7.3.1611
 * The output of
   * `lxc-start --version`: `2.0.8`
   * `lxc-checkconfig`:

```
# lxc-checkconfig 
Kernel configuration not found at /proc/config.gz; searching...
Kernel configuration found at /boot/config-3.10.0-514.21.1.el7.x86_64
--- Namespaces ---
Namespaces: enabled
Utsname namespace: enabled
Ipc namespace: enabled
Pid namespace: enabled
User namespace: enabled
newuidmap is not installed
newgidmap is not installed
Network namespace: enabled
Multiple /dev/pts instances: enabled

--- Control groups ---
Cgroup: enabled
Cgroup clone_children flag: enabled
Cgroup device: enabled
Cgroup sched: enabled
Cgroup cpu account: enabled
Cgroup memory controller: enabled
Cgroup cpuset: enabled

--- Misc ---
Veth pair device: enabled
Macvlan: enabled
Vlan: enabled
Bridges: enabled
Advanced netfilter: enabled
CONFIG_NF_NAT_IPV4: enabled
CONFIG_NF_NAT_IPV6: enabled
CONFIG_IP_NF_TARGET_MASQUERADE: enabled
CONFIG_IP6_NF_TARGET_MASQUERADE: enabled
CONFIG_NETFILTER_XT_TARGET_CHECKSUM: enabled
FUSE (for use with lxcfs): enabled

--- Checkpoint/Restore ---
checkpoint restore: enabled
CONFIG_FHANDLE: enabled
CONFIG_EVENTFD: enabled
CONFIG_EPOLL: enabled
CONFIG_UNIX_DIAG: enabled
CONFIG_INET_DIAG: enabled
CONFIG_PACKET_DIAG: enabled
CONFIG_NETLINK_DIAG: enabled
File capabilities: enabled

Note : Before booting a new kernel, you can check its configuration
usage : CONFIG=/path/to/config /usr/bin/lxc-checkconfig
```

   * `uname -a`: `Linux lxc-test 3.10.0-514.21.1.el7.x86_64 #1 SMP Thu May 25 17:04:51 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
   * `cat /proc/self/cgroup`:

```
# cat /proc/self/cgroup 
11:pids:/
[...]

PR #1892, opened 2017-10-31
mknod /dev/ptmx instead of bind mounting it

jordemort:
This compensates for a problem in versions of systemd older than 219, which causes units with "PrivateDevices=yes" to fail to start with a "226/NAMESPACE" error.

Fixes https://github.com/lxc/lxc/issues/1623

It's my understanding that the original issue was left unfixed because systemd < 219 is relatively ancient, in terms of systemd versions, and that it had been fixed upstream already (https://github.com/systemd/systemd/pull/3254) - however, we have existing container setups based on Debian Jessie, which ships systemd 215, and this is preventing us from migrating to a newer version of LXC.

It looked like the original approach with a symlink was unacceptable because it confused apparmor, wheras the bind mount approach breaks older versions of systemd that didn't handle private /dev correctly. I've replaced the bind mount with a call to `mknod`; in a nutshell, this patch calls `stat` on `/dev/pts/ptmx` and then creates a separate-but-identical device node at '/dev/ptmx`. This works for us; I'm hoping you'll find this somewhat compromised approach acceptable in the name of keeping Jessie-based containers working.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1446,10 +1446,7 @@ static int lxc_setup_devpts(int num_pts)
 
 	/* Create doppelganger of /dev/pts/ptmx at /dev/ptmx */
 	ret = mknod("/dev/ptmx", ptmx.st_mode, ptmx.st_rdev);
-	if (ret) {
-		SYSERROR("failed to mknod \"/dev/ptmx\"");
-		/* Fall through to symlink code below */
-	} else {
+	if (!ret) {
 		/* Set ownership and mode on doppelganger */
 		ret = chown("/dev/ptmx", ptmx.st_uid, ptmx.st_gid);
 		if (ret)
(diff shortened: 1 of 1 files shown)


jordemort:
Note that Jessie also ships with units that contain `PrivateDevices=yes` (`systemd-hostnamed` is what ended up biting us here, specifically) so it's not simply a matter of "just don't do that there" - this problem breaks the default install of the distribution.

stgraber:
Hmm, won't this break nesting and fully unprivileged containers where mknod isn't allowed?

jordemort:
@stgraber I'm not sure about nesting but yeah it would definitely break if mknod is not allowed. The symlink fallback code is still there; I could maybe put the bind mount stuff back as an additional fallback to the mknod?

jordemort:
I've added the bind mount back as a fallback. So now it:
  1. Tries mknod
  2. Tries bind mount
  3. Tries symlink

I also restructured things a little bit in the mknod stuff to make things more consistent with what was already there, and updated a couple comments.

brauner:
You should use am_root in the handler and if you are create a second console device otherwise execute the bind code and then the symlink code. Sorry, on my phone so not really detailed. :)

src/lxc/conf.c
```diff
@@ -1433,6 +1434,40 @@ static int lxc_setup_devpts(int num_pts)
 		DEBUG("removed existing \"/dev/ptmx\"");
 	}
 
+	/* Read attributes of /dev/pts/ptmx */
+	ret = stat("/dev/pts/ptmx", &ptmx);
+	if (ret) {
```
brauner: Please make that `(if ret < 0)`

src/lxc/conf.c
```diff
@@ -1433,6 +1434,40 @@ static int lxc_setup_devpts(int num_pts)
...
+		SYSERROR("failed to stat \"/dev/pts/ptmx\"");
+		return -1;
+	} else if (!S_ISCHR(ptmx.st_mode)) {
```
brauner: Since you return in the  `if` branch above the `else if` is not really needed. Avoid the indentation and just perform the check unconditionally. Meaning:

```
if (/* char dev check fails */)
       return error;
```

src/lxc/conf.c
```diff
@@ -1433,6 +1434,40 @@ static int lxc_setup_devpts(int num_pts)
...
+		SYSERROR("\"/dev/pts/ptmx\" isn't a character device!");
+		return -1;
+	}
+
+	/* Create doppelganger of /dev/pts/ptmx at /dev/ptmx */
+	ret = mknod("/dev/ptmx", ptmx.st_mode, ptmx.st_rdev);
+	if (!ret) {
```
brauner: The handler records - on startup - whether you are root or not. So you should either give this fun an additional arg to inform it whether it is root or not or pass down the handler itself and. Then you can check here and only perform the `mknod()` syscall when you are root. This saves a context switch.

src/lxc/conf.c
```diff
@@ -1433,6 +1434,40 @@ static int lxc_setup_devpts(int num_pts)
...
+		ret = chmod("/dev/ptmx", ptmx.st_mode);
+		if (ret)
+		{
+			SYSERROR("failed to chmod \"/dev/ptmx\"");
+			return -1;
+		}
+		/* Our work is done here */
+		DEBUG("made copy of \"/dev/pts/ptmx\" at \"/dev/ptmx\"");
+		return 0;
+	} else {
```
brauner: Again, the else is not needed since since you are returning anyway.

jordemort:
Thanks for the feedback, @brauner! I'm not sure if I'll have time to get back to it today but I'll be working on this this week.

brauner:
@jordemort, np. Thanks for the patch! :)

brauner:
Haven't heard back in  while. Please re-open a new PR if still something you need.

PR #1896, opened 2017-11-02
Fix compilation on toolchain without prlimit

ffontaine:
Some toolchains which are not bionic like uclibc does not support
prlimit or prlimit64. In this case, return an error.


brauner:
Hm, including our own implementation is not sufficient here?

brauner:
Because we have one because of bionic.

ffontaine:
But I could update the patch to include a check on prlimit64 in configure.ac. If this function is available and not prlimit, try your implementation otherwise return an error.

brauner:
Sounds good.

brauner:
jenkins: test this please

brauner:
Looks like Android is happy as well. Merging.

PR #1901, opened 2017-11-06
lxc/start.c: Fix legacy PR_{G,S}ET_NO_NEW_PRIVS handling

jacmet:
The configure checks for these use AC_CHECK_DECLS, which define the symbol
to 0 if not available - So adjust the code to match.  From the autoconf
manual:

https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Generic-Declarations.html)

For each of the symbols (comma-separated list), define HAVE_DECL_symbol (in
all capitals) to ‘1’ if symbol is declared, otherwise to ‘0’.

Signed-off-by: Peter Korsgaard <peter@korsgaard.com>

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -52,15 +52,15 @@
 #include <sys/capability.h>
 #endif
 
-#ifndef HAVE_DECL_PR_CAPBSET_DROP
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

PR #1908, opened 2017-11-08
lxc-test-unpriv: fix the overlayfs mount error

Cypresslin:
This patch fixes the missing workdir issue for the overlayfs mount command in
the lxc-test-unpriv test.

Bug link: https://bugs.launchpad.net/bugs/1730915

### DIFF
--- a/src/tests/lxc-test-unpriv
+++ b/src/tests/lxc-test-unpriv
@@ -39,8 +39,8 @@ if modprobe -q overlayfs; then
 
         mount -t tmpfs none ${MOUNTDIR}
 
-        mkdir "${MOUNTDIR}/lowerdir" "${MOUNTDIR}/upperdir" "${MOUNTDIR}/overlayfs"
-        mount -t overlayfs -o lowerdir="${MOUNTDIR}/lowerdir",upperdir="${MOUNTDIR}/upperdir" none "${MOUNTDIR}/overlayfs"
+        mkdir ${MOUNTDIR}/{lowerdir,upperdir,workdir,overlayfs}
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

src/tests/lxc-test-unpriv
```diff
@@ -39,8 +39,8 @@ if modprobe -q overlayfs; then
 
         mount -t tmpfs none ${MOUNTDIR}
 
-        mkdir "${MOUNTDIR}/lowerdir" "${MOUNTDIR}/upperdir" "${MOUNTDIR}/overlayfs"
-        mount -t overlayfs -o lowerdir="${MOUNTDIR}/lowerdir",upperdir="${MOUNTDIR}/upperdir" none "${MOUNTDIR}/overlayfs"
+        mkdir ${MOUNTDIR}/{lowerdir,upperdir,workdir,overlayfs}
+        mount -t overlayfs -o lowerdir="${MOUNTDIR}/lowerdir",upperdir="${MOUNTDIR}/upperdir",workdir="${MOUNTDIR}/workdir" none "${MOUNTDIR}/overlayfs"
```
brauner: Thanks for the patch. One nit though. What happens on older kernels whose overlay module does not support the workdir option? Should the logic be:
- try with workdir option first and if that fails try without the workdir option?

stgraber: Yeah, looks like this is going to break older kernels, just a reminder that LXC 2.0.x still supports kernels from 2.6.32 onwards so this does need fixing as the ultimate target for this fix is 2.0.x

Cypresslin: Hi, thanks for catching this. 

To this test itself, the script will skip this overlayfs test part on older kernels (tested on Trusty 3.13, which needs the overlayfs module) with an if statement:

    if modprobe -q overlayfs; then

So I think as long as this module check criteria is valid, this change should be fine?

brauner: Ok, cool.

PR #1912, opened 2017-11-09
commands: clear console log file

brauner:
When a "clear" request is sent to the console ringbuffer we should truncate the
console log file as well.

### DIFF
--- a/src/lxc/console.c
+++ b/src/lxc/console.c
@@ -660,6 +660,10 @@ int lxc_console_create(struct lxc_conf *conf)
 		goto err;
 	}
 
+	/* Only open the console log file in O_APPEND if no console ringbuffer
+	 * has been requested. Otherwise the log file will be opened when the
+	 * a request is sent to dump the contents of the ringbuffer to disk.
+	 */
 	if (console->log_path && console->log_size <= 0) {
 		console->log_fd = lxc_unpriv(open(console->log_path, O_CLOEXEC | O_RDWR | O_CREAT | O_APPEND, 0600));
 		if (console->log_fd < 0) {
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -23,6 +23,7 @@
 
 #include "config.h"
 
+#include <caps.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <malloc.h>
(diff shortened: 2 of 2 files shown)


brauner:
jenkins: test this please

brauner:
jenkins: test this please

hallyn:
How is the log file intended to be used?

I would expect something like:

close(log file);
rename(logfile, logfile.N);
open(log file);

to happen on clear instead.  Does that make any sense at all?

brauner:
Hm, wouldn't that lead to a similar problem in the long run that we'd accumulate log files and fill the disk?
Also, I'd think that if a user requests a `clear` they thereby signal that they don't care about the content of the ringbuffer anymore so there's probably not a good reason to keep it around. But I'm open to your suggestion if you think it's better.

src/lxc/console.c
```diff
@@ -660,6 +660,10 @@ int lxc_console_create(struct lxc_conf *conf)
 		goto err;
 	}
 
+	/* Only open the console log file in O_APPEND if no console ringbuffer
+	 * has been requested. Otherwise the log file will be opened when the
```
hallyn: "the a request"

hallyn:
So long as we are rotating the log files and allow demanding a new log file, user space can safely remove old files (after backing them up or not) if disk starts getting full.  The real problem is when there's only a single log file and user space can't truncate it because lxc is still writing to it.

I'm picturing a set of remote containers, where I read the log file through cmd and then say 'clear' because I don't want to see the rest through the command interface; but I still want the persistent logs on disk to go back through later.

We seem to have different ideas about how this would be used - let's see what @stgraber thinks.

hallyn:
ping @stgraber

stgraber:
Right, there are a few different use cases here I think.

For LXD, we don't want to make it possible for a container user to fill the disk, so in this case I'd want a dump of the ringbuffer to disk on exit, therefore limiting the amount of potential disk usage to the size of the ringbuffer.

In a more traditional LXC case, I definitely can see @hallyn's use case of wanting to retain console history by having LXC rotate the console when it gets cleared rather than just truncate it.


Looking at config, right now we have:

 - lxc.console.logfile (path to text file to dump console to)
 - lxc.console.path (path to use as the /dev/console device)
 - lxc.console.logsize (size of ring buffer)

I suspect we may want this to become:

 - lxc.console.buffer.size (size of ring buffer)
 - lxc.console.buffer.logfile (where to dump the ringbuffer to disk)

 - lxc.console.logfile (path to text file to dump console to)
 - lxc.console.path (path to use as the /dev/console device)
 - lxc.console.rotate (boolean to control rotation on clear, if false, nothing happens and we keep appending)

brauner:
Closing in favor of #1925 .

PR #1917, opened 2017-11-10
console: report detach message on demand

brauner:
When users pass -1 there's there won't be an escape sequence to exit the
console so no need to print a misleading info message about how to detach.


stgraber:
Cool, I was wondering about that when I looked at the previous one, we'll want to backport this to stable branches.

PR #1918, opened 2017-11-10
lxccontainer: enable daemonized execute

brauner:
This enables daemonized execute containers with our minimal init running as
pid one and the requested program running as second pid.


brauner:
@exxo.

PR #1919, opened 2017-11-11
console: use correct escape sequence check

brauner:


hallyn:
Going by the comment in console.h, only -1 should technically be ignored, though I don't know what 0 would really mean.

PR #1929, opened 2017-11-19
commands: fix state socket implementation

brauner:
The old state socket implementation used the client fd for the currently
running command socket itself but did so in a bogus way such that the actual
client could be accidently handled in the mainloop itself. This is faulty and
problematic. For example, consider the case where the container doesn't respond
to a shutdown request. Now the process requesting the shutdown exists. If we
now subsequently kill the container with a stop command the monitor would write
the STOPPED message into the client socket but the original process has exited
so the mainloop handles the request and would thus receive nonsense:

```
root@conventiont|~
> lxc-start -n a1
root@conventiont|~
> lxc-stop -n a1
^C
root@conventiont|~
> lxc-stop -n a1 -k
lxc-stop: a1: commands.c: lxc_cmd_rsp_recv: 165 File too large - Response data for command "stop" is too long: 12641 bytes > 8192
```

This commit uses a sockepair() for each client.

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -231,7 +231,7 @@ static int lxc_cmd_send(const char *name, struct lxc_cmd_rr *cmd,
 	int client_fd;
 	ssize_t ret = -1;
 
-	client_fd = lxc_cmd_connect(name, lxcpath, hashed_sock_name);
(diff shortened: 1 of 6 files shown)


brauner:
Closing in favor of #1931.

PR #1933, opened 2017-11-21
lxc_init: set the control terminal in the child session

3XX0:
Additionally, the console seems to be [execute only (0111)](https://github.com/lxc/lxc/blob/master/src/lxc/conf.c#L1536) not sure if it's intended or not.


brauner:
jenkins: test this please

brauner:
> Additionally, the console seems to be execute only (0111) not sure if it's intended or not.

It doesn't matter in this case since this file is only used as a mount point not by itself.

PR #1934, opened 2017-11-21
commands: improve and simplify locking + lxccontainer: add reboot2() API extension

brauner:
lxccontainer: add reboot2() API extension
This adds reboot2() as a new API extension. This function properly wait until a
reboot succeeded. It takes a timeout argument. When set to > 0 reboot2() will
block until the timeout is reached, if timeout is set to zero reboot2() will
not block, if set to -1 reboot2() will block indefinitly.

The struct state_client gets rename to lxc_state_client since it's more in line
with other declarations. It also gets moved from the lxc_handler to the
lxc_conf struct so that the state clients waiting for reboots don't get
deallocated on reboot since the handler is deallocated on reboot.

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -932,9 +932,7 @@ int lxc_cmd_add_state_client(const char *name, const char *lxcpath,
 		return state;
 	}
 
-	process_lock();
 	ret = lxc_cmd(name, &cmd, &stopped, lxcpath, NULL);
-	process_unlock();
 	if (ret < 0) {
 		ERROR("%s - Failed to execute command", strerror(errno));
 		return -1;
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -904,44 +904,35 @@ int lxc_cmd_add_state_client(const char *name, const char *lxcpath,
 	    },
 	};
 
-	/* Lock the whole lxc_cmd_add_state_client_callback() call to ensure
-	*  that lxc_set_state() doesn't cause us to miss a state.
-	 */
-	process_lock();
 	/* Check if already in requested state. */
 	state = lxc_getstate(name, lxcpath);
 	if (state < 0) {
-		process_unlock();
 		TRACE("%s - Failed to retrieve state of container", strerror(errno));
 		return -1;
 	} else if (states[state]) {
-		process_unlock();
 		TRACE("Container is %s state", lxc_state2str(state));
 		return state;
 	}
 
 	if ((state == STARTING) && !states[RUNNING] && !states[STOPPING] && !states[STOPPED]) {
-		process_unlock();
 		TRACE("Container is in %s state and caller requested to be "
 		      "informed about a previous state", lxc_state2str(state));
 		return state;
 	} else if ((state == RUNNING) && !states[STOPPING] && !states[STOPPED]) {
-		process_unlock();
 		TRACE("Container is in %s state and caller requested to be "
 		      "informed about a previous state", lxc_state2str(state));
 		return state;
 	} else if ((state == STOPPING) && !states[STOPPED]) {
-		process_unlock();
 		TRACE("Container is in %s state and caller requested to be "
 		      "informed about a previous state", lxc_state2str(state));
 		return state;
 	} else if ((state == STOPPED) || (state == ABORTING)) {
-		process_unlock();
 		TRACE("Container is in %s state and caller requested to be "
 		      "informed about a previous state", lxc_state2str(state));
 		return state;
 	}
 
+	process_lock();
 	ret = lxc_cmd(name, &cmd, &stopped, lxcpath, NULL);
 	process_unlock();
 	if (ret < 0) {
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -128,6 +128,9 @@ static int lxc_cmd_rsp_recv(int sock, struct lxc_cmd_rr *cmd)
 	if (ret < 0) {
 		WARN("%s - Failed to receive response for command \"%s\"",
 		     strerror(errno), lxc_cmd_str(cmd->req.cmd));
+		if (errno == ECONNRESET)
+			return -ECONNRESET;
+
 		return -1;
 	}
 	TRACE("Command \"%s\" received response", lxc_cmd_str(cmd->req.cmd));
@@ -318,6 +321,8 @@ static int lxc_cmd(const char *name, struct lxc_cmd_rr *cmd, int *stopped,
 	}
 
 	ret = lxc_cmd_rsp_recv(client_fd, cmd);
+	if (ret == -ECONNRESET)
+		*stopped = 1;
 out:
 	if (!stay_connected || ret <= 0)
 		close(client_fd);
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -1176,6 +1176,10 @@ static void lxc_cmd_fd_cleanup(int fd, struct lxc_handler *handler,
 		lxc_list_del(cur);
 		free(cur->elem);
 		free(cur);
+		/* No need to walk the whole list. If we found the state client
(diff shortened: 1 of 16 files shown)


brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

src/lxc/commands.c
```diff
@@ -904,44 +904,35 @@ int lxc_cmd_add_state_client(const char *name, const char *lxcpath,
 	    },
 	};
 
-	/* Lock the whole lxc_cmd_add_state_client_callback() call to ensure
-	*  that lxc_set_state() doesn't cause us to miss a state.
-	 */
-	process_lock();
 	/* Check if already in requested state. */
 	state = lxc_getstate(name, lxcpath);
```
hallyn: ...  but the state could change between lxc_getstate and later checks, no?

src/lxc/start.c
```diff
@@ -341,14 +341,10 @@ static int lxc_serve_state_clients(const char *name,
 	struct state_client *client;
 	struct lxc_msg msg = {.type = lxc_msg_state, .value = state};
 
-	process_lock();
-
-	/* Only set state under process lock held so that we don't cause
-	*  lxc_cmd_add_state_client() to miss a state.
```
hallyn: TBH given the comment here I don't understand why the process_unlock happens before lxc_monitor_send_state().  It seems like the set_state and send_state should happen atomically together.

src/lxc/commands.c
```diff
@@ -914,24 +914,6 @@ int lxc_cmd_add_state_client(const char *name, const char *lxcpath,
 		return state;
 	}
 
-	if ((state == STARTING) && !states[RUNNING] && !states[STOPPING] && !states[STOPPED]) {
```
hallyn: ...  I'm not quite sure what this code was originally meant to do?

src/lxc/start.c
```diff
@@ -222,6 +225,20 @@ int lxc_check_inherited(struct lxc_conf *conf, bool closeall,
 		    (i < len_fds && fd == fds_to_ignore[i]))
 			continue;
 
+		/* Keep state clients that wait on reboots. */
+		lxc_list_for_each(cur, &conf->state_clients) {
+			struct lxc_state_client *client = cur->elem;
```
hallyn: Seems like this should be a separate helper?

src/lxc/conf.h
```diff
@@ -248,6 +248,11 @@ enum lxchooks {
 
 extern char *lxchook_names[NUM_LXC_HOOKS];
 
+struct lxc_state_client {
```
hallyn: Can you explain in the commit message for future readers why the lxc_state_client gets moved from handler to conf?

brauner:
On Fri, Nov 24, 2017 at 02:40:21AM +0000, Serge Hallyn wrote:
> hallyn commented on this pull request.
> 
> 
> 
> > @@ -904,44 +904,35 @@ int lxc_cmd_add_state_client(const char *name, const char *lxcpath,
>  	    },
>  	};
>  
> -	/* Lock the whole lxc_cmd_add_state_client_callback() call to ensure
> -	*  that lxc_set_state() doesn't cause us to miss a state.
> -	 */
> -	process_lock();
>  	/* Check if already in requested state. */
>  	state = lxc_getstate(name, lxcpath);
> 
> ...  but the state could change between lxc_getstate and later checks, no?

I think the right thing to do is to remove this check altogether. Instead, we
check the state by looking at handler->state directly when we have taken the
lock on the state client list in lxc_add_state_client(), i.e. in the callback.

> 
> > @@ -341,14 +341,10 @@ static int lxc_serve_state_clients(const char *name,
>  	struct state_client *client;
>  	struct lxc_msg msg = {.type = lxc_msg_state, .value = state};
>  
> -	process_lock();
> -
> -	/* Only set state under process lock held so that we don't cause
> -	*  lxc_cmd_add_state_client() to miss a state.
> 
> TBH given the comment here I don't understand why the process_unlock happens before lxc_monitor_send_state().  It seems like the set_state and send_state should happen atomically together.

This is the legacy lxc-monitord binary which doesn't use the command socket at
all.

> 
> > @@ -914,24 +914,6 @@ int lxc_cmd_add_state_client(const char *name, const char *lxcpath,
>  		return state;
>  	}
>  
> -	if ((state == STARTING) && !states[RUNNING] && !states[STOPPING] && !states[STOPPED]) {
> 
> ...  I'm not quite sure what this code was originally meant to do?
> 
> > @@ -222,6 +225,20 @@ int lxc_check_inherited(struct lxc_conf *conf, bool closeall,
>  		    (i < len_fds && fd == fds_to_ignore[i]))
>  			continue;
>  
> +		/* Keep state clients that wait on reboots. */
> +		lxc_list_for_each(cur, &conf->state_clients) {
> +			struct lxc_state_client *client = cur->elem;
> 
> Seems like this should be a separate helper?

I'm not sure this licenses a separate helper.

src/lxc/commands.c
brauner: I think the right thing to do is to remove this check altogether. Instead, we
check the state by looking at handler->state directly when we have taken the
lock on the state client list in lxc_add_state_client(), i.e. in the callback.

src/lxc/conf.h
brauner: Yeah, good idea.

src/lxc/start.c
brauner: This is the legacy lxc-monitord binary which doesn't use the command socket at
all.

src/lxc/start.c
brauner: I'm not sure this licenses a separate helper.

brauner:
@hallyn updated.

src/lxc/lxccontainer.c
```diff
@@ -1771,24 +1771,52 @@ static bool lxcapi_create(struct lxc_container *c, const char *t,
...
+		state = lxc_cmd_sock_rcv_state(sock, 0);
+		close(sock);
+		if (state != RUNNING)
+			return false;
+		TRACE("Container rebooted");
+		retv = true;
+	} else {
+		TRACE("Received state \"%s\" instead of expected \"RUNNING\"",
+		      lxc_state2str(ret));
+	}
```
stgraber: Hmm, so am I missing something or is this turning reboot() from a non-blocking API call to a blocking API call?

Unlike shutdown() there is no timeout argument here for the caller to indicate that they just want the signal sent but not wait for something to actually happen.

brauner: So probably better to add a `reboot2()` that does both jobs. API-wise it'll be pretty nice to finally having a blocking reboot call...

stgraber: Yeah, I'd be fine with keeping reboot() be just signal sending and reboot2() be signal sending + wait with a timeout.

src/lxc/conf.h
brauner: done

brauner:
Ready for another review //cc @hallyn.

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins seems to be messed up

brauner:
jenkins: test this please

brauner:
jenkins: test this please

hallyn:
Just browsed through the .diff again, lgtm.

brauner:
@hallyn, this should be good to go. :)

brauner:
@stgraber, since @hallyn basically acked this we should land this.

PR #1936, opened 2017-11-22
lxc-test-unpriv: check user existence before removing it

Cypresslin:
Check the test user (lxcunpriv) before calling deluser command,
otherwise it will print unnecessary error message:
/usr/sbin/deluser: The user 'lxcunpriv' does not exist.


brauner:
jenkins: test this please

PR #1940, opened 2017-11-23
lxccontainer.c: Use calloc instead of malloc/memset

marcosps:

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -4408,11 +4408,10 @@ static int do_lxcapi_migrate(struct lxc_container *c, unsigned int cmd,
 	 * to initialize.
 	 */
 	if (size < sizeof(*opts)) {
-		valid_opts = malloc(sizeof(*opts));
+		valid_opts = calloc(1, sizeof(*opts));
 		if (!valid_opts)
 			return -ENOMEM;
 
-		memset(valid_opts, 0, sizeof(*opts));
 		memcpy(valid_opts, opts, size);
 	}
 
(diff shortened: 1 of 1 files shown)


stgraber:
Can you check that calloc also exists in bionic, uclibc and musl?

marcosps:
@stgraber per my quick search, yes:

https://github.com/kraj/uClibc/blob/master/libc/stdlib/malloc/calloc.c
https://android.googlesource.com/platform/bionic/+/master/linker/linker_memory.cpp#74
https://github.com/rofl0r/musl/blob/master/src/malloc/calloc.c

Does lxc have some setup to test code compilation with different environments with such c libraries?

stgraber:
We have tests for bionic, not for the others

brauner:
We usually stay away from `calloc()` since it's interface is a little weird and it actually only provides performance benefits if you allocate quite a bit of memory usually at least around 1 MB or something. In this case `calloc()` employs `mmap()` to retrieve zeroed pages and also *can* reuse a single page as long as some memory is not touced. But for small allocations there's really not a difference. Even if `calloc()` is smart here and somehow avoids `memset()` work the `criu` codepath here needs to make very sure that the whole struct is set to 0 in case callers pass in a smaller struct. Furthermore the comments here match very well with the explicit `memset()`. So I'd prefer to nack this actually.

marcosps:
@brauner ok, no problem. I'll try other places to contribute :)

hallyn:
> @brauner ok, no problem. I'll try other places to contribute :)

Hi,

I hope you mean other places in the lxc code :)

marcosps:
@hallyn sure, sorry for the dubious phrase. I liked the lxc idea, so I'm trying to learn more and to contribute with code and tests :)

PR #1941, opened 2017-11-23
Improve lxc-oci.in

flx42:
Remaining improvements (will be separate PRs):
- Handle uid/gid (`USER` in a Dockerfile), this value can be non-numeric so it might require some discussions.
- Handle layer caching.


brauner:
jenkins: test this please

PR #1945, opened 2017-11-24
start.c: always handing uid and gid

tanyifeng:
Currently, when we configured `lxc.init_cmd=/bin/bash` and uid, gid which is not 0, running `lxc-start` will not execute `/bin/bash` with specified uid, gid, because we  ignore handing uid, gid in [do_start](https://github.com/lxc/lxc/blob/dc3de872519390f5c359586b9f07defb262bcf64/src/lxc/start.c#L1012).

Maybe lxc do not allow us to start a container without root permission since its default cmd is `/sbin/init`, I think remove this constraint is better for user to configure container.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -776,6 +776,9 @@ static int do_start(void *data)
 	char path[PATH_MAX];
 	int devnull_fd = -1;
 	struct lxc_handler *handler = data;
+	bool have_cap_setgid;
+	uid_t new_uid;
+	gid_t new_gid;
 
 	if (sigprocmask(SIG_SETMASK, &handler->oldmask, NULL)) {
 		SYSERROR("Failed to set signal mask.");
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

src/lxc/start.c
```diff
@@ -1009,30 +1009,28 @@ static int do_start(void *data)
...
-		#endif
-		if (lxc_list_empty(&handler->conf->id_map) && have_cap_setgid) {
-			if (lxc_setgroups(0, NULL) < 0)
-				goto out_warn_father;
-		}
-
-		if (lxc_switch_uid_gid(new_uid, new_gid) < 0)
+	bool have_cap_setgid;
+	uid_t new_uid = handler->conf->init_uid;
+	gid_t new_gid = handler->conf->init_gid;
```
brauner: Please move this to the top of the function. We don't use middle-of-scope declarations. :)

tanyifeng: fixed

brauner:
Applied to master in https://github.com/lxc/lxc/commit/928b1f04fc0c7579e19822e0f8a620ba121d7f97 with just a minor change to the commit message. Thanks, @tanyifeng. :)

PR #1952, opened 2017-11-28
lxc_helper.h: Avoid duplicated sethostname

marcosps:
Create new lxc_helper.h header to hold the implementation of
sethostname, instead of duplicating code in conf.c and lxc_unshare.c

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -122,19 +122,6 @@ static int pivot_root(const char * new_root, const char * put_old)
 extern int pivot_root(const char * new_root, const char * put_old);
 #endif
 
-/* Define sethostname() if missing from the C library */
-#ifndef HAVE_SETHOSTNAME
-static int sethostname(const char * name, size_t len)
-{
-#ifdef __NR_sethostname
-	return syscall(__NR_sethostname, name, len);
-#else
-	errno = ENOSYS;
-	return -1;
-#endif
-}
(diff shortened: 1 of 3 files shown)


marcosps:
I have a question about the copyright statement in the new header file, is that right? I added myself as author, but @stgraber wrote the original code in both conf.c and lxc_unshare.c. If this is wrong, I can remove my name from it and add Stephan instead.

brauner:
The definition should simply go into `utils.h`.

src/include/lxc_helper.h
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#include <sys/syscall.h>
+
+/* Define sethostname() if missing from the C library */
+#ifndef HAVE_SETHOSTNAME
```
brauner: The definition should simply go into utils.h so no need for a new file.

marcosps:
Sorry, pushed by mistake. Still needs more work.

marcosps:
Now it works as expected.

brauner:
jenkins: test this please

brauner:
Thanks!

PR #1955, opened 2017-11-28
Fix errors in configuration file parsing

flx42:
- One memory leak
- One stack overflow
- Multiple out-of-bounds accesses

### DIFF
--- a/src/lxc/log.c
+++ b/src/lxc/log.c
@@ -533,6 +533,17 @@ extern int lxc_log_syslog(int facility)
 		lxc_log_category_lxc.appender = &log_appender_syslog;
 		return 0;
 	}
+
+	appender = lxc_log_category_lxc.appender;
+	/* Check if syslog was already added, to avoid creating a loop */
+	while (appender) {
+		if (appender == &log_appender_syslog) {
+			/* not an error: openlog re-opened the connection */
+			return 0;
+		}
+		appender = appender->next;
+	}
+
 	appender = lxc_log_category_lxc.appender;
 	while (appender->next != NULL)
 		appender = appender->next;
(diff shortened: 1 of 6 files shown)


brauner:
I'd like to have tests for the latter two cases but if you'd prefer I can just add them. A test for the second to last commit should be added to https://github.com/lxc/lxc/blob/master/src/tests/parse_config_file.c and for the last commit to https://github.com/lxc/lxc/blob/master/src/tests/lxc-test-utils.c .

brauner:
jenkins: test this please

flx42:
No problem, I'll add the tests.

flx42:
Oops, messed up the commits, will fix, sorry.

flx42:
@brauner PTAL.
While adding the tests, I noticed an infinite loop in the logging code, I added another patch for this: d964d06b888dad4e2a87e0975a6ee71874bc681a

With the old `lxc_config_net_hwaddr`, valgrind reports 8 memory errors when running the test program.
With the new `lxc_config_net_hwaddr`, only one memory error. But it comes from `parse_byte_size_string`

brauner:
@flx42, nicely spotted! Thanks for the close look and the patches. I'm going to do a review now.

src/lxc/log.c
```diff
@@ -533,6 +533,17 @@ extern int lxc_log_syslog(int facility)
...
 		return 0;
 	}
+
+	appender = lxc_log_category_lxc.appender;
+	/* Check if syslog was already added, to avoid creating a loop */
+	while (appender) {
+		if (appender == &log_appender_syslog) {
+			/* not an error: openlog re-opened the connection */
+			return 0;
+		}
```
brauner: For future reference, when there's only a single line after an if-condition we don't enclose it in brackets. :) But no need to bikeshed on this. .)

brauner:
jenkins: test this please

brauner:
Thanks!

Issue #494, opened 2015-04-21
"lxc-unshare -s USER -u uid" not working


brauner:
It seems that `lxc-unshare -s USER -u uid` is not working. Running e.g.

    lxc-unshare -s USER -u 101000 /bin/bash

will not result in a new shell with `uid 101000`. I'm on an `Archlinux` box with `kernel 4.0.0`, `lxc 1.1.2`, `lxcfs 0.7`, `cgmanager 0.36`. Here is the output of `strace -o FILE lxc-unshare -s USER -u 101000 /bin/bash`:

    execve("/usr/bin/lxc-unshare", ["lxc-unshare", "-s", "USER", "-u", "1002", "/bin/bash"], [/* 27 vars */]) = 0
    brk(0)                                  = 0x1e2b000
    access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
    open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
    fstat(3, {st_mode=S_IFREG|0644, st_size=67616, ...}) = 0
    mmap(NULL, 67616, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7fbf7709e000
    close(3)                                = 0
    open("/usr/lib/liblxc.so.1", O_RDONLY|O_CLOEXEC) = 3
    read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0000\5\1\0\0\0\0\0"..., 832) = 832
    fstat(3, {st_mode=S_IFREG|0755, st_size=517672, ...}) = 0
    mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fbf7709d000
    mmap(NULL, 2613040, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7fbf76c10000
    mprotect(0x7fbf76c8b000, 2093056, PROT_NONE) = 0
    mmap(0x7fbf76e8a000, 16384, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x7a000) = 0x7fbf76e8a000
    close(3)                                = 0
    open("/usr/lib/libpthread.so.0", O_RDONLY|O_CLOEXEC) = 3
[...]

PR #1979, opened 2017-11-29
lxc_unshare: Add uid_mapping when creating userns

marcosps:
Change conf.c to export function write_id_mapping, which will now be
called inside main function of lxc_unshare.c.

This is required because setuid syscalls only permits a new userns to
set a new uid if the uid of parameter is mapped inside the ns using
uid_map file[1]. So, just after the clone invocation, map the uid passed as
parameter into the newly created user namespace, and put the current uid
as the ID-outside-ns. After the mapping is done, setuid call succeeds.

Closes: #494

[1] https://elixir.free-electrons.com/linux/latest/source/kernel/user_namespace.c#L286


src/lxc/tools/lxc_unshare.c
```diff
@@ -112,6 +117,9 @@ static int do_start(void *arg)
 	int want_default_mounts = start_arg->want_default_mounts;
 	const char *want_hostname = start_arg->want_hostname;
 
+	/* waiting until uid maps is set */
+	read(wait_fd, &val, sizeof(char));
```
brauner: You can't really ignore the return value from `read()`/`write()`. You need to assign it and do something with it you could `printf()` a harmless info or direct output of `fprintf()` to `/dev/null`:

```
	mv -f $depbase.Tpo $depbase.Po
../../../src/lxc/tools/lxc_unshare.c: In function ‘main’:
../../../src/lxc/tools/lxc_unshare.c:276:8: error: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Werror=unused-result]
   write(wait_fd, &val, 8);
        ^
../../../src/lxc/tools/lxc_unshare.c: In function ‘do_start’:
../../../src/lxc/tools/lxc_unshare.c:121:6: error: ignoring return value of ‘read’, declared with attribute warn_unused_result [-Werror=unused-result]
  read(wait_fd, &val, sizeof(char));
```

marcosps:
Fixed the problem with read/write in this new version.

brauner:
jenkins: test this please

src/lxc/tools/lxc_unshare.c
```diff
@@ -44,6 +45,10 @@
 #include "network.h"
 #include "utils.h"
 
+
+static int wait_fd = -1;
+static uint64_t val = 1;
```
brauner: Please don't make them global variables. Open the eventfd in main() and pass it as an argument to lxc_clone() by giving the args struct that is passed to main an additional argument. Same thing goes for the sync value.

src/lxc/tools/lxc_unshare.c
```diff
@@ -112,6 +117,14 @@ static int do_start(void *arg)
 	int want_default_mounts = start_arg->want_default_mounts;
 	const char *want_hostname = start_arg->want_hostname;
 
+	if (start_arg->setuid) {
+		/* waiting until uid maps is set */
+		if (read(wait_fd, &val, sizeof(val)) == -1) {
```
brauner: Nit, please use:

    ret = read();
    if (ret < 0) {
            /* stuff happening */
    }

as we do it all over the codebase.

brauner: And please close the wait_fd on error conditions. It's just good practice.

src/lxc/tools/lxc_unshare.c
```diff
@@ -241,12 +254,43 @@ int main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 
+	if (start_arg.setuid) {
+		wait_fd = eventfd(0, EFD_CLOEXEC);
+		if (wait_fd == -1) {
```
brauner: As I said, above, declare `wait_fd` at the top of main and pass `wait_fd` as arg in start_arg.

src/lxc/tools/lxc_unshare.c
```diff
@@ -241,12 +254,43 @@ int main(int argc, char *argv[])
...
 	}
 
+	if (start_arg.setuid) {
+		/* enough space to accommodate uids */
+		char *umap = (char *)alloca(100);
+
+		/* create new uid mapping using current UID and the one
+		 * specified as parameter
+		 */
+		if (snprintf(umap, 100, "%d %d 1" , *(start_arg.uid), getuid()) < 0) {
```
brauner: Please use:

    ret = snprintf();
    if (ret < 0 || ret >= 100) {
            /* handle error */
    }

src/lxc/tools/lxc_unshare.c
```diff
@@ -241,12 +254,43 @@ int main(int argc, char *argv[])
...
+			fprintf(stderr, "snprintf failed");
+			exit(EXIT_FAILURE);
+		}
+
+		if (write_id_mapping(ID_TYPE_UID, pid, umap, strlen(umap))) {
```
brauner: Iirc, writing to `/proc/<pid>/uid_map` requires the string to include a final `\n`.

marcosps:
I hope that all issues pointed by you were solved in this version.

marcosps:
Let me know if you find any other code that needs more changes.

marcosps:
@brauner let me know if you find any other code that needs more changes.

src/lxc/tools/lxc_unshare.c
```diff
@@ -241,12 +256,49 @@ int main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 
+	if (start_arg.setuid) {
+		start_arg.wait_fd = eventfd(0, EFD_CLOEXEC);
+		if (start_arg.wait_fd < 0) {
+			perror("eventfd");
```
brauner: Nit: Why `perror()` instead of `fprintf()` like in all the other error paths? :)

marcosps: oops, fixed :)

brauner:
jenkins: test this please

brauner:
lgtm, but let's see if Android is happy with `eventfd()`.

stgraber:
jenkins: test this please

stgraber:
Looks good on Android: https://jenkins.linuxcontainers.org/view/All/job/lxc-build-android/2568/

marcosps:
This testsuit error is not related to my changes. So the patch is good to go?

PR #1980, opened 2017-12-01
OCI layer caching + misc fixes

flx42:


hallyn:
jenkins: test this please

hallyn:
Thanks - patchset looks good.

flx42:
Caching will be enabled by default with this patchset (provided skopeo is recent enough), is that fine for you?

hallyn:
yeah i think it's good.  where does skopeo cache the images by default?

flx42:
It doesn't, you need to use `dest-shared-blob-dir` with a path.
Previously, we downloaded all the layers to a temp directory.

PR #1981, opened 2017-12-01
utils: declare sethostname() static inline

brauner:


brauner:
@stgraber, fixes Android.

PR #1987, opened 2017-12-02
Started to remove dependencies from /tools subfolder 

RicardoSanchezA:
Added arguments.{c,h} and utils.{c,h} to /tools and removed dependencies and internal logging from "lxc_cgroup.c".


src/lxc/tools/utils.c
```diff
@@ -0,0 +1,9 @@
+
+#include "tools/utils.h"
```
brauner: Both files, `utils.{c,h}` need to gain a license a top anything else as well. You can add your name and email there but please also include the others of the original function in the license header. An example would be:

```
/*
 * Copyright © 2017 Ricardo Sanchez A <ricardo.sanchez@utexas.edu>.
 *
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License version 2, as
 * published by the Free Software Foundation.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License along
 * with this program; if not, write to the Free Software Foundation, Inc.,
 * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
```

However, I'm not an attorney. :)

RicardoSanchezA:
So, I modified `src/lxc/Makefile.am` to reflect the movement of `arguments.{c,h}` from `src/lxc` into `srx/lxc/tools`. I also modified all of the tools that use `arguments.{c,h}` so that they now include `tools/arguments.h`. Moreover, I removed the use of internal symbols from `tools/arguments.{c,h}`.

One thing that I haven't done is move `lxc_init.c` up to `src/lxc` because I am not completely sure of what arguments I need to parse for this binary. By looking at the code of `lxc_init.c` I realized that it doesn't really use any arguments provided to the program, not even the name of the container. Thus, it seems that `lxc-init` doesn't even need an argument parser, is this correct or am I misunderstanding something?

RicardoSanchezA:
Wait. I just realized that the `Makefile.am` is incomplete. I'm working on it and will post a fix soon.

src/lxc/tools/lxc_attach.c
```diff
@@ -36,7 +36,7 @@
 #include <lxc/lxccontainer.h>
 
 #include "attach.h"
-#include "arguments.h"
+#include "tools/arguments.h"
```
brauner: That's actually unnecessary you should just include `"arguments.h"`. What I meant in the last comment is what you did in `Makefile.am` above, namely prefixing the references to `arguments.{c,h}` for each tool with `tools/`.
The benefit of doing it just in the Makefile and not in each binary is that you have a single place to change when you decide to change paths. For example, when you were to move everything from `tools/` to `cmd/` then you would only need to change `Makefile.am` and not each binary.

RicardoSanchezA:
Ok. Since I messed up this pull request, I'll close it and make a new one with the required changes.

Issue #723, opened 2015-12-12
Improve error handling


brauner:
Improve error handling in lxc

As we have discussed in a few mails error handling in lxc should be improved.
Here are two different suggestions.

We should start by defining a set of reusable error codes which are modelled
after traditional C error codes. They should be put in a common header
`lxc_error.h`. An example could be:

```
    #define ELXCIDMAP 1 /* Indicates missing or invalid id mapping. /*
```

Associated with it we can have a really small set of error handling functions.
Something in the spirit of `strerror()` e.g. `lxc_strerror()` which turns the error
code into a human readable string that gives useful information about the error.

This should be the smallest common denominator for an error checking system. From
this point onward I see two possible ways to proceed.

(1) Simple and fast error handling:

We keep an `error_num` variable in `struct lxc_container *current_config` which
contains the last error code which we can report back to the user.

`error_num` will only be set by the api functions defined in lxccontainer.c.

(2) Sophisticated error handling:

The more full-blown approach would probably be to define error codes per module.
That means we would have a simple `struct lxc_error` in `struct lxc_container
*current_config` with a field for a module and an associated error code. A module
here would be e.g. bdev.h. `lxc_strerror()` would then take two arguments: module
and error code and return an appropriate human readable string.

Icing:

When this is working we can discuss if we want to add a `struct lxc_list
*error_list` that is a list of strings or error codes of all the things that went
wrong. (In the spirit of what @hallyn suggested.)

hallyn:
Both of those might be fine, but let me complicate it with another proposal :) Feel free to shoot down:

1. Classify our error messages into two types: leaf and non-leaf.

2. Keep a variable-sized lxc_conf->error_msg

3. when we get a non-leaf error message, for instance we just called
   open_file(), which returned a failure, we do (pseudo-code)

   	current_config->error_msg = "%s: %s" % (our_new_message, current_config->error_msg)

2. When we get a leaf error message (like, inability to open a file), we
   a. call lxc_error_dump() which prints out current_config->error_msg
   b. set current_config->error_msg to just our new error message.

3. lxc_error_dump() would nicely format the message so that each set of related messages is clearly related.

AustinReichert:
Here's the PR for the progress I've made on this, I would appreciate feedback :) https://github.com/lxc/lxc/pull/1997

brauner:
Our error handlig has been improved substantially and I still intend to rip out some of the stuff @AustinReichert did and integrate some of it into liblxc.

PR #1997, opened 2017-12-04
skeleton for error handling

AustinReichert:
Requesting review for skeleton error handling for https://github.com/lxc/lxc/issues/723

THIS IS NOT READY TO BE MERGED.

The following is copied from lxc_error.h for ease of access

Overview 
1. Error codes defined based on grepping for "error" in tools/

2. lxc_error_dump() has intelligence to know when/if to dump errors and concat,
    or just concatenate, thus only needing to add one lxc_error_dump call per
    error to capture, and all that needs to be known is what type of error it is.

3. lxc_error_concat() is called by lxc_error_dump and has logic for appropriately
    concatenating past/new error strings. Used '~' as delimiter since it's the least
    used ASCII character, reduces need for escaping. If not sufficient can use, "~~" :)

4. lxc_error_handle() can be called to look at the most recent error_num and run
    some code to "handle" it in some way, then return success/failure

5. main() and test_dump() are only there for testing at this point, it will be removed

Example:
```
if (something_did_not_work) {
     char *lxc_error = "<some-error>";
     SYSERROR("%s", lxc_error);
     lxc_error_dump(c, lxc_error, <LXC_ERROR_CODE>);
}
```

Issues:
1. Can only store errors where a struct lxc_container *c is accessible, this
    means lots of liblxc files cannot use this. How to resolve this/is this a
    problem? Shouldn't the container try to handle errors from functions it can't
    see inside of?

 2. Would like to be able to identify which file the error occurred in. i.e.
    "lxc_attach: Out of memory", but not sure how to do it (aside from manually).
    Trying to figure out a way to use my_args.progname, but any other suggestions
    welcome.

 3. Need a better ordering for macros, and way to determine which are considered
    leaf vs non-leaf

 4. Need to figure out how/which errors should/can be handled/ignored.

 5. Incorporate into Makefile and build, handle dependencies and decide #includes

 6. 3 memory leaks from lxc_error_concat

 7. Need to better understand how the internal logging mechanism and the error handling complement 
     each other, the error dumping from the error handling seems redundant if we have the logging mechanism


brauner:
Slowly picking bits and pieces from this branch.

PR #2001, opened 2017-12-05
Moved arguments.{c,h} to tools/ and updated Makefile

RicardoSanchezA:
@brauner 
So, I decided to keep it simple in this pull request and will eventually make more as Yoshi and I keep removing dependencies from more tools. For now, I simply moved `arguments.{c,h}` to the `tools/` subfolder and updated `Makefile.am` to reflect this file movements (i.e. remove `arguments.{c,h}` from the liblxc sources and add it to the sources of each tool).

Regarding the argument parser for `lxc_init.c`, I think I was completely misunderstanding what we are supposed to do. The argument parser for `lxc_init.c` will basically be a subset of the parser in `arguments.c` (i.e. it will only parse the arguments that are required for `lxc_init.c` to work), right? 
Or am I still getting it wrong? If so, please enlighten me since I would really like to contribute to this project and help you out to get rid of dependencies in the tools. :) 

Thanks!


brauner:
jenkins: test this please

brauner:
Adapted the commit message and pushed to master in a144d2883021e67d22412e11fb6d2ee2a6439798 . So closing here. :)

brauner:
> Regarding the argument parser for lxc_init.c, I think I was completely misunderstanding what we are supposed to do. The argument parser for lxc_init.c will basically be a subset of the parser in arguments.c 
(i.e. it will only parse the arguments that are required for lxc_init.c to work), right?

Yes, that is correct. `lxc_init.c` should be moved from `tools/` one level up, stripped of any dependency on `arguments.{c,h}` and then receive a minimal parser that just parses the arguments `lxc_init.c` currently receives.

RicardoSanchezA:
@brauner 
Okay, sorry about that. I'll make sure to use shorter commit messages and triple check that the code is compiling before making pull requests.

More on the `lxc_init.c` argument parser: since the code for it won't be that short, do you think it's a reasonable idea to create `.c` and `.h` files (something like `lxc_init_arguments.{c,h}`) to store the code of the parser in there or should we keep it simple and just add all of the parser code in `lxc_init.c`?

brauner:
On Tue, Dec 05, 2017 at 05:59:30AM -0800, Ricardo Sanchez wrote:
> @brauner 
> Okay, sorry about that. I'll make sure to use shorter commit messages and triple check that the code is compiling before making pull requests.

No worries, I wasn't trying to reprimand you. :) That's simply stuff one
picks up along the way.

> 
> More on the `lxc_init.c` argument parser: since the code for it won't be that short, do you think it's a reasonable idea to create `.c` and `.h` files (something like `lxc_init_arguments.{c,h}`) to store the code of the parser in there or should we keep it simple and just add all of the parser code in `lxc_init.c`?

I would leave all the code in the same file. The reasoning behind this
is:
- we don't expect that code to be used from any other place so you can
  make declare all functions in there as static
- keeps the whole thing self-contained
- sparses you adapting the Makefile :) (The less time spent with the
  build environment the better. :))

Thanks!
Christian

PR #2004, opened 2017-12-05
start: set loglevel correctly

### LINKED ISSUE #2003: CLI log options don't correctly override the configuration file
I think this comment is currently incorrect:
https://github.com/lxc/lxc/blob/a144d2883021e67d22412e11fb6d2ee2a6439798/src/lxc/conf.h#L290-L297

I tried using the `LXC_LOG_LEVEL` feature in my hook (since b8f88d9bb2371d2e11b9c2fbc0a72d8e9608042e), but I was always getting `NOTSET` when using `lxc-execute --logpriority=DEBUG`. However, setting `lxc.log.level` in the container configuration file worked fine. Do we agree it's a bug?

I believe we are missing something like this:
https://github.com/flx42/lxc/commit/36b881fa9310438fc6ac2942b70b0a2e6202fd6f
However, this is no fully correct. If you specify a log file in the configuration, this file will still be created (but will be empty). It's because the log file is opened immediately when the configuration file is parsed:
https://github.com/lxc/lxc/blob/a144d2883021e67d22412e11fb6d2ee2a6439798/src/lxc/confile.c#L1246-L1251

Should we extract the log initialization from this function? As a side-effect, I believe it will also avoid opening all log files when doing a simple `lxc-ls`.

brauner:
We want the loglevel that the user specified when starting the container
because it overrides the default one set in the config.

Closes #2003.

Reported-by: Felix Abecassis <fabecassis@nvidia.com>

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -572,6 +572,7 @@ on_error:
 
 int lxc_init(const char *name, struct lxc_handler *handler)
 {
+	const char *loglevel;
 	struct lxc_conf *conf = handler->conf;
 
 	lsm_init();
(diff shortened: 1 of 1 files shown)


brauner:
Jenkins failed to pull from github for both i386 builders. I want to see at least one clean i386 build --> restarting.

brauner:
jenkins: test this please

brauner:
@stgraber, good to go.

PR #2008, opened 2017-12-06
add --share-$NS= support to lxc-execute

tych0:
This is basically just hoisting the functionality out of lxc-start, so lxc-execute can use it too.

### DIFF
--- a/src/lxc/lxc_init.c
+++ b/src/lxc/lxc_init.c
@@ -92,6 +92,86 @@ static struct arguments my_args = {
 	.shortopts = short_options
 };
 
+static void prevent_forking(void)
+{
+	FILE *f;
+	char name[PATH_MAX], path[PATH_MAX];
+	int ret;
+
+	f = fopen("/proc/self/cgroup", "r");
+	if (!f) {
+		SYSERROR("opening /proc/self/cgroup");
+		return;
+	}
+
+	while (!feof(f)) {
+		int fd;
+
+		if (2 != fscanf(f, "%*d:%[^:]:%s", name, path)) {
+			ERROR("didn't scan the right number of things");
+			goto out;
+		}
+
+		if (strcmp(name, "pids"))
+			continue;
+
+		ret = snprintf(name, sizeof(name), "/sys/fs/cgroup/pids/%s/pids.max", path);
+		if (ret < 0 || ret >= sizeof(path)) {
+			ERROR("failed snprintf");
+			goto out;
+		}
+
+		fd = open(name, O_WRONLY);
+		if (fd < 0) {
+			SYSERROR("open");
+			goto out;
+		}
+
+		if (write(fd, "1", 1) != 1)
+			SYSERROR("write");
+
+		close(fd);
+		break;
+	}
+
+out:
+	fclose(f);
+}
+
+static void kill_children(pid_t pid)
+{
+	FILE *f;
+	char path[PATH_MAX];
+	int ret;
+
+	ret = snprintf(path, sizeof(path), "/proc/%d/task/%d/children", pid, pid);
+	if (ret < 0 || ret >= sizeof(path)) {
+		ERROR("failed snprintf");
+		return;
+	}
+
+	f = fopen(path, "r");
+	if (!f) {
+		SYSERROR("couldn't open %s", path);
+		return;
+	}
+
+	while (!feof(f)) {
+		pid_t pid;
+
+		if (fscanf(f, "%d ", &pid) != 1) {
+			ERROR("couldn't scan pid");
+			fclose(f);
+			return;
+		}
+
+		kill_children(pid);
+		kill(pid, SIGKILL);
+	}
+
+	fclose(f);
+}
+
 int main(int argc, char *argv[])
 {
 	int i, ret;
@@ -258,18 +338,28 @@ int main(int argc, char *argv[])
 		case SIGTERM:
 			if (!shutdown) {
 				shutdown = 1;
-				ret = kill(-1, SIGTERM);
-				if (ret < 0)
-					DEBUG("%s - Failed to send SIGTERM to "
-					      "all children", strerror(errno));
+				prevent_forking();
+				if (getpid() != 1) {
+					kill_children(getpid());
+				} else {
+					ret = kill(-1, SIGTERM);
+					if (ret < 0)
+						DEBUG("%s - Failed to send SIGTERM to "
+						      "all children", strerror(errno));
+				}
 				alarm(1);
 			}
 			break;
 		case SIGALRM:
-			ret = kill(-1, SIGKILL);
-			if (ret < 0)
-				DEBUG("%s - Failed to send SIGKILL to all "
-				      "children", strerror(errno));
+			prevent_forking();
+			if (getpid() != 1) {
+				kill_children(getpid());
+			} else {
+				ret = kill(-1, SIGTERM);
+				if (ret < 0)
+					DEBUG("%s - Failed to send SIGTERM to "
+					      "all children", strerror(errno));
+			}
 			break;
 		default:
 			ret = kill(pid, was_interrupted);
--- a/src/lxc/lxc_init.c
+++ b/src/lxc/lxc_init.c
@@ -46,6 +46,9 @@
 #define OPT_USAGE 0x1000
 #define OPT_VERSION OPT_USAGE - 1   
 
+#define QUOTE(macro) #macro
+#define QUOTEVAL(macro) QUOTE(macro)
+
 lxc_log_define(lxc_init, lxc);
 
 static sig_atomic_t was_interrupted = 0;
@@ -95,40 +98,49 @@ static struct arguments my_args = {
 static void prevent_forking(void)
 {
 	FILE *f;
-	char name[PATH_MAX], path[PATH_MAX];
+	char name[MAXPATHLEN], path[MAXPATHLEN];
 	int ret;
 
 	f = fopen("/proc/self/cgroup", "r");
 	if (!f) {
-		SYSERROR("opening /proc/self/cgroup");
+		SYSERROR("Failed to open \"/proc/self/cgroup\"");
 		return;
 	}
 
 	while (!feof(f)) {
-		int fd;
+		int fd, i;
 
-		if (2 != fscanf(f, "%*d:%[^:]:%s", name, path)) {
-			ERROR("didn't scan the right number of things");
+		if (1 != fscanf(f, "%*d:%" QUOTEVAL(MAXPATHLEN) "s", name)) {
+			ERROR("Failed to parse \"/proc/self/cgroup\"");
 			goto out;
+		}
+		path[0] = 0;
+
+		for (i = 0; i < sizeof(name); i++) {
+			if (name[i] == ':') {
+				name[i] = 0;
+				strncpy(path, name + i + 1, sizeof(path));
+				break;
+			}
 		}
 
 		if (strcmp(name, "pids"))
 			continue;
 
 		ret = snprintf(name, sizeof(name), "/sys/fs/cgroup/pids/%s/pids.max", path);
-		if (ret < 0 || ret >= sizeof(path)) {
-			ERROR("failed snprintf");
+		if (ret < 0 || (size_t)ret >= sizeof(path)) {
+			ERROR("Failed to create string");
 			goto out;
 		}
 
 		fd = open(name, O_WRONLY);
 		if (fd < 0) {
-			SYSERROR("open");
+			SYSERROR("Failed to open \"%s\"", name);
 			goto out;
 		}
 
 		if (write(fd, "1", 1) != 1)
-			SYSERROR("write");
+			SYSERROR("Failed to write to \"%s\"", name);
 
 		close(fd);
 		break;
@@ -145,14 +157,14 @@ static void kill_children(pid_t pid)
 	int ret;
 
 	ret = snprintf(path, sizeof(path), "/proc/%d/task/%d/children", pid, pid);
-	if (ret < 0 || ret >= sizeof(path)) {
-		ERROR("failed snprintf");
+	if (ret < 0 || (size_t)ret >= sizeof(path)) {
+		ERROR("Failed to create string");
 		return;
 	}
 
 	f = fopen(path, "r");
 	if (!f) {
-		SYSERROR("couldn't open %s", path);
+		SYSERROR("Failed to open %s", path);
 		return;
 	}
 
@@ -160,7 +172,7 @@ static void kill_children(pid_t pid)
 		pid_t pid;
 
 		if (fscanf(f, "%d ", &pid) != 1) {
-			ERROR("couldn't scan pid");
+			ERROR("Failed to retrieve pid");
 			fclose(f);
 			return;
 		}
@@ -337,10 +349,12 @@ int main(int argc, char *argv[])
 		case SIGPWR:
 		case SIGTERM:
 			if (!shutdown) {
+				pid_t mypid = getpid();
+
 				shutdown = 1;
 				prevent_forking();
-				if (getpid() != 1) {
-					kill_children(getpid());
+				if (mypid != 1) {
+					kill_children(mypid);
 				} else {
 					ret = kill(-1, SIGTERM);
 					if (ret < 0)
@@ -350,10 +364,12 @@ int main(int argc, char *argv[])
 				alarm(1);
 			}
 			break;
-		case SIGALRM:
+		case SIGALRM: {
+			pid_t mypid = getpid();
+
 			prevent_forking();
-			if (getpid() != 1) {
-				kill_children(getpid());
+			if (mypid != 1) {
+				kill_children(mypid);
 			} else {
 				ret = kill(-1, SIGTERM);
 				if (ret < 0)
@@ -361,6 +377,7 @@ int main(int argc, char *argv[])
 					      "all children", strerror(errno));
 			}
 			break;
+		}
 		default:
 			ret = kill(pid, was_interrupted);
 			if (ret < 0)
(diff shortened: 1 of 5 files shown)


src/lxc/tools/lxc_execute.c
```diff
@@ -73,6 +77,10 @@ static const struct option my_longopts[] = {
 	{"define", required_argument, 0, 's'},
 	{"uid", required_argument, 0, 'u'},
 	{"gid", required_argument, 0, 'g'},
+	{"share-net", required_argument, 0, OPT_SHARE_NET},
+	{"share-ipc", required_argument, 0, OPT_SHARE_IPC},
+	{"share-uts", required_argument, 0, OPT_SHARE_UTS},
+	{"share-pid", required_argument, 0, OPT_SHARE_PID},
```
hallyn: Ok, so here I get scared, though, because if you look at src/lxc/tools/lxc_init.c, when it gets a signal it likes to kill -1.  So if you share-pid with host pid 1, then sending sigterm to the container will try to send that to the host init.

Now for unpriv containers that shoudl be fine as it won't be allowed to kill -1, but for privileged containers that's a bad thing.

Not sure what to do about it though.  Just let the user shoot themselves in the foot?

src/lxc/tools/lxc_start.c
```diff
@@ -284,27 +278,8 @@ int main(int argc, char *argv[])
...
-			key = "lxc.namespace.uts";
-		else if (i == LXC_NS_PID)
-			key = "lxc.namespace.pid";
-		else
-			continue;
-
-		if (!c->set_config_item(c, key, value))
-			goto out;
-	}
+	if (!lxc_ns_setup(&my_args, c))
```
hallyn: how about lxc_ns_sharing_setup() to be more specific?

src/lxc/tools/lxc_execute.c
hallyn: Maybe we update lxc_init to not kill -1 if it's share-pid?

src/lxc/tools/lxc_start.c
brauner: `lxc_setup_shared_ns()` (just to throw one more into the mix :))

src/lxc/tools/lxc_execute.c
brauner: @hallyn, if there's a sensible way to do this then sure.

hallyn: It would have to detect that its vpid != 1.

src/lxc/tools/lxc_start.c
hallyn: yeah I like that better.

src/lxc/tools/lxc_execute.c
tych0: Ok, I'll add a patch on top to do this.

src/lxc/tools/lxc_start.c
tych0: Sounds good, I'll rename it.

tych0:
Sorry for the delay, I just pushed the updates now.

FWIW, before this patch I got:

```
strace: Process 3110 attached
wait4(-1, 0x7fff125a664c, 0, NULL)      = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGPWR {si_signo=SIGPWR, si_code=SI_USER, si_pid=3172, si_uid=0} ---
rt_sigreturn({mask=[]})                 = -1 EINTR (Interrupted system call)
kill(4294967295, SIGTERM)               = -1 EACCES (Permission denied)
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=3113, si_uid=0, si_status=0, si_utime=0, si_stime=0} ---
rt_sigreturn({mask=[]})                 = -1 EACCES (Permission denied)
alarm(1)                                = 0
wait4(-1, [{WIFEXITED(s) && WEXITSTATUS(s) == 0}], 0, NULL) = 3113
alarm(1)                                = 1
wait4(-1, 0x7fff125a664c, 0, NULL)      = -1 ECHILD (No child processes)
exit_group(0)                           = ?
+++ exited with 0 +++
```

And it didn't kill my entire host, but it *did* kill the actual child task. So I'm not sure what the kernel's checks are exactly, but we might have been safe without this.

hallyn:
I think that's apparmor's rule only allowing signal to your own profile.

hallyn:
Does kill_children need to be arrested^W recursive?

brauner:
So what worries me here is the possibility that we can be forced into an infinite loop in `kill_children()` since another process could always fork of a new process. This can't happen with`kill(-1, SIGTERM)` since the `kill` syscall will take a read lock `tasklist_lock` and when another process tries to `fork()` it will have to take a write lock `tasklist_lock`. `copy_process()` when it succeeds with the write lock will check for any pending signals and fail if there are any trying to restart the process. But since the parent will exit on a deadly signal the `fork()` will fail. When we're relying on doing this in userspace we're opening us up to `fork()` bombs I fear.

brauner:
I mean I'm not nacking this but we should probably at least document this in `lxc.container.conf.sgml.in` unless I'm thinking wrong here.

brauner:
Fwiw, if we had container IDs like they are discussed you could make a case for having a `kill2()` syscall that kills processes per-container ID. :)

hallyn:
> So what worries me here is the possibility that we can be forced into an infinite loop in `kill_children()` since another process could always fork of a new process.

If there were a way for init to talk to lxc, it could request lxc to
freeze the whole container then kill -1 each task in it.

brauner:
> If there were a way for init to talk to lxc, it could request lxc to
freeze the whole container then kill -1 each task in it.

I was thinking the same thing but `freezer` won't be in `cgroup v2`. :)

hallyn:
> Fwiw, if we had container IDs like they are discussed you could make a case for having a `kill2()` syscall that kills processes per-container ID. :)

Or,

init could set the number of pids to 1 in pids cgroup, so no more
processes can be forked?

hallyn:
> > If there were a way for init to talk to lxc, it could request lxc to
> freeze the whole container then kill -1 each task in it.
> 
> I was thinking the same thing but `freezer` won't be in `cgroup v2`. :)

That's insane.

brauner:
> Or,

> init could set the number of pids to 1 in pids cgroup, so no more
processes can be forked?

That sounds more reasonable to me.

brauner:
> Quoting Christian Brauner (notifications@github.com):
> > If there were a way for init to talk to lxc, it could request lxc to
> freeze the whole container then kill -1 each task in it.
> 
> I was thinking the same thing but `freezer` won't be in `cgroup v2`. :)
> 
> That's insane.

Let me double check.

brauner:
> > Quoting Christian Brauner (notifications@github.com):
> > > If there were a way for init to talk to lxc, it could request lxc to
> > freeze the whole container then kill -1 each task in it.
> > 
> > I was thinking the same thing but `freezer` won't be in `cgroup v2`. :)
> > 
> > That's insane.

> Let me double check.

According to kernel's git history it seems to be planned to be in cgroup v2 so yes, freezer would be good and pids would also be a good place.

brauner:
> If there were a way for init to talk to lxc, it could request lxc to
freeze the whole container then kill -1 each task in it.

I'm still uncomfortable with this idea. I really don't want the `lxc-init` binary to be able to talk to liblxc. It should be as dumb as possible. Also, it would require a `freeze()` + `deliver_signals()` + `unfreeze()`. I'm also not clear how liblxc would freeze the container without also freezing `lxc-init`.

hallyn:
> > If there were a way for init to talk to lxc, it could request lxc to
> freeze the whole container then kill -1 each task in it.
> 
> I'm still uncomfortable with this idea. I really don't want the `lxc-init` binary to be able to talk to liblxc. It should be as dumb as possible. Also, it would require a `freeze()` + `deliver_signals()` + `unfreeze()`. I'm also not clear how liblxc would freeze the container without also freezing `lxc-init`.

Yes, I prefer the pids cgroup approach.  It seems acceptable to me to
do that and tell people if they dont' have pids cgroup then we'll do
best-effort killing the children the way Tycho's patch now does.

tych0:
The freezer is quite unreliable and can cause kernel deadlocks on some filesystems (we went through this with NFS).

Yes, I think kill_children needs to be recursive. I don't think fork bombs will be an issue due to libc's buffering, but I can read pids first and only loop that many times. I'll make the changes.

hallyn:
Quoting Tycho Andersen (notifications@github.com):
> The freezer is quite unreliable and can cause kernel deadlocks on some filesystems (we went through this with NFS).
> 
> Yes, I think kill_children needs to be recursive. I don't think fork bombs will be an issue due to libc's buffering, but I can read pids first and only loop that many times. I'll make the changes.

Ok - but please do use the pids cgroup if it's available as well?

tych0:
Ok, I updated it to try and write pids.max = 1 before it starts killing stuff, and made kill_children recursive. Worth noting that my default lxc-execute invocation for testing didn't have the pids cgroup mounted, so that change will be a no-op for most invocations.

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+	f = fopen("/proc/self/cgroup", "r");
+	if (!f) {
+		SYSERROR("opening /proc/self/cgroup");
+		return;
+	}
+
+	while (!feof(f)) {
+		int fd;
+
+		if (2 != fscanf(f, "%*d:%s:%s", name, path)) {
```
hallyn: Any concerns about length of path?

src/lxc/tools/arguments.h
```diff
@@ -160,6 +161,11 @@ struct lxc_arguments {
 #define OPT_VERSION OPT_USAGE - 1
 #define OPT_RCFILE OPT_USAGE - 2
 
+#define OPT_SHARE_NET OPT_USAGE + 1
```
hallyn: (not sure) if these are now in a shared file, should these become OPT_USAGE -3 etc , so that OPT_USAGE + 1 is always usable by a shared file?

src/lxc/lxc_init.c
tych0: Anwyay, the only change was to this line, and it's now updated.

src/lxc/tools/arguments.h
tych0: Happy to do it either way, just let me know.

hallyn: Yes, please change it.

src/lxc/lxc_init.c
hallyn: ... do you trust PATH_MAX?

tych0: Sure, it's POSIX. If you can't trust the headers, none of this works.

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
 
+static void prevent_forking(void)
+{
+	FILE *f;
+	char name[PATH_MAX], path[PATH_MAX];
+	int ret;
+
+	f = fopen("/proc/self/cgroup", "r");
+	if (!f) {
+		SYSERROR("opening /proc/self/cgroup");
```
brauner: Please make this:

    SYSERROR("Failed to open \"/proc/self/cgroup\"");

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+		return;
+	}
+
+	while (!feof(f)) {
+		int fd;
+
+		if (2 != fscanf(f, "%*d:%[^:]:%s", name, path)) {
```
brauner: This will fail on systems that have the unified hierarchy enabled and where the unified hierarchy entry under `/proc/self/cgroup` comes before the `pids` cgroup entry:

    0::/user.slice/user-1000.slice/session-1.scope

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+			ERROR("didn't scan the right number of things");
```
brauner: Please make this:

    ERROR("Failed to parse \"/proc/self/cgroup\"");

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+			goto out;
+		}
+
+		if (strcmp(name, "pids"))
+			continue;
+
+		ret = snprintf(name, sizeof(name), "/sys/fs/cgroup/pids/%s/pids.max", path);
+		if (ret < 0 || ret >= sizeof(path)) {
+			ERROR("failed snprintf");
```
brauner: Please make this:

    ERROR("Failed to create string");

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+			goto out;
+		}
+
+		fd = open(name, O_WRONLY);
+		if (fd < 0) {
+			SYSERROR("open");
```
brauner: Please make this:

    SYSERROR("Failed to open \"%s\"", name);

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+			goto out;
+		}
+
+		if (write(fd, "1", 1) != 1)
+			SYSERROR("write");
```
brauner: Please make this:

    ERROR("Failed to write to \"%s\"", name);

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+
+static void kill_children(pid_t pid)
+{
+	FILE *f;
+	char path[PATH_MAX];
+	int ret;
+
+	ret = snprintf(path, sizeof(path), "/proc/%d/task/%d/children", pid, pid);
+	if (ret < 0 || ret >= sizeof(path)) {
+		ERROR("failed snprintf");
```
brauner: Please make this:

    ERROR("Failed to create string");

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+		return;
+	}
+
+	f = fopen(path, "r");
+	if (!f) {
+		SYSERROR("couldn't open %s", path);
```
brauner: Please make this:

    SYSERROR("Failed to open \"%s\"", path);

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+		return;
+	}
+
+	while (!feof(f)) {
+		pid_t pid;
+
+		if (fscanf(f, "%d ", &pid) != 1) {
+			ERROR("couldn't scan pid");
```
brauner: Please make this:

    ERROR("Failed to retrieve pid");

src/lxc/lxc_init.c
```diff
@@ -258,19 +338,29 @@ int main(int argc, char *argv[])
 		case SIGTERM:
 			if (!shutdown) {
 				shutdown = 1;
+				prevent_forking();
+				if (getpid() != 1) {
+					kill_children(getpid());
```
brauner: New `glibc` version have thankfully removed their insane `getpid()` userspace cache but this also means this is actually performing two syscalls. I know this is minor but I'd like `getpid()` be called once and record the `pid` in a variable.

src/lxc/tools/arguments.c
```diff
@@ -289,3 +290,34 @@ int lxc_arguments_str_to_int(struct lxc_arguments *args, const char *str)
...
+			key = "lxc.namespace.ipc";
+		else if (i == LXC_NS_UTS)
+			key = "lxc.namespace.uts";
+		else if (i == LXC_NS_PID)
+			key = "lxc.namespace.pid";
+		else
+			continue;
+
+		if (!c->set_config_item(c, key, value)) {
+			fprintf(stderr, "failed to set %s\n", key);
```
brauner: Please make this:

    ERROR("Failed to set \"%s = %s\"", key, value);

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+		if (2 != fscanf(f, "%*d:%[^:]:%s", name, path)) {
+			ERROR("didn't scan the right number of things");
+			goto out;
+		}
+
+		if (strcmp(name, "pids"))
+			continue;
+
+		ret = snprintf(name, sizeof(name), "/sys/fs/cgroup/pids/%s/pids.max", path);
+		if (ret < 0 || ret >= sizeof(path)) {
```
brauner: Some compilers under certain settings will complain about comparing different integer types so please use:

    if (ret < 0 || (size_t)ret >= sizeof(path))

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
...
+}
+
+static void kill_children(pid_t pid)
+{
+	FILE *f;
+	char path[PATH_MAX];
+	int ret;
+
+	ret = snprintf(path, sizeof(path), "/proc/%d/task/%d/children", pid, pid);
+	if (ret < 0 || ret >= sizeof(path)) {
```
brauner: Some compilers under certain settings will complain about comparing different integer types so please use:

    if (ret < 0 || (size_t)ret >= sizeof(path))

src/lxc/lxc_init.c
```diff
@@ -92,6 +92,86 @@ static struct arguments my_args = {
 	.shortopts = short_options
 };
 
+static void prevent_forking(void)
+{
+	FILE *f;
+	char name[PATH_MAX], path[PATH_MAX];
```
brauner: We actually use an internal macro `MAXPATHLEN` all across the codebase. Would be good if you could use this here too. :)

brauner:
@tych0, I like you too. :)

tych0:
:)

PR #2010, opened 2017-12-07
confile: enable tune host's OOM preferences

tanyifeng:

### DIFF
--- a/src/lxc/conf.h
+++ b/src/lxc/conf.h
@@ -106,6 +106,16 @@ struct lxc_sysctl {
 };
 
 /*
+ * Defines a structure to configure proc filesystem at runtime.
+ * @filename : the proc filesystem will be configured without the "lxc.proc" prefix
+ * @value    : the value to set
+ */
+struct lxc_proc {
+	char *filename;
+	char *value;
+};
+
+/*
  * id_map is an id map entry.  Form in confile is:
  * lxc.idmap = u 0    9800 100
  * lxc.idmap = u 1000 9900 100
@@ -383,6 +393,9 @@ struct lxc_conf {
 
 	/* sysctls */
 	struct lxc_list sysctls;
+
+	/* procs */
+	struct lxc_list procs;
 };
 
 int write_id_mapping(enum idtype idtype, pid_t pid, const char *buf,
@@ -443,5 +456,7 @@ extern int run_script(const char *name, const char *section, const char *script,
 extern int in_caplist(int cap, struct lxc_list *caps);
 extern int setup_sysctl_parameters(struct lxc_list *sysctls);
 extern int lxc_clear_sysctls(struct lxc_conf *c, const char *key);
+extern int setup_proc_filesystem(struct lxc_list *procs, pid_t pid);
+extern int lxc_clear_procs(struct lxc_conf *c, const char *key);
 
 #endif /* __LXC_CONF_H */
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -142,6 +142,7 @@ lxc_config_define(tty_max);
 lxc_config_define(tty_dir);
 lxc_config_define(uts_name);
 lxc_config_define(sysctl);
+lxc_config_define(proc);
 
 static struct lxc_config_t config[] = {
                                            /* REMOVE in LXC 3.0 */
@@ -243,6 +244,7 @@ static struct lxc_config_t config[] = {
 	{ "lxc.tty.max",                   false,                  set_config_tty_max,                     get_config_tty_max,                     clr_config_tty_max,                   },
 	{ "lxc.uts.name",                  false,                  set_config_uts_name,                    get_config_uts_name,                    clr_config_uts_name,                  },
 	{ "lxc.sysctl",                    false,                  set_config_sysctl,                      get_config_sysctl,                      clr_config_sysctl,                    },
+	{ "lxc.proc",                      false,                  set_config_proc,                        get_config_proc,                        clr_config_proc,                      },
 
 	/* [START]: REMOVE IN LXC 3.0 */
 	{ "lxc.pts",                       true,                   set_config_pty_max,                     get_config_pty_max,                     clr_config_pty_max,                   },
@@ -1559,6 +1561,55 @@ on_error:
 		free(sysctl_elem->value);
 		free(sysctl_elem);
 	}
+	return -1;
+}
+
+static int set_config_proc(const char *key, const char *value,
+			    struct lxc_conf *lxc_conf, void *data)
+{
+	const char *subkey;
+	struct lxc_list *proclist = NULL;
+	struct lxc_proc *procelem = NULL;
+
+	if (lxc_config_value_empty(value))
+		return clr_config_proc(key, lxc_conf, NULL);
+
+	if (strncmp(key, "lxc.proc.", sizeof("lxc.proc.") -1) != 0)
+		return -1;
+
+	subkey = key + sizeof("lxc.proc.") - 1;
+	if (*subkey == '\0')
+		return -EINVAL;
+
+	proclist = malloc(sizeof(*proclist));
+	if (!proclist)
+		goto on_error;
+
+	procelem = malloc(sizeof(*procelem));
+	if (!procelem)
+		goto on_error;
+	memset(procelem, 0, sizeof(*procelem));
+
+	procelem->filename = strdup(subkey);
+	procelem->value = strdup(value);
+
+	if (!procelem->filename || !procelem->value)
+		goto on_error;
+
+	proclist->elem = procelem;
+
+	lxc_list_add_tail(&lxc_conf->procs, proclist);
+
+	return 0;
+
+on_error:
+	free(proclist);
+	if (procelem) {
+		free(procelem->filename);
+		free(procelem->value);
+		free(procelem);
+	}
+
 	return -1;
 }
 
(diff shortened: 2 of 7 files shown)


src/lxc/confile.c
```diff
@@ -929,6 +931,24 @@ static int set_config_init_gid(const char *key, const char *value,
 	return 0;
 }
 
+static int set_config_oom_score_adj(const char *key, const char *value,
+			       struct lxc_conf *lxc_conf, void *data)
+{
+	int oom_score_adj;
+
+	if (lxc_config_value_empty(value)) {
+		lxc_conf->oom_score_adj = 0;
```
brauner: You should just call your `clr_config_oom_score_adj()` function here and return:

```
if (lxc_config_value_empty(value))
        return clr_config_oom_score_adj()`
```

src/lxc/start.c
```diff
@@ -174,6 +174,34 @@ static bool preserve_ns(int ns_fd[LXC_NS_MAX], int clone_flags, pid_t pid)
 	return false;
 }
 
+static int set_oom_score_adj(int oom_score_adj, pid_t pid)
+{
+	int ret;
+	FILE *f;
+/* 5 /proc + 21 /int_as_str + 14 /oom_score_adj + 1 \0 */
+#define OOM_PATH_LEN 45
```
brauner: Nit: Please make this `__OOM_PATH_LEN`. I usually make `#defines` in headers without `__` prefixes and `#defines` local to a file with `__`.

src/lxc/start.c
```diff
@@ -174,6 +174,34 @@ static bool preserve_ns(int ns_fd[LXC_NS_MAX], int clone_flags, pid_t pid)
...
+	char path[OOM_PATH_LEN];
+
+	ret = snprintf(path, OOM_PATH_LEN, "/proc/%d/oom_score_adj", pid);
+	if (ret < 0 || (size_t)ret >= OOM_PATH_LEN)
```
brauner: Nit: remove the `(size_t)` cast since `OOM_PATH_LEN` is an `int` anyway. :)

src/lxc/start.c
```diff
@@ -174,6 +174,34 @@ static bool preserve_ns(int ns_fd[LXC_NS_MAX], int clone_flags, pid_t pid)
...
+		return -EFBIG;
+
+	f = fopen(path, "w");
+	if (!f) {
+		ERROR("Failed to open %s", path);
```
brauner: Can you quote the path, please, aka `ERROR("Failed to open \"%s\", path)` it's something we now do with most strings in our log. :)

src/lxc/start.c
```diff
@@ -174,6 +174,34 @@ static bool preserve_ns(int ns_fd[LXC_NS_MAX], int clone_flags, pid_t pid)
...
+		return -1;
+	}
+
+	if (fprintf(f, "%d", oom_score_adj) < 0) {
```
brauner: Nit/coding-ocd: Please don't check the function call in the `if`-condition if it's not a boolean. Use
```
ret = fprintf();
if (ret < 0)
       /* handle error */
```

src/lxc/start.c
```diff
@@ -174,6 +174,34 @@ static bool preserve_ns(int ns_fd[LXC_NS_MAX], int clone_flags, pid_t pid)
...
+		ERROR("Error writing oom score");
+		fclose(f);
+		return -1;
+	}
+
+	fclose(f);
```
brauner: The multiple `fclose()` calls here can probably be replaced by a single `fclose()` right after you called `fprintf()` since you're not using the file after the `fprintf()` anymore anyway. Just makes it a little cleaner. :)

```
ret = fprintf(f, /* stuff */);
fclose(f);
if (ret < 0) {
       ERROR("Error writing oom score");
       return -1;
}
return 0;

```

src/lxc/start.c
```diff
@@ -1386,6 +1414,11 @@ static int lxc_spawn(struct lxc_handler *handler)
 		goto out_delete_net;
 	}
 
+	if (set_oom_score_adj(handler->conf->oom_score_adj, handler->pid)) {
```
brauner: Please do:

```
ret = set_oom_score_adj();
if (ret < 0) 
        /* handle error */
```

tanyifeng:
Thanks for these comments, I modified the key to 'lxc.oom.preference'， this is convenient and more readable if we want to add another similar key, such as 'lxc.oom.disable' to disable oom killer.

stgraber:
Don't really have a better suggestion since this isn't grouped under anything at the kernel level.
No-new-privs could have been put under a generic lxc.prctl namespace, but that's not the case for oom adj.

brauner:
First, sorry for the delay but we've been discussing this a little. My main problem with this approach is that we could end up with a bunch of special keys for things that are not actually a - let's call it - "module" at the kernel level. This is something which I want to avoid. Making this `lxc.oom.*` doesn't really help the situation I think.
Here's a suggestion: Instead of using `lxc.oom.*` we introduce a more generic `lxc.proc.*` configuration key similar to `lxc.sysctl.`. This key can be used to address any file under `/proc/<container-pid>/*` including `oom_score_adj`. Does that sound reasonable?
It wouldn't be that much work since you can basically just use @lifeng68's code for introducing `lxc.sysctl.*` and @Blub's code for `lxc.prlimit.*` as inspiration.

tanyifeng:
Good idea, I pushed a new commits with key `lxc.proc` which is similar to `lxc.sysctl.*`.

src/lxc/conf.h
```diff
@@ -105,6 +105,16 @@ struct lxc_sysctl {
 	char *value;
 };
 
+/*
+ * Defines a structure to configure proc filesystem at runtime.
+ * @filename : the proc filesystem will be configured without the "lxc.sysctl" prefix
```
brauner: s/lxc.sysctl/lxc.proc/

src/lxc/confile.c
```diff
@@ -1562,6 +1564,61 @@ static int set_config_sysctl(const char *key, const char *value,
...
+	char *token = "lxc.proc.";
+	struct lxc_list *proclist = NULL;
+	struct lxc_proc *procelem = NULL;
+
+	if (lxc_config_value_empty(value))
+		return clr_config_proc(key, lxc_conf, NULL);
+
+	subkey = strstr(key, token);
+	if (!subkey)
+		return -1;
```
brauner: Ah, I see you took the `set_config_cgroup()` call as inspiration. :) I don't like the `strstr()` that is used in there. This can all be done without using `strlen()` and friends. I mean, any reasonable compiler should optimize this anyway but at the same time I like to see it done in the code correctly without relying on the compiler being smart. :) So this should likely be:

```
---
 src/lxc/confile.c | 23 +++++++++--------------
 1 file changed, 9 insertions(+), 14 deletions(-)

diff --git a/src/lxc/confile.c b/src/lxc/confile.c
index 78c09f32..9df885b8 100644
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -1567,40 +1567,35 @@ on_error:
 static int set_config_proc(const char *key, const char *value,
 			    struct lxc_conf *lxc_conf, void *data)
 {
-	char *subkey;
-	char *token = "lxc.proc.";
+	const char *subkey;
 	struct lxc_list *proclist = NULL;
 	struct lxc_proc *procelem = NULL;
 
 	if (lxc_config_value_empty(value))
 		return clr_config_proc(key, lxc_conf, NULL);
 
-	subkey = strstr(key, token);
-	if (!subkey)
+	if (strncmp(key, "lxc.proc.", sizeof("lxc.proc.") -1) != 0)
 		return -1;
 
-	if (!strlen(subkey))
-		return -1;
+	subkey = key + sizeof("lxc.proc.") - 1;
 
-	if (strlen(subkey) == strlen(token))
-		return -1;
-
-	subkey += strlen(token);
+	if (*subkey != '\0')
+		return -EINVAL;
 
 	proclist = malloc(sizeof(*proclist));
 	if (!proclist)
-		goto out;
+		goto on_error;
 
 	procelem = malloc(sizeof(*procelem));
 	if (!procelem)
-		goto out;
+		goto on_error;
 	memset(procelem, 0, sizeof(*procelem));
 
 	procelem->filename = strdup(subkey);
 	procelem->value = strdup(value);
 
 	if (!procelem->filename || !procelem->value)
-		goto out;
+		goto on_error;
 
 	proclist->elem = procelem;
 
@@ -1608,7 +1603,7 @@ static int set_config_proc(const char *key, const char *value,
 
 	return 0;
 
-out:
+on_error:
 	free(proclist);
 	if (procelem) {
 		free(procelem->filename);
-- 
2.14.1
```

src/lxc/confile.c
```diff
@@ -3414,6 +3471,40 @@ static int get_config_sysctl(const char *key, char *retv, int inlen,
...
+
+	if (!retv)
+		inlen = 0;
+	else
+		memset(retv, 0, inlen);
+
+	if (!strcmp(key, "lxc.proc"))
+		get_all = true;
+	else if (!strncmp(key, "lxc.proc.", 11))
+		key += 11;
```
brauner: Please make this:

```
---
 src/lxc/confile.c | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/src/lxc/confile.c b/src/lxc/confile.c
index 9df885b8..bb5d339b 100644
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -3479,10 +3479,10 @@ static int get_config_proc(const char *key, char *retv, int inlen,
 	else
 		memset(retv, 0, inlen);
 
-	if (!strcmp(key, "lxc.proc"))
+	if (strcmp(key, "lxc.proc") == 0)
 		get_all = true;
-	else if (!strncmp(key, "lxc.proc.", 11))
-		key += 11;
+	else if (strncmp(key, "lxc.proc.", sizeof("lxc.proc.") - 1) == 0)
+		key += sizeof("lxc.proc.") - 1;
 	else
 		return -1;
 
-- 
2.14.1
```

src/lxc/start.c
```diff
@@ -1386,6 +1386,12 @@ static int lxc_spawn(struct lxc_handler *handler)
 		goto out_delete_net;
 	}
 
+	if (!lxc_list_empty(&conf->procs)) {
+		ret = setup_proc_filesystem(&conf->procs, handler->pid);
```
brauner: I like that you placed it in `lxc_spawn()`. This runs after `lxc_setup()` has already finished in the child, if I remember correctly. If so then this is fine.

src/lxc/confile.c
tanyifeng:
```
-	subkey += strlen(token);
+	if (*subkey != '\0')
+		return -EINVAL;
```
Is this should be 
```
-	subkey += strlen(token);
+	if (*subkey == '\0')
+		return -EINVAL;
```

tanyifeng:
I have fixed all problems.

brauner:
jenkins: test this please

brauner:
Thanks!

PR #2012, opened 2017-12-08
lxc: Distinguish pthread_mutex_unlock error messages

marcosps:
The same message exists in lxclock.c and cgmanager.c, so print the
filename along with the message.

Before this patch:
lxc-destroy -n u1
pthread_mutex_unlock returned:1 Operation not permitted

After this patch:
xc-destroy -n u1
lxclock.c: pthread_mutex_unlock returned:1 Operation not permitted


hallyn:
Looks good to me.

hallyn:
jenkins: test this please

PR #2015, opened 2017-12-08
hooks: add mount hook to configure access to NVIDIA GPUs

flx42:
This hook requires the nvidia-container-cli tool provided by libnvidia-container:
https://github.com/nvidia/libnvidia-container

For containers that do not have CUDA_VERSION or NVIDIA_VISIBLE_DEVICES
set in the environment, the hook will be a no-op.

To enable in the configuration file:
lxc.hook.mount = /usr/local/share/lxc/hooks/nvidia

cc @3XX0


brauner:
jenkins: test this please

hallyn:
Interesting - so this hook will actually break if/when we stop sending the hook args as argv.

flx42:
@hallyn well, it's not merged, so we can still fix our hooks. I would be fine to do something like:
```
if [ -n "${LXC_HOOK_ARGS}" ]; then
   $(getopt .... -- "${LXC_HOOK_ARGS}")
else
   $(getopt .... -- "${@}")
fi
```
But it would require agreeing on the name of the environment variable.

hooks/nvidia
```diff
@@ -0,0 +1,224 @@
...
+if [ -n "${NVIDIA_DISABLE_REQUIRE+x}" ]; then
+    if [ $(parse_bool "${NVIDIA_DISABLE_REQUIRE}") = "true" ]; then
+        CLI_DISABLE_REQUIRE="true"
+    fi
+fi
+
+if [ -z "${CLI_DEBUG}" ]; then
+    if [ "${LXC_LOG_LEVEL}" = "DEBUG" ] || [ "${LXC_LOG_LEVEL}" = "TRACE" ]; then
+        rootfs_path="${LXC_ROOTFS_PATH#*:}"
+        hookdir="${rootfs_path/%rootfs/hook}"
```
flx42: @hallyn I was wondering if this path was fine for you, it's for the log file of `nvidia-container-cli`.
If the rootfs is `dir:/a/b/c/rootfs`, we will write the log in a new directory above: `/a/b/c/hook/nvidia.log`

hallyn:
Indeed it would require that :) 

@brauner - opinion?

flx42:
By the way, what is the issue with passing the hook arguments as `$@`?

brauner:
> By the way, what is the issue with passing the hook arguments as $@?

The issue is that older liblxc versions used to pass a - by now - useless set of arguments in all cases. So they are always there when they clearly don't need to. What we want is to remove them from being passed explicitly so that uses can directly pass the arguments they want without having to swallow the default arguments before. That is just way cleaner.

brauner:
> Indeed it would require that :)
> 
> @brauner - opinion?

This is about the name of the environment variable? I'm fine with `LXC_HOOK_ARGS`? @RicardoSanchezA are you and Yoshinobu working still working on this?

flx42:
Ok, I understand better now.
So, to summarize, today if I have the following in my config file:
```
lxc.hook.mount = /usr/local/share/lxc/hooks/nvidia --debug /tmp/nvidia.log
```
The arguments passed to my hook will be:

| Arg | Value |
| --- | --- |
| $1 | --debug |
| $2 | /tmp/nvidia.log |
| $3 | mycontainer |
| $4 | lxc |
| $5 | mount |

I still want to have `$1` and `$2` be unchanged.
Since we already have `LXC_NAME=mycontainer`, if we simply add `LXC_HOOK_TYPE=mount` , then lxc can stop injecting any additional argument on the command line after my own (replacing `$4` would probably not be very useful).

brauner:
I sent patches to get rid of the args in #2017 since we need to do it soon and it seemed no one was working on this.

brauner:
> Ok, I understand better now.
So, to summarize, today if I have the following in my config file:
> ```
> lxc.hook.mount = /usr/local/share/lxc/hooks/nvidia --debug /tmp/nvidia.log
> ```
> The arguments passed to my hook will be:

> | Arg | Value |
> | --- | --- |
> | $1 | --debug |
> | $2 | /tmp/nvidia.log |
> | $3 | mycontainer |
> | $4 | lxc |
> | $5 | mount |

Fwiw, I think they will be:

| Arg | Value |
| --- | --- |
| $1 | mycontainer |
| $2 | lxc |
| $3 | mount |
| $4 | --debug |
| $5 | /tmp/nvidia.log |

> I still want to have `$1` and `$2` be unchanged.

Given the ordering you still want `$4` and `$5` be unchanged. That shouldn't be a problem. The way I currently did it in my branch is essentially:

- introduce `lxc.hook.version`.
  - This can be either set to `0` or `1`. If this is set to `0` then `$1`, `$2`, `$3` will be passed as arguments as before. If this is set to `1` then `$1`, `$2`, `$3` will be available in an environment variable `LXC_HOOK_ARGS`. All other arguments that people specify to their hooks will be passed as arguments and are unaffected by `lxc.hooks.version`. This also means that in your script you could test whether `LXC_HOOK_ARGS` exist - if you wanted to - to detect whether the old-style or new-style argument style is used.
  - Setting `lxc.hook.version` will also have effect on passing arguments to the stop hook. @Blub made it such that for each namespace the `stop` hook would receive a list of arguments of the form `mnt:/proc/<pid>/fd/<nr> net:/proc/<pid>/fd/<nr>`. When `lxc.hook.version` is set to `1` then these namespace file descriptors for the container will be set as individual environment variable (per namespace). That is they will be available as:

    ```
    LXC_MNT_NS=/proc/<pid>/fd/<nr>
    LXC_NET_NS=/proc/<pid>/fd/<nr>
    .
    .
    .
    ```

I'm currently contemplating whether it would make sense to also set `LXC_HOOK_VERSION` so that scripts could check for the hook version that is used.

brauner:
> I'm currently contemplating whether it would make sense to also set LXC_HOOK_VERSION so that scripts could check for the hook version that is used.

Without consulting the container's config that is. Hm, I think I'll push a patch for this on top of my branch and see what @hallyn thinks.

flx42:
> Fwiw, I think they will be [...]

That's not what I'm seeing. I'm seeing my own args first, then the ones injected by lxc.

>  If this is set to 1 then $1, $2, $3 will be available in an environment variable LXC_HOOK_ARGS

So, `LXC_HOOK_ARGS=mycontainer lxc mount`? Doesn't seem very convenient. It's easy to split the variable in bash. But if your hook is in C, it won't be very pleasant to have to split the string manually.
Why not just add `LXC_HOOK_TYPE=mount` then?

brauner:
On Mon, Dec 11, 2017 at 05:49:56PM +0000, Felix Abecassis wrote:
> > Fwiw, I think they will be [...]
> 
> That's not what I'm seeing. I'm seeing my own args first, then the ones injected by lxc.
> 
> >  If this is set to 1 then $1, $2, $3 will be available in an environment variable LXC_HOOK_ARGS
> 
> So, `LXC_HOOK_ARGS=mycontainer lxc mount`? Doesn't seem very convenient. It's easy to split the variable in bash. But if your hook is in C, it won't be very pleasant to have to split the string manually.
> Why not just add `LXC_HOOK_TYPE=mount` then?

Yes, that is the better solution I guess. So you'd essentially have:

LXC_HOOK_VERSION=0
LXC_NAME=<container-name>
LXC_HOOK_TYPE=<hook-type>
LXC_HOOK_SECTION=<section>

Would you prefer that?

flx42:
Yes, that seems better, from my very limited experience as a hook developer for LXC :)

hallyn:
Quoting Felix Abecassis (notifications@github.com):
> Yes, that seems better, from my very limited experience as a hook developer for LXC :)

Agreed, easier to parse.

brauner:
Did we make a call on what to do with the network scripts? Do they keep on receiving the arguments or do we pass them as env variables as well?

brauner:
The following arguments are passed to the script:  container  name  and
              config section name (net) Additional arguments depend on the config
              section employing a script hook; the following are used by the net‐
              work    system:    execution    context    (up),    network    type
              (empty/veth/macvlan/phys), Depending on  the  network  type,  other
              arguments  may  be  passed:  veth/macvlan/phys.  And finally (host-
              sided) device name.

If we pass this as env variables then we need:

    LXC_NET_SCRIPT_TYPE=up
    LXC_NET_TYPE=empty
    LXC_NET_HOST_NAME=bla

probably more.

brauner:
Here's a suggestion for the network hooks:

```
run_script(handler->name, "net", netdev->upscript, "up", "veth", veth1, (char*) NULL);
LXC_NET_TYPE=veth
LXC_NET_SCRIPT_TYPE=up
LXC_SECTION=net
LXC_NET_PARENT=netdev->link
LXC_NET_PEER=veth1

run_script(handler->name, "net", netdev->upscript, "up", "macvlan", netdev->link, (char*) NULL);
LXC_NET_TYPE=macvlan
LXC_NET_SCRIPT_TYPE=up
LXC_SECTION=net
LXC_NET_PARENT=netdev->link

run_script(handler->name, "net", netdev->upscript, "up", "phys", netdev->link, (char*) NULL);
LXC_NET_TYPE=phys
LXC_NET_SCRIPT_TYPE=up
LXC_SECTION=net
LXC_NET_PARENT=netdev->link

run_script(handler->name, "net", netdev->upscript, "up", "empty", (char*) NULL);
LXC_NET_TYPE=phys
LXC_NET_SCRIPT_TYPE=up
LXC_SECTION=empty
```

@stgraber, @hallyn does that make sense?

hallyn:
> @stgraber, @hallyn does that make sense?

Yup

brauner:
See #2026.

flx42:
Updated hook for handling `lxc.hook.version` (or the absence thereof):
https://github.com/flx42/lxc/blob/64bcf8facdd8a51aec33e8fa5a4c420ccfc1a715/hooks/nvidia#L98-L104

brauner:
Labeling this as "blocked" until #2026 is merged. :)

flx42:
We have a minor fix to do to this PR. I'll tell you when it's ready for merging :)

brauner:
@flx42, sure np. :)

flx42:
@brauner ready to merge.
And yes, it's signed off :) The test failed.

brauner:
jenkins: test this please

brauner:
@flx42, I'm running `shellcheck` now. ;)

brauner:
@flx42, can you run `shellcheck --shell sh hooks/nvidia`. We don't strictly require POSIX shell compliance but we strongly encourage it. If this is something you can do and want put the effort in that would be great. If you think it only makes sense to check for bash compliance then run `shellcheck hooks/nvidia`. :)

flx42:
Ah well, a lot of the templates are already using `bash`, I thought it was fine. It would be quite painful to do that in pure `/bin/sh`.
I did fixes in the script after running `shellcheck`, but I blame someone else for the warnings on the `in_userns` function.

brauner:
jenkins: test this please

brauner:
LGTM, will merge once the test-suite is happy.

brauner:
@flx42, @3XX0, thanks guys. :) Good work. High-fives overall. :)

PR #2016, opened 2017-12-09
lxc-oci misc improvements

flx42:


hallyn:
Can you leave a comment above getcwd? Or actually perhaps rename it cwd_from_oci or something, bc getcwd has a different implicit meaning.

Otherwise Lgtm, thanks.

flx42:
Added a comment for now, to be consistent with `getenv`. We will probably need a cleanup pass soon to reorganize this template anyway.

PR #2020, opened 2017-12-11
start: intelligently use clone() on ns sharing

brauner:
When I first solved this problem I went for a fork() + setns() + clone() model.
This works fine but has unnecessary overhead for a couple of reasons:

- doing a full fork() including copying file descriptor table and virtual
  memory
- using pipes to retrieve the pid of the second child (the actual container
  process)

This can all be avoided by being a little smart in how we employ the clone()
syscall:

- using CLONE_VM will let us get rid of using pipes since we can simply write
  to the handler because we share the memory with our parent
- using CLONE_VFORK will also let us get rid of using pipes since the execution
  of the parent is suspended until the child returns
- using CLONE_VM will not cause virtual memory to be copied
- using CLONE_FILES will not cause the file descriptor table to be copied

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -1157,43 +1157,41 @@ void resolve_clone_flags(struct lxc_handler *handler)
 		INFO("Inheriting pid namespace");
 }
 
-static pid_t lxc_fork_attach_clone(int (*fn)(void *), void *arg, int flags)
+/* Note that this function is used with clone(CLONE_VM). Some glibc versions
+ * used to reset the pid/tid to -1 when CLONE_VM was used without CLONE_THREAD.
+ * But since the memory between parent and child is shared on CLONE_VM this
+ * would invalidate the getpid() cache that glibc used to maintain and so
+ * getpid() in the child would return the parent's pid. This is all fixed in
+ * newer glibc versions where the getpid() cache is removed and the pid/tid is
+ * not reset anymore.
+ * However, if for whatever reason you - dear commiter - somehow need to get the
+ * pid of the dummy intermediate process for do_share_ns() you need to call
+ * syscall(__NR_getpid) directly. The next lxc_clone() call does not employ
+ * CLONE_VM and will be fine.
+ */
+static inline int do_share_ns(void *arg)
 {
-	int i, r, ret;
-	pid_t pid, init_pid;
+	int i, flags, ret;
 	struct lxc_handler *handler = arg;
 
-	pid = fork();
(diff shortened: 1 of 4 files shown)


src/lxc/start.c
```diff
@@ -1288,28 +1286,28 @@ static int lxc_spawn(struct lxc_handler *handler)
...
 		 * set up correctly.
 		 */
-		flags &= ~CLONE_NEWNET;
+		handler->on_clone_flags &= ~CLONE_NEWNET;
 	}
 
-	if (fork_before_clone)
-		handler->pid = lxc_fork_attach_clone(do_start, handler, flags | CLONE_PARENT);
+	if (share_ns)
+		ret = lxc_clone(do_share_ns, handler, CLONE_VFORK | CLONE_VM | CLONE_FILES);
```
hallyn: You say "Some glibc versions used to reset the pid/tid to -1 when CLONE_VM was used without CLONE_THREAD". How old are those versions?

brauner: Not super old. But it shouldn't matter since we don't care about getpid() in the intermediate process.

brauner: I remember Andrei Vagin having this issue on glibc 2.24. The pid cache got removed in 2.25 I think.

brauner: But they actually had important stuff in the child fun.

src/lxc/start.c
```diff
@@ -1157,43 +1157,41 @@ void resolve_clone_flags(struct lxc_handler *handler)
 		INFO("Inheriting pid namespace");
 }
 
-static pid_t lxc_fork_attach_clone(int (*fn)(void *), void *arg, int flags)
+/* Note that this function is used with clone(CLONE_VM). Some glibc versions
+ * used to reset the pid/tid to -1 when CLONE_VM was used without CLONE_THREAD.
+ * But since the memory between parent and child is shared on CLONE_VM this
+ * would invalidate the getpid() cache that glibc used to maintain and so
+ * getpid() in the child would return the parent's pid. This is all fixed in
+ * newer glibc versions where the getpid() cache is removed and the pid/tid is
```
hallyn: How new?

brauner: commit c579f48edba88380635ab98cb612030e3ed8691e
Author: Adhemerval Zanella <adhemerval.zanella@linaro.org>
Date:   Mon Oct 10 15:08:39 2016 -0300

    Remove cached PID/TID in clone

PR #2022, opened 2017-12-11
conf: avoid spawning unnecessary subshells

3XX0:
Signed-off-by: Jonathan Calmels <jcalmels@nvidia.com>

I'm not sure about the implications of this, this should be fine but @brauner might know better :)

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -355,10 +355,11 @@ static int run_script_argv(const char *name, const char *section,
 
 	size += strlen(hook) + 1;
 
+	size += strlen("exec");
 	size += strlen(script);
 	size += strlen(name);
 	size += strlen(section);
-	size += 3;
+	size += 4;
 
 	if (size > INT_MAX)
 		return -1;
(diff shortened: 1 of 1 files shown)


brauner:
I don't see obvious reasons to nack it. :) @stgraber?

brauner:
jenkins: test this please

src/lxc/conf.c
```diff
@@ -355,10 +355,11 @@ static int run_script_argv(const char *name, const char *section,
 
 	size += strlen(hook) + 1;
 
+	size += strlen("exec");
```
brauner: That escaped me that should be a static `sizeof("exec")`. Fixing this in #2026 since I need to touch that code anyway. :)

3XX0: If you want to use `sizeof` instead, it should be `sizeof("exec") - 1`. The compiler should be smart enough to do the right thing though

brauner: > If you want to use sizeof instead, it should be sizeof("exec") - 1.

No, not really I'm just leaveing it as `sizeof("exec")` since this automatically accounts for the added `' '`. In my branch I'm doing a `size++` after each `strlen()`. In this case I'm just going to leave the `sizeof()` and not do a `size++`. But feel free to comment on the #2026. :)

> The compiler should be smart enough to do the right thing though

Yeah, I agree but I like the code to reflect that this is information that is available statically.

3XX0: Right, cleaner indeed ;)

PR #2030, opened 2017-12-12
start: fix cgroup namespace preservation

brauner:
Prior to this patch we raced with a very short-lived init process. Essentially,
the init process could exit before we had time to record the cgroup namespace
causing the container to abort and report ABORTING to the caller when it
actually started just fine. Let's not do this.

(This uses syscall(SYS_getpid) in the the child to retrieve the pid just in case
we're on an older glibc version and we end up in the namespace sharing branch
of the actual lxc_clone() call.)

Additionally this fixes the shortlived tests. They were faulty so far and
should have actually failed because of the cgroup namespace recording race but
the ret variable used to return from the function was not correctly
initialized. This fixes it.
Furthermore, the shortlived tests used the c->error_num variable to determine
success or failure but this is actually not correct when the container is
started daemonized.

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -829,7 +829,7 @@ out_unlock:
  */
 static void do_restore(struct lxc_container *c, int status_pipe, struct migrate_opts *opts, char *criu_version)
 {
-	int fd;
+	int fd, ret;
 	pid_t pid;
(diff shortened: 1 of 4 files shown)


stgraber:
@hallyn can you take a quick look at this one?

PR #2031, opened 2017-12-13
conf.c: add relative option for lxc.mount.entry

tanyifeng:

### DIFF
--- a/doc/lxc.container.conf.sgml.in
+++ b/doc/lxc.container.conf.sgml.in
@@ -1066,6 +1066,17 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
               <option>optional</option> don't fail if mount does not work.
               <option>create=dir</option> or <option>create=file</option>
               to create dir (or file) when the point will be mounted.
+              <option>relative</option> source path is taken to be relative to
+              the mounted container root. For instance,
+             </para>
+<screen>
+dev/null proc/kcore none bind,relative 0 0
+</screen>
+             <para>
+              Will expand dev/null to ${<option>LXC_ROOTFS_MOUNT</option>}/dev/null,
+              and mount it to proc/kcore inside the container.
+             </para>
+             <para>
             </para>
           </listitem>
         </varlistentry>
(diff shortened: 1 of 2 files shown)


brauner:
@stgraber, @hallyn, please share your thoughts. :)

brauner:
And yes, technically you could do this with one of the mount hooks bug I think doing this in the config itself might be worth it.

brauner:
So, @stgraber and I have discussed this today little bit and we think the correct solution is basically what I outlined in 1.:

`lxc.mount.entry` should gain an option `relative` which indicates that a given source is relative to the containers `lxc.rootfs.mount`. This would then let you specify:

    lxc.mount.entry = proc/kcore proc/kcore none bind,ro,relative 0 0

`liblxc` would then parse `relative` and see that  `proc/kcore` is taken to be relative to `lxc.rootfs.mount` and would expand this to (on Ubuntu) `/usr/lib/x86_64-linux-gnu/lxc/proc/kcore` and thus re-mount `/proc/kcore` read-only inside the container.
This requires that the mount entry setup in `lxc_setup()` be moved *after* filling the autodev otherwise you cannot do things like:

    lxc.mount.entry = /dev/null proc/kcore none,bind,ro,relative 0 0

without an additional `lxc.mount.entry` for `/dev/null`.

@stgraber, @hallyn how does that sounds?

tanyifeng:
If we use `lxc.mount.entry`, how can we distinguish file and directories since we can not mount `/dev/null` to a file. Just like function `mask_path` do.

brauner:
> If we use `lxc.mount.entry`, how can we distinguish file and directories since we can not mount `/dev/null` to a file. Just like function `mask_path` do.

In your original patch you detected it with `ENOTDIR` and then overmounted with a `tmpfs`? I would argue that if users want to mask paths they must know the reason for hiding those paths which usually should involve knowing whether it's a file or directory. :) If we hear a lot of complaints we can discuss adding another `liblxc` specific option.

tanyifeng:
Sorry for the delay, I've changed the way of implementing option `relative` for `lxc.mount.entry` as you said.  It's a simply and reasonable way to resolve masked path and readonly path.  I'm not sure moving `setup_mount_entries` after `lxc_fill_autodev` is properly done or not. Please make sure that. Thanks.

brauner:
jenkins: test this please

brauner:
Thanks!

brauner:
@tanyifeng, if you have a minute you could send a follow-up PR that makes the new `int relative` argument to `mount_entry() ` a `bool`. :)

PR #2034, opened 2017-12-14
utils: use lxc_raw_clone() in run_command()

brauner:
run_command() is called in contexts where threads explicitly hold a lock while
fork()ing. Currently, this just affects the legacy cgmanager cgroup driver.
Here's what's happening when we use fork():

1. cgm_chown() calls cgm_dbus_connect()
2. cgm_dbus_connect() calls cgm_lock():
   Now the thread holds an explicit lock on the mutex
3. cgm_chown() calls chown_cgroup()
4. chown_cgroup() calls userns_exec_1()
5. userns_exec_1() forks with an explicit lock on the mutex being held
6. pthread_atfork() handlers get run including the prepare() handler:

        #ifdef HAVE_PTHREAD_ATFORK __attribute__((constructor))
        static void process_lock_setup_atfork(void)
        {
                pthread_atfork(cgm_lock, cgm_unlock, cgm_unlock);
        }
        #endif

   thus trying to acquire the mutex that is being explicitly held in the
   parent. If we were using recursive locks then the parent would now
   hold two locks but since I don't see us using them I guess we're
   simply getting undefined behavior.

There are multiple ways to solve this problem. They are all not very nice. One
solution is to use interposition wrapper for pthread_atfork() but that is
rather tricky since we need to have wrappers for the pthread_atfork() callbacks
and need to identify our caller so that we can make a decision whether we
should execute the callback or not. If this were a generic problem I'd say we
go for this solution but as this only affects the legacy cgmanager driver we
don't really care and I'd much rather enforce that any future code does not
take an explicit lock during a fork(). That sounds like a bad idea in the first
place. So simply switch from using fork() to clone() which does not run
pthread_atfork() handlers. If push comes to shove we might just go for doing
the clone() syscall directly via syscall(SYS_clone, ...).

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -2230,7 +2230,7 @@ int run_command(char *buf, size_t buf_size, int (*child_fn)(void *), void *args)
 		return -1;
 	}
 
-	child = fork();
+	child = lxc_raw_clone(0);
 	if (child < 0) {
 		close(pipefd[0]);
 		close(pipefd[1]);
--- a/src/lxc/namespace.c
+++ b/src/lxc/namespace.c
@@ -24,10 +24,12 @@
 #include <alloca.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <sched.h>
 #include <signal.h>
 #include <unistd.h>
 #include <sys/param.h>
 #include <sys/stat.h>
+#include <sys/syscall.h>
 #include <sys/types.h>
 
 #include "log.h"
@@ -59,8 +61,7 @@ pid_t lxc_clone(int (*fn)(void *), void *arg, int flags)
 	pid_t ret;
 
 #ifdef __ia64__
-	ret = __clone2(do_clone, stack,
-		       stack_size, flags | SIGCHLD, &clone_arg);
+	ret = __clone2(do_clone, stack, stack_size, flags | SIGCHLD, &clone_arg);
 #else
 	ret = clone(do_clone, stack  + stack_size, flags | SIGCHLD, &clone_arg);
 #endif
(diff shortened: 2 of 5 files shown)


src/lxc/utils.c
```diff
@@ -2215,11 +2215,42 @@ int lxc_unstack_mountpoint(const char *path, bool lazy)
...
+	/* Close the read-end of the pipe. */
+	close(args->pipefd[0]);
+
+	/* Redirect std{err,out} to write-end of the pipe. */
+	ret = dup2(args->pipefd[1], STDOUT_FILENO);
+	if (ret >= 0)
+		ret = dup2(args->pipefd[1], STDERR_FILENO);
+
+	/* Close the write-end of the pipe. */
+	close(args->pipefd[1]);
```
hallyn: ..  unlikely, but if pipefd[1] were 1 or 2, you would end up closing it here.

brauner:
I actually want to change this patch and use `syscall(__NR_syscall)` directly to ensure that no `pthread_atfork()` handlers are run. I'm also thinking about declaring this function with an attribute like `no_pthread_atfork` or something such that future users/developers/reviewers know right away that the context such functions are called in matters.

brauner:
@hallyn, ready for another review.

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
@hallyn, the failure on one builder is a valid failure but completely unrelated to this patch. Basically, between informing all clients that the container is now `STOPPED` and closing the command socket we have a very subtle race if the init process is short lived. We can report `STOPPED` to the caller which may call `c->start()` again and get `ECONNREFUSED` because it tried to connect to the command socket before we closed it. I'll just lock the command socket I think.

brauner:
https://github.com/lxc/lxc/pull/2039 should fix the race I mentioned before.

PR #2035, opened 2017-12-14
criu: add feature check capability

adrianreber:
For migration optimization features like pre-copy or post-copy migration the support cannot be determined by simply looking at the CRIU version. Features like that depend on the architecture/kernel/criu combination and CRIU offers a feature checking interface to query if it is supported.

This adds a LXC interface to query CRIU for those feature via the migrate() API call. For the recent pre-copy migration support in LXD this can be used to automatically detect if pre-copy migration should be used.

In addition to the existing migrate() API commands this adds a new command: MIGRATE_FEATURE_CHECK'.

The migrate_opts{} structure is extended by the member features_to_check which is a bitmask defining which CRIU features should be queried.

Currently only the querying of the features FEATURE_MEM_TRACK and FEATURE_LAZY_PAGES is supported.

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -653,6 +653,104 @@ err:
 }
 
 /*
+ * Function to check if the checks activated in 'features_to_check' are
+ * available with the current architecture/kernel/criu combination.
+ *
+ * Parameter features_to_check is a bit mask of all features that should be
+ * checked (see feature check defines in lxc/lxccontainer.h).
+ *
+ * If the return value is true, all requested features are supported. If
+ * the return value is false the features_to_check parameter is updated
+ * to reflect which features are available. '0' means no feature but
+ * also that something went totally wrong.
+ *
+ * Some of the code flow of criu_version_ok() is duplicated and maybe it
+ * is a good candidate for refactoring.
+ */
+bool __criu_check_feature(uint64_t *features_to_check)
+{
+	pid_t pid;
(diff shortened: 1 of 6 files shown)


brauner:
jenkins: test this please

src/lxc/criu.c
```diff
@@ -652,6 +652,106 @@ static void exec_criu(struct criu_opts *opts)
...
+				exit(1);
+
+			null_stdfds();
+
+			execvp("criu", args);
+			SYSERROR("Failed to exec \"criu\"");
+			exit(1);
+		}
+
+		ret = lxc_wait_for_pid_status(pid);
```
brauner: I think you should use `wait_for_pid(pid)` which will return `0` on success and `-1` on error. The function you're calling right now returns the status that `waitpid(pid, &status, ...)` stores in `status` and this needs to be inspected with appropriate macros.

brauner:
@tych0, do you have an opinion on this? :)

tych0:
Looks reasonable to me. Another option would be to just try to migrate using the feature, and then fail if it doesn't work ("ask for forgiveness not permission"), but if it seems better to do this, that's fine.

stgraber:
Looks reasonable to me.

adrianreber:
Force pushed with ```wait_for_pid(pid)```.

brauner:
jenkins: test this please

PR #2036, opened 2017-12-14
commands: fix race when open()/close() cmd socket

brauner:
When we report STOPPED to a caller and then close the command socket it is
technically possible - and I've seen this happen on the test builders - that a
container start() right after a wait() will receive ECONNREFUSED because it
called open() before we close(). So let's lock setting the STOPPED state.


brauner:
Damn, we're not going to be in the same thread... That's annoying.

PR #2039, opened 2017-12-14
commands: fix race when open()/close() cmd socket

brauner:
When we report STOPPED to a caller and then close the command socket it is
technically possible - and I've seen this happen on the test builders - that a
container start() right after a wait() will receive ECONNREFUSED because it
called open() before we close(). So for all new state clients simply close the
command socket. This will inform all state clients that the container is
STOPPED and also prevents a race between a open()/close() on the command socket
causing a new process to get ECONNREFUSED because we haven't yet closed the
command socket.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -753,12 +753,22 @@ void lxc_fini(const char *name, struct lxc_handler *handler)
 
 	cgroup_destroy(handler);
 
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

brauner:
@hallyn, this should fix the race.

PR #2043, opened 2017-12-16
tree-wide: use _exit() in child

brauner:
- When fork()ing we shouldn't run any exit() handlers as they can mess with the
  parent state.
- exit() is not thread-safe so replace it with _exit() inside library
  codepaths.


hallyn:
Jenkins: test this please

brauner:
Should be good to go.

PR #2044, opened 2017-12-16
tools: add UNPRIVILEGED field in fancy output mode

brauner:

### DIFF
--- a/src/lxc/tools/lxc_ls.c
+++ b/src/lxc/tools/lxc_ls.c
@@ -71,6 +71,7 @@ struct ls {
 	double swap;
 	bool autostart;
 	bool running;
+	bool unprivileged;
 };
 
 /* Keep track of field widths for printing. */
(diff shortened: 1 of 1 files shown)


brauner:
@hallyn, might not be worth it so feel free to nack it. Just found it helpful.

hallyn:
Do you think this might mess up any people's scripts?  Shouldn't so long as it's appended as a new last column...

What about instead of 'unprivileged', showing the value to which container root is mapped?  Seems more useful, but maybe less new-user-friendly.

brauner:
Ok, `aa_change_profile()` writes to `/proc`. Sigh...

hallyn:
Ok technically since we (aim to) support containers without a root uid mapping, it seems useful to separately show 'unprivileged' and 'container root mapping'.  Illl merge this.

brauner:
Oh sorry, I would've been happy to merge this I was just hadn't come around to it yet.

PR #2049, opened 2017-12-18
start: reap intermediate process

brauner:
When we inherit namespaces we need to reap the attaching process.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -1399,14 +1399,28 @@ static int lxc_spawn(struct lxc_handler *handler)
 	/* The cgroup namespace gets unshare()ed not clone()ed. */
 	handler->on_clone_flags &= ~CLONE_NEWCGROUP;
 
-	if (share_ns)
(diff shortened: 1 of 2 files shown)


src/lxc/start.c
```diff
@@ -1399,14 +1399,28 @@ static int lxc_spawn(struct lxc_handler *handler)
...
 		handler->pid = lxc_clone(do_start, handler, handler->on_clone_flags);
-	if (handler->pid < 0 || ret < 0) {
-		SYSERROR("Failed to clone a new set of namespaces.");
+	}
+	if (handler->pid < 0) {
+		SYSERROR(LXC_CLONE_ERROR);
 		goto out_delete_net;
 	}
+
 	TRACE("Cloned child process %d", handler->pid);
```
hallyn: So...  in the share_ns case, this will tell us that the child process is pid 0?

hallyn: Oh, nm, I see.

brauner: Hm, why, `CLONE_VM` is used.

Issue #1678, opened 2017-07-05
archlinux in unprivileged lxc cannot start


ss1h2a3tw:
Seems to be that current mainline systemd have some new operations that unprivileged lxc
will blocks.
It works fine in privileged lxc.
Can confirm that systemd 232-8 still works well
# Required information

 * Distribution: archlinux
 * The output of
   * `lxc-start --version` 2.0.8
   * `lxc-checkconfig` all green
   * `uname -a` 4.11.7-1-userns (the linux-userns in AUR)
   * `cat /proc/self/cgroup`                                      
> 10:memory:/lxc                                                                      
9:perf_event:/lxc                         
8:freezer:/lxc  
7:net_cls,net_prio:/lxc                                                
6:devices:/lxc                                              
5:cpuset:/lxc                                              
4:blkio:/lxc                                                   
3:pids:/lxc                                           
2:cpu,cpuacct:/lxc                                           
1:name=systemd:/user.slice/user-1000.slice/session-c1.scope
0::/user.slice/user-1000.slice/session-c1.scope
   * `cat /proc/1/mounts`
>proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
sys /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
dev /dev devtmpfs rw,nosuid,relatime,size=61761116k,nr_inodes=15440279,mode=755 0 0
run /run tmpfs rw,nosuid,nodev,relatime,mode=755 0 0
/dev/sda2 / btrfs rw,relatime,space_cache,subvolid=257,subvol=/root 0 0
securityfs /sys/kernel/security securityfs rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /dev/shm tmpfs rw,nosuid,nodev 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000 0 0
tmpfs /sys/fs/cgroup tmpfs rw,mode=755 0 0
cgroup /sys/fs/cgroup/unified cgroup2 rw,nosuid,nodev,noexec,relatime 0 0
cgroup /sys/fs/cgroup/systemd cgroup rw,nosuid,nodev,noexec,relatime,xattr,name=systemd 0 0
pstore /sys/fs/pstore pstore rw,nosuid,nodev,noexec,relatime 0 0
[...]

PR #2055, opened 2017-12-20
cgfsng: Add new macro to print errors

marcosps:
At this point, macros such DEBUG or ERROR does not take effect because
this code is called from cgroup_ops_init(cgroup.c), which runs with
__attribute__((constructor)), before any log level is set form any tool
like lxc-start, so these messages are lost.

For now on, use the same LXC_DEBUG_CGFSNG environment variable to
control these messages.


marcosps:
This helped me to understand better what happens with bug #1678 which also happens in Fedora 27.

brauner:
jenkins: test this please

PR #2057, opened 2017-12-22
start: simplify cgroup namespace preservation

brauner:
Since we are now dumpable we can open /proc/<child-pid>/ns/cgroup so let's
avoid the overhead of sending around fds.


brauner:
@hallyn, one more bugfix. :)

brauner:
Should be fairly uncontroversial.

PR #2065, opened 2018-01-01
lxccontainer: revert set_running_config_item()

brauner:
- As discussed we will have a proper API extension that will allow updating
  various parts of a running container. The prior approach wasn't a good idea.

- Revert this is not a problem since we haven't released any version with the
  set_running_config_item() API extension.

- I'm not simply reverting so that master users can still call into new
  liblxc's without crashing the container. This is achieved by keeping the
  commands callback struct member number identical.


brauner:
@stgraber, this should be a fairly uncontroversial merge. :)

PR #2066, opened 2018-01-02
Support configurations without root mapping

### LINKED ISSUE #2033: [RFE] Support configurations without root mapping
Let's say I'm only interested in running unprivileged application containers.
Also, I don't want to be `root` inside the container since I want to bind-mount my `$HOME` folder and have everything work similarly as outside the container as a user (i.e., it's just a way to just swap the distro I'm using).

I could achieve this by having the following in `default.conf`:
```
lxc.idmap = u 1000 1000 1
lxc.idmap = g 1000 1000 1

lxc.init.uid = 1000
lxc.init.gid = 1000
```
However, there seems to be a bunch of places in the code where it is rejected. For example, this is the first problem I'm having:
```
$ lxc-create -t download -n xenial -- --dist ubuntu --release xenial --arch amd64
lxc-create: xenial: conf.c: chown_mapped_root: 2825 No uid mapping for container root
```
I've quickly removed all these tests in the code and it seems to work, but it's a pretty huge change that touches many parts of the code, so I'll let @brauner handle it ;). If you think that's a valid use case, of course.

brauner:
Closes #2033.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1267,25 +1267,41 @@ next:
 	return r;
 }
 
+struct generic_userns_exec_data {
+	struct cgfsng_handler_data *d;
+	struct lxc_conf *conf;
+	uid_t origuid; /* target uid in parent namespace */
+	char *path;
+};
+
 static int rmdir_wrapper(void *data)
 {
-	char *path = data;
+	struct generic_userns_exec_data *arg = data;
+	uid_t nsuid = (arg->conf->root_nsuid_map != NULL) ? 0 : arg->conf->init_uid;
+	gid_t nsgid = (arg->conf->root_nsgid_map != NULL) ? 0 : arg->conf->init_gid;
 
-	if (setresgid(0,0,0) < 0)
+	if (setresgid(nsgid, nsgid, nsgid) < 0)
 		SYSERROR("Failed to setgid to 0");
-	if (setresuid(0,0,0) < 0)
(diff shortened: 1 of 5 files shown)


brauner:
@flx42, does that cover your use-case?

flx42:
@brauner I'll test this patch set ASAP, thanks!

flx42:
It will still require a root mapping for `lxc-create`, right?
May I ask which configuration you used for the container? I was not able to make it work, will dig more into it tomorrow.

brauner:
On Wed, Jan 03, 2018 at 03:26:56AM +0000, Felix Abecassis wrote:
> It will still require a root mapping for `lxc-create`, right?

Yes, that's what @hallyn was suggesting. @hallyn, opinions on that
front? :)

> May I ask which configuration you used for the container? I was not able to make it work, will dig more into it tomorrow.

To run the container itself I just added:

lxc.idmap = u 1000 1000 1
lxc.idmap = g 1000 1000 1
lxc.init.uid = 1000
lxc.init.gid = 1000
lxc.hook.mount =

To the container's config file. I suspect that you didn't use the
"lxc.hook.mount=" option. This is required to make sure that the lxcfs
hook is not run. The lxcfs hook uses mount with the --types option which
is not allowed unless you're root in the respective user namespace.
Otherwise it should work. Does that work for you?

brauner:
@flx42, here's a POC:
[asciicast](https://asciinema.org/a/155311)

PR #2067, opened 2018-01-03
conf: write "deny" to /proc/[pid]/setgroups

brauner:
When fully unprivileged users run a container that only maps their own {g,u}id
and they do not have access to setuid new{g,u}idmap binaries we will write the
idmapping directly. This however requires us to write "deny" to
/proc/[pid]/setgroups otherwise any write to /proc/[pid]/gid_map will be
denied.

On a sidenote, this patch enables fully unprivileged containers. If you now set
lxc.net.[i].type = empty no privilege whatsoever is required to run a container.

Enhances #2033.

Cc: Felix Abecassis <fabecassis@nvidia.com>
Cc: Jonathan Calmels <jcalmels@nvidia.com>

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1284,7 +1284,7 @@ static int rmdir_wrapper(void *data)
 		SYSERROR("Failed to setgid to 0");
 	if (setresuid(nsuid, nsuid, nsuid) < 0)
 		SYSERROR("Failed to setuid to 0");
-	if (setgroups(0, NULL) < 0)
+	if (setgroups(0, NULL) < 0 && errno != EPERM)
 		SYSERROR("Failed to clear groups");
 
 	return cgroup_rmdir(arg->path);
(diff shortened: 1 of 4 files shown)


flx42:
> On a sidenote, this patch enables fully unprivileged containers. If you now set
lxc.net.[i].type = empty no privilege whatsoever is required to run a container.

This is great, thanks!

src/lxc/cgroups/cgfsng.c
```diff
@@ -1284,7 +1284,7 @@ static int rmdir_wrapper(void *data)
 		SYSERROR("Failed to setgid to 0");
 	if (setresuid(nsuid, nsuid, nsuid) < 0)
 		SYSERROR("Failed to setuid to 0");
-	if (setgroups(0, NULL) < 0)
+	if (setgroups(0, NULL) < 0 && errno != EPERM)
```
hallyn: So we'll ignore failures to write due to eperm.  Is that ok?

brauner: It should only ever mean that there's a "deny" in /proc/[pid]/setgroups but I don't want to open this and read it every time. But we could.

PR #2075, opened 2018-01-05
lxc_get_version() should show the "-devel" suffix

stgraber:
As this is actively looked for by go-lxc and LXD.


stgraber:
Should be good to go

PR #2077, opened 2018-01-08
Add return check for 'lxc_cmd_get_name'

lifeng68:
If 'lxc_cmd_get_name' failed and return with NULL, this will cause a segment fault when calling ‘add_to_array’.


brauner:
jenkins: test this please

brauner:
Thanks!

PR #2085, opened 2018-01-16
[console] return wrong console logsize

duguhaotian:
get_config_console_logsize want console.buffer_size not c->autodev


brauner:
jenkins: test this please

PR #2086, opened 2018-01-16
lxc-alpine: allow retaining sys_ptrace per container

kunkku:
It would be convenient to be able to toggle sys_ptrace on a per container basis. Removing the option from the common configuration file would have an effect on deployed containers, though.

Any better ideas to achieve this? Most of the templates do not drop this capability, so there is no good precedent.


brauner:
@kunkku, hm, you mean when you restart the container? Hm, not really. I can see the advantage of toggling `sys_ptrace` per container.

brauner:
There's no real advantage to disallowing ptrace. The only scenario where this is relevant is when you attach to the container and you ptrace the attached process but we have made this pretty secure. We fixed a CVE for this a long while back. So I think this is fine.

PR #2089, opened 2018-01-17
lxccontainer: restore blocking wait()

### LINKED ISSUE #2027: 'lxc-wait' command returns error code immediate if container is stopped
Hi,

$subj

[$]> lxc-start --version                                                                                                                     
2.1.1

STR:
-stop container
-run lxc-wait -n \<container name\> --state=RUNNING

Here is an example
```
[#]> lxc-info -sn container1
State:          STOPPED
[$]> lxc-wait -n container1 --state=RUNNING --timeout=600
[$]> echo $?
1
```

The issue is not reproducible on lxc-2.0.X. Also 'lxc-wait' command works for other statuses if the container is started. Please advice.

brauner:
Closes #2027.
Closes lxc/go-lxc#98.

### DIFF
--- a/src/lxc/freezer.c
+++ b/src/lxc/freezer.c
@@ -32,45 +32,57 @@
 #include <sys/param.h>
 
 #include "error.h"
-#include "state.h"
-#include "monitor.h"
 #include "log.h"
 #include "lxc.h"
+#include "monitor.h"
+#include "parse.h"
+#include "state.h"
 
 lxc_log_define(lxc_freezer, lxc);
 
 lxc_state_t freezer_state(const char *name, const char *lxcpath)
 {
+	int ret;
 	char v[100];
-	if (lxc_cgroup_get("freezer.state", v, 100, name, lxcpath) < 0)
+
+	ret = lxc_cgroup_get("freezer.state", v, 100, name, lxcpath);
+	if (ret < 0)
 		return -1;
 
-	if (v[strlen(v)-1] == '\n')
-		v[strlen(v)-1] = '\0';
+	v[99] = '\0';
+	v[lxc_char_right_gc(v, strlen(v))] = '\0';
+
 	return lxc_str2state(v);
 }
 
 static int do_freeze_thaw(int freeze, const char *name, const char *lxcpath)
 {
+	int ret;
 	char v[100];
 	const char *state = freeze ? "FROZEN" : "THAWED";
 
-	if (lxc_cgroup_set("freezer.state", state, name, lxcpath) < 0) {
-		ERROR("Failed to freeze %s:%s", lxcpath, name);
+	ret = lxc_cgroup_set("freezer.state", state, name, lxcpath);
+	if (ret < 0) {
+		ERROR("Failed to freeze %s", name);
 		return -1;
 	}
-	while (1) {
-		if (lxc_cgroup_get("freezer.state", v, 100, name, lxcpath) < 0) {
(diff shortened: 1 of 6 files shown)


src/lxc/state.c
```diff
@@ -115,13 +115,26 @@ extern int lxc_wait(const char *lxcname, const char *states, int timeout,
...
+		state = lxc_cmd_sock_get_state(lxcname, lxcpath, s, timeout);
+		if (state >= 0)
+			break;
+
+		if (errno != ECONNREFUSED) {
+			SYSERROR("Failed to receive state from monitor");
+			return -1;
+		}
+
+		sleep(1);
```
hallyn: Shouldn't the sleep happen after the timeout check?

src/lxc/freezer.c
```diff
@@ -32,45 +32,55 @@
...
 	char v[100];
-	if (lxc_cgroup_get("freezer.state", v, 100, name, lxcpath) < 0)
+
+	ret = lxc_cgroup_get("freezer.state", v, 100, name, lxcpath);
+	if (ret < 0)
 		return -1;
 
-	if (v[strlen(v)-1] == '\n')
-		v[strlen(v)-1] = '\0';
+	v[lxc_char_right_gc(v, strlen(v))] = '\0';
```
hallyn: if v happens to not be null-terminated, i.e. v is 100 'a's, this will write a \0 to v[100] which is invalid, right?

(near as I can tell v will be filled in by lxc_read_from_file which will read 100 bytes, without null-terminating)

src/lxc/freezer.c
```diff
@@ -32,45 +32,55 @@
...
+			ERROR("Failed to get freezer state of %s", name);
 			return -1;
 		}
-		if (v[strlen(v)-1] == '\n')
-			v[strlen(v)-1] = '\0';
-		if (strncmp(v, state, strlen(state)) == 0) {
-			if (name)
-				lxc_monitor_send_state(name, freeze ? FROZEN : THAWED, lxcpath);
+
+		v[lxc_char_right_gc(v, strlen(v))] = '\0';
```
hallyn: (Same here as above - you could theoretically write to v[100])

src/lxc/freezer.c
```diff
@@ -76,8 +78,9 @@ static int do_freeze_thaw(int freeze, const char *name, const char *lxcpath)
 		v[lxc_char_right_gc(v, strlen(v))] = '\0';
 
 		ret = strncmp(v, state, strlen(state));
-		if (ret != 0) {
-			lxc_monitor_send_state(name, freeze ? FROZEN : THAWED, lxcpath);
+		if (ret == 0) {
```
hallyn: ...  why is inverting the check for strncmp return value and then doing the same thing valid here?  Was it wrong before?

src/lxc/state.c
brauner: fixed

src/lxc/freezer.c
brauner: fixed

src/lxc/freezer.c
brauner: fixed

src/lxc/freezer.c
brauner: This looks like a comment on an old version?

brauner: Ah no, you were right. My diff was wrong.

Issue #2088, opened 2018-01-17
One question about rotate of console log file


duguhaotian:
https://github.com/lxc/lxc/blob/e7d7dd73d2d16da7acccf16c22c4cb3c1af594eb/src/lxc/commands.c#L1026

Now, if i want to rotate my console log file, I need to call API **lxc_cmd_console_log**. 
This implements, if i want to limit my log file size. I need **a while loop** to check file size, and if log file size is big than limit, then i  call this API to rotate my log file.
Some scenes, I can not limit my log files. For example, when i check size of log file, container writing large logs. And then when i call API to rotate, the real size of log file, maybe is large than my limit.

And, ringbuffer, now , like log file. Better than log file, but just can limit log use disk space.

The big problem is **Now, we can not auto rotate!!!**

Why do not implement rotate of  console log file  at **lxc_console_cb_con** function.
Like below:

`
 int lxc_console_cb_con(int fd, uint32_t events, void *data, struct lxc_epoll_descr *descr)
 {

    .....   ........
    if (fd == console->peer)
        w = lxc_write_nointr(console->master, buf, r);

    w_rbuf = w_log = 0;
    if (fd == console->master) {
        /* write to peer first */
        if (console->peer >= 0)
            w = lxc_write_nointr(console->peer, buf, r);

        /* write to console ringbuffer */
        if (console->buffer_size > 0)
            w_rbuf = lxc_ringbuf_write(&console->ringbuf, buf, r);

        /* write to console log */
        if (console->log_fd >= 0) {
             /* STEP1, check log file size*/
             /* STEP2, if large than limit, begin rotate*/
             w_log = lxc_write_nointr(console->log_fd, buf, r);
        }
    }
    ..... ......

 }
`

Maybe,  we can user ringbuffer to replace console log file.
First, set ringbuffersize as limit of one log file size. For example, we limit one log file is 8KB, and then we set ringbuffersize is 8KB.
Second, if ringbuffer is full, trigger that write buffer data into new log file. Check number of log files, if number bigger than limit of log file number, and then trigger that rotate log file, or just delete the oldest one.

`
 int lxc_console_cb_con(int fd, uint32_t events, void *data, struct lxc_epoll_descr *descr)
 {

    .....   ........
    if (fd == console->peer)
        w = lxc_write_nointr(console->master, buf, r);

    w_rbuf = w_log = 0;
    if (fd == console->master) {
        /* write to peer first */
        if (console->peer >= 0)
[...]

PR #2091, opened 2018-01-18
Introducing lxc.console.buffer.rotate

duguhaotian:
#2088  add lxc.console.buffer.rotate

### DIFF
--- a/doc/lxc.container.conf.sgml.in
+++ b/doc/lxc.container.conf.sgml.in
@@ -834,6 +834,20 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
         <varlistentry>
           <term>
+            <option>lxc.console.buffer.rotate</option>
+          </term>
+          <listitem>
+            <para>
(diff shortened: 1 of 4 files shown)


duguhaotian:
@brauner I have two doubts.
First , now ,my implements, will let buffer log file increase, no limit. Do we need limit this file size less than  **lxc.console.buffer.size** ???
Second, **lxc.console.buffer.rotate** now is 0 or 1. Do we need let it big than 1???

duguhaotian:
And this PR,  will make **write_logfile** function in **lxc_cmd_console_log_callback** meaningless. Because rotate always to do this when ringbuffer is full.

duguhaotian:
Hi, is this patch necessary ? @brauner

brauner:
I'm thinking about a different design for this but I want to discuss this with the other maintainers first. Basically, I'd like to introduce `lxc.console.size` and drop `lxc.console.buffer.logfile `entirely.

brauner:
I'm working on a solution to this.

brauner:
Closing in favor of https://github.com/lxc/lxc/pull/2167 .

PR #2092, opened 2018-01-18
Create console when the rootfs is NULL

lifeng68:
Steps to reproduce the problem：

1. Create a container with -t none
`lxc-create -n test1 -t none `
2. Modify the config of container test1 as follow
```
$ cat /var/lib/lcrd/lcr/test1/config
lxc.init_cmd = echo test
lxc.console.logfile = /home/out.log
```

3.Start the container test1,and check the output file `/home/out.log`, It is empty.

We should set the config` lxc.autodev = 0`, if we want to create console when the rootfs is NULL. Because so that we could see the allocated /dev/pts/x device and use it to bind mount the /dev/console.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1569,7 +1569,7 @@ static int lxc_setup_dev_console(const struct lxc_rootfs *rootfs,
 	if (console->path && !strcmp(console->path, "none"))
 		return 0;
 
-	ret = snprintf(path, sizeof(path), "%s/dev/console", rootfs->mount);
+	ret = snprintf(path, sizeof(path), "%s/dev/console", rootfs->path ? rootfs->mount : "");
 	if (ret < 0 || (size_t)ret >= sizeof(path))
 		return -1;
 
@@ -1583,12 +1583,6 @@ static int lxc_setup_dev_console(const struct lxc_rootfs *rootfs,
 			return -ret;
 		} else {
 			DEBUG("cleared all (%d) mounts from \"%s\"", ret, path);
-		}
-
-		ret = unlink(path);
-		if (ret < 0) {
-			SYSERROR("error unlinking %s", path);
-			return -errno;
 		}
 	}
 
@@ -1610,7 +1604,7 @@ static int lxc_setup_dev_console(const struct lxc_rootfs *rootfs,
 		return -errno;
 	}
 
-	if (safe_mount(console->name, path, "none", MS_BIND, 0, rootfs->mount) < 0) {
+	if (safe_mount(console->name, path, "none", MS_BIND, 0, rootfs->path ? rootfs->mount : "") < 0) {
 		ERROR("failed to mount '%s' on '%s'", console->name, path);
 		return -1;
 	}
@@ -1623,11 +1617,14 @@ static int lxc_setup_ttydir_console(const struct lxc_rootfs *rootfs,
 				    const struct lxc_console *console,
 				    char *ttydir)
 {
-	int ret;
+	int ret, fd;
 	char path[MAXPATHLEN], lxcpath[MAXPATHLEN];
 
+	if (console->path && !strcmp(console->path, "none"))
+		return 0;
+
 	/* create rootfs/dev/<ttydir> directory */
-	ret = snprintf(path, sizeof(path), "%s/dev/%s", rootfs->mount, ttydir);
+	ret = snprintf(path, sizeof(path), "%s/dev/%s", rootfs->path ? rootfs->mount : "", ttydir);
 	if (ret < 0 || (size_t)ret >= sizeof(path))
 		return -1;
 
@@ -1638,7 +1635,7 @@ static int lxc_setup_ttydir_console(const struct lxc_rootfs *rootfs,
 	}
  	DEBUG("Created directory for console and tty devices at \"%s\"", path);
 
-	ret = snprintf(lxcpath, sizeof(lxcpath), "%s/dev/%s/console", rootfs->mount, ttydir);
+	ret = snprintf(lxcpath, sizeof(lxcpath), "%s/dev/%s/console", rootfs->path ? rootfs->mount : "", ttydir);
 	if (ret < 0 || (size_t)ret >= sizeof(lxcpath))
 		return -1;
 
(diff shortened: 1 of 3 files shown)


brauner:
@lifeng68, I think it was done because we didn't bind-mount t the pty that is created via `lxc_console_create()` over `/dev/console` because `/dev/console` is the host's character device. But I don't see a real problem with allowing this. If users want to use the host's `/dev/console` in this case they could just set `lxc.console.path = none`.
However, what you need to do is to adapt `lxc_setup_console()` in `conf.c` to not touche `/dev/console` when the rootfs is shared. That is, it shouldn't try to remove `/dev/console` when it already exists. If the container is created with a new mount namespace it should just unmount all mounts on top of `/dev/console` but  never delete `/dev/console`.

lifeng68:
@brauner , I have modified the ` lxc_setup_console()` to bind mount `/dev/console` instead of unlink it.
But i found that we should set the config `lxc.autodev = 0`. if we want to create console when the rootfs is NULL. Because so that we could see the allocated `/dev/pts/x` device and use it to bind mount the `/dev/console`.

src/lxc/start.c
```diff
@@ -472,7 +472,7 @@ int lxc_set_state(const char *name, struct lxc_handler *handler,
 int lxc_poll(const char *name, struct lxc_handler *handler)
 {
 	int ret;
-	bool has_console = (handler->conf->rootfs.path != NULL);
+	struct lxc_console *console = &handler->conf->console;
```
brauner: Hm, shouldn't this still check whether `strcmp(handler->conf->console->rootfs.path, "none") != 0`?

src/lxc/conf.c
```diff
@@ -1569,7 +1569,7 @@ static int lxc_setup_dev_console(const struct lxc_rootfs *rootfs,
 	if (console->path && !strcmp(console->path, "none"))
 		return 0;
 
-	ret = snprintf(path, sizeof(path), "%s/dev/console", rootfs->mount);
+	ret = snprintf(path, sizeof(path), "%s/dev/console", rootfs->path ? rootfs->mount : "");
```
brauner: Could you please perform this check atop the function once:

    char rootfs_path = rootfs->path ? rootfs->mount : "";

and then just use `rootfs_path` everywhere, please?

src/lxc/conf.c
```diff
@@ -1650,110 +1647,57 @@ static int lxc_setup_ttydir_console(const struct lxc_rootfs *rootfs,
...
 
-	/* create symlink from rootfs /dev/console to '<ttydir>/console' */
-	ret = snprintf(lxcpath, sizeof(lxcpath), "%s/console", ttydir);
-	if (ret < 0 || (size_t)ret >= sizeof(lxcpath))
-		return -1;
-
-	ret = unlink(path);
-	if (ret && errno != ENOENT) {
-		SYSERROR("error unlinking %s", path);
+	if (chmod(console->name, S_IXUSR | S_IXGRP | S_IXOTH)) {
```
brauner: That's probably a line from me but could you please do:
```
ret = chmod(console->name, S_IXUSR | S_IXGRP | S_IXOTH);
if (ret < 0) {
        /* handle error */
}
```

src/lxc/conf.c
```diff
@@ -1650,110 +1647,57 @@ static int lxc_setup_ttydir_console(const struct lxc_rootfs *rootfs,
...
-		SYSERROR("failed to create symlink for console from \"%s\" to \"%s\"", lxcpath, path);
+	/* bind mount console->name to '/dev/<ttydir>/console' */
+	if (safe_mount(console->name, lxcpath, "none", MS_BIND, 0, rootfs->path ? rootfs->mount : "") < 0) {
+		ERROR("failed to mount '%s' on '%s'", console->name, lxcpath);
 		return -1;
 	}
+	DEBUG("mounted \"%s\" onto \"%s\"", console->name, lxcpath);
+
+	/* bind mount '/dev/<ttydir>/console'  to '/dev/console'  */
+	if (safe_mount(lxcpath, path, "none", MS_BIND, 0, rootfs->path ? rootfs->mount : "") < 0) {
```
brauner: ret = safe_mount(lxcpath, path, "none", MS_BIND, 0, rootfs->path ? rootfs->mount : "");

src/lxc/conf.c
```diff
@@ -1650,110 +1647,57 @@ static int lxc_setup_ttydir_console(const struct lxc_rootfs *rootfs,
...
+	if (chmod(console->name, S_IXUSR | S_IXGRP | S_IXOTH)) {
+		SYSERROR("failed to set mode '0%o' to '%s'", S_IXUSR | S_IXGRP | S_IXOTH, console->name);
 		return -errno;
 	}
 
-	ret = symlink(lxcpath, path);
-	if (ret < 0) {
-		SYSERROR("failed to create symlink for console from \"%s\" to \"%s\"", lxcpath, path);
+	/* bind mount console->name to '/dev/<ttydir>/console' */
+	if (safe_mount(console->name, lxcpath, "none", MS_BIND, 0, rootfs->path ? rootfs->mount : "") < 0) {
```
brauner: ret = safe_mount(console->name, lxcpath, "none", MS_BIND, 0, rootfs->path ? rootfs->mount : "");

src/lxc/conf.c
lifeng68: Fixed

src/lxc/conf.c
lifeng68: Fixed

src/lxc/conf.c
lifeng68: Fixed

src/lxc/conf.c
lifeng68: Fixed

src/lxc/start.c
lifeng68: Fixed

lifeng68:
@brauner 
`One questions: you're using a bind-mount for /dev/console when /dev/<ttydir>/console is requested instead of a symlink. Any particular reason for that and do we suspect breakage because of that?`
If the containers share the rootfs with the host, the symlink is not isolated by mount namespace, so If we start multiple containers, the `/dev/console` will be covered.
 
`Could you please start all error strings with a capital letter? This is a convention we have been using for a while now. The codebase is slowly adapting to this.`
I fix the codes related to this PR in this patch, i can send another patch to fix the other codes in `conf.c`

brauner:
>> Could you please start all error strings with a capital letter? This is a convention we have been using for a while now. The codebase is slowly adapting to this.
>
> I fix the codes related to this PR in this patch, i can send another patch to fix the other codes in conf.c

I mean you can if you want to but I was just referring to your own commit. :)

brauner:
jenkins: test this please

PR #2094, opened 2018-01-19
Propagate exit code for app containers

tych0:
Unfortunately, liblxc didn't do a very good job of keeping track of the exit codes of the tasks it spawned. Here's a series that addresses that. See patch notes for explanations on individual patches, hopefully they present clear arguments, and this will be mostly uncontroversial :)

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -1823,7 +1823,7 @@ int __lxc_start(const char *name, struct lxc_handler *handler,
 	}
 
 	lxc_monitor_send_exit_code(name, status, handler->lxcpath);
-	err =  lxc_error_set_and_log(handler->pid, status);
+	lxc_error_set_and_log(handler->pid, status);
 
 out_fini:
 	lxc_delete_network(handler);
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -323,6 +323,27 @@ static int signal_handler(int fd, uint32_t events, void *data,
 	if (ret == 0 && info.si_pid == hdlr->pid)
 		hdlr->init_died = true;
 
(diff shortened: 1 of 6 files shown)


src/lxc/start.c
```diff
@@ -323,6 +323,27 @@ static int signal_handler(int fd, uint32_t events, void *data,
...
+		case CLD_STOPPED:
+			hdlr->exit_status = info.si_status << 8 | 0x7f;
+			break;
+		case CLD_CONTINUED:
+			/* Huh? The waitid() told us it's dead *and* continued? */
+			WARN("Init %d dead and continued?", hdlr->pid);
+			hdlr->exit_status = 1;
+			break;
+		default:
+			ERROR("Unknown si_code: %d", hdlr->init_died);
```
hallyn: But should you still set hdlr->exit_status here?

hallyn:
What might be neat (maybe I'll write it) woudl be a testcase that actually spawns containers designed to exit(0), exit(1), ..., and be killed with various signals, and verify that lxc-execute exits the way we want.

src/lxc/start.c
hallyn: I'll wait for a reply to this before I merge, otherwise looks good.

tych0: Yeah, I'll set it to 1 so someone will see it and know to look for something weird.

tych0:
Ok, should be good to go. I can look at writing some testcases tomorrow if you want.

brauner:
Thanks guys! Uhhh, and test-cases would be welcome indeed! :)
Note that moving forward we this repo should never use the `lxc-*` tools in test-scripts since they will be moved to a separate repo. :)

PR #2097, opened 2018-01-19
storage: treat return value from ops->destroy as int

mikemccracken:
r->ops->destroy() returns an int, -1 on error.
When assigned to a bool, this becomes true and hides errors.

ret will always be true - on error, it's -1 - a true bool - or it is zero and thus set to true.

Signed-off-by: Michael McCracken <mikmccra@cisco.com>


tych0:
+1, confusing but legit, I thought through it too :)

PR #2113, opened 2018-01-27
seccomp.c: Use return instead of attribution and return

marcosps:


brauner:
jenkins: test this please

PR #2119, opened 2018-02-01
lsm: do not #ifdefine

brauner:
Since we stopped linking against AppArmor we can now remove the conditional
compilation.


brauner:
jenkins: test this please

PR #2131, opened 2018-02-05
lsm.c: Remove constructor attribute

marcosps:
Only start.c uses makes use of lsm_init, and calls it exlpicity, so we
can safety remove __attribute__((constructor)) of it.


brauner:
Yeah, but the point is that other codepaths such as `criu.c` and `attach.c` need initialized `lsm` support as well so the constructor should stay.

marcosps:
Didn't notice that. Thanks.

PR #2133, opened 2018-02-07
selinux.c: Simplify label_set

marcosps:
If there is no label in the argument or in the config, and we can't use
the default, return error.

If we can use the default, later on we compare the label with
"unconfined_t", which is the same as DEFAUL_LABEL.

We can simplify it by checking if there is not label (in argument and in
conf), and return error if we can't use the default, or return 0 if we
the default applies. After this change, DEFAULT_LABEL can also be
removed, as it is not used anymore.

### DIFF
--- a/src/lxc/lsm/selinux.c
+++ b/src/lxc/lsm/selinux.c
@@ -76,12 +76,12 @@ static int selinux_process_label_set(const char *inlabel, struct lxc_conf *conf,
 {
 	const char *label = inlabel ? inlabel : conf->lsm_se_context;
 	if (!label) {
-		if (use_default)
-			label = DEFAULT_LABEL;
-		else
+		if (!use_default)
 			return -1;
+		label = DEFAULT_LABEL;
 	}
-	if (!strcmp(label, "unconfined_t"))
+
+	if (!strcmp(label, DEFAULT_LABEL))
 		return 0;
 
 	if (on_exec) {
--- a/src/lxc/lsm/selinux.c
+++ b/src/lxc/lsm/selinux.c
@@ -63,8 +63,8 @@ static char *selinux_process_label_get(pid_t pid)
  * selinux_process_label_set: Set SELinux context of a process
  *
  * @label   : label string
- * @conf    : the container configuration to use @label is NULL
- * @default : use the default context if label is NULL
+ * @conf    : the container configuration to use if @label is NULL
+ * @default : use the default context if @label is NULL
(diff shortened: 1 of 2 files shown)


src/lxc/lsm/selinux.c
```diff
@@ -76,13 +74,10 @@ static int selinux_process_label_set(const char *inlabel, struct lxc_conf *conf,
...
 	if (!label) {
-		if (use_default)
-			label = DEFAULT_LABEL;
-		else
+		if (!use_default)
 			return -1;
-	}
-	if (!strcmp(label, "unconfined_t"))
 		return 0;
+	}
```
brauner: But then you leave `label` uninitialized that'll cause undefined behavior further down.

brauner: Oh, I see.

brauner: But you're not handling the case where the `label` that is passed in is `unconfined_t`. So the old logic still is correct.

marcosps: Ouch... true. I'll update the patch leaving the comparison with unconfined_t after the label check.

marcosps:
@brauner so, I was my mistake. Now the function is simpler, and the rationale remains the same. There is now a second patch just to fix the documentation of lsm functions.

src/lxc/lsm/selinux.c
```diff
@@ -76,12 +76,12 @@ static int selinux_process_label_set(const char *inlabel, struct lxc_conf *conf,
...
-		if (use_default)
-			label = DEFAULT_LABEL;
-		else
+		if (!use_default)
 			return -1;
+		label = DEFAULT_LABEL;
 	}
-	if (!strcmp(label, "unconfined_t"))
+
+	if (!strcmp(label, DEFAULT_LABEL))
```
brauner: You shouldn't use `DEFAULT_LABEL` here. If we ever change `DEFAULT_LABEL` to something other than `unconfined_t` we'd want the code after this check to be executed. I'm taking your patch and adapt it and then push it to master. No need to rework it for you.

marcosps: OK, thanks!

brauner:
Closed in https://github.com/lxc/lxc/commit/ca20a3b35087b2cfd3cf18cfeaa982effdfd44b5 .

brauner:
Thanks!

PR #2136, opened 2018-02-07
Add two auto mount options:

keloyang:
With cgroup:ro(rw/mixed) option, we have some restrictions:
   * general containers(non system containers with cgroup namespace and sys_admin) have no cgroup (/sys/fs/cgroup is empty)
   * container's /sys/fs/cgroup has lxc/containersid, sometimes it's not a issue, but it is strange and the  os also make cgroup subsystem directly in /sys/fs/cgroup.

What I do?
Add two auto mount options:
   *  cgroup-self:ro   partial mount by lxc itself, no lxc/containerid exist in /sys/fs/cgroup r/o
   *  cgroup-self:rw   partial mount by lxc itself, no lxc/containerid exist in /sys/fs/cgroup r/w

Signed-off-by: Shukui Yang <yangshukui@huawei.com>

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -2048,18 +2048,24 @@ static bool cgfsng_mount(void *hdata, const char *root, int type)
 	bool retval = false;
 	struct lxc_handler *handler = hdata;
 	struct cgfsng_handler_data *d = handler->cgroup_data;
-	bool has_cgns = false, has_sys_admin = true;
+	bool has_cgns = false, wants_force_mount = false;
 
 	if ((type & LXC_AUTO_CGROUP_MASK) == 0)
 		return true;
 
-	has_cgns = cgns_supported();
-	if (!lxc_list_empty(&handler->conf->keepcaps))
-		has_sys_admin = in_caplist(CAP_SYS_ADMIN, &handler->conf->keepcaps);
-	else
-		has_sys_admin = !in_caplist(CAP_SYS_ADMIN, &handler->conf->caps);
+	if (type & LXC_AUTO_CGROUP_FORCE){
+		type &= ~LXC_AUTO_CGROUP_FORCE;
+		wants_force_mount = true;
+	}
+	if (!wants_force_mount){
+		if (!lxc_list_empty(&handler->conf->keepcaps))
+			wants_force_mount = !in_caplist(CAP_SYS_ADMIN, &handler->conf->keepcaps);
+		else
+			wants_force_mount = in_caplist(CAP_SYS_ADMIN, &handler->conf->caps);
+	}
 
-	if (has_cgns && has_sys_admin)
+	has_cgns = cgns_supported();
+	if (has_cgns && !wants_force_mount)
 		return true;
 
 	tmpfspath = must_make_path(root, "/sys/fs/cgroup", NULL);
@@ -2096,7 +2102,7 @@ static bool cgfsng_mount(void *hdata, const char *root, int type)
 			goto bad;
 		}
 
-		if (has_cgns && !has_sys_admin) {
+		if (has_cgns && wants_force_mount) {
 			/* If cgroup namespaces are supported but the container
 			 * will not have CAP_SYS_ADMIN after it has started we
 			 * need to mount the cgroups manually.
--- a/src/lxc/cgroups/cgfsng.c.rej
+++ b/src/lxc/cgroups/cgfsng.c.rej
@@ -1,0 +1,53 @@
+diff a/src/lxc/cgroups/cgfsng.c b/src/lxc/cgroups/cgfsng.c	(rejected hunks)
+@@ -1960,7 +1960,7 @@ static int do_secondstage_mounts_if_needed(int type, struct hierarchy *h,
+ 
+ 	if (flags & MS_RDONLY) {
+ 		ret = mount(sourcepath, cgpath, "cgroup",
+-			    MS_REMOUNT | flags | MS_RDONLY, NULL);
++			    MS_REMOUNT | flags, NULL);
+ 		if (ret < 0) {
+ 			SYSERROR("Failed to remount \"%s\" ro", cgpath);
+ 			free(sourcepath);
+@@ -1975,24 +1975,27 @@ static int do_secondstage_mounts_if_needed(int type, struct hierarchy *h,
+ 
+ static int mount_cgroup_cgns_supported(int type, struct hierarchy *h, const char *controllerpath)
+ {
+-	 int ret;
+-	 char *controllers = NULL;
+-	 char *fstype = "cgroup2";
+-	 unsigned long flags = 0;
++	int ret;
++	char *controllers = NULL;
++	char *fstype = "cgroup2";
++	unsigned long flags = 0;
+ 
+-	 flags |= MS_NOSUID;
+-	 flags |= MS_NOEXEC;
+-	 flags |= MS_NODEV;
+-	 flags |= MS_RELATIME;
++	flags |= MS_NOSUID;
++	flags |= MS_NOEXEC;
++	flags |= MS_NODEV;
++	flags |= MS_RELATIME;
+ 
+-	 if (type == LXC_AUTO_CGROUP_RO || type == LXC_AUTO_CGROUP_FULL_RO)
+-		 flags |= MS_RDONLY;
++	if (type & LXC_AUTO_CGROUP_FORCE){
++		type &= ~LXC_AUTO_CGROUP_FORCE;
++	}
++	if (type == LXC_AUTO_CGROUP_RO || type == LXC_AUTO_CGROUP_FULL_RO)
++		flags |= MS_RDONLY;
+ 
+-	 if (h->version != CGROUP2_SUPER_MAGIC) {
+-		 controllers = lxc_string_join(",", (const char **)h->controllers, false);
+-		 if (!controllers)
+-			 return -ENOMEM;
+-		 fstype = "cgroup";
++	if (h->version != CGROUP2_SUPER_MAGIC) {
++		controllers = lxc_string_join(",", (const char **)h->controllers, false);
++		if (!controllers)
++			return -ENOMEM;
++		fstype = "cgroup";
+ 	}
+ 
+ 	ret = mount("cgroup", controllerpath, fstype, flags, controllers);
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -1706,26 +1706,30 @@ static int set_config_mount_auto(const char *key, const char *value,
 		int mask;
 		int flag;
 	} allowed_auto_mounts[] = {
-	    { "proc",              LXC_AUTO_PROC_MASK,   LXC_AUTO_PROC_MIXED         },
-	    { "proc:mixed",        LXC_AUTO_PROC_MASK,   LXC_AUTO_PROC_MIXED         },
-	    { "proc:rw",           LXC_AUTO_PROC_MASK,   LXC_AUTO_PROC_RW            },
-	    { "sys",               LXC_AUTO_SYS_MASK,    LXC_AUTO_SYS_MIXED          },
-	    { "sys:ro",            LXC_AUTO_SYS_MASK,    LXC_AUTO_SYS_RO             },
-	    { "sys:mixed",         LXC_AUTO_SYS_MASK,    LXC_AUTO_SYS_MIXED          },
-	    { "sys:rw",            LXC_AUTO_SYS_MASK,    LXC_AUTO_SYS_RW             },
-	    { "cgroup",            LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_NOSPEC      },
-	    { "cgroup:mixed",      LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_MIXED       },
-	    { "cgroup:ro",         LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_RO          },
-	    { "cgroup:rw",         LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_RW          },
-	    { "cgroup-full",       LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_FULL_NOSPEC },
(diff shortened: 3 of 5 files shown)


src/lxc/cgroups/cgfsng.c
```diff
@@ -1947,7 +1947,7 @@ static int do_secondstage_mounts_if_needed(int type, struct hierarchy *h,
 
 	sourcepath = must_make_path(h->mountpoint, h->base_cgroup,
 				    container_cgroup, NULL);
-	if (type == LXC_AUTO_CGROUP_RO)
+	if (type == LXC_AUTO_CGROUP_RO || type == LXC_AUTO_CGROUP_RO)
```
brauner: That is redundant. Did you mean to use a different option?

src/lxc/cgroups/cgfsng.c
```diff
@@ -2023,10 +2024,13 @@ static bool cgfsng_mount(void *hdata, const char *root, int type)
 		has_sys_admin = in_caplist(CAP_SYS_ADMIN, &handler->conf->keepcaps);
 	else
 		has_sys_admin = !in_caplist(CAP_SYS_ADMIN, &handler->conf->caps);
+	if (type != LXC_AUTO_CGROUP_SELF_RO &&
+		type != LXC_AUTO_CGROUP_SELF_RW){
```
brauner: Can you move the if statement onto one line, please:

```
if (type != LXC_AUTO_CGROUP_SELF_RO && type != LXC_AUTO_CGROUP_SELF_RW){
```

keloyang:
> One questions: is the intention for cgroup-self to work with both, cgroup namespace and without cgroup namespaces

@brauner yes, this pr do with both. thanks for your review and I have updated.

brauner:
@keloyang, thanks. I'm going to grab your branch and do some testing on it.

brauner:
Hm, the interaction with cgroup namespaces is not yet correct because you mount the full host path. That should never be the case. I'll try to track this down.

brauner:
I think to get what you want you at least need the following on top your your current patch:

```
diff --git a/src/lxc/cgroups/cgfsng.c b/src/lxc/cgroups/cgfsng.c
index 529567c1..84b7ebcc 100644
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1985,7 +1985,8 @@ static int mount_cgroup_cgns_supported(int type, struct hierarchy *h, const char
 	 flags |= MS_NODEV;
 	 flags |= MS_RELATIME;
 
-	 if (type == LXC_AUTO_CGROUP_RO || type == LXC_AUTO_CGROUP_FULL_RO)
+	 if (type == LXC_AUTO_CGROUP_RO || type == LXC_AUTO_CGROUP_FULL_RO ||
+	     type == LXC_AUTO_CGROUP_SELF_RO)
 		 flags |= MS_RDONLY;
 
 	 if (h->version != CGROUP2_SUPER_MAGIC) {
@@ -2013,7 +2014,7 @@ static bool cgfsng_mount(void *hdata, const char *root, int type)
 	bool retval = false;
 	struct lxc_handler *handler = hdata;
 	struct cgfsng_handler_data *d = handler->cgroup_data;
-	bool has_cgns = false, has_sys_admin = true;
+	bool has_cgns = false, wants_manual_mount = false;
 	char *container_cgroup = "";
 
 	if ((type & LXC_AUTO_CGROUP_MASK) == 0)
@@ -2021,14 +2022,17 @@ static bool cgfsng_mount(void *hdata, const char *root, int type)
 
 	has_cgns = cgns_supported();
 	if (!lxc_list_empty(&handler->conf->keepcaps))
-		has_sys_admin = in_caplist(CAP_SYS_ADMIN, &handler->conf->keepcaps);
+		wants_manual_mount = !in_caplist(CAP_SYS_ADMIN, &handler->conf->keepcaps);
 	else
-		has_sys_admin = !in_caplist(CAP_SYS_ADMIN, &handler->conf->caps);
-	if (type != LXC_AUTO_CGROUP_SELF_RO && type != LXC_AUTO_CGROUP_SELF_RW){
-		if (has_cgns && has_sys_admin )
+		wants_manual_mount = in_caplist(CAP_SYS_ADMIN, &handler->conf->caps);
+	if (type != LXC_AUTO_CGROUP_SELF_RO &&
+	    type != LXC_AUTO_CGROUP_SELF_RW) {
+		if (has_cgns && !wants_manual_mount)
 			return true;
 
 		container_cgroup = d->container_cgroup;
+	} else {
+		wants_manual_mount = true;
 	}
 	tmpfspath = must_make_path(root, "/sys/fs/cgroup", NULL);
 
@@ -2064,7 +2068,7 @@ static bool cgfsng_mount(void *hdata, const char *root, int type)
 			goto bad;
 		}
 
-		if (has_cgns && !has_sys_admin) {
+		if (has_cgns && wants_manual_mount) {
 			/* If cgroup namespaces are supported but the container
 			 * will not have CAP_SYS_ADMIN after it has started we
 			 * need to mount the cgroups manually.
```

brauner:
Ok, I think we really only want and need `cgroup-ns:ro` and `cgroup-ns:rw` all other cases are already correctly handled by `cgroup:{mixed,ro,rw}`. If only `cgroup-ns` is specified it should default to `cgroup-ns:rw`.

brauner:
Let's call the new options `cgroup:ro:force` and `cgroup:rw:force`.

brauner:
@keloyang, when you have something I need to review, please ping me because I'm not getting notified when you push changes. :)

brauner:
@keloyang, what's your use-case for adding `cgroup-direct` it seems that this is only really useful when running unprivileged containers and privileged containers that drop cap_sys_admin. Containers that retain cap_sys_admin can simply umount the `cgroup-direct` mounts and call `mount -t cgroup <controller-name> -o <controller-name> /sys/fs/cgroup/<controller-name>` at which point they see the full cgroup hierarchies leading up to their own.

src/lxc/confile.c
```diff
@@ -1758,22 +1758,35 @@ static int set_config_mount_auto(const char *key, const char *value,
...
 	    { "sys:ro",                  LXC_AUTO_SYS_MASK,    LXC_AUTO_SYS_RO                                   },
 	    { "sys:mixed",               LXC_AUTO_SYS_MASK,    LXC_AUTO_SYS_MIXED                                },
 	    { "sys:rw",                  LXC_AUTO_SYS_MASK,    LXC_AUTO_SYS_RW                                   },
+
 	    { "cgroup",                  LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_NOSPEC                            },
 	    { "cgroup:mixed",            LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_MIXED                             },
 	    { "cgroup:ro",               LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_RO                                },
 	    { "cgroup:rw",               LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_RW                                },
+
 	    { "cgroup:force",            LXC_AUTO_CGROUP_MASK, LXC_AUTO_CGROUP_NOSPEC|LXC_AUTO_CGROUP_FORCE      },
```
brauner: Hm, is that really necessary. Seems to me that we should only support `cgroup:<mount-type>:force`.

brauner: I'm not nacking it I'm really just asking. :)

keloyang: @brauner 
https://github.com/lxc/lxc/blob/master/src/lxc/cgroups/cgfsng.c#L2070 why we need judge has_cgns？ if a container has no sys_admin, it can't mount by itself, so I think lxc should mount cgroup for container if it has no sys_admin. WDYT?

What is it the purpose of judge has_cgns，Is it to make /sys/fs/cgroup has the dir of  lxc/containerid ?

On the other hand, cg_mount_in_cgroup_namespace make container has no lxc/containerid in /sys/fs/cgroup,but do_secondstage_mounts_if_needed has lxc/containerid in /sys/fs/cgroup, Is it a issue ?

I add all cgroup options with force, and also cgroup-direct to make both cg_mount_in_cgroup_namespace and do_secondstage_mounts_if_needed has no lxc/containerid in container's /sys/fs/cgroup.

brauner:
> @brauner
https://github.com/lxc/lxc/blob/master/src/lxc/cgroups/cgfsng.c#L2070 why we need judge has_cgns？ if a container has no sys_admin, it can't mount by itself, so I think lxc should mount cgroup for container if it has no sys_admin. WDYT?

> What is it the purpose of judge has_cgns，Is it to make /sys/fs/cgroup has the dir of lxc/containerid ?

When cgroup namespaces are supported the container's cgroup `lxc/containerid` is not allowed to be visible inside the container as this would be information leak. If you unshare your cgroup namespace and you look at `/proc/1/cgroup` from inside the container you will see:

```chb@conventiont|~
> lxc-attach a10
/ # cat /proc/1/cgroup
12:memory:/
11:pids:/
10:perf_event:/
9:net_cls,net_prio:/
8:freezer:/
7:hugetlb:/
6:rdma:/
5:cpuset:/
4:cpu,cpuacct:/
3:blkio:/
2:devices:/
1:name=systemd:/
0::/
```
If you were to mount with the name of the container's cgroup you would leak the name of the host cgroup into the container. While this is not a big deal it's way cleaner to not do it. So the check needs to be there.

> On the other hand, cg_mount_in_cgroup_namespace make container has no lxc/containerid in /sys/fs/cgroup,but do_secondstage_mounts_if_needed has lxc/containerid in /sys/fs/cgroup, Is it a issue ?

When no cgroup namespaces are supported it is perfectly fine to show the path leading up to the container's cgroup inside the container because the container can gather that information anyway from `/proc/1/cgroup`. In this case it would show something like:

```chb@conventiont|~
> lxc-attach a10
/ # cat /proc/1/cgroup
12:memory:/lxc/bla
11:pids:/lxc/bla
10:perf_event:/lxc/bla
9:net_cls,net_prio:/lxc/bla
8:freezer:/lxc/bla
7:hugetlb:/lxc/bla
6:rdma:/lxc/bla
5:cpuset:/lxc/bla
4:cpu,cpuacct:/lxc/bla
3:blkio:/lxc/bla
2:devices:/lxc/bla
1:name=systemd:/lxc/bla
0::/lxc/bla
```
So there's no need to hide it and that's why `cgroup-direct` doesn't give you any additional security at all. In fact with `cgroup:{rw,ro,mixed}` the cgroups leading up to the container's cgroup will be empty and will not be actual cgroups so you can't do any harm. Now, if you're root and have sys_admin in a container you can even mount cgroups the way you want and manipulate them short of using an LSM to prevent that so again there's no security to gain from `cgroup-direct`.

> I add all cgroup options with force, and also cgroup-direct to make both cg_mount_in_cgroup_namespace and do_secondstage_mounts_if_needed has no lxc/containerid in container's /sys/fs/cgroup.

As I said `cgroup-direct` is not needed since it doesn't give you any additional security when used without cgroup namespaces. The container can gather the information from `/proc/1/cgroup` anyway and if it is privileged with sys_admin will be able to remount cgroups anyway. If cgroup namespaces are enabled `cgroup-direct` is in fact the wrong way to mount cgroups and will cause information leak. So I suggest you simply add the `force` option like you did in your first commit and drop `cgroup-direct` completely.

keloyang:
@brauner thanks for you reply.
the purpose of removing lxc/containerid is not for security, but for compatibility.
for example, if users has  programs which want to read some infomation from cgroup in os, they need access /sys/fs/cgroup/cgroup-subsystem,  but in lxc, they need access  /sys/fs/cgroup/lxc/containerid/cgroup-subsystem, so if we migrate programs from os to the container, we need change all of the programs.

brauner:
> the purpose of removing lxc/containerid is not for security, but for compatibility.
for example, if users has programs which want to read some infomation from cgroup in os, they need access /sys/fs/cgroup/cgroup-subsystem, but in lxc, they need access /sys/fs/cgroup/lxc/containerid/cgroup-subsystem, so if we migrate programs from os to the container, we need change all of the programs.

Ok, have you considered using `cgroup-full`?

keloyang:
@brauner  Do you mean we change the action of cgroup-full and make container's /sys/fs/cgroup has no lxc/containerid if user select cgroup-full ?

brauner:
> @brauner Do you mean we change the action of cgroup-full and make container's /sys/fs/cgroup has no lxc/containerid if user select cgroup-full ?

Hm, what `cgroup:{mixed,ro,rw}` will mount empty directories leading up to the containers cgroup whereas `cgroup-full:{mixed,ro,rw}` will mount the full hierarchy leading up to the containers cgroup into the container at which point you will effectively have a populated cgroup tree. I was wondering whether that was enough when a program was looking at the cgroups. But I guess you want the program to take the limits from the cgroup it actually resides in.

brauner:
If this is a real-use case we're going to make it happen. It's just figuring out a nice way to do it. You could send the `force` option as a first patch and then we can discuss the other addition? Wdyt?

keloyang:
@brauner I have remove the second patch.

src/lxc/cgroups/cgfsng.c.rej
```diff
@@ -0,0 +1,53 @@
+diff a/src/lxc/cgroups/cgfsng.c b/src/lxc/cgroups/cgfsng.c	(rejected hunks)
```
brauner: I think you don't want to commit this, right? :)

brauner:
Closed in https://github.com/lxc/lxc/pull/2170 .

brauner:
Thanks!

src/lxc/cgroups/cgfsng.c.rej
keloyang: @brauner It's my fault, thanks for https://github.com/lxc/lxc/pull/2170.

brauner:
On Feb 21, 2018 03:08, "keloyang" <notifications@github.com> wrote:

*@keloyang* commented on this pull request.
------------------------------

In src/lxc/cgroups/cgfsng.c.rej
<https://github.com/lxc/lxc/pull/2136#discussion_r169518839>:

> @@ -0,0 +1,53 @@
+diff a/src/lxc/cgroups/cgfsng.c b/src/lxc/cgroups/cgfsng.c	(rejected hunks)

@brauner <https://github.com/brauner> It's my fault, thanks for #2170
<https://github.com/lxc/lxc/pull/2170>.

No worries, thanks for the patch. :)

PR #2139, opened 2018-02-08
signal child in case of failure when running in userns

tych0:
In the case where we fail (e.g. if there are no subuids for the current
user) to run something in the userns, let's signal the child so that they
can exit. Simply closing the pipe causes a hang:

int main(void)
{
	int p[2];
	pid_t pid;
	int status;

	if (pipe(p) < 0) {
		perror("pipe");
		return 1;
	}

	pid = fork();
	if (pid < 0) {
		perror("fork");
		return 1;
	}

	if (pid == 0) {
		char c;

		if (read(p[0], &c, 1) != 1) {
			perror("read");
			exit(1);
		}

		exit(0);
	}

	close(p[1]);

	if (waitpid(pid, &status, 0) != pid) {
		perror("waitpid");
		return 1;
	}

	if (!WIFEXITED(status) || WEXITSTATUS(status)) {
		printf("wrong return code\n");
		return 1;
	}

	return 0;
}

How did this ever work? :)

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3737,6 +3737,9 @@ static int run_userns_fn(void *data)
 	/* Close read end of the pipe. */
 	close(d->p[0]);
 
+	if (c == '0')
+		return 1;
+
 	if (d->fn_name)
 		TRACE("calling function \"%s\"", d->fn_name);
 	/* Call function to run. */
@@ -4026,11 +4029,11 @@ int userns_exec_full(struct lxc_conf *conf, int (*fn)(void *), void *data,
 	int p[2];
 	struct id_map *map;
 	struct lxc_list *cur;
-	char c = '1';
 	int ret = -1;
 	struct lxc_list *idmap = NULL, *tmplist = NULL;
 	struct id_map *container_root_uid = NULL, *container_root_gid = NULL,
 		      *host_uid_map = NULL, *host_gid_map = NULL;
+	bool error = true;
 
 	ret = pipe(p);
 	if (ret < 0) {
@@ -4172,12 +4175,19 @@ int userns_exec_full(struct lxc_conf *conf, int (*fn)(void *), void *data,
 	}
 
 	/* Tell child to proceed. */
-	if (write(p[1], &c, 1) != 1) {
+	if (write(p[1], "1", 1) != 1) {
 		SYSERROR("failed telling child process \"%d\" to proceed", pid);
 		goto on_error;
 	}
 
+	error = false;
+
 on_error:
+	/* Try to wake the child up and tell them we failed. */
+	if (error && write(p[1], "0", 1) != 1) {
+		ERROR("error telling the child about our error");
+	}
+
 	/* Wait for child to finish. */
 	if (pid > 0)
 		ret = wait_for_pid(pid);


brauner:
> In the case where we fail (e.g. if there are no subuids for the current
user) to run something in the userns, let's signal the child so that they
can exit. Simply closing the pipe causes a hang:
> ```
> int main(void)
> {
>	int p[2];
>	pid_t pid;
>	int status;
>
>	if (pipe(p) < 0) {
>		perror("pipe");
>		return 1;
>	}
>
>	pid = fork();
>	if (pid < 0) {
>		perror("fork");
>		return 1;
>	}
>
>	if (pid == 0) {
>		char c;
>

You're hanging cause you're not closing the read-end of the pipe in the child. A
`close(p[1])` is missing.

>		if (read(p[0], &c, 1) != 1) {
>			perror("read");
>			exit(1);
>		}
>
>		exit(0);
>	}
>
>	close(p[1]);
>
>	if (waitpid(pid, &status, 0) != pid) {
>		perror("waitpid");
>		return 1;
>	}
>
>	if (!WIFEXITED(status) || WEXITSTATUS(status)) {
>		printf("wrong return code\n");
>		return 1;
>	}
>
>	return 0;
> }
```

**Comment by brauner** [comment]
These functions do heavy duty work all of the time. This is especially true for `userns_exec_1()` which is called multiple times when you start and stop a container. If this is really a problem for `userns_exec_full()` which is only ever called when a container's storage is created or deleted it must be a massive problem for `userns_exec_1()` the fact that I've never seen a single issue about hanging makes me wonder. What exactly is the problem you're seeing?

**Comment by tych0** [comment]
Ah, so it is :). Ok, well, I'm getting a hang after:

```
lxc-destroy: x1: conf.c: userns_exec_full: 4093 No mapping for container root found
^C
```

The child is stuck in read(), and the parent is stuck in wait(), so this code seems very suspicious to me.

**Comment by brauner** [comment]
@tych0, right, then the problem simply is that `goto on_error` in 

	if (!container_root_uid || !container_root_gid) {
		ERROR("No mapping for container root found");
		goto on_error;
	}

jumps to the `wait()` at the end of the function without closing the write end of the pipe at which point you hang. So the solution is to close the write-end of the pipe when doing the goto.

**Comment by brauner** [comment]
:)

**Comment by tych0** [comment]
Ok, I'll send a patch for that if you want. I'm getting on a plane now, so
it'll be a bit before I get to it.

**Comment by brauner** [comment]
@tych0, thanks! If it's not too much work it would be cool if you could fix up `userns_exec_1()` as well. It's likely just rearranging the cleanup statement after the `goto` label. But if you don't have time to do `userns_exec_1()` as well I'll push it on top your commit.

**Comment by brauner** [comment]
Yeah, I think it's super obvious that:

```
on_error:
	/* Wait for child to finish. */
	if (pid > 0)
		status = wait_for_pid(pid);

	if (p[0] != -1)
		close(p[0]);
	close(p[1]);

	if (status < 0)
		ret = -1;

	return ret;
```

should be

```
on_error:
	if (p[0] != -1)
		close(p[0]);
	close(p[1]);

	/* Wait for child to finish. */
	if (pid > 0)
		status = wait_for_pid(pid);

	if (status < 0)
		ret = -1;

	return ret;
```

Not that I wrote that code in the first place...

tych0:
See #2149

PR #2148, opened 2018-02-09
pam: create writable cgroups for unpriv users

### LINKED ISSUE #1307: move pam_cgfs to liblxc when preparing for lxc 3.0
https://github.com/lxc/lxcfs/issues/158

brauner:
This moves pam_cgfs from the LXCFS repo into the LXC repo. This will allow us
to share a bunch of code between the cgroup backends and the pam module. The
next step obviously is to share code.

Closes #1307.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -646,6 +646,35 @@ else
 fi
 AM_CONDITIONAL([IS_BIONIC], [test "x$is_bionic" = "xyes"])
 
+AC_ARG_WITH(
+	[pamdir],
+	[AS_HELP_STRING([--with-pamdir=PATH],[Specify the directory where PAM modules are stored,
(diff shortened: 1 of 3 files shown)


brauner:
@stgraber, if we still want to do this it requires libpam to be installed on the test-runners.

stgraber:
I somehow don't remember us discussing this :)
libpam-cgfs is kinda odd under either lxcfs or lxc but I guess it's most needed by lxc so may indeed make more sense here.

@hallyn what do you think?

marcosps:
That sounds good! Now Fedora can have a seamless integration with unprivileged LXC containers :)

stgraber:
Reverted this change as it broke quite a few things. Before we let it back in, we need it under a --enable-pam option or something like that.

PR #2149, opened 2018-02-09
fix userns helper error handling

tych0:
In both of these cases if there is actually an error, we won't close the
pipe and the api call will hang. Instead, let's be sure to close the pipe
before waiting, so that it doesn't hang.


brauner:
I'm ignoring the download errors.

PR #2158, opened 2018-02-12
Fix lxc-console hang

lifeng68:
The variable 'descr' is mistakenly covered with 'descr_console'.

Steps to reproduce the problem：

1. Create and start an container test1
2. use lxc-console to attach the container, it hang


brauner:
jenkins: test this please

PR #2160, opened 2018-02-12
[confile] Add config for populate devices.

lifeng68:
Now the lxc provide the autodev hook to mknod in the container. 
I think we should provide a way to make the same thing in config.

### DIFF
--- a/doc/lxc.container.conf.sgml.in
+++ b/doc/lxc.container.conf.sgml.in
@@ -1605,6 +1605,29 @@ dev/null proc/kcore none bind,relative 0 0
     </refsect2>
 
     <refsect2>
+      <title>Populate devices</title>
+      <para>
+        Specify the devices on host to be populated into the container.
+        For instance, you can populate the block device '/dev/sda' on host into the container.
+      </para>
+      <variablelist>
+        <varlistentry>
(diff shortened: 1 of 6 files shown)


brauner:
So one thing that immediately comes to mind is where this setup step is run. You're currently running it in `lxc_setup()` which runs in the container's namespaces. But this prevents unprivileged containers that are started by the root user from creating device nodes. So the setup process should likely take place somewhere else. We might have this work by doing it from the parent and having a helper process attach to the containers mount namespace and create the device nodes for us. That should work. But we need to discuss the sensibility of this whole approach first. And I want to admit up front: not a huge fan of the name `lxc.populate.device`. ;)

hallyn:
Was there a good reason to not just hook up the c->add_device_node and c->remove_device_node to a confile.c entry?

brauner:
Yeah, we should use those two API functions. I just realized they need to be tweaked wrt to the `devices` cgroup. Meaning, we need to handle the unified hierarchy. But I can take care of that once this patchset lands.

brauner:
So we'd still be interested in this but there hasn't been movement on this in a lot of months so I'm closing for now.

PR #2167, opened 2018-02-15
console: simplify console logging

brauner:

@duguhaotian, this is my proposal.

### DIFF
--- a/src/lxc/console.c
+++ b/src/lxc/console.c
@@ -204,6 +204,152 @@ void lxc_console_signal_fini(struct lxc_tty_state *ts)
 	free(ts);
 }
 
+static int lxc_console_truncate_log_file(struct lxc_console *console)
+{
+	/* be very certain things are kosher */
+	if (!console->log_path || console->log_fd < 0)
+		return -EBADF;
+
+	return lxc_unpriv(ftruncate(console->log_fd, 0));
+}
+
+static int lxc_console_rotate_log_file(struct lxc_console *console)
+{
+	int ret;
+	size_t len;
+	char *tmp;
+
+	/* rotate the console log file */
+	if (!console->log_path || console->log_rotate == 0)
+		return -EOPNOTSUPP;
+
+	/* be very certain things are kosher */
+	if (console->log_fd < 0)
+		return -EBADF;
+
+	len = strlen(console->log_path) + sizeof(".1");
+	tmp = alloca(len);
+
+	ret = snprintf(tmp, len, "%s.1", console->log_path);
+	if (ret < 0 || (size_t)ret >= len)
+		return -EFBIG;
+
+	close(console->log_fd);
+	console->log_fd = -1;
+	ret = lxc_unpriv(rename(console->log_path, tmp));
+	if (ret < 0)
+		return ret;
+
+	return lxc_console_create_log_file(console);
+}
+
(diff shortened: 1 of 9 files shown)


src/lxc/console.c
```diff
@@ -583,7 +746,7 @@ void lxc_console_delete(struct lxc_console *console)
 
 	ret = lxc_console_write_ringbuffer(console);
```
duguhaotian: Do we need this?  lxc_console_cb_con write console log into ringbuffer and log file, i think log file have the same logs with ringbuffer.
```
lxc_console_cb_con() {
     if (console->buffer_size > 0)
 			w_rbuf = lxc_ringbuf_write(&console->ringbuf, buf, r);
     if (console->log_fd > 0)
			w_log = lxc_console_write_log_file(console, buf, r);
}
```

brauner: > Do we need this? lxc_console_cb_con write console log into ringbuffer and log file, i think log file have the same logs with ringbuffer.

If `console->log_fd > 0` then the user wants us to write to the logfile. If they only care about the ringbuffer they can just choose to not use a console logfile.

brauner: Oh, sorry I misread your comment. The idea is that if the container e.g. aborts you get a dump of the last ringbuffer contents.

stgraber:
@hallyn can you take a look at this one?

src/lxc/console.c
```diff
@@ -204,6 +204,152 @@ void lxc_console_signal_fini(struct lxc_tty_state *ts)
...
+	return lxc_unpriv(ftruncate(console->log_fd, 0));
+}
+
+static int lxc_console_rotate_log_file(struct lxc_console *console)
+{
+	int ret;
+	size_t len;
+	char *tmp;
+
+	/* rotate the console log file */
```
hallyn: This comment seems superfluous

src/lxc/console.c
```diff
@@ -204,6 +204,152 @@ void lxc_console_signal_fini(struct lxc_tty_state *ts)
...
+		 * he's requesting us to do weird stuff.
+		 */
+		if (console->log_rotate > 0 || console->log_size > 0)
+			return -EINVAL;
+
+		/* I mean, sure log wherever you want to. */
+		return lxc_write_nointr(console->log_fd, buf, bytes_read);
+	}
+
+	/* check how much space we have left */
```
hallyn: The useful variable name makes the comment unnecessary.

src/lxc/console.c
```diff
@@ -701,10 +701,53 @@ static int lxc_console_peer_default(struct lxc_console *console)
...
+
+	/* There's not log file where we can dump the ringbuffer to. */
+	if (console->log_fd < 0)
+		return 0;
+
+	/* The log file is simply appended to. */
+	if (console->log_size == 0)
+		return 0;
+
+	/* The log file is rotated. */
```
hallyn: Sorry, why do you not dump the final contents on exit if the user does not want log files rotated?

src/lxc/console.c
brauner: removed

src/lxc/console.c
brauner: removed

src/lxc/console.c
brauner: You're right, we should dump it.

PR #2183, opened 2018-02-22
confile: make reading config file thread-safe

brauner:
When multiple threads read the same config file (e.g. the global default
config) the file pointer seems to be modified by the various threads. There's
an obvious solution to this problem: m{un}map().


brauner:
The analysis was wrong, the spirit correct. I'm going to send a thread hardening branch.

PR #2196, opened 2018-02-28
terminal: improvements and cleanup

brauner:

### DIFF
--- a/src/lxc/terminal.c
+++ b/src/lxc/terminal.c
@@ -1026,7 +1026,7 @@ int lxc_console(struct lxc_container *c, int ttynum,
 		int stdinfd, int stdoutfd, int stderrfd,
 		int escape)
 {
-	int ret, ttyfd, masterfd;
+	int masterfd, ret, ttyfd;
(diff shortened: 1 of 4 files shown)


src/lxc/terminal.c
```diff
@@ -206,7 +206,10 @@ void lxc_terminal_signal_fini(struct lxc_terminal_state *ts)
 	if (ts->sigfd >= 0) {
 		close(ts->sigfd);
 
-		if (sigprocmask(SIG_SETMASK, &ts->oldmask, NULL) < 0)
+		int ret;
```
hallyn: So...  I'm pretty sure this was in our codingstyle, but do we really want to do this?  It's longer and arguably not clearer.

I'll merge it if you really want, but I prefer the old style here.

brauner: No, I can just remove it.

brauner: removed :)

hallyn:
Thanks - I didn't look at every commit, but looks very nice.  Will merge when tests are done.

PR #2211, opened 2018-03-12
usernsexec: init log fd

tych0:
lxc-usernsexec uses some functions (e.g. lxc_map_ids()), which are part of
the lxc library and thus use the WARN etc. macros to emit log messages.
However, it doesn't initialize the log in any way, so these messages go
into the ether.

lxc-usernsexec currently has no log parameters, so let's just log these to
stderr. Someone can do something fancier later if they want.

### DIFF
--- a/src/lxc/cmd/lxc_usernsexec.c
+++ b/src/lxc/cmd/lxc_usernsexec.c
@@ -53,6 +53,8 @@
 #define MS_SLAVE (1 << 19)
 #endif
 
+extern int lxc_log_fd;
+
 int unshare(int flags);
 
 static void usage(const char *name)
@@ -273,6 +275,8 @@ int main(int argc, char *argv[])
 	char buf[1];
 	int pipe1[2],  /* child tells parent it has unshared */
 	    pipe2[2];  /* parent tells child it is mapped and may proceed */
+
+	lxc_log_fd = STDERR_FILENO;
 
 	memset(ttyname0, '\0', sizeof(ttyname0));
 	memset(ttyname1, '\0', sizeof(ttyname1));


hallyn:
LGTM.  I wonder whether we should do this in a __init fn.

src/lxc/cmd/lxc_usernsexec.c
```diff
@@ -53,6 +53,8 @@
 #define MS_SLAVE (1 << 19)
 #endif
 
+extern int lxc_log_fd;
```
brauner: I fail to see how this is supposed to work so @hallyn I leave this to you.

hallyn:
@brauner you've overhauled this code (or overseen some overhaul) recently, so I'll wait for you to take a look rather than merge it.

src/lxc/cmd/lxc_usernsexec.c
hallyn: D'oh!

brauner: I mean, I'm not completely sure if a) overwriting `extern int lxc_log_fd` is save b) if this actually logs something when it is set to `STDERR_FILENO`.

hallyn: Can you elaborate?  It's impolite, but lxc_log_fd is not defined static, so lxc_usernsexec.c should be able to fetch it.

tych0: I can confirm that it does log stuff :). Is there some reason it wouldn't be safe? Nothing else is initializing it.

tych0: I could add a lxc_set_log_fd() or something if you want.

hallyn: lxc_set_log_fd() sounds worthwhile, or perhaps it should be handled through lxc_log_init().

But for now I think this is fine.

brauner: Oh I'm not objecting. I just don't feel comfortable acking this by merging it. :)

hallyn: Ok, thanks for clarifying.  IIRC there is someone looking at a logging overhaul, but if not, maybe I'll consider pursuing that for lxc 4.0.

hallyn:
Thanks Tycho.  Hate losing log msgs.

brauner:
> Thanks Tycho. Hate losing log msgs.

I hate non-uniform user-facing log messages. All that usually gets me is "stop bikeshedding". :)

PR #2212, opened 2018-03-13
cgroups: don't escape if we're not real root

tych0:
If we're host unpriv but root in our userns, we can't really escape
cgroups. Let's switch the cgroup escape test to reflect this.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1368,7 +1368,7 @@ static char *cg_unified_get_current_cgroup(void)
 	bool will_escape;
 	char *copy = NULL;
 
-	will_escape = (geteuid() == 0);
+	will_escape = !am_host_unpriv();
 	if (will_escape)
 		basecginfo = read_file("/proc/1/cgroup");
 	else


brauner:
Cool, thanks! Just waiting for the test-suite. :)

src/lxc/cgroups/cgfsng.c
```diff
@@ -1368,7 +1368,7 @@ static char *cg_unified_get_current_cgroup(void)
 	bool will_escape;
 	char *copy = NULL;
 
-	will_escape = (geteuid() == 0);
+	will_escape = !am_host_unpriv();
```
hallyn: Wait I don't think this is right.
This has to do with whether the rootfs was fully configured or not.  Not with whether you're host root.

hallyn: @brauner The idea behind cgfsng.c was to simplify the code by making assumptions about /sys/fs/cgroup layout.  Part of that is that root is allowed to write to some of it's /sys/fs/cgroup root cgroups.  When that's not the case, we  were supposed to fall back to cgfs.c.  There we could do more complicated logic, which is what we need here.  In particular: if root is allowed to write to /sys/fs/cgroup/freezer, use that.  If it isn't, then we could try the current cgroup and stay under that hierarchy.  But actually before all that we should also check for lxc.cgroup.dir - which we can't check in the __constructor cgroup_init.  Now we can probably handle all this nicely in cgfsng.c - and really since we are not choosing other modules any more, we don't need to do these checks from the __constructor before we have a container config.  BUT I don't wnat to rush this.  For now let's stick to "set up your rootfs in the way cgfsng.c expects", and let's design the next version of cgfs.  Next version should perhaps have per-container cgroup driver;  honor lxc.cgroup.dir and lxc.cgroup.use (which is global config) there;  and handle "do I have the rights to create cgroups" more intelligently.

brauner: Hm, @hallyn please revert it then.

PR #2214, opened 2018-03-14
tools: fix usage of boolean function set_config_item

### LINKED ISSUE #2073: tools: remove last internal symbols
The lxc tools should only rely on API symbols. Some work has been done but the remaining tools need to be stripped from internal symbols as well. Here's some context from a mail I wrote to someone interested in doing the work a while back:

The next step is to rip out all other dependencies. I assume you
already know what I mean by "internal symbols" but just in case: an internal
symbol in this sense is anything that isn't exported in the following three
files:

1. attach_options.h
2. lxccontainer.h
3. version.h

When you install the headers (on Ubuntu this is lxc-dev) you will usually find
those three files under:

/usr/include/lxc/

Of course, they are also available in our source repo.

The most interesting file is probably "lxccontainer.h" as it basically defines
our public API. But note that anything which is marked "private" in these files
e.g.

	/* private fields */
	/*!
	 * \private
	 * Name of container.
	 */
	char *name;

in "lxccontainer.h" is not to be touched by external tools either.

Currently our tools make wildly liberal use of internal symbols. They call
functions from e.g. utils.h, conf.h etc. All of this needs to go. One would
probably make a utils{c,h} file in the tools/ directory
which holds all helpers for all tools. This file as well, should not depend on
anything else but public API. This task will likely see some copy and pasting
but it will also involve writing a bunch of new helpers in order to remove all
dependencies.

flx42:

The command `lxc-execute --define lxc.environment='foo=bar' -n ubuntu --` was failing silently, with exit return code 1.

git bisect:
```
791e7a73a9bca8dbde19afb49f172a901ce0dd43 is the first bad commit
commit 791e7a73a9bca8dbde19afb49f172a901ce0dd43
Date:   Fri Jan 12 15:31:03 2018 +0100

    tools: move lxc-execute to API symbols only
    
    Closes #2073.
    
```

### DIFF
--- a/src/lxc/tools/lxc_execute.c
+++ b/src/lxc/tools/lxc_execute.c
@@ -194,8 +194,8 @@ int main(int argc, char *argv[])
 		}
 	}
 
-	ret = lxc_config_define_load(&defines, c);
-	if (ret) {
+	bret = lxc_config_define_load(&defines, c);
+	if (!bret) {
 		lxc_container_put(c);
 		exit(EXIT_FAILURE);
 	}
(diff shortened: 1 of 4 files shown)


flx42:
thanks @brauner !

brauner:
@flx42, thanks for the patch! :)

Issue #2215, opened 2018-03-14
error compiling with clang-5.0: comparison of unsigned expression < 0 is always false


igalic:
When trying to compile the current  master (0b628094f80a038aa6b31e9e3b6401f378073478) with clang-5.0 the following error is encountered:

```
conf.c:2366:16: error: comparison of unsigned expression < 0 is always false [-Werror,-Wtautological-compare]
        for (i = 0; i < sizeof(caps_opt)/sizeof(caps_opt[0]); i++) {
                    ~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
Makefile:1795: recipe for target 'liblxc_la-conf.lo' failed
make[3]: *** [liblxc_la-conf.lo] Error 1
make[3]: Leaving directory '/home/igalic/src/lxc/lxc/src/lxc'
```

the full(er) context of the code is here:

```c
static int parse_cap(const char *cap)
{
	char *ptr = NULL;
	size_t i;
	int capid = -1;

	if (!strcmp(cap, "none"))
		return -2;

	for (i = 0; i < sizeof(caps_opt)/sizeof(caps_opt[0]); i++) {

		if (strcmp(cap, caps_opt[i].name))
			continue;

		capid = caps_opt[i].value;
		break;
	}
```

brauner:
That very much smells like a compiler bug:

```
conf.c:2366:16: error: comparison of unsigned expression < 0 is always false [-Werror,-Wtautological-compare]
        for (i = 0; i < sizeof(caps_opt)/sizeof(caps_opt[0]); i++) {
                    ~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

the warning doesn't make any sense to me unless I'm being dense.

igalic:
@brauner given that i didn't install libcap-dev, this warning is legit
(pr comming up ;)

brauner:
@igalic, cool. :) The warning just seems very weird to me.

igalic:
fixed ^

PR #2216, opened 2018-03-14
conf: fix clang warning when building w/o libcap

igalic:
when compiling lxc with clang-5.0 parse_cap()'s main loop will produce a
warning about a tautological comparision (#2215).

By moving the result of computation into a variable (end) this is no
longer a constant expression. clang-5.0 does not do dataflow analysis at
this point, so it is, to quote someone from #llvm, "morally equivalent"
to casting `(int)i`.

in addition, we also clean up the #if HAVE_LIBCAP to no longer need
its #else branch!

Signed-off-by: Igor Galić <igor.galic@automatic-server.com>

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -205,8 +205,8 @@ static struct mount_opt propagation_opt[] = {
 	{ NULL,            0, 0                    },
 };
 
-#if HAVE_LIBCAP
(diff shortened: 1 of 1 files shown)


igalic:
@brauner as promised

brauner:
jenkins: test this please

brauner:
@igalic thanks. :)

PR #2217, opened 2018-03-14
conf: do not run the "mount" hooks twice

flx42:
Regression introduced by 8353b4c90ed18e570521134f2c60bef56a082b55

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3368,7 +3368,7 @@ int lxc_setup(struct lxc_handler *handler)
 	}
 
 	ret = run_lxc_hooks(name, "mount", lxc_conf, NULL);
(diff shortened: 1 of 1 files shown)


brauner:
@flx42, excellent catch! Thanks!

brauner:
jenkins: test this please

flx42:
Luckily `nvidia-container-cli` crashed and burned when called twice on the same rootfs.
I guess this can also happen if you declare our hook twice in the config, is this a case hooks should try to guard against?

brauner:
@flx42, yeah, if you declare the same hook twice it get's run twice but there's nothing that LXC would be able to handle because it cannot easily tell if users do this intentionally or not.

PR #2235, opened 2018-03-25
conf: simplify autodev

brauner:
This function was way more syscall heavy than it needed to be.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1203,28 +1203,28 @@ static int mount_autodev(const char *name, const struct lxc_rootfs *rootfs,
 	return 0;
 }
 
-struct lxc_devs {
+struct lxc_device_node {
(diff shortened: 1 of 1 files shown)


src/lxc/conf.c
```diff
@@ -1235,7 +1235,7 @@ static int lxc_fill_autodev(const struct lxc_rootfs *rootfs)
 	if (!dir_exists(path))
 		return 0;
 
-	INFO("Populating \"/dev\"");
+	INFO("Populating devtmpfs");
```
stgraber: It's specifically NOT a devtmpfs though, right?

brauner: Technically it's a tmpfs, yes. I just thought that devtmpfs is nicer. But I can update or you can merge now and I'll send a small update later

PR #2247, opened 2018-03-31
confile: expand lxc.environment

brauner:
When a bare environment variable is specified then retrieve the value from the
current environment. For example, setting

lxc.environment = PATH

will cause LXC to inherit the value of PATH from the current environment.

Suggested-by: Jonathan Calmels <jcalmels@nvidia.com>


stgraber:
Misses documentation

brauner:
@stgraber, fixed.

PR #2258, opened 2018-04-04
fix signal sending in lxc.init

tych0:
The problem here is that these two clauses were ordered backwards: we first
check if the signal came from not the init pid, and if it did, then we give
a notice and return. The comment notes that this is intended to protect
against SIGCHLD, but we don't in fact know if the signal is a SIGCHLD yet,
because that's tested in the next hunk.

The symptom is that if I e.g. send SIGTERM from the outside world to the
container init, it ignores it and gives this notice. If we re-order these
clauses, it forwards non SIGCHLD signals, and ignores SIGCHLD signals from
things that aren't the real container process.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -380,18 +380,18 @@ static int signal_handler(int fd, uint32_t events, void *data,
 		return hdlr->init_died ? LXC_MAINLOOP_CLOSE : 0;
 	}
 
+	if (siginfo.ssi_signo != SIGCHLD) {
+		kill(hdlr->pid, siginfo.ssi_signo);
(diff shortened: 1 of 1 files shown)


brauner:
Is there anything in git history/blame that explains why we did this in the order?

hallyn:
There was probably some concern about signals coming from other containers.  For instance, it could happen through a write to a file along with F_SETOWN.

hallyn:
Commit 3a9e949f6d4cad085d645b2efbc126cdbc48773a by Tycho seems to have moved that block, and added the comment  :)

+1 from me.

hallyn:
Merged because really that commit did introduce the regression.

Thanks, Tycho.

tych0:
Heh, whoops :)

PR #2259, opened 2018-04-06
rootfs pinning: On NFS, make file hidden but don't delete it

gjaekel:
On NFS, avoid random names of the root pin file due to "NFS silly renaming" but use a fixed hidden name instead.

IMHO this is a better solution because one is able to relate the anyway vislble file to a purpose.

Signed-off-by: <G.Jaekel@DNB.DE>

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -539,10 +539,11 @@ int run_script(const char *name, const char *section, const char *script, ...)
 }
 
 /* pin_rootfs
- * if rootfs is a directory, then open ${rootfs}/lxc.hold for writing for
+ * if rootfs is a directory, then open ${rootfs}/.lxc-keep for writing for
  * the duration of the container run, to prevent the container from marking
  * the underlying fs readonly on shutdown. unlink the file immediately so
- * no name pollution is happens
+ * no name pollution is happens.
+ * don't unlink on NFS to avoid random named stale handles.
  * return -1 on error.
  * return -2 if nothing needed to be pinned.
  * return an open fd (>=0) if we pinned it.
@@ -552,6 +553,7 @@ int pin_rootfs(const char *rootfs)
 	int fd, ret;
 	char absrootfs[MAXPATHLEN], absrootfspin[MAXPATHLEN];
 	struct stat s;
+	struct statfs sfs;
 
 	if (rootfs == NULL || strlen(rootfs) == 0)
 		return -2;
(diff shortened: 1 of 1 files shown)


brauner:
@gjaekel, thanks for the patch however, you forgot to sign of your commit. If you go into the branch and do:

    git commit --amed -s
    git pus -f <from> <to>

all should be good.

brauner:
Let me ask a few question on the patch. On the mailing list [1] you said:


> Serge, reject the patch because it introduces a varying name for each use of the rootfs. I agree that one name will be enough. But the ugly NFS quirk with the visible file for a stale handle for each unlinked file will remain.

What exactly is the problem with NFS silly rename? The file will be hidden, right, i.e. it will be `.nfsXXXXX` or something. Is it only the fact that this file is around that is bothering you or is this causing an actual issue?

[1]: https://lists.linuxcontainers.org/pipermail/lxc-devel/2018-March/017544.html

gjaekel:
You're welcome!

My eyes "problem" is that such a file is just "hidden" in the terms of Unix, .i.e. it will be "still seen" by a certain class of operational filesystem usecases, such as cloning, archiving, snapshots and so on. But if you copy such a silly renamed file, the copy will be a plain vanilla file without any "magic" - it will remain and you can't distinguish between later on. In addition, by getting renamed to random, you loose the information what "the hell" this file is used for. Or may get "more and more of the same" in your backup. This might be called the "actuall issues".

And also take it from the other side and ask yourself what is the reason to use the "unlink trick" to make this file really invisible on common filesystems.

If you would use NFS als the one-and-only like me, you're facing this "problem" from time to time, again -- because it's not common known, of corse. And if you have to deal with it, you would try to minimize this.

Please notice, that I implement this as a "special" behavior that just will be active on NFS and will not affect the very most of other users.  Without further changes the file will even not be deleted on container shutdown. But this bother me much less than the current state. Of course, we may add an "remove" operation on shutdown.

gjaekel:
Sorry, I don't find a way to retroactively change a commit comment using GitHub's web interface yet.

brauner:
Ok, as I said you can:

    git commit --amed -s
    git push -f ...

brauner:
If you can't find a way yourself either close this branch and open a new one or simply tell me what I should use as your Signed-off-by:

Guido Jäkel <me@somewhere.com>

or something. The mail address should be reachable.

gjaekel:
Thank you, As we start a discussion here, i want to avoid to drop this. You may use

Signed-off-by: G.Jaekel@DNB.DE

as in the first comment. I edited that, but with no influence of the commit comment itself.

brauner:
@gjaekel, no worries. Thanks for going through the effort. :) @hallyn, you're fine with the patch and the reasoning for it?

gjaekel:
@hallyn: 
A) Would you like to have a remove at shutdown? I may try to add an unconditional, "stateless" `ret = remove();` at the right place™. I don't look at this yet, therefore I don't know if it's required to save `absrootfspin` in a global context to archive this.

B) On NFS, this file then may be more easily -- due to the fixed name -- used to pass informations between the container and host. But in principle, this also holds if it's "silly renamed" to my knowledge.

gjaekel:
If you agree in "agile development", you may accept this pull and I may add the remove at shutdown later in a second sprint.

A "well known named" file might be (miss)used to exchange data between host and container, e.g. some host environment information to be known inside the container. Or to signal states of the container or it's applications to the host. I just want to point out that this might be in the domain of security, but I don't see a "hole" here.

hallyn:
Oh - we've never audited the boundary to try and prevent container to host information leaks, so while that's interesting for some use cases, it certainly shouldn't block this :)

Will merge, thanks!

PR #2260, opened 2018-04-06
Fix temp file creation.

thmo:
lxc_make_tmpfile() uses mkstemp() internally, and thus expects the
template to contain 'XXXXXX' and be writable.

The existing code in make_anonymous_mount_file() did not work in
case the memfd_create() syscall was not available.

Furthermore, mkstemp() modifies its template argument, hence it
should not be a constant, or undefined behavior can happen. Fixed
both occurrences.


brauner:
@thmo, thanks for spotting this!
I'm just going to grab your branch and make a few coding style changes and merge!

thmo:
Applied via #2261, thanks.

PR #2261, opened 2018-04-06
Fix temp file creation

brauner:
lxc_make_tmpfile() uses mkstemp() internally, and thus expects the
template to contain 'XXXXXX' and be writable.

The existing code in make_anonymous_mount_file() did not work in
case the memfd_create() syscall was not available.

Furthermore, mkstemp() modifies its template argument, hence it
should not be a constant, or undefined behavior can happen. Fixed
both occurrences.

//cc @thmo I tried to push directly to your branch but I think you opted out of this feature. So I'm doing a new pr here. It's really just coding OCD stuff. :)

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2286,9 +2286,12 @@ FILE *make_anonymous_mount_file(struct lxc_list *mount)
 
 	fd = memfd_create(".lxc_mount_file", MFD_CLOEXEC);
 	if (fd < 0) {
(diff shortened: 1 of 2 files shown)


brauner:
Your signed-off by is retained.

brauner:
Closes  #2260.

thmo:
LGTM

brauner:
jenkins: test this please

PR #2270, opened 2018-04-11
attach: try to always drop supplementary groups

brauner:

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -873,7 +873,7 @@ static int attach_child_main(struct attach_clone_payload *payload)
 	}
 
 	ret = lxc_setgroups(0, NULL);
-	if (ret < 0)
+	if (ret < 0 && errno != EPERM)
 		goto on_error;
 
 	if ((init_ctx->container && init_ctx->container->lxc_conf &&


brauner:
@hallyn, that should hopefully address your concerns. :)

src/lxc/attach.c
```diff
@@ -873,7 +873,7 @@ static int attach_child_main(struct attach_clone_payload *payload)
 	}
 
 	ret = lxc_setgroups(0, NULL);
-	if (ret < 0)
+	if (ret < 0 && errno != EPERM)
```
hallyn: Sorry to cause more work - but I do think it's worth adding a debug msg in the case where it failed with EPERM.

On the one hand we don't want to completely break containers who are under deny_setgroup.  OTOH if i'm attaching to a container and it unexpectedly fails to drop some groups that i didn't want in the container, that's actually a big deal i guess.

brauner: Back when I implemented the `lxc_setgroups()` function I made it such that it always logs :):
```
/* Simple covenience function which enables uniform logging. */
int lxc_setgroups(int size, gid_t list[])
{
	if (setgroups(size, list) < 0) {
		SYSERROR("Failed to setgroups().");
		return -errno;
	}
	NOTICE("Dropped additional groups.");

	return 0;
}
```

PR #2273, opened 2018-04-13
conf: fix net type checks in run_script_argv()

aither64:
I noticed that with `lxc.hook.version = 1` set, environment variable `LXC_NET_PEER` wasn't set for veth interfaces in `lxc.net.[i].script.{up,down}` hooks. It seems to me that all net type check conditions in `run_script_argv()` should test for string equality, not inequality.


brauner:
jenkins: test this please

brauner:
@aither64, excellent catch. Surprised that no one noticed this yet.

PR #2279, opened 2018-04-15
do_lxcapi_create: set umask

kunkku:
Always use 022 as the umask when creating the rootfs directory and
executing the template. A too loose umask may cause security issues.
A too strict umask may cause programs to fail inside the container.


brauner:
jenkins: test this please

brauner:
lgtm

brauner:
But isn't the `umask` already set in `do_create_container_dir`?

kunkku:
Yes, but it's reset right after the mkdir for the container directory. The rootfs directory and any subdirectories created by the template script may get wrong permissions.

brauner:
The `rootfs` should be owned by `755` usually. Your system's `rootfs` will have the same permissions.

PR #2281, opened 2018-04-15
seccomp: handle arch inversion - The Architecture Strikes Back

brauner:


brauner:
@pcmoore, sending a big than you from over here. :)

PR #2236, opened 2018-03-25
Allow passing action scripts to CRIU

ehmm:
Add option to lxc-checkpoint to allow passing of action script to CRIU from lxc-checkpoint command

PR #2287, opened 2018-04-24
Also pass action scripts to CRIU on checkpointing

thyth:
I noticed that support for passing action scripts to CRIU via lxc-checkpoint was added in #2236 , but the action script is only passed to CRIU in restore mode. This change adds passing of the action script in checkpoint mode as well.


brauner:
jenkins: test this please

brauner:
Thanks! :)

PR #2291, opened 2018-04-26
Fix tool_utils.c build when HAVE_SETNS is unset

pkun:
When HAVE_SETNS is unset the tool_utils.c can't be build because it can't find setns(). The inline setns() is defined in lxc/utils.h but isn't defined in lxc/tools/tool_utils.h. The patch fixes it. Another minor patch fixes headers in lxc/tools to use their own defines to disable double including ( __LXC_TOOLS_UTILS_H).


brauner:
Thanks for the patch but before I can even look at it you need to sign-off your commits kernel DCO style. Should be as easy as `git commit -s`.

PR #2293, opened 2018-04-26
Fix tool_utils.c build when HAVE_SETNS is unset

pkun:
Add inline setns() function to tool_utils.h. Without it
tool_utils.c can't be build when HAVE_SETNS is unset.


brauner:
jenkins: test this please

brauner:
Thanks!

PR #2300, opened 2018-05-02
Mount injection API 

LizaTretyakova:
Hi there!
This is an initial step for moving the mount injection stuff from LXD to LXC. 
In this PR, the new `lxc_container` API method `mount` is introduced, alongside with the supporting config option `shmounts` for `lxc.mount.auto` (because in order to enable mount injection in the running container, a shared mount point is established between on a startup) and the necessary utils.

UPD: also introduce the `umount` API method.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2714,6 +2714,8 @@ struct lxc_conf *lxc_conf_init(void)
 	new->lsm_aa_profile = NULL;
 	new->lsm_se_context = NULL;
 	new->tmp_umount_proc = false;
+	new->lxc_shmount.path_host = NULL;
+	new->lxc_shmount.path_cont = NULL;
 
 	/* if running in a new user namespace, init and COMMAND
 	 * default to running as UID/GID 0 when using lxc-execute */
@@ -4041,6 +4043,8 @@ void lxc_conf_free(struct lxc_conf *conf)
 	lxc_clear_procs(conf, "lxc.proc");
 	free(conf->cgroup_meta.dir);
 	free(conf->cgroup_meta.controllers);
+	free(conf->lxc_shmount.path_host);
+	free(conf->lxc_shmount.path_cont);
 	free(conf);
 }
 
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -648,22 +648,21 @@ unsigned long add_required_remount_flags(const char *s, const char *d,
 #endif
 }
 
-static int add_shmount_to_list(struct lxc_conf *conf) {
+static int add_shmount_to_list(struct lxc_conf *conf)
+{
 	char new_mount[MAXPATHLEN];
 	/* Offset for the leading '/' since the path_cont
-	 * is absolute inside the container */
-	int ret = -1, offset = 1;
+	 * is absolute inside the container.
+	 */
+	int offset = 1, ret = -1;
 
-	ret = snprintf(new_mount, sizeof(new_mount), "%s %s none bind,create=dir 0 0",
-				   conf->shmount.path_host, conf->shmount.path_cont + offset);
+	ret = snprintf(new_mount, sizeof(new_mount),
+		       "%s %s none bind,create=dir 0 0", conf->shmount.path_host,
+		       conf->shmount.path_cont + offset);
 	if (ret < 0 || (size_t)ret >= sizeof(new_mount))
 		return -1;
 
-	ret = add_elem_to_mount_list(new_mount, conf);
-	if (ret < 0)
-		ERROR("Failed to add new mount \"%s\" to the config", new_mount);
-
-	return ret;
+	return add_elem_to_mount_list(new_mount, conf);
 }
 
 static int lxc_mount_auto_mounts(struct lxc_conf *conf, int flags, struct lxc_handler *handler)
@@ -806,7 +805,7 @@ static int lxc_mount_auto_mounts(struct lxc_conf *conf, int flags, struct lxc_ha
 		int ret = add_shmount_to_list(conf);
 		if (ret < 0) {
 			ERROR("Failed to add shmount entry to container config");
-			return ret;
+			return -1;
 		}
 	}
 
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -650,19 +650,13 @@ unsigned long add_required_remount_flags(const char *s, const char *d,
 
 static int add_shmount_to_list(struct lxc_conf *conf) {
 	char new_mount[MAXPATHLEN];
-	size_t len_mount;
 	/* Offset for the leading '/' since the path_cont
 	 * is absolute inside the container */
 	int ret = -1, offset = 1;
 
-	/* +1 for the separating whitespace */
-	len_mount = strlen(conf->shmount.path_host) + 1
-			+ strlen(conf->shmount.path_cont) - offset
-			+ sizeof(" none bind,create=dir 0 0") - 1;
-
-	ret = snprintf(new_mount, len_mount + 1, "%s %s none bind,create=dir 0 0",
+	ret = snprintf(new_mount, sizeof(new_mount), "%s %s none bind,create=dir 0 0",
 				   conf->shmount.path_host, conf->shmount.path_cont + offset);
-	if (ret < 0 || (size_t)ret >= len_mount + 1)
+	if (ret < 0 || (size_t)ret >= sizeof(new_mount))
 		return -1;
 
 	ret = add_elem_to_mount_list(new_mount, conf);
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -656,12 +656,12 @@ static int add_shmount_to_list(struct lxc_conf *conf) {
 	int ret = -1, offset = 1;
 
 	/* +1 for the separating whitespace */
-	len_mount = strlen(conf->lxc_shmount.path_host) + 1
-			+ strlen(conf->lxc_shmount.path_cont) - offset
+	len_mount = strlen(conf->shmount.path_host) + 1
+			+ strlen(conf->shmount.path_cont) - offset
 			+ sizeof(" none bind,create=dir 0 0") - 1;
 
 	ret = snprintf(new_mount, len_mount + 1, "%s %s none bind,create=dir 0 0",
-				   conf->lxc_shmount.path_host, conf->lxc_shmount.path_cont + offset);
+				   conf->shmount.path_host, conf->shmount.path_cont + offset);
 	if (ret < 0 || (size_t)ret >= len_mount + 1)
 		return -1;
 
@@ -2747,8 +2747,9 @@ struct lxc_conf *lxc_conf_init(void)
 	new->lsm_aa_profile = NULL;
 	new->lsm_se_context = NULL;
 	new->tmp_umount_proc = false;
-	new->lxc_shmount.path_host = NULL;
-	new->lxc_shmount.path_cont = NULL;
+	new->tmp_umount_proc = 0;
+	new->shmount.path_host = NULL;
+	new->shmount.path_cont = NULL;
 
 	/* if running in a new user namespace, init and COMMAND
 	 * default to running as UID/GID 0 when using lxc-execute */
@@ -4076,8 +4077,8 @@ void lxc_conf_free(struct lxc_conf *conf)
 	lxc_clear_procs(conf, "lxc.proc");
 	free(conf->cgroup_meta.dir);
 	free(conf->cgroup_meta.controllers);
-	free(conf->lxc_shmount.path_host);
-	free(conf->lxc_shmount.path_cont);
+	free(conf->shmount.path_host);
+	free(conf->shmount.path_cont);
 	free(conf);
 }
 
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -23,6 +23,7 @@
 
 #define _GNU_SOURCE
 #include "config.h"
+#include "confile.h"
 
 #include <arpa/inet.h>
 #include <dirent.h>
@@ -647,6 +648,30 @@ unsigned long add_required_remount_flags(const char *s, const char *d,
 #endif
 }
 
+static int add_shmount_to_list(struct lxc_conf *conf) {
+	char new_mount[MAXPATHLEN];
+	size_t len_mount;
+	/* Offset for the leading '/' since the path_cont
+	 * is absolute inside the container */
+	int ret = -1, offset = 1;
+
+	/* +1 for the separating whitespace */
+	len_mount = strlen(conf->lxc_shmount.path_host) + 1
+			+ strlen(conf->lxc_shmount.path_cont) - offset
+			+ sizeof(" none bind,create=dir 0 0") - 1;
+
+	ret = snprintf(new_mount, len_mount + 1, "%s %s none bind,create=dir 0 0",
+				   conf->lxc_shmount.path_host, conf->lxc_shmount.path_cont + offset);
+	if (ret < 0 || (size_t)ret >= len_mount + 1)
+		return -1;
+
+	ret = add_elem_to_mount_list(new_mount, conf);
+	if (ret < 0)
+		ERROR("Failed to add new mount \"%s\" to the config", new_mount);
+
+	return ret;
+}
+
 static int lxc_mount_auto_mounts(struct lxc_conf *conf, int flags, struct lxc_handler *handler)
 {
 	int i, r;
@@ -780,6 +805,14 @@ static int lxc_mount_auto_mounts(struct lxc_conf *conf, int flags, struct lxc_ha
 						cg_flags)) {
 			SYSERROR("Failed to mount \"/sys/fs/cgroup\"");
 			return -1;
+		}
+	}
+
+	if (flags & LXC_AUTO_SHMOUNTS_MASK) {
+		int ret = add_shmount_to_list(conf);
+		if (ret < 0) {
+			ERROR("Failed to add shmount entry to container config");
+			return ret;
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


brauner:
jenkins: test this please

brauner:
jenkins: ok to test

src/lxc/start.c
```diff
@@ -1549,6 +1613,21 @@ static int lxc_spawn(struct lxc_handler *handler)
 		}
 	}
 
+	if (conf->lxc_shmount.path_host) {
+		if(!conf->lxc_shmount.path_cont) {
```
brauner: Would probably be nicer if you do a:
```
 if (conf->lxc_shmount.path_host || !conf->lxc_shmount.path_cont)
```

src/lxc/utils.c
```diff
@@ -1159,16 +1152,30 @@ int detect_shared_rootfs(void)
...
+/*
+ * Detect whether / is mounted MS_SHARED.  The only way I know of to
+ * check that is through /proc/self/mountinfo.
+ * I'm only checking for /.  If the container rootfs or mount location
+ * is MS_SHARED, but not '/', then you're out of luck - figuring that
+ * out would be too much work to be worth it.
+ */
+int detect_shared_rootfs(void)
+{
+	if(is_shared_mountpoint("/"))
```
brauner: Missing space between `if` and `(`. :)

src/lxc/conf.c
```diff
@@ -630,6 +631,39 @@ unsigned long add_required_remount_flags(const char *s, const char *d,
...
+	/* +1 for the termintating '\0' */
+	new_mount = malloc(len_mount + 1);
+	if (!new_mount) {
+		SYSERROR("Not enough memory");
+		return -ENOMEM;
+	}
+
+	ret = snprintf(new_mount, len_mount + 1, "%s %s none bind,create=dir 0 0",
+				   conf->lxc_shmount.path_host, conf->lxc_shmount.path_cont + offset);
+	if (ret < 0 || ret >= len_mount + 1) {
```
brauner: You should cast `(size_t)ret`.

src/lxc/conf.c
```diff
@@ -630,6 +631,39 @@ unsigned long add_required_remount_flags(const char *s, const char *d,
 #endif
 }
 
+static int add_shmount_to_list(struct lxc_conf *conf) {
+	char *new_mount;
+	size_t len_mount;
+	/* Offset for the leading '/' since the path_cont
+	 * is absolute inside the container */
```
brauner: This is guaranteed because we control the path, right?

src/lxc/conf.c
```diff
@@ -630,6 +631,39 @@ unsigned long add_required_remount_flags(const char *s, const char *d,
...
+	int ret = -1, offset = 1;
+
+	/* +1 for the separating whitespace */
+	len_mount = strlen(conf->lxc_shmount.path_host) + 1
+			+ strlen(conf->lxc_shmount.path_cont) - offset
+			+ sizeof(" none bind,create=dir 0 0") - 1;
+	/* +1 for the termintating '\0' */
+	new_mount = malloc(len_mount + 1);
+	if (!new_mount) {
```
brauner: Again, I wonder if we should be lazy bastard and use `MAXPATHLEN`. Hm...

src/lxc/lxccontainer.c
```diff
@@ -4691,6 +4712,136 @@ static bool do_lxcapi_restore(struct lxc_container *c, char *directory, bool ver
...
+	size_t len;
+	int ret = -1, fd = -EBADF;
+
+	if (!c || !c->lxc_conf) {
+		ERROR("Container or configuration is NULL");
+		return -EINVAL;
+	}
+
+	len = strlen(c->lxc_conf->lxc_shmount.path_host) + sizeof("/lxcmountXXXXXX") - 1;
+	template = malloc(len + 1);
```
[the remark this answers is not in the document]
LizaTretyakova: Sounds not bad to me. Probably, if `len` is more than `MAXPATHLEN`, something went wrong anyway :) And, I guess, calling `malloc` and `free` has its overhead (apart from the risk not to call `free`).

src/lxc/conf.c
LizaTretyakova: Yes, here https://github.com/lxc/lxc/pull/2300/files#diff-2f3344d3cb9700d11140decc560aaacaR1796. We could probably mention it somewhere else, closer to the declaration of the struct or something, so that if we ever want to change this we didn't forget the (undocumented) requirement.

src/lxc/lxccontainer.c
```diff
@@ -4691,6 +4712,136 @@ static bool do_lxcapi_restore(struct lxc_container *c, char *directory, bool ver
...
+	pid_t pid, init_pid;
+	size_t len;
+	int ret = -1, fd = -EBADF;
+
+	if (!c || !c->lxc_conf) {
+		ERROR("Container or configuration is NULL");
+		return -EINVAL;
+	}
+
+	len = strlen(c->lxc_conf->lxc_shmount.path_host) + sizeof("/lxcmountXXXXXX") - 1;
```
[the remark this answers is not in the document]
LizaTretyakova: I wonder if we should do such checks before other `strlen`s as well, like here https://github.com/lxc/lxc/pull/2300/files#diff-9064c4ae46015f2c78f9deff397d67c7R642 or here https://github.com/lxc/lxc/pull/2300/files#diff-b0b78d751da8af17e80738fb43409c80R4788?

src/lxc/start.c
LizaTretyakova: I don't think so because the enclosing `if` is bigger than just this check. Or you mean to do the `if (conf->lxc_shmount.path_host)` twice?

brauner: We usually just try to avoid nesting `if`-conditions too much. Not very pleasant to read usually. :)

LizaTretyakova: Yeah, right

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

src/lxc/lxccontainer.c
```diff
@@ -4691,6 +4712,119 @@ static bool do_lxcapi_restore(struct lxc_container *c, char *directory, bool ver
...
+	len = strlen(c->lxc_conf->lxc_shmount.path_host) + sizeof("/.lxcmount_XXXXXX") - 1;
+
+	ret = snprintf(template, len + 1, "%s/.lxcmount_XXXXXX", c->lxc_conf->lxc_shmount.path_host);
+	if (ret < 0 || (size_t)ret >= len + 1) {
+		SYSERROR("Error writing shmounts tempdir name");
+		goto out;
+	}
+
+	/* Create a temporary dir under the shared mountpoint */
+	sret = mkdtemp(template);
```
brauner: I just realized that your API addition will currently only work for directory-based mounts. Here you need to create either a temporary directory or a temporary file depending on whether the source of the mount is a file or a directory.
Your tests don't check for bind-mounting a file into a container. The test that looks like it does it is:
```
ret = c->mount(c, "/dev/urandom", "/mnt/mount_injection_test_urandom", "devtmpfs", 0, NULL, &mnt);
```
But it's not mounting a device actually since your telling the kernel to mount `devtmpfs` which is a directory-based mount. That means the kernel will also ignore the `/dev/urandom` source path. It will be listed as the source but it actually isn't. :)

The test that you likely want in addition to the `devtmpfs` one after changing the new API function is:
```
ret = c->mount(c, "/dev/urandom", "/mnt/mount_injection_test_urandom", NULL, MS_BIND, NULL, &mnt);
```

brauner: You can use 

https://github.com/lxc/lxc/blob/9ab207cad74f77ce643d98334107949ade97463b/src/lxc/utils.c#L2388
to create the temporary file.

brauner: Also, I would prefer it if the temporary directory or temporary file you're creating as the mount target would have permission `0000` instead of `0700` and `0600` since we don't need any perms at all. :) Unless this doesn't work.

LizaTretyakova: Somehow I didn't find any analogues of `mkdir_p()` for files so I adopted some nice utils from LXD code [here](https://github.com/lxc/lxd/blob/master/lxd/main_forkmount.go#L51) (and also their logic regarding permissions of the files being created inside the container with 0755).

LizaTretyakova:
Looks good to me

src/lxc/conf.h
```diff
@@ -365,6 +368,13 @@ struct lxc_conf {
 
 	/* procs */
 	struct lxc_list procs;
+
+	struct lxc_shmount {
```
hallyn: adding 'lxc_' prefix isn't helpful since this is in lxc codebase.  How about shmount_path or something?

src/lxc/conf.c
```diff
@@ -630,6 +631,30 @@ unsigned long add_required_remount_flags(const char *s, const char *d,
...
+	/* Offset for the leading '/' since the path_cont
+	 * is absolute inside the container */
+	int ret = -1, offset = 1;
+
+	/* +1 for the separating whitespace */
+	len_mount = strlen(conf->lxc_shmount.path_host) + 1
+			+ strlen(conf->lxc_shmount.path_cont) - offset
+			+ sizeof(" none bind,create=dir 0 0") - 1;
+
+	ret = snprintf(new_mount, len_mount + 1, "%s %s none bind,create=dir 0 0",
```
hallyn: you never ensured len_mount < MAXPATHLEN

src/lxc/confile.c
```diff
@@ -1779,6 +1787,14 @@ static int set_config_mount_auto(const char *key, const char *value,
 
 		lxc_conf->auto_mounts &= ~allowed_auto_mounts[i].mask;
 		lxc_conf->auto_mounts |= allowed_auto_mounts[i].flag;
+		if (is_shmounts) {
+			lxc_conf->lxc_shmount.path_host = strdup(token + (sizeof("shmounts:") - 1));
```
hallyn: check for strdup failure here and below

src/lxc/lxccontainer.c
```diff
@@ -4698,6 +4719,119 @@ static bool do_lxcapi_restore(struct lxc_container *c, char *directory, bool ver
...
+			ERROR("Failed to enter mount namespace");
+			_exit(EXIT_FAILURE);
+		}
+
+		suff = strrchr(template, '/');
+		if (!suff)
+			_exit(EXIT_FAILURE);
+
+		len = strlen(c->lxc_conf->lxc_shmount.path_cont) + sizeof("/.lxcmount_XXXXXX") - 1;
+		ret = snprintf(path, len + 1, "%s%s", c->lxc_conf->lxc_shmount.path_cont, suff);
```
hallyn: You're letting the *source* length dictate the length to be printed.  The size of the target (path) should be used.

src/lxc/lxccontainer.c
```diff
@@ -4698,6 +4719,299 @@ static bool do_lxcapi_restore(struct lxc_container *c, char *directory, bool ver
...
+		close(fd);
+
+	return ret;
+}
+
+WRAP_API_6(int, lxcapi_mount, const char *, const char *, const char *,
+	   unsigned long, const void *, struct lxc_mount *)
+
+static int do_lxcapi_umount(struct lxc_container *c, const char *target,
+			    unsigned long mountflags, struct lxc_mount *mnt)
```
LizaTretyakova: NIT, but maybe it'd be better to rename `mountflags` to just `flags` since no mount is performed here

src/lxc/conf.c
Blub: According to `snprintf(3)`, it returns the number of bytes which _would_ have been printed if the buffer had been large enough, so maybe skip the length calculation, pass `sizeof(new_mount)` and verify the result is < that?

src/lxc/lxccontainer.c
brauner: Yes, agreed. I wonder if we should also have a custom struct `struct lxc_umount`?

src/lxc/lxccontainer.c
```diff
@@ -4698,6 +4719,299 @@ static bool do_lxcapi_restore(struct lxc_container *c, char *directory, bool ver
...
+	}
+
+	ret = mkdir(dest, 0000);
+	if (ret < 0) {
+		SYSERROR("Failed to create directory \"%s\"", dest);
+		return -1;
+	}
+
+	return 0;
+}
```
Blub: These two helpers are quite large and also quite similar. I wonder if maybe one suffices:
```C
static int ensure_dummy(const char *dest, mode_t mode)
{
	int ret;
	struct stat sb;

	ret = stat(dest, &sb);
	if (ret == 0) {
		if (!S_ISDIR(sb.st_mode) == !S_ISDIR(mode))
			return 0;

		ret = S_ISDIR(sb.st_mode) ? rmdir(dest) : unlink(dest);
		if (ret < 0) {
			SYSERROR("Failed to remove \"%s\"", dest);
			return -1;
		}
	}

	ret = S_ISDIR(mode) ? mkdir(dest, 0000) : mknod(dest, S_IFREG, 0);
	if (ret < 0) {
		SYSERROR("Failed to create destination \"%s\"", dest);
		return -1;
	}

	return 0;
}
```

At the same time I wonder if we should even bother with the verbose checking and just shorten it to:
```C
static int ensure_dummy(const char *dest, mode_t mode)
{
	int ret;
	struct stat sb;

	ret = stat(dest, &sb);
	if (ret == 0) {
		if (!S_ISDIR(sb.st_mode) == !S_ISDIR(mode))
			return 0;

		// If neither works, mount mkdir/mknod will fail anyway
		(void)rmdir(dest);
		(void)unlink(dest);
	}

	ret = S_ISDIR(mode) ? mkdir(dest, 0000) : mknod(dest, S_IFREG, 0);
	if (ret < 0) {
		SYSERROR("Failed to create destination \"%s\"", dest);
		return -1;
	}

	return 0;
}
```

or even just

```C
static int ensure_dummy(const char *dest, mode_t mode)
{
	int ret;
	struct stat sb;

	(void)rmdir(dest);
	(void)unlink(dest);
	if (S_ISDIR(mode))
		ret = mkdir(dest, 0000);
	else
		ret = mknod(dest, S_IFREG, 0);
	if (ret < 0) {
		SYSERROR("Failed to create destination \"%s\"", dest);
		return -1;
	}

	return 0;
}
```

Simply because AFAICT the shared mount directory is supposed to be internal to lxc and whether the removal of a previously different type of mount point entry failed, or simply its creation, won't usually point out any particular error that wouldn't be obvious from a failing mkdir/mknod already anyway?

src/lxc/lxccontainer.c
```diff
@@ -4698,6 +4719,299 @@ static bool do_lxcapi_restore(struct lxc_container *c, char *directory, bool ver
...
+		return -EINVAL;
+	}
+
+	if (!c->lxc_conf->lxc_shmount.path_host) {
+		ERROR("Host path to shared mountpoint must be specified in the config\n");
+		return -EINVAL;
+	}
+	len = strlen(c->lxc_conf->lxc_shmount.path_host) + sizeof("/.lxcmount_XXXXXX") - 1;
+
+	ret = snprintf(template, len + 1, "%s/.lxcmount_XXXXXX", c->lxc_conf->lxc_shmount.path_host);
```
Blub: This `snprintf` + `strlen` pattern is the same as the one I previously commented on, the strlen part could be simplified away I think.

src/lxc/lxccontainer.c
brauner: I prefer your last version, though in terms of coding style I'd prefer if this were:
```C
static int ensure_dummy(const char *dest, mode_t mode)
{
	int ret;

	(void)remove(dest);

	if (S_ISDIR(mode))
		ret = mkdir(dest, 0000);
	else
		ret = mknod(dest, S_IFREG | 0000, 0);
	if (ret < 0) {
		SYSERROR("Failed to create dummy mount target \"%s\"", dest);
		return -1;
	}

	return 0;
}
```

src/lxc/lxccontainer.c
```diff
@@ -4698,6 +4719,299 @@ static bool do_lxcapi_restore(struct lxc_container *c, char *directory, bool ver
...
+		sret = mkdtemp(template);
+		if (!sret) {
+			SYSERROR("Could not create shmounts temporary dir");
+			goto out;
+		}
+
+		ret = chmod(template, 0000);
+		if (ret < 0) {
+			SYSERROR("Could not chmod shmounts temporary dir \"%s\"", template);
+			goto out;
```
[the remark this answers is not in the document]
brauner: Yeah, we probably don't need to care. I just like `0000` permission model that's why I suggested it to @LizaTretyakova but yeah, we can do without it.

src/lxc/lxccontainer.c
Blub: Ah yes, there's `remove()`, as well :-)

src/lxc/lxccontainer.c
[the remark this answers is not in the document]
Blub: I think what mostly matters are the permission of the shared mount directory itself. You don't want someone injecting symlinks there ;-)

src/lxc/lxccontainer.c
LizaTretyakova: Actually, in this case `ensure_dummy` is rather small, and it's probably easier to make it just a part of the `create_mount_target`.

src/lxc/lxccontainer.c
LizaTretyakova: Hmm. After I thought about it, I cannot think of something largely different which user would want to pass to `umount` but not to `mount`. What might it be?

brauner:
@LizaTretyakova, I think you need to break-out your rebasing skills. :) Looks like my recent cgroup rework introduced a merge conflict. :)

LizaTretyakova:
@brauner Sure ;)
hg

brauner:
Time's up, merging this now.

PR #2311, opened 2018-05-09
execute: don't bind in lxc.init.static

tych0:
Instead of bind mounting in the init and execing it, let's just use
fexecve. There are at least two reasons to prefer this:

* We don't pollute the users rootfs (even if only temporarily). Even if we
  remove the file, it still changes the mtime of /tmp (or wherever we would
  bind it), which users who are sensitive to any changes in the rootfs will
  see.
* It allows us to get rid of a whole bunch of code.


hallyn:
How heavily did you test this?  I definitely like this - but I worry about the note in the manpage about /proc having to be available for it to work, and the fact that we mess with /proc quite a bit :)  Note that when setting up apparmor we don't even assume proc is available.

tych0:
The execveat syscall went into kernel 3.19, so I think the bits about proc don't apply in most cases. What's liblxc's minimum supported kernel version?

tych0:
Oh, but glibc uses /proc until 2.26. Well, I tested it on a system running 2.23, and it worked. I haven't actually tested it on a system with 2.27, but it should work better then, right? :)

tych0:
I guess it probably won't (?) work if the user doesn't specify to mount /proc in their container. So maybe what we should just do is call the raw syscall, if the minimum supported kernel version is >= 3.19,

brauner:
`execveat()` was introduced at different points for different architectures. By that I mean `x86` and friends got it around `3.19`, `alpha` with `4.2`, `hppa` and `microblaze` with `4.0`. So 
```
#ifdef __NR_execveat
```
will yield true but you still might get `ENOSYS`. In this case you should fallback to using `/proc`.

tych0:
...but do we care about microblaze? Sounds like there's no real advantage to using that vs. libc, then.

brauner:
So I know this might be one of my crazy things but I'd like to keep supporting all architectures that upstream supports. We didn't think people would care a lot about `mips` either now we have had patches to add all kinds of `seccomp` support for it.
I'm in favor of using `fexecve()` though. I think it's a huge improvement but we need to handle the execute case without `/proc`; as annoying as it is. Application contains are definitely a usecase where sometimes `/proc` might not be mounted. If the `execveat()` syscall returns `ENOSYS` then `glibc` will try do to:
```
 char buf[sizeof "/proc/self/fd/" + sizeof (int) * 3];
  __snprintf (buf, sizeof (buf), "/proc/self/fd/%d", fd);
  /* We do not need the return value.  */
  __execve (buf, argv, envp);
```

tych0:
Sounds like we should just drop this then, and I'll hack around it in some other way. Without /proc or execveat, you still need all the bind mounting logic.

PR #2320, opened 2018-05-14
Fix resource leak CID #1248106

simos:
In function is_wlan() @ network.c, there is a file handle FILE* f.
There is a single exit point where the file handle is not fclose()-ed.
Elsewhere it is OK.


brauner:
@simos, please use  the following form for all coverity based commits:
```
commit ba1de6dbfe630d3ed49cede91846602c5f497f48
Date:   Sun Apr 29 00:50:50 2018 +0200

    coverity: #1435198

    Unchecked return value

```

PR #2323, opened 2018-05-15
Fixed resource leak in userns_exec_full()

simos:
In function userns_exec_full() @ conf.c, there is the following
pointer that is used in a double-linked list, struct lxc_list *idmap = NULL;

This pointer is malloc()-ed, and then additional objects are added to
the list using lxc_list_add_tail().
At the end of the function, the dynamically allocated memory is freed with lxc_free_idmap(idmap);

lxc_free_idmap(idmap) does not free memory of the initial memory
allocation for "idmap", therefore there is a memory leak.

The function lxc_free_idmap() is used in another place as well, and at
that place it does not free() the initial pointer (correct behaviour).
Therefore, there is a need for a free() in the function
userns_exec_full() @ conf.c.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -4082,8 +4082,10 @@ struct lxc_list *get_minimal_idmap(struct lxc_conf *conf)
 	return idmap;
 
 on_error:
-	if (idmap)
+	if (idmap) {
 		lxc_free_idmap(idmap);
+		free(idmap);
+	}
 	if (container_root_uid)
 		free(container_root_uid);
 	if (container_root_gid)


brauner:
Please use the following form for all coverity fixes:
```
commit cef701ede302a70f52bb1951f2dec2c0f88c2e5f
Date:   Mon Apr 30 12:18:23 2018 +0200

    coverity: #1435263

    Use after free

```

The information to be put in the commit message is gathered from the `CID` and `Type` columns.

brauner:
jenkins: test this please

PR #2332, opened 2018-05-15
capabilities: raise ambient capabilities

brauner:
Suggested-by: Jonathan Calmels <jcalmels@nvidia.com>

### DIFF
--- a/src/lxc/caps.c
+++ b/src/lxc/caps.c
@@ -42,6 +42,27 @@ lxc_log_define(lxc_caps, lxc);
 #define PR_CAPBSET_READ 23
 #endif
 
+/* Control the ambient capability set */
+#ifndef PR_CAP_AMBIENT
+#define PR_CAP_AMBIENT 47
+#endif
+
+#ifndef PR_CAP_AMBIENT_IS_SET
+#define PR_CAP_AMBIENT_IS_SET 1
+#endif
+
+#ifndef PR_CAP_AMBIENT_RAISE
+#define PR_CAP_AMBIENT_RAISE 2
+#endif
+
+#ifndef PR_CAP_AMBIENT_LOWER
+#define PR_CAP_AMBIENT_LOWER 3
+#endif
+
+#ifndef PR_CAP_AMBIENT_CLEAR_ALL
+#define PR_CAP_AMBIENT_CLEAR_ALL 4
+#endif
+
 int lxc_caps_down(void)
 {
 	cap_t caps;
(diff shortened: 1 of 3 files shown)


brauner:
//cc @3XX0, you definitely need a custom mount binary though. :) But I could get the following hook to run successfully with this patch:
```
lxc.hook.mount =
lxc.hook.mount = /home/chb/files/Cground/test
```
where `/home/chb/files/Cground/test.c` contains:
```
int main()
{
        int ret = mount("/dev/console", "/usr/lib/x86_64-linux-gnu/lxc/home/ubuntu/.profile", NULL, MS_BIND, NULL);
        if (ret < 0) {
                fprintf(stderr, "%s\n", strerror(errno));
                exit(EXIT_FAILURE);
        }


        exit(EXIT_SUCCESS);
}
```

3XX0:
Seems to work fine with the nvidia hook after I relaxed our checks on uid 0 :tada: 
We should probably restore the ambient caps after the start hook no? Otherwise an unprivileged container process will keep all capabilities (see below).

```
% lxc-execute cuda bash

I have no name!@cuda:/$ nvidia-smi -L          
GPU 0: GeForce GTX 1080 (UUID: GPU-4265621a-3886-6d0e-bcec-d99039a76278)
GPU 1: GeForce GTX 970 (UUID: GPU-c0ddf4f7-2a0a-718b-61a4-d26fe655b826)

I have no name!@cuda:/$ cat /proc/self/uid_map 
      1000       1000          1

I have no name!@cuda:/$ cat /proc/self/status | grep '^Cap'
CapInh: 0000003fffffffff
CapPrm: 0000003fffffffff
CapEff: 0000003fffffffff
CapBnd: 0000003fffffffff
CapAmb: 0000003fffffffff
```

brauner:
@3XX0, caps should now be lowered. :) Please test. :)

brauner:
If you want more fancy useages of the `mount` command itself it might be worth to strip:
https://github.com/karelzak/util-linux/blob/master/sys-utils/mount.c
clean of any root checks and unneeded options such as `--no-mtab` and add a minimal `lxc-mount` that ships with the lxc shared library. It should really only support mounting, no printing of stuff, nothing fancy.

src/lxc/caps.c
```diff
@@ -126,6 +147,113 @@ int lxc_caps_up(void)
...
+
+	cap_names = cap_to_text(caps, NULL);
+	if (!cap_names)
+		goto out;
+
+	TRACE("Raised %s in inheritable and ambient capability set", cap_names);
+
+out:
+
+	cap_free(cap_names);
```
hallyn: shouldn't this just be free(cap_names) ?

src/lxc/caps.c
```diff
@@ -126,6 +147,113 @@ int lxc_caps_up(void)
...
+		return -1;
+	}
+
+	for (cap = 0; cap <= CAP_LAST_CAP; cap++) {
+		cap_flag_value_t flag;
+
+		ret = cap_get_flag(caps, cap, CAP_PERMITTED, &flag);
+		if (ret < 0) {
+			if (errno == EINVAL) {
+				INFO("Last supported cap was %d", cap - 1);
```
hallyn: Don't you want to store the fact that you found the last supported cap here?

Otherwise if lxc was compiled with a CAP_LAST_CAP > the kernel was, then you'll get EPERM below when you try to raise the ambient bit.

src/lxc/caps.c
brauner: According to the manpage you're supposed to use `cap_free()`:
```
 cap_to_text()  converts the capability state in working storage identified
       by cap_p into a nul-terminated human-readable string.  This function allo‐
       cates any memory necessary to contain the string, and returns a pointer to
       the string.  If the pointer len_p is not NULL,  the  function  shall  also
       return the full length of the string (not including the nul terminator) in
       the location pointed to by len_p.  The capability state in  working  stor‐
       age,  identified  by  cap_p,  is  completely  represented in the character
       string.  When the  capability  state  in  working  storage  is  no  longer
       required,  the  caller  should  free  any  releasable  memory  by  calling
       cap_free() with the returned string pointer as an argument.
```

src/lxc/caps.c
brauner: Ah, excellent, missed that. Thanks, Serge!

brauner:
> Can you think of any case where a site would want to run with non-root uid, but want hooks to be run without privilege? I can't, but if there is any such use case then we need to clearly document that they will have to somehow drop the caps. If not then no problem.

Not really. :)

src/lxc/caps.c
hallyn: Hah - good call, thanks.

PR #2333, opened 2018-05-17
[log] support use container name as log_vmname

duguhaotian:
enhance readability of logs, use container name as log_vmname.
Now we use log config to implement.
But if we use API in mutl-threads, this function will failure.

We can use thread-local variable to store container name,
when we call lxc_container_new set this thread-local variable,
and when we call lxc_container_put to free this thread-local variable.

Signed-off-by: duguhaotian <duguhaotian@gmail.com>

PR #2338, opened 2018-05-20
log: enable per-thread container name prefix

brauner:
When using the LXC API multi-thread and users initialize a log:

struct lxc_log log;
log.name = "my-log";
lxc_log_init(&log);

all threads will have the same "my-log" prefix even though thy might call
lxc_container_new() in separate threads. There is currently no easy way to
handle per-thread container name prefixes.
To handle this carry a reference to the name of the container in struct
lxc_conf and if no log.name was set, use it by default. This way each thread
will get the container it is currently working on as a log-prefix.

Reported-by: duguhaotian <duguhaotian@gmail.com>

### DIFF
--- a/src/lxc/conf.h
+++ b/src/lxc/conf.h
@@ -228,6 +228,8 @@ struct lxc_state_client {
 };
 
 struct lxc_conf {
+	/* Pointer to the name of the container. Do not free! */
+	const char *name;
 	int is_execute;
 	char *fstab;
 	unsigned int tty;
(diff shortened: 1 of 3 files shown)


brauner:
Closes #2333.
@duguhaotian, can you check whether that is an acceptable solution to you?

brauner:
jenkins: test this please

stgraber:
jenkins: test this please

duguhaotian:
@brauner Yes, It's good solution.

PR #2345, opened 2018-05-24
Seccomp fixes

flx42:
The most important commit is eacebcc3cb8cf5d29556c85fe826096723475242, which fixes an integer overflow while parsing the argument filters.


brauner:
Thanks!

PR #2362, opened 2018-05-30
support tls in cross-compile

duguhaotian:
AC_RUN_IFELSE will fail in cross-compile,
we can use AC_COMPILE_IFELSE replace.

Signed-off-by: duguhaotian <duguhaotian@gmail.com>

### DIFF
--- a/config/tls.m4
+++ b/config/tls.m4
@@ -5,7 +5,7 @@
 AC_DEFUN([LXC_CHECK_TLS],
 [
     AC_MSG_CHECKING(for TLS)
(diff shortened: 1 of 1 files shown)


brauner:
@duguhaotian, thanks for the patch but I have a formal complaint. :) Can you please sign-off your commits with your clear name and a valid email address that ideally stays the same. Over the last couple of months we've received prs from you from all kinds of emails and names. I don't care what email you use as long as it is a valid one and it's your clear name (doesn't matter if it's the latin alphabet or the chinese alphabet :)).

PR #2363, opened 2018-05-30
conf: va_end was not called.

2xsec:
Hello,

Function 'va_end' was not called before conf.534 inside function 'run_script'.
Function 'va_start' was called at conf.c:528.


brauner:
Thanks!

PR #2364, opened 2018-05-30
confile: pointer 'retv' is dereferenced.

2xsec:
Hello,

If 'retv' pointer is null, snprintf can derefernce a null pointer.
snprintf allows null pointer with length = 0. In case of this, calling snprintf and other codes are inefficient.

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -2899,9 +2899,9 @@ static int get_config_personality(const char *key, char *retv, int inlen,
 	int fulllen = 0;
 
 	if (!retv)
-		inlen = 0;
-	else
-		memset(retv, 0, inlen);
+		return -1;
+
+	memset(retv, 0, inlen);
 
 #if HAVE_SYS_PERSONALITY_H
 	int len = 0;
@@ -2979,9 +2979,9 @@ static int __get_config_cgroup_controller(const char *key, char *retv,
 	bool get_all = false;
 
 	if (!retv)
-		inlen = 0;
-	else
-		memset(retv, 0, inlen);
+		return -1;
+
+	memset(retv, 0, inlen);
 
 	if (version == CGROUP2_SUPER_MAGIC) {
 		global_token = "lxc.cgroup2";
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

src/lxc/confile.c
```diff
@@ -3079,9 +3079,9 @@ static int get_config_idmaps(const char *key, char *retv, int inlen,
 	char buf[__LXC_IDMAP_STR_BUF];
 
 	if (!retv)
-		inlen = 0;
-	else
-		memset(retv, 0, inlen);
+		return -1;
```
brauner: The idea is that users can pass in `NULL` for `retv` in which case we report back the number of bytes that need to be allocated for the config item in question to be retrieved, i.e.:
```
	/*!
	 * \brief Retrieve the value of a config item.
	 *
	 * \param c Container.
	 * \param key Name of option to get.
	 * \param[out] retv Caller-allocated buffer to write value of \p key
	 * into (or \c NULL to determine length of value).
	 * \param inlen Length of \p retv (may be zero).
	 *
	 * \return Length of config items value, or < 0 on error.
	 *
	 * \note The caller can (and should) determine how large a buffer to allocate for
	 *  \p retv by initially passing its value as \c NULL and considering the return value.
	 *  This function can then be called again passing a newly-allocated suitably-sized buffer.
	 * \note If \p retv is NULL, \p inlen is ignored.
	 * \note If \p inlen is smaller than required, nothing will be written to \p retv and still return
	 *  the length of config item value.
	 */
	int (*get_config_item)(struct lxc_container *c, const char *key, char *retv, int inlen);
```
Your change would break this use-case. :)

2xsec: Yes, you are right. However, As far as I know, C99 allows str to be null but SUSv2 is not.
Anyway, thanks for your review.

brauner: @2xsec, ah, I see what you're getting at here. Let me think for a sec.

brauner: SUSv2 shouldn't be in use anymore but we can make it safe by passing in a non-null buffer but 0 as size.

2xsec: OK, I'll try to make it.

brauner: Oh, sorry I jumped the gun in #2365. :) I wait for your comment next time. :)

2xsec: That's OK^^

brauner:
Closed in favor of #2365. Thanks!

PR #2365, opened 2018-05-30
confile: improve strprint()

brauner:
POSIX specifies [1]:
"If the value of n is zero on a call to snprintf(), nothing shall be written,
the number of bytes that would have been written had n been sufficiently large
excluding the terminating null shall be returned, and s may be a null pointer."

But in case there are any non-sane libcs out there that do actually dereference
the buffer when when 0 is passed as length to snprintf() let's give them a
dummy buffer.

[1]: The Open Group Base Specifications Issue 7, 2018 edition
     IEEE Std 1003.1-2017 (Revision of IEEE Std 1003.1-2008)
     Copyright © 2001-2018 IEEE and The Open Group

CC: Donghwa Jeong <dh48.jeong@samsung.com>


brauner:
Closes #2364.
//cc @2xsec

PR #2370, opened 2018-06-01
Patch lxc-update-config

jsurloppe:
The current script doesn't generate a valid configuration for
lxc.network.ipv4 key, it lacking an .address part which lead to:

parse.c: lxc_file_for_each_line: 58 Failed to parse config: lxc.net.0.ipv4 = 192.168.10.101/24

Signed-off-by: Julien Surloppe <julien@surloppe.fr>


brauner:
Thanks!

PR #2373, opened 2018-06-04
conf: fix misreading issue of proc mountinfo

2xsec:
Hello,

While a container reads mountinfo from proc fs, the mountinfo can be changed by kernel anytime.

It caused the critical issues on our device. So, we changed some logic for reading mountinfo.

Please review it. Thanks.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3262,7 +3262,7 @@ void remount_all_slave(const char *name)
 	fclose(f);
 	free(line);
 
-	if (ret && remove("/tmp/lxc_mountinfo.txt") < 0)
+	if (ret && remove(path) < 0)
 		SYSERROR("Failed to remove \"%s\"", path);
 }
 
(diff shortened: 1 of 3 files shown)


brauner:
Yup, doing now.

2xsec:
The mountinfo path is wrong to remove it. I'm sorry.

brauner:
@2xsec, no need to apologize for stuff like that. :) Thanks for the patch. I have a slightly different approach. that avoids a lingering temporary file. Would you please review #2374 and see if that works for you?

2xsec:
OK, I'll review and test it on our device.

brauner:
Depends on what device we're talking about and what kernel it is. If it is a post `2.6.33` kernel things should work fine. Pre-`2.6.33` kernels only allow `sendfile()` to a socket fd...

brauner:
Sorry, there's a missing `lseek()` in the branch. Will fix now.

2xsec:
We've applied our patch on TizenTV devices(3.0, 4.0). If your patch works fine, we'll apply it to TizenTV 5.0 with kernel 4.x. Thanks.

brauner:
Cool! Note, I pushed a small addition onto the branch I just sent so you might need to re-apply.

PR #2378, opened 2018-06-05
Revert "seccomp: make do_resolve_add_rule() more strict"

### LINKED ISSUE #2376: seccomp support is now too strict
The OCI runtime spec has a [section](https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config-linux.md#seccomp) on seccomp. 

It's not well-defined yet what must happen when a syscall can not be resolved, but I've asked the question: https://github.com/opencontainers/runtime-spec/issues/972

In the meantime, the current seccomp support in LXC is too strict to support OCI configurations in the wild. For instance, here's the list of syscalls used by Docker for their seccomp profile:
https://github.com/moby/moby/blob/master/profiles/seccomp/default.json

1) libseccomp might be too old to know about some syscalls as I mentioned in the issue above. This will be blocked here since dfddc8aa7ef3362212f8394995088a5f525730dd:
https://github.com/lxc/lxc/blob/2095b6d83acbb61bfd3c4a33dd88a6532148a23b/src/lxc/seccomp.c#L500-L504
2) Some syscalls in this list don't make sense for `x86_64`, like `chown32`. This will be blocked here since dfddc8aa7ef3362212f8394995088a5f525730dd:
https://github.com/lxc/lxc/blob/2095b6d83acbb61bfd3c4a33dd88a6532148a23b/src/lxc/seccomp.c#L506-L509
3) Conversely, some syscalls only make sense for `x86_64`, like `epoll_ctl_old`. In this case it fails because the call is `exact`:
https://github.com/lxc/lxc/blob/2095b6d83acbb61bfd3c4a33dd88a6532148a23b/src/lxc/seccomp.c#L530-L531
We resolve the syscall for the native arch, but in libseccomp it will be translated to a pseudo-syscall and it will fail since it's `strict` mode:
https://github.com/seccomp/libseccomp/blob/8ad3638ea9023c3948976dfadebd1554380a31c9/src/arch-x86.c#L331-L334

For 2) and 3),  the issue is that the OCI spec the arch is not per syscall. But libseccomp seems to allow that with the non `add_exact` calls and "pseudo-syscalls".

Should we...
* Revert dfddc8aa7ef3362212f8394995088a5f525730dd?
* Switch to non `exact` calls?
* Revert dfddc8aa7ef3362212f8394995088a5f525730dd and switch to non `exact` calls under some conditions? For instance failing to whitelist a syscall is probably more acceptable than failing to blacklist. Or maybe behind a configuration flag?

cc @brauner @hallyn

brauner:
This reverts commit dfddc8aa7ef3362212f8394995088a5f525730dd.

Closes #2376.

### DIFF
--- a/src/lxc/seccomp.c
+++ b/src/lxc/seccomp.c
@@ -500,13 +500,13 @@ bool do_resolve_add_rule(uint32_t arch, char *line, scmp_filter_ctx ctx,
 	nr = seccomp_syscall_resolve_name(line);
 	if (nr == __NR_SCMP_ERROR) {
 		WARN("Failed to resolve syscall \"%s\"", line);
-		WARN("This syscall will NOT be blacklisted");
+		WARN("This syscall will NOT be handled by seccomp");
 		return true;
 	}
 
 	if (nr < 0) {
 		WARN("Got negative return value %d for syscall \"%s\"", nr, line);
-		WARN("This syscall will NOT be blacklisted");
+		WARN("This syscall will NOT be handled by seccomp");
 		return true;
 	}
 
--- a/src/lxc/seccomp.c
+++ b/src/lxc/seccomp.c
@@ -500,12 +500,14 @@ bool do_resolve_add_rule(uint32_t arch, char *line, scmp_filter_ctx ctx,
 	nr = seccomp_syscall_resolve_name(line);
 	if (nr == __NR_SCMP_ERROR) {
 		WARN("Failed to resolve syscall \"%s\"", line);
-		return false;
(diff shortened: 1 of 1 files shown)


brauner:
//cc @Blub

flx42:
I think the error message is/was misleading. It's not always blacklisting, the syscalls might fail to be whitelisted too (if default action is "errno").

brauner:
@flx42, replaced the misleading error messages. Thanks!

brauner:
@stgraber, ready to go

PR #2382, opened 2018-06-07
Coverity

2xsec:
Hello,

some deadcodes were removed.

Thanks.


brauner:
That should really be just one commit, so please squash them. :) Additionally, we have a very specific form for coverity commits. The commit message should always look like this:
```
    coverity: #1435747

    Dereference before null check

    Signed-off-by: Foo Bar <foo@bar.io>
```

2xsec:
Oh, I'm sorry. I'll keep the rules. Thanks!

PR #2405, opened 2018-06-14
configure: check for getgrgid_r()

brauner:


src/include/getgrgid_r.c
```diff
@@ -0,0 +1,306 @@
+#define _GNU_SOURCE
```
stgraber: This file probably should have the same comment about coming from musl and if it's not licensed identically to liblxc, the license header should be there too.

brauner: done

PR #2419, opened 2018-06-20
secure coding: strcat => strncat

2xsec:
Hello,

secure coding: strcat => strncat

Thanks.


brauner:
jenkins: test this please

brauner:
Please note that `strncat()` is equally annoying as `strncpy()` as it doesn't guarantee `\0`-termination when the buffer is to small.

2xsec:
Yes, you are right. So, we checked buffer size of all dst buffers.

If you have any issues, please let us know.

Thanks.

brauner:
To be honest I'd prefer to have a helper for this like the kernel does:
```C
ssize_t strlcat(char *dest, const char *src, size_t count)
{
	size_t dsize = strlen(dest);
	size_t len = strlen(src);
	size_t res = dsize + len;

	/* This would be a bug */
        if (dsize >= count)
                return -EINVAL;

	dest += dsize;
	count -= dsize;
	if (len >= count)
		len = count-1;
	memcpy(dest, src, len);
	dest[len] = 0;
	return (ssize_t)res;
}
```

2xsec:
OK. No problem^^

Could you add strlcat() implementation like strlcpy()?

Thanks.

stgraber:
Helper has been merged

brauner:
jenkins: test this please

brauner:
Hm, the merge commit shouldn't be in there. :)

2xsec:
Oh, I'm sorry. I'll do pull request again tomorrow.

brauner:
We should also check for truncation. :)
```
ret = strlcat(pname, file, sizeof(pname));
if (ret >= sizeof(pname))
                   goto -ENAMETOOLONG; /* or something sensible */
```

brauner:
No need to apologize. :) Thanks for all the work!

2xsec:
The merge commit is removed.

We reviewed return size of strlcat() you requested and fix it.

If you have additional comments, please let us know.

Thanks^^

brauner:
jenkins: test this please

src/lxc/cgroups/cgfsng.c
```diff
@@ -1195,19 +1203,23 @@ static bool cg_unified_create_cgroup(struct hierarchy *h, char *cgname)
...
+		add_controllers = must_realloc(add_controllers, full_len + 1);
+
+		if (h->controllers[0] == *it)
+			add_controllers[0] = '\0';
+
+		(void)strlcat(add_controllers, "+", full_len + 1);
+		(void)strlcat(add_controllers, *it, full_len + 1);
+
+		if ((it + 1) && *(it + 1))
+			(void)strlcat(add_controllers, " ", full_len + 1);
```
brauner: I like these cleanups! :)

2xsec: I'll keep in mind^^

PR #2423, opened 2018-06-20
btrfs: s/strncat()/strlcat()/g

brauner:


brauner:
@2xsec just had to do this fix because this breaks compilation with some newer gcc versions. :) The rest is there for you. :)

PR #2426, opened 2018-06-22
monitor: change exit() => _exit() system call in child process

2xsec:
Hello,

_exit() system call is safe in child process.

Thanks.


brauner:
jenkins: test this please

PR #2434, opened 2018-06-27
tools: fix quiet option is not working

2xsec:
Hello,

Currently, a quiet option of tools is not working because of using fprintf directly.

So, we change fprintf or print to lxc_error log with some cleanups.

And lxc_sys_error is added to print errno string additionally.

Thanks.

### DIFF
--- a/src/lxc/tools/arguments.h
+++ b/src/lxc/tools/arguments.h
@@ -29,6 +29,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <sys/types.h>
+#include <sys/param.h>
 #include <lxc/lxccontainer.h>
 
 struct lxc_arguments;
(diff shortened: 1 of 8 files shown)


brauner:
jenkins: test this please

brauner:
I once did all the work to decouple the tools (e.g `lxc-start`, `lxc-stop`, ... from the shared library such that they do not share any functions because we thought that we'd move them into another repo. Now that we've decided against this it's super annoying that no code is shared anymore as we're duplicating so much stuff. :)

2xsec:
Yes, there are some duplicated source codes in tools. Especially, lxc_monitor.

So, I thought it was strange.

brauner:
@2xsec, feel free to change this back. :)

2xsec:
So, do you want to share the source codes? Could you share the releated commits?

brauner:
Yeah, sharing would make sense now that we're keeping the tools in the repo. The issue to track the prior work was https://github.com/lxc/lxc/issues/2073 and there's also a bunch of references to related pull requests.

2xsec:
I think we can do that.

PR #2435, opened 2018-06-27
[RFC] conf: introduce lxc.rootfs.managed

brauner:
This introduces a new config key lxc.rootfs.managed which can be used to
indicate whether this LXC instance is managing the container storage. If LXC is
not managing the storage then LXC will not modify the container storage.
For example, an API call to c->destroy(c) will then run any destroy hooks but
will not destroy the actual rootfs (Unless, of course, the hook does so behind
LXC's back.).

CC: Wolfgang Bumiller <w.bumiller@proxmox.com>
CC: Stéphane Graber <stgraber@ubuntu.com>
CC: Serge Hallyn <serge@hallyn.com>
CC: 2xsec <dh48.jeong@samsung.com>

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2760,6 +2760,7 @@ struct lxc_conf *lxc_conf_init(void)
 		free(new);
 		return NULL;
 	}
+	new->rootfs.managed = true;
 	new->logfd = -1;
 	lxc_list_init(&new->cgroup);
 	lxc_list_init(&new->cgroup2);
--- a/src/lxc/conf.h
+++ b/src/lxc/conf.h
@@ -154,6 +154,7 @@ struct lxc_tty_info {
  * @options    : mount options
  * @mountflags : the portion of @options that are flags
  * @data       : the porition of @options that are not flags
+ * @managed    : whether it is managed by LXC
  */
 struct lxc_rootfs {
 	char *path;
@@ -162,6 +163,7 @@ struct lxc_rootfs {
 	char *options;
 	unsigned long mountflags;
 	char *data;
+	bool managed;
 };
 
 /*
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -141,6 +141,7 @@ lxc_config_define(no_new_privs);
 lxc_config_define(personality);
 lxc_config_define(prlimit);
 lxc_config_define(pty_max);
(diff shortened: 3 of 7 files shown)


brauner:
//cc @Blub, @2xsec

stgraber:
`lxc.rootfs.managed` maybe? to keep everything namespaced together

brauner:
Yeah, might be a good idea.

brauner:
@stgraber, updated.

src/lxc/conf.c
```diff
@@ -2645,6 +2645,7 @@ struct lxc_conf *lxc_conf_init(void)
 		free(new);
 		return NULL;
 	}
+	new->rootfs.managed = true;
```
Blub: Shouldn't `lxc_conf_init()` set this to the same as the `clear` method?

brauner: The other way around actually. The default should be that it's managed. So clear should set to `true` most likely.

brauner: Unless that is to confusing?

brauner: *too

Blub: Whichever is less of a change to the previous behavior I guess?

brauner: `lxc.rootfs.managed` is supposed to mean "the rootfs is managed by this lxc instance" so the default should be `true`.

stgraber: Would calling it `lxc.rootfs.external` and reversing the logic make it less confusing maybe?

brauner: Yeah, but I like the `managed` part and `external` is harder to grasp. Usually the `lxc.rootfs.managed` config key will not be set anyway. It will only be set when set to `false`.

brauner:
Updated.

Blub:
LGTM, the only question I have is what to do about (other) files in `/var/lib/lxc/$container`, eg the config file. Currently an `lxc-destroy` on a container which otherwise sits in `/var/lib/lxc` regularly won't remove anything when `lxc.rootfs.managed=0`, iow. it's nop. I suppose it makes sense if they're setup only via liblxc rather than an actual dir in `/var/lib/lxc`, just thought I'd ask.

brauner:
Yeah, my current idea is to basically treat the directory as a no-go area for destroy when `managed` is set to `0`.

PR #2438, opened 2018-06-29
autodev: adapt to changes in Linux 4.18

brauner:
Starting with commit
`55956b59df33 ("vfs: Allow userns root to call mknod on owned filesystems.")`
Linux will allow mknod() in user namespaces for userns root if CAP_MKNOD is
available.
However, these device nodes are useless since
```
static struct super_block *alloc_super(struct file_system_type *type, int flags,
                                       struct user_namespace *user_ns)
{
        /* <snip> */

        if (s->s_user_ns != &init_user_ns)
                s->s_iflags |= SB_I_NODEV;

        /* <snip> */
}
```

will set the SB_I_NODEV flag on the filesystem. When a device node created in
non-init userns is open()ed the call chain will hit:
```
bool may_open_dev(const struct path *path)
{
        return !(path->mnt->mnt_flags & MNT_NODEV) &&
                !(path->mnt->mnt_sb->s_iflags & SB_I_NODEV);
}
```
which will cause an EPERM because the device node is located on an fs
owned by non-init-userns and thus doesn't grant access to device nodes due to
SB_I_NODEV.

The solution is straightforward. Unless you're real root you should bind-mount
device nodes.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1238,6 +1238,7 @@ struct lxc_device_node {
 };
 
 static const struct lxc_device_node lxc_devices[] = {
+	{ "console", S_IFCHR | S_IRWXU | S_IRWXG | S_IRWXO, 1, 5 },
 	{ "full",    S_IFCHR | S_IRWXU | S_IRWXG | S_IRWXO, 1, 7 },
 	{ "null",    S_IFCHR | S_IRWXU | S_IRWXG | S_IRWXO, 1, 3 },
 	{ "random",  S_IFCHR | S_IRWXU | S_IRWXG | S_IRWXO, 1, 8 },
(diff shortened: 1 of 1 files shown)


stgraber:
@brauner does systemd need fixing for that too?

brauner:
@stgraber, yeah, I already wrote you a message on irc because of that. :)

brauner:
User on systems with a non-patched LXC versions can bind-mount the for devices nodes over the created device nodes at which point they can go back to a working state.

brauner:
jenkins: test this please

aarondanielson:
> User on systems with a non-patched LXC version can bind-mount the four devices nodes over the created device nodes at which point they can go back to a working state.

@brauner , can you unpack this comment a bit please?  I'm running into this issue.

brauner:
@aarondanielson, if you'r on an unpatched LXC and can't upgrade or backport then you can add:
```
lxc.mount.entry = /dev/full dev/full none bind,optional,create=file 0 0
lxc.mount.entry = /dev/null dev/null none bind,optional,create=file 0 0
lxc.mount.entry = /dev/random dev/random none bind,optional,create=file 0 0
lxc.mount.entry = /dev/tty dev/tty none bind,optional,create=file 0 0
lxc.mount.entry = /dev/urandom dev/urandom none bind,optional,create=file 0 0
lxc.mount.entry = /dev/zero dev/zero none bind,optional,create=file 0 0
```
to the config at which point it should work again.

PR #2440, opened 2018-06-30
terminal: security fixes

brauner:

### DIFF
--- a/src/lxc/terminal.c
+++ b/src/lxc/terminal.c
@@ -584,6 +584,18 @@ static int lxc_terminal_peer_proxy_alloc(struct lxc_terminal *terminal,
 		goto on_error;
 	}
 
+	ret = fd_cloexec(terminal->proxy.master, true);
+	if (ret < 0) {
+		SYSERROR("Failed to set FD_CLOEXEC flag on proxy terminal master");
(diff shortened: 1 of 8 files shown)


src/lxc/terminal.c
```diff
@@ -584,6 +584,18 @@ static int lxc_terminal_peer_proxy_alloc(struct lxc_terminal *terminal,
 		goto on_error;
 	}
 
+	ret = fcntl(terminal->proxy.master, F_SETFD, FD_CLOEXEC);
```
Blub: I always wonder - given the fds come from `openpty` where you don't pass flags and don't exactly know which flags they have, if this shouldn't be a read-modify-write as stated in the `fcntl(ep)` manpage's `Application Usage` section.

brauner: You mean retrieve the flags, set the `FD_CLOEXEC` flag and pass the flags back down?

brauner: Care to push a patch on top of my branch?

brauner: Hm, or send me one :D

brauner:
@Blub, added `fd_cloexec()`.

PR #2441, opened 2018-07-02
coverity: #1425816

2xsec:
Hello,

fix "Explicit null dereferenced".

This is false positive because the optarg value is not null.

But coverity can not detect this. So, I add a contidtion to avoid this case.

Thanks.


brauner:
jenkins: test this please

brauner:
Thanks!

PR #2442, opened 2018-07-02
tree-wide: remove unneded log prefixes

brauner:
Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

PR #2446, opened 2018-07-03
Logging and cgroup fixes

brauner:

Closes #2442.

### DIFF
--- a/src/lxc/af_unix.c
+++ b/src/lxc/af_unix.c
@@ -40,7 +40,7 @@
 #include "include/strlcpy.h"
 #endif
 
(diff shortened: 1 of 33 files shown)


2xsec:
@brauner I checked the name of log define prefixs are the same with tools. I didn't check it. These commits will be merged? If that, there is no problem.

brauner:
@2xsec, the idea is to a) make all log prefixes unique (for building a static liblxc) and b) to remove the unneeded "lxc_" prefix. :)

2xsec:
@brauner I got it. :)

PR #2448, opened 2018-07-04
log: remove the trailing .

2xsec:
Hello,

The trailing . are removed in log.c.

Thanks.


brauner:
Thanks!

PR #2470, opened 2018-07-16
confile: add lxc.monitor.signal.pdeath

brauner:
Set the signal to be sent to the container's init when the lxc monitor exits.
By default it is set to SIGKILL which will cause all container processes to be
killed when the lxc monitor process dies.
To ensure that containers stay alive even if lxc monitor dies set this to 0.


brauner:
> You're still setting it to SIGKILL while printing the configured value in the error message AFAIC.
But shouldn't we leave SIGKILL in the current places for during the startup process and only set the signal to the configured value at the end of a successful startup?

Good call(s).

PR #2473, opened 2018-07-16
doc: Translate lxc.monitor.signal.pdeath into Japanese in lxc.contain…

tenforward:
…er.conf(5)


brauner:
Thanks! :)

PR #2475, opened 2018-07-17
conf: improve rootfs setup

brauner:

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1140,84 +1140,6 @@ on_error:
 	return ret;
 }
 
-static int setup_rootfs_pivot_root(const char *rootfs)
-{
-	int ret;
-	int newroot = -1, oldroot = -1;
-
-	oldroot = open("/", O_DIRECTORY | O_RDONLY);
-	if (oldroot < 0) {
-		SYSERROR("Failed to open old root directory");
-		return -1;
-	}
-
-	newroot = open(rootfs, O_DIRECTORY | O_RDONLY);
-	if (newroot < 0) {
-		SYSERROR("Failed to open new root directory");
-		goto on_error;
-	}
-
-	/* change into new root fs */
-	ret = fchdir(newroot);
-	if (ret < 0) {
-		SYSERROR("Failed to change to new rootfs \"%s\"", rootfs);
-		goto on_error;
-	}
-
-	/* pivot_root into our new root fs */
-	ret = pivot_root(".", ".");
-	if (ret < 0) {
-		SYSERROR("Failed to pivot_root()");
-		goto on_error;
-	}
-
-	/* At this point the old-root is mounted on top of our new-root. To
-	 * unmounted it we must not be chdir'd into it, so escape back to
-	 * old-root.
-	 */
-	ret = fchdir(oldroot);
-	if (ret < 0) {
-		SYSERROR("Failed to enter old root directory");
-		goto on_error;
-	}
-
-	/* Make oldroot rslave to make sure our umounts don't propagate to the
-	 * host.
-	 */
-	ret = mount("", ".", "", MS_SLAVE | MS_REC, NULL);
-	if (ret < 0) {
-		SYSERROR("Failed to make oldroot rslave");
-		goto on_error;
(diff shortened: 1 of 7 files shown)


stgraber:
@hallyn can you review this one? It's a bit long for me to read through while sprinting :)

src/lxc/conf.c
```diff
@@ -1508,53 +1432,143 @@ int prepare_ramfs_root(char *root)
...
+		SYSERROR("Failed to open old root directory");
+		return -errno;
 	}
 
-	if (detect_ramfs_rootfs()) {
-		DEBUG("Detected that container is on ramfs");
+	newroot = open(rootfs, O_DIRECTORY | O_RDONLY);
+	if (newroot < 0) {
+		SYSERROR("Failed to open new root directory");
+		ret = -errno;
```
Blub: These assignments should happen before `SYSERROR` to avoid `errno` being clobbered before. Although when we're already using the `SYS*` macros `-1` would probably be sufficient anyway.

src/lxc/conf.c
```diff
@@ -1423,33 +1344,36 @@ static int lxc_setup_rootfs(struct lxc_conf *conf)
...
 	char *p, *p2;
 	char buf[LXC_LINELEN], nroot[PATH_MAX];
 	FILE *f;
+	char *root = rootfs->mount;
 
-	if (!realpath(root, nroot))
-		return -1;
+	if (!realpath(root, nroot)) {
+		SYSERROR("Failed to resolve \"%s\"", root);
+		return -errno;
```
Blub: Similarly here `errno` could be clobbered and -1 is enough with the addition of the `SYSERROR`.

src/lxc/confile.c
```diff
@@ -2073,7 +2073,27 @@ static int set_config_rootfs_mount(const char *key, const char *value,
 static int set_config_rootfs_options(const char *key, const char *value,
 				     struct lxc_conf *lxc_conf, void *data)
 {
-	return set_config_string_item(&lxc_conf->rootfs.options, value);
+	int ret;
+	unsigned long pflags = 0;
+	struct lxc_rootfs *rootfs = &lxc_conf->rootfs;
+
+	ret = set_config_string_item(&rootfs->options, value);
```
Blub: I think we should do this last, otherwise we can get `-EINVAL` while already having `value` replaced.

src/lxc/confile.c
```diff
@@ -2073,7 +2073,27 @@ static int set_config_rootfs_mount(const char *key, const char *value,
...
+	if (ret < 0)
+		return -EINVAL;
+
+	ret = parse_mntopts(value, &rootfs->mountflags, &rootfs->data);
```
Blub: When moving the assignment down, this should be parsed into a temporary value first. Otherwise we get mixed values.
Iow. first do all the checking, then do the assignments together.

src/lxc/confile.c
```diff
@@ -2073,7 +2073,27 @@ static int set_config_rootfs_mount(const char *key, const char *value,
...
+	if (ret < 0)
+		return -EINVAL;
+
+	ret = parse_propagationopts(value, &pflags);
+	if (ret < 0) {
+               free(rootfs->data);
+		return -EINVAL;
+       }
```
Blub: Indentation error.

src/lxc/conf.h
```diff
@@ -150,14 +150,18 @@ struct lxc_tty_info {
  * optionals pivot_root, rootfs mount paths
  * @path       : the rootfs source (directory or device)
  * @mount      : where it is mounted
- * @options    : mount options
  * @bev_type   : optional backing store type
+ * @options    : mount options
+ * @mountflags : the portion of @options that are flags
```
Blub: I wonder if it makes sense to add these as part of a larger commit like this without mentioning it in the commit message. I currently only see them parsed out but not actually used?

brauner:
@Blub, updated.

Blub:
You missed two clobbered `errno`s. LGTM otherwise, I pushed a fixup on top, not sure if you prefer it squashed in.

PR #2481, opened 2018-07-19
apparmor: Allow /usr/lib* paths for mount and pivot_root

hwoarang:
openSUSE Leap 15 is using --libdir=/usr/lib64 when building for
x86_64 so we need to allow this path in the apparmor profiles.

Link: https://bugzilla.opensuse.org/show_bug.cgi?id=1099239
Signed-off-by: Markos Chandras <mchandras@suse.de>

### DIFF
--- a/config/apparmor/abstractions/start-container
+++ b/config/apparmor/abstractions/start-container
@@ -9,8 +9,8 @@
   ptrace,
 
   # currently blocked by apparmor bug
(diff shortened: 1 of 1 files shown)


brauner:
I think this can be safely done. At least I see no obvious reasons why we shouldn't be able to allow this. @Blub, any thoughts?

brauner:
Thanks! :)

PR #2484, opened 2018-07-21
attach: fix return value & cleanups

2xsec:
Hello,

attach: fix return value & cleanups

Thanks.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -108,18 +108,19 @@ static struct lxc_proc_context_info *lxc_proc_get_context_info(pid_t pid)
 
 	proc_file = fopen(proc_fn, "r");
 	if (!proc_file) {
-		SYSERROR("Could not open %s.", proc_fn);
+		SYSERROR("Could not open %s", proc_fn);
 		goto on_error;
 	}
 
 	info = calloc(1, sizeof(*info));
 	if (!info) {
-		SYSERROR("Could not allocate memory.");
+		SYSERROR("Could not allocate memory");
 		fclose(proc_file);
 		return NULL;
 	}
 
 	found = false;
+
 	while (getline(&line, &line_bufsz, proc_file) != -1) {
 		ret = sscanf(line, "CapBnd: %llx", &info->capability_mask);
 		if (ret != EOF && ret == 1) {
@@ -132,9 +133,8 @@ static struct lxc_proc_context_info *lxc_proc_get_context_info(pid_t pid)
 	fclose(proc_file);
 
 	if (!found) {
-		SYSERROR("Could not read capability bounding set from %s.",
-			 proc_fn);
-		errno = ENOENT;
+		ERROR("Could not read capability bounding set from %s",
+		      proc_fn);
 		goto on_error;
 	}
 
@@ -156,6 +156,7 @@ static inline void lxc_proc_close_ns_fd(struct lxc_proc_context_info *ctx)
 	for (i = 0; i < LXC_NS_MAX; i++) {
 		if (ctx->ns_fd[i] < 0)
 			continue;
+
 		close(ctx->ns_fd[i]);
 		ctx->ns_fd[i] = -EBADF;
 	}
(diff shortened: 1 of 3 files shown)


src/lxc/attach.c
```diff
@@ -189,6 +190,7 @@ static void lxc_proc_put_context_info(struct lxc_proc_context_info *ctx)
 static int in_same_namespace(pid_t pid1, pid_t pid2, const char *ns)
 {
 	int ns_fd1 = -1, ns_fd2 = -1, ret = -1;
+	int saved_errno = errno;
```
Blub: I'd assign this to 0 here, since `errno` has an undefined value and this is for the `-EINVAL` case which for which `errno` has no defined value anyway. Or, if you move the assignment to `out:` (as I describe in the other comment) you can leave it uninitialized.

src/lxc/attach.c
```diff
@@ -199,52 +201,60 @@ static int in_same_namespace(pid_t pid1, pid_t pid2, const char *ns)
 		if (errno == ENOENT)
 			return -EINVAL;
 
+		saved_errno = errno;
```
Blub: This is not wrong, but all but one cases want to assign `errno`, so it would also be okay to move this down to after `out:`.  The caller would know whether `errno` makes sense from the return value anyway.

src/lxc/attach.c
```diff
@@ -199,52 +201,60 @@ static int in_same_namespace(pid_t pid1, pid_t pid2, const char *ns)
...
-	int i, ret;
+	int i;
 
 	for (i = 0; i < LXC_NS_MAX; i++) {
 		if (ctx->ns_fd[i] < 0)
 			continue;
 
-		ret = setns(ctx->ns_fd[i], ns_info[i].clone_flag);
-		if (ret < 0) {
+		if (setns(ctx->ns_fd[i], ns_info[i].clone_flag) < 0) {
```
Blub: According to [1] the old version was the preferred way in this case. @brauner should have a stronger opinion here. We do have both in the code base.
[1] https://github.com/lxc/lxc/blob/master/CODING_STYLE.md#functions-not-returning-booleans-must-assign-return-value-before-performing-checks

src/lxc/attach.c
```diff
@@ -658,14 +678,14 @@ static bool fetch_seccomp(struct lxc_container *c, lxc_attach_options_t *options
 
 	/* Remove current setting. */
 	if (!c->set_config_item(c, "lxc.seccomp.profile", "") &&
-	    !c->set_config_item(c, "lxc.seccomp", "")) {
```
Blub: @brauner, I've been meaning to ask whether the single-line-if exception in the coding style also holds for cases like this where the condition spans across multiple lines?

brauner: Yeah, it does. )

src/lxc/attach.c
brauner: Yes, assigning is the preferred way. :)

2xsec: The ret variable is assigned to stack area. but it is used only 1 time in function. So, I thought it is inefficient because a return value is stored in register and use it directly. But this is the coding rule, I'll change it. :)
Thanks for your review.

brauner: Yeah, these are corner cases so I won't really back it based on that but it's preferred in general. So you can leave it if you want. :)

src/lxc/attach.c
2xsec: @brauner in case of this, brace is needed?

brauner: No, no braces needed.

src/lxc/attach.c
2xsec: @Blub Yes, it looks better. Thanks for your review.

src/lxc/attach.c
2xsec: @Blub Yes, it looks better. Thanks for your review.

brauner:
jenkins: test this please

brauner:
Thanks!

PR #2489, opened 2018-07-22
change log macro of error case from lxc_ambient_caps_up/down

2xsec:
Hello,

change log macro of error case from lxc_ambient_caps_up/down.

because all errno strings are printed in lxc_amvient_caps_up / down() functions.

Thanks.


brauner:
jenkins: test this please

brauner:
Thanks!

PR #2503, opened 2018-07-31
doc: Add incompatibility with network type=none into Japanese man

tenforward:
Update for commit e4b3e36


brauner:
Thanks!

PR #2514, opened 2018-08-07
pam_cgfs: remove duplicated utils

2xsec:
Hello,

Some duplicated utils used in pam_cgfs are removed.

Thakns.


stgraber:
We had a build issue of some kind where those symbols were missing from the cgfs binary as it doesn't and can't look against liblxc.

2xsec:
@brauner @stgraber pam_cgfs.c is separated library. I didn't check it. Thanks for your review.

PR #2515, opened 2018-08-08
netns: allocate network namespace id

brauner:
Start to allocate a new network namespace id for each container.

Relates to lxc/lxd#4831.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -1818,6 +1818,13 @@ static int lxc_spawn(struct lxc_handler *handler)
 	} else {
 		handler->nsfd[LXC_NS_NET] = ret;
 		DEBUG("Preserved net namespace via fd %d", ret);
+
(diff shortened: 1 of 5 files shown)


src/lxc/start.c
```diff
@@ -1818,6 +1818,13 @@ static int lxc_spawn(struct lxc_handler *handler)
 	} else {
 		handler->nsfd[LXC_NS_NET] = ret;
 		DEBUG("Preserved net namespace via fd %d", ret);
+
+		ret = lxc_netns_set_nsid(handler->nsfd[LXC_NS_NET]);
+		if (ret < 0) {
+			ERROR("Failed to allocate new network namespace id: %d", ret);
```
stgraber: Do you know when this got introduced? Are we limiting backward compat by making this fatal?

brauner: Since we don't request an ACK from the kernel it should work on all kernels. I've verified on `3.13`.

PR #2516, opened 2018-08-09
tests: containertests: fix dereference pointer c

2xsec:
Hello,

fix dereference pointer c in containertests

### DIFF
--- a/src/lxc/caps.c
+++ b/src/lxc/caps.c
@@ -310,7 +310,7 @@ int lxc_caps_init(void)
 	return 0;
 }
 
-static int _real_caps_last_cap(void)
+static long int _real_caps_last_cap(void)
 {
 	int fd, result = -1;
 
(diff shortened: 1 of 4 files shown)


brauner:
jenkins: test this please

src/lxc/attach.c
```diff
@@ -314,6 +314,9 @@ static int lxc_attach_drop_privs(struct lxc_proc_context_info *ctx)
 	int cap, last_cap;
 
 	last_cap = lxc_caps_last_cap();
+	if (last_cap < 0 || last_cap > INT_MAX)
```
brauner: `las_cap` can't be greater than `INT_MAX` since it is an int itself. :) You could make `last_cap` a `long int` and perform that check.

2xsec: @brauner Yes, you're right. I'll check it again. Thanks for your review. :)

brauner:
Thanks! :)

PR #2534, opened 2018-08-16
Checkpatch

tcharding:
@brauner commented on another PR that `src/lxc/cmd/lxc_usernsexec.c` could do with some cleaning up.  I parsed `src/lxc/cmd/*.c` with the kernel 'linter' `scripts/checkpatch.pl`.  Here are just fixes to the most obvious ERROR messages.  Just doing a few patches in this PR to see that I'm following protocol.

Code builds but is untested.  I see you guys have CI on your side, are there tests I should be running locally before raising pull requests?

thanks,
Tobin.

### DIFF
--- a/src/lxc/cmd/lxc_init.c
+++ b/src/lxc/cmd/lxc_init.c
@@ -54,7 +54,7 @@
 
 lxc_log_define(lxc_init, lxc);
 
-static sig_atomic_t was_interrupted = 0;
+static sig_atomic_t was_interrupted;
 
 static void interrupt_handler(int sig)
 {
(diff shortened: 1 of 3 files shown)


brauner:
Yep, that sounds good. Yeah, we follow kernel coding style. I once thought about vendoring `checkpatch.pl` in the `LXC` tree.

tcharding:
Man you do **not** want that script in your tree, it is a monster :)

brauner:
> Man you do **not** want that script in your tree, it is a monster :)

Heh. :) We should still point to it in the `CODING_STYLE.md` doc though. Just mention that it should be used to check for obvious style errors.

brauner:
Thanks!

PR #2535, opened 2018-08-16
log: add logging tools for commands; lxc-usernsexec: cleanup and bugfixes

brauner:

### DIFF
--- a/src/lxc/cmd/lxc_usernsexec.c
+++ b/src/lxc/cmd/lxc_usernsexec.c
@@ -42,6 +42,8 @@
 #include <grp.h>
 
 #include "conf.h"
+#include "list.h"
+#include "log.h"
 #include "namespace.h"
 #include "utils.h"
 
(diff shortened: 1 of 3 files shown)


src/lxc/cmd/lxc_usernsexec.c
```diff
@@ -106,51 +109,60 @@ static void opentty(const char *tty, int which)
...
 
 /*
- * given a string like "b:0:100000:10", map both uids and gids
- * 0-10 to 100000 to 100010
+ * Given a string like "b:0:100000:10", map both uids and gids 0-10 to 100000
+ * to 100010
  */
 static int parse_map(char *map)
 {
+	long host_id, i, ns_id, range, ret;
```
stgraber: why is ret a long here?

brauner: Accident. Let me fix.

brauner: fixed

brauner:
@2xsec, you can add them too if you want. :) I do wonder though whether both `tools` and `cmd`s couldn't just use the standard `log` macros directly. Probably possible since they are now once again integrated into the main LXC tree.

PR #2537, opened 2018-08-16
Remove empty README file

tcharding:
README file was moved to markdown format but the empty original file
was never removed.

Remove empty README file.

Fixes: dc46727ff721 ("repo: add new README")


brauner:
I think we can't remove the empty README because it is required by autotools.

tcharding:
You are right @brauner, autotools does need the `README` file to exist.  In an attempt to save you from having to go through this with every over zealous new contributor I came up with the following idea (force pushed to this branch).  Put the README content in `README` and create a symlink to it called `README.md`.  This seems to satisfy autotools and to function as expected with emacs (correct mode seleciton) and github (renders markdown).

Feel free to close this though if you are happy having the empty `README`.

PR #2540, opened 2018-08-17
cmd: Fix up checkpatch warnings

tcharding:
This is the remainder of the fixes found by checkpatch for `cmd/`.  The warnings left are either contentious or not demanded by our coding style.

- Use parenthesis around complex macro i.e. `(foo + 1)` instead of `foo + 1`.
- Use 'void' instead of empty parameter list i.e. `static int fn(void)` .
- Do not use braces for single statement block.
- Fix all whitespace issues as a single patch since these are easy to review.
- Use 'const' for static string constant.
- Remove unnecessary whitespace in string (separate patch because this only subjectively better).
- Put trailing */ on a separate line (standard comment format).
- Remove accidental duplicate semicolon 
- Do not use comparison to NULL i.e. `if (!foo)` instead of `if (foo == NULL)`.

All patches build cleanly but no further testing done.

thanks,
Tobin.

### DIFF
--- a/src/lxc/cmd/lxc_init.c
+++ b/src/lxc/cmd/lxc_init.c
@@ -550,9 +550,8 @@ static int arguments_parse(struct arguments *args, int argc,
 	args->argc = argc - optind;
 
 	/* If no lxcpath was given, use default */
-	if (!args->lxcpath) {
+	if (!args->lxcpath)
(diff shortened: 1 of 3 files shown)


tcharding:
woops, this is only source (`.c`) files.

brauner:
jenkins: test this please

brauner:
Thanks! :)

Issue #2523, opened 2018-08-12
lxc-autostart does not signal failures using exit status codes


thedrow:
# Required information

Applicable in any environment and is reproducible at least at 3.0+ but probably in earlier versions as well.

# Issue description

I just set up a systemd service to run my unprivileged containers on boot.
The containers failed to start correctly but the service is still marked as active since the command line [always exits with a 0 status code](https://github.com/lxc/lxc/blob/master/src/lxc/tools/lxc_autostart.c#L507) unless a CLI parsing option occurred, initializing the logger fails or if there are no defined containers to start.
This makes retries and error reporting much harder.

I suggest we reserve exit code 2 for partial failures (not all containers were started/shutdown etc) and exit code 3 for full failures. We can alternatively reuse exit code 1 if that's the right path in your opinion.

# Steps to reproduce

Create the following systemd service in /home/<user>/.config/systemd/user/lxc-autostart.service:

```ini
[Unit]
Description=Start LXC containers on boot

[Service]
Type=oneshot
ExecStart=/usr/bin/lxc-autostart
ExecStop=/usr/bin/lxc-autostart --shutdown
RemainAfterExit=yes

[Install]
WantedBy=default.target
```

Type `systemctl --user daemon-reload && systemctl --user enable lxc-autostart`.

Reboot.

The service will fail to start since dependency order was not declared and lxc has not been set up correctly yet.
Note that this is probably the wrong systemd configuration and the service should be configured as a system service in order to boot correctly.

# Information to attach

N/A

stgraber:
Sure, having lxc-autostart fail when it failed to start some containers is reasonable.

systemd distros shouldn't be using lxc-autostart directly though, they should use lxc-autostart with `--list` to get the list of containers and then spawn individual instances of a `lxc` unit for each of those.
That way you get per-container units with their own status and retry policy.

Upstream already ships a `lxc@.service` unit which then allows you to do `systemctl start lxc@blah` to start container `blah`, but the main `lxc.service` unit itself doesn't seem to be making use of that, so that'd be something to improve if someone has the interest to do that.

thedrow:
@stgraber How do you prefer the exit codes to behave?

PR #2542, opened 2018-08-17
tools: Indicate container startup failure

tcharding:
When running `lxc-autostart` we do not currently indicate failure to start containers, either partial failure (i.e. some of the containers failed to start) or total failure (i.e. all of the containers failed to start).

(We specifically use -1 instead of EXIT_FAILURE.  Even if EXIT_FAILURE is defined as 1 this is congruent with 'total failure'.)

Indicate container startup failure.  For total failure return -1, for partial failure return -2.

Fixes: #2523 

Tested using systemd service described in the issue (#2523).  Before applying this patch:
```
$ systemctl --user start lxc-autostart

$ systemctl --user status lxc-autostart
● lxc-autostart.service - Start LXC containers on boot
   Loaded: loaded (/home/tobin/.config/systemd/user/lxc-autosta
   Active: active (exited) since Fri 2018-08-17 17:07:49 AEST;
```
After applying this patch:
```
$ systemctl --user start lxc-autostart
Job for lxc-autostart.service failed because the control process exited with error code.
See "systemctl --user status lxc-autostart.service" and "journalctl --user -xe" for details.
```
Relevant lines of output from `journalctl --user -xe` 
```
Aug 17 17:08:32 eros systemd[1615]: lxc-autostart.service: Main process exited, code=exited, status=1/FAILURE
Aug 17 17:08:32 eros systemd[1615]: lxc-autostart.service: Failed with result 'exit-code'.
Aug 17 17:08:32 eros systemd[1615]: Failed to start Start LXC containers on boot.
```

### DIFF
--- a/src/lxc/tools/lxc_autostart.c
+++ b/src/lxc/tools/lxc_autostart.c
@@ -318,7 +318,7 @@ static int toss_list(struct lxc_list *c_groups_list)
 
 int main(int argc, char *argv[])
 {
-	int count = 0, i = 0, ret = 0;
+	int count = 0, failed = 0, i = 0, ret = 0;
 	struct lxc_list *cmd_group;
 	struct lxc_container **containers = NULL;
 	struct lxc_list **c_groups_lists = NULL;
@@ -491,11 +491,15 @@ int main(int argc, char *argv[])
 
 	}
 
-	/* clean up any lingering detritus */
+	/* clean up any lingering detritus, if container exists here
+	 * then it must have failed to start.
+	 */
+	failed = 0;
 	for (i = 0; i < count; i++) {
-		if (containers[i])
+		if (containers[i]) {
+			failed++;
 			lxc_container_put(containers[i]);
-
+		}
 		if (c_groups_lists && c_groups_lists[i])
 			toss_list(c_groups_lists[i]);
 	}
@@ -503,6 +507,11 @@ int main(int argc, char *argv[])
 	free(c_groups_lists);
 	toss_list(cmd_groups_list);
 	free(containers);
+
+	if (failed == count)
+		exit(1);	/* Total failure */
+	else if (failed > 0)
+		exit(2);	/* Partial failure */
 
 	exit(EXIT_SUCCESS);
 }


src/lxc/tools/lxc_autostart.c
```diff
@@ -504,5 +508,10 @@ int main(int argc, char *argv[])
 	toss_list(cmd_groups_list);
 	free(containers);
 
+	if (failed == count)
+		return -1;	/* Total failure */
```
brauner: Hm, what I gather from #2523 this should be:
```
if (failed == count)
        exit(1);
else if (failed > 0)
        exit(2);
```

tcharding: Cool, thanks for the review.  Will fix and re-push.

src/lxc/tools/lxc_autostart.c
```diff
@@ -318,7 +318,7 @@ static int toss_list(struct lxc_list *c_groups_list)
 
 int main(int argc, char *argv[])
 {
-	int count = 0, i = 0, ret = 0;
```
tcharding: Is there some reason this is not written as
```
int count, i, ret = 0;
```
or
```
int count, ret, i = 0;
```

src/lxc/tools/lxc_autostart.c
```diff
@@ -491,11 +491,15 @@ int main(int argc, char *argv[])
 
 	}
 
-	/* clean up any lingering detritus */
+	/* clean up any lingering detritus, if container exists here
+	 * then it must have failed to start.
+	 */
+	failed = 0;
```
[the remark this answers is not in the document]
tcharding: I initialized at declaration time to try and fit in with the current code.  I _really_ wanted the initialization here so it is close to the usage, especially since all we do is increment the var it seems cleaner to me to have the initialization close to the code that uses it.  I'm happy to put the initialization where ever though if you have a preference.

thanks

tcharding:
Thanks for the snippet above @brauner, sorry to take your time up to write it.  You have just witnessed my number one current shortcoming as a developer that I need to work on - I tend to get excited and rush things resulting in little mistakes like this.

thanks man,
Tobin.

src/lxc/tools/lxc_autostart.c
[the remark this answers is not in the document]
tcharding: It wast the initialization of `i` at declaration time that threw me, hence I initialized `failed` in similar fashion.

brauner:
Thanks.

Issue #1872, opened 2017-10-23
storage/loop.c:  File exists - Failed to seek to set new loop file size


panders23:
# Required information
 * Distribution: archlinux
 * Distribution version: ~
 * The output of
   * `lxc-start --version`
```
2.1.0-devel
```
   * `lxc-checkconfig`
```
--- Namespaces ---Namespaces: enabled
Utsname namespace: enabled
Ipc namespace: enabled
Pid namespace: enabled
User namespace: missing
Network namespace: enabled

--- Control groups ---
Cgroups: enabled
Cgroup v1 mount points: /sys/fs/cgroup/systemd
/sys/fs/cgroup/freezer
/sys/fs/cgroup/cpuset
/sys/fs/cgroup/memory
/sys/fs/cgroup/net_cls,net_prio
/sys/fs/cgroup/pids
/sys/fs/cgroup/devices/sys/fs/cgroup/blkio
/sys/fs/cgroup/cpu,cpuacct
/sys/fs/cgroup/perf_event

Cgroup v2 mount points: 
/sys/fs/cgroup/unified

Cgroup v1 clone_children flag: enabled
Cgroup device: enabled
Cgroup sched: enabled
Cgroup cpu account: enabled
Cgroup memory controller: enabled
Cgroup cpuset: enabled

--- Misc ---
Veth pair device: enabled, loaded
Macvlan: enabled, not loaded
Vlan: enabled, not loaded
```
   * `uname -a`
```
Linux host1.local 4.13.9-1-ARCH #1 SMP PREEMPT Sun Oct 22 09:07:32 CEST 2017 x86_64 GNU/Linux
```
   * `cat /proc/self/cgroup`
```
10:perf_event:/
9:cpu,cpuacct:/
8:blkio:/
7:devices:/user.slice
6:pids:/user.slice/user-1000.slice/session-c2.scope
5:net_cls,net_prio:/
4:memory:/
3:cpuset:/
2:freezer:/
1:name=systemd:/user.slice/user-1000.slice/session-c2.scope
0::/user.slice/user-1000.slice/session-c2.scope
```
   * `cat /proc/1/mounts`
```
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
devtmpfs /dev devtmpfs rw,nosuid,size=8129384k,nr_inodes=2032346,mode=755 0 0
[...]

PR #1705, opened 2017-07-15
storage: rework storage drivers

brauner:
Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

PR #2545, opened 2018-08-18
storage/loop.c: integer overflow

ljagiello:
The issue was introduced in PR (https://github.com/lxc/lxc/pull/1705):

Previous code:
```
  if (lseek(fd, size, SEEK_SET) < 0) {
    SYSERROR("Error seeking to set new loop file size");
    close(fd);
    return -1;
  }
```
New code:
```
  int fd, ret;

  [...]

  ret = lseek(fd, size, SEEK_SET);
  if (ret < 0) {
    SYSERROR("Failed to seek to set new loop file size for loop "
       "file \"%s\"", path);
    close(fd);
    return -1;
  }
```

Based on http://man7.org/linux/man-pages/man2/lseek.2.html:
> Upon successful completion, lseek() returns the resulting offset location as measured in bytes from the beginning of the file.

Because `size` is `uint64_t`, it's very easy to overflow `res` and as a result the code will success randomly (if overflow is >0 it will pass it it's <0 it will fail).

Example from strace:
```
[...]
[pid  5561] lseek(4, 53687091200, SEEK_SET) = 53687091200
[pid  5561] close(4)                    = 0
[pid  5561] exit_group(1)               = ?
[pid  5561] +++ exited with 1 +++
[...]
```

This PR fix issues (https://github.com/lxc/lxc/issues/1872).

Signed-off-by: Lukasz Jagiello <lukasz@wikia-inc.com>

### DIFF
--- a/src/lxc/storage/loop.c
+++ b/src/lxc/storage/loop.c
@@ -297,6 +297,7 @@ int loop_umount(struct lxc_storage *bdev)
 static int do_loop_create(const char *path, uint64_t size, const char *fstype)
 {
 	int fd, ret;
+	off_t ret_size;
 	char cmd_output[MAXPATHLEN];
 	const char *cmd_args[2] = {fstype, path};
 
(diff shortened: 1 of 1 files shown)


brauner:
You're missing the signed-off-by line that we require. :)
Also, please use a dedicated `off_t size` return variable. :)

brauner:
jenkins: test this please

brauner:
Thanks!

PR #2548, opened 2018-08-18
cmd: lxc-user-nic: change log macro & cleanups

2xsec:
Hello,

log macros using strerror are changed to CMD_SYSERROR macro.

Thanks.

### DIFF
--- a/src/lxc/cmd/lxc_user_nic.c
+++ b/src/lxc/cmd/lxc_user_nic.c
@@ -46,6 +46,7 @@
 #include <sys/types.h>
 
 #include "config.h"
+#include "log.h"
 #include "namespace.h"
 #include "network.h"
 #include "parse.h"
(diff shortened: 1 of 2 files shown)


src/lxc/cmd/lxc_user_nic.c
```diff
@@ -1330,5 +1323,5 @@ int main(int argc, char *argv[])
 		host_veth_ifidx);
 	free(newname);
 	free(nicname);
-	exit(EXIT_SUCCESS);
+	_exit(EXIT_SUCCESS);
```
brauner: Since you're calling `_exit()` and `_exit()` doesn't flush output streams you need to call `fflush(stdout)` before the `_exit()`.

2xsec: @brauner generally, we use _exit() when terminating child process forked by parent process using fork() syscall. I checked your commit("lxc-usernsexec: cleanup and bugfixes"). _exit() call is used all cases. Is it prefer to use _exit() call for all cases? :)

brauner:
jenkins: test this please

brauner:
Thanks!

PR #2549, opened 2018-08-19
cmd: lxc-usernsexec fixes + macro: move declarations + config_utils: macvlan fixes

brauner:

### DIFF
--- a/src/lxc/cmd/lxc_usernsexec.c
+++ b/src/lxc/cmd/lxc_usernsexec.c
@@ -21,25 +21,25 @@
  * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
+
 #include "config.h"
 
+#include <errno.h>
+#include <fcntl.h>
+#include <grp.h>
+#include <libgen.h>
+#include <pwd.h>
+#include <sched.h>
+#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
-#include <sched.h>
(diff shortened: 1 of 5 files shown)


src/lxc/cmd/lxc_usernsexec.c
```diff
@@ -228,8 +229,20 @@ static int read_default_map(char *fnam, int which, char *user)
 			return -1;
 		}
 
-		newmap->hostid = atol(p1 + 1);
-		newmap->range = atol(p2 + 1);
+		ret = lxc_safe_ulong(p1 + 1, &newmap->hostid);
+		if (ret < 0) {
+			fclose(fin);
+			free(line);
```
Blub: Like the `!tmp` case below this should also `free(newmap)` - perhaps we should have an `int result` on top and just `break` here (and set it to `0` before the break at the loop's bottom) - or the usual goto...

src/lxc/cmd/lxc_usernsexec.c
```diff
@@ -223,46 +223,37 @@ static int read_default_map(char *fnam, int which, char *user)
...
 
 		tmp = malloc(sizeof(*tmp));
-		if (!tmp) {
-			fclose(fin);
-			free(line);
-			free(newmap);
-			return -1;
-		}
+		if (!tmp)
+			goto on_error;
```
Blub: This now needs to set `ret = -1;` as it's 0 from the `lxc_safe_ulong()` call succeeding above.

brauner: ffs

brauner: fixed

PR #2550, opened 2018-08-20
storage: exit() => _exit()

2xsec:
Hello,

1. storage: exit() => _exit()
    => When exec syscall is failed, child process needs to use _exit().

2. tools: lxc-wait: add default log priority & cleanups

Thanks.


brauner:
Thanks!

Issue #2471, opened 2018-07-16
error handling cleanups


Blub:
We have a bunch of functions throughout with inconsistent error return values, mixing `-1` with and without a valid `errno` with returning negative errno codes. Eg. some confile.c functions can return `-EINVAL` as well as just `-1`.
We should clean these up to follow the coding style[1] document.
In other words: in our own functions, if not in conflict with a successful return value, always return a negative error code (iow, if the error was triggered by something setting `errno`, simply `return -errno;`), otherwise actually set `errno` if needed. (Eg. when a non-NULL pointer is to be returned on success, return `NULL` after setting `errno`).

[1] https://github.com/lxc/lxc/blob/master/CODING_STYLE.md#return-error-codes

brauner:
@2xsec, this might be something you might be interested in doing. Probably a good idea to either do a commit for each function or for each file.

2xsec:
@brauner, Of course. I'll check it.

brauner:
@2xsec, no pressure just in case you're interested. You might also consider jumping into our irc channel #lxc-dev on Freenode. But of course, that's completely optional too. :)

2xsec:
@brauner This is no pressure for me.:) Please recommend a good tool for irc.:)

brauner:
Hm, we usually either use `irssi` or `weechat`. But there are graphical or browser based irc clients as well. It depends on whether you want to be connected constantly or only on demand. Most of us have a persistent irc session running in tmux on a remote server so that we can simply attach to it every day and see what we missed and also have logs of all interactions.
Personally, I use ZNC as irc bouncer but that is probably not needed.

Blub:
Updated my intial comment a bit in response to #2551. I think this part might benefit from a few examples in the coding style doc, or some more situation distinctions.

brauner:
@Blub, cool. From what I gathered from our recent irc discussion you'd prefer to set `errno` instead of returning `-<errno-value>` from functions, right. I'll let you review branches such as https://github.com/lxc/lxc/pull/2551 since you have a better grasp of what you're aiming for. :)

Blub:
I don't remember saying that ;-). Setting `errno` was for functions where the return type conflicts with the error codes (eg pointers.). Weren't you entertaining the idea of using `PTR_ERR()`? That one I'm not so fond of ;-).

brauner:
[...]

PR #2551, opened 2018-08-20
cmd: Clean up function return values

tcharding:
In response to issue #2471 this PR cleans up function return values for code error paths resulting from failed stdlib function calls.

- Return `-EINVAL` instead of -1 for invalid command line options.
- Return `-EINVAL` instead of -1 for invalid function args.
- Save and return `-errno` for failed stdlib function calls.

For stdlib functions that return a single error code just return the error code directly, comment if not obvious.

Code is untested.

(This PR only does `src/lxc/cmd/`, if this is correct I can do the rest of `src/lxc`.  Assuming that @2xsec is not working on this anymore).

thanks,
Tobin.

### DIFF
--- a/src/lxc/Makefile.am
+++ b/src/lxc/Makefile.am
@@ -322,6 +322,7 @@ if ENABLE_COMMANDS
 init_lxc_SOURCES = cmd/lxc_init.c
 lxc_monitord_SOURCES = cmd/lxc_monitord.c
 lxc_user_nic_SOURCES = cmd/lxc_user_nic.c \
+		       log.c log.h \
 		       namespace.c namespace.h \
 		       network.c network.h \
 		       parse.c parse.h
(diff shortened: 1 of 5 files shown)


Blub:
I'll need to look through those a bit more. A while ago we talked about this in irc a bit but haven't written anything down. Basically, whether it makes sense to return -1 or a specific code comes down to whether the caller can actually make decisions based on the error type. For utility functions there may be many different callers, so error codes are usually a good idea, for specific code (eg. private to start.c or conf.c), it's mostly about failing or not failing (the "binary" way as the coding style text would say). In the latter case the `SYS*` macros often already show the error (eg. if it came from a syscall, otherwise one should try to use the regular ERROR/... macros to print useful diagnostics) and the caller probably won't be using the exact code as much. In either case: one function should only be returning either always an error code or always -1. A quick glance through the changes showed a few which are still mixed, but I haven't yet gone through them in detail to figure out which way is better.

brauner:
@tcharding, you've rebased one commit too far. That's why there's a merge conflict now. You're changing the hash of an already existing commit with this. :)

tcharding:
> @tcharding, you've rebased one commit too far. That's why there's a merge conflict now. You're changing the hash of an already existing commit with this. :)

Woops, its hard to get good help eh :)

Closing this since @2xsec and @Blub are already across it.

PR #2559, opened 2018-08-21
start: Fix net iface remaining issue

ssup2:
When creating container that has multiple net ifaces fails
because of wrong net config, lxc doesn't remove previously created
net ifaces.

Solve this issue with changing return path in lxc_spawn().

Signed-off-by: Jungsub Shin jungsub_shin@tmax.co.kr


brauner:
jenkins: test this please

brauner:
Thanks for spotting this!

PR #2560, opened 2018-08-21
conf: fix devpts mounting when fully unprivileged

brauner:

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1625,8 +1625,10 @@ static const struct id_map *find_mapped_nsid_entry(struct lxc_conf *conf,
 static int lxc_setup_devpts(struct lxc_conf *conf)
 {
 	int ret;
-	char default_devpts_mntopts[] = "gid=5,newinstance,ptmxmode=0666,mode=0620";
+	char **opts;
 	char devpts_mntopts[256];
+	char *mntopt_sets[5];
+	char default_devpts_mntopts[256] = "gid=5,newinstance,ptmxmode=0666,mode=0620";
 
 	if (conf->pty_max <= 0) {
 		DEBUG("No new devpts instance will be mounted since no pts "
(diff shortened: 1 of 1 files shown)


src/lxc/conf.c
```diff
@@ -1652,29 +1659,27 @@ static int lxc_setup_devpts(struct lxc_conf *conf)
...
-			}
-		}
+	/* gid=5 && max= */
+	mount_options[0] = devpts_mntopts;
+	/* !gid=5 && max= */
+	mount_options[1] = devpts_mntopts + sizeof("gid=5");
+	/* gid=5 && !max= */
+	mount_options[2] = default_devpts_mntopts;
+	/* !gid=5 && !max= */
+	mount_options[2] = default_devpts_mntopts + sizeof("gid=5");
```
Blub: Should be `[3]`, and then `mount_options` needs to be resized to 5 to not break the loop condition below.

src/lxc/conf.c
```diff
@@ -1625,8 +1625,15 @@ static const struct id_map *find_mapped_nsid_entry(struct lxc_conf *conf,
 static int lxc_setup_devpts(struct lxc_conf *conf)
 {
 	int ret;
-	char default_devpts_mntopts[] = "gid=5,newinstance,ptmxmode=0666,mode=0620";
+	char **it;
 	char devpts_mntopts[256];
+	char default_devpts_mntopts[256] = "gid=5,newinstance,ptmxmode=0666,mode=0620";
+	char *mount_options[4] = {
```
Blub: I think we can skip the initialization here since it's filled in single block anyway, the last `NULL` can also be added to the assignments below.

src/lxc/conf.c
brauner: Thanks, added!

src/lxc/conf.c
brauner: Yup, done.

PR #2569, opened 2018-08-23
execute: skip lxc-init logging when unprivileged

brauner:
Unprivileged app containers will not be able to open the passed in
/proc/1/fd/<idx> log path and will thus currently fail completely as soon as
any log level or log file is passed.

Reported-by: Mrinal Dhillon <mdhillon@juniper.net>


tych0:
LGTM

PR #2574, opened 2018-08-27
 confile: add lxc.cgroup.keep

brauner:
This adds the new lxc.cgroup.keep config key. The key can be used to instruct
LXC to not escape to never escape to the root cgroup. This makes it easy for
users to adhere to restrictions enforced by cgroup2 and systemd. Specifically,
this makes it possible to run LXC containers as systemd services.

Note that cgroup v1 is considered legacy and will not see additional
controllers being added to it. This means that it is safe to use
lxc.cgroup.keep as config key since there is no "keep" controller. The only way
a conflict can be introduced is if the user is creating a named controller. I
think this case can be safely ignored since it is super rare and also the users
problem.

Cc: Felix Abecassis <fabecassis@nvidia.com>
Cc: Jonathan Calmels <jcalmels@nvidia.com>

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1261,7 +1261,7 @@ int lxc_attach(const char *name, const char *lxcpath,
 		if (options->attach_flags & LXC_ATTACH_MOVE_TO_CGROUP) {
 			struct cgroup_ops *cgroup_ops;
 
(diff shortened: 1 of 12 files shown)


brauner:
//cc @3XX0, @flx42 :)

PR #2577, opened 2018-08-29
tree-wide: use sizeof on static arrays

Blub:
I think this way it's easier with future changes, unless someone changes an array to a pointer, but then all uses must be checked anyway, and we already have both versions in the code.


brauner:
Perfect!

PR #2584, opened 2018-09-03
commands: switch to setting errno and returning -1

brauner:


src/lxc/commands.c
```diff
@@ -311,24 +294,27 @@ static int lxc_cmd(const char *name, struct lxc_cmd_rr *cmd, int *stopped,
...
 
 	ret = lxc_cmd_rsp_recv(client_fd, cmd);
-	if (ret == -ECONNRESET)
+	if (ret < 0 && errno == ECONNRESET)
 		*stopped = 1;
 
 	if (!stay_connected || ret <= 0)
-		if (client_fd >= 0)
+		if (client_fd >= 0) {
+			saved_errno = errno;
```
brauner: Actually, the saved_errno can be made local to the if branch.

brauner: But no showstopper so can be merged as is.

src/lxc/commands.c
```diff
@@ -147,10 +147,12 @@ static int lxc_cmd_rsp_recv(int sock, struct lxc_cmd_rr *cmd)
 
 		rspdata = malloc(sizeof(*rspdata));
 		if (!rspdata) {
-			ERROR("Failed to allocate response buffer for command \"%s\"",
-			      lxc_cmd_str(cmd->req.cmd));
-			return -ENOMEM;
+			errno = ENOMEM;
+			SYSERROR("Failed to allocate response buffer for command \"%s\"",
```
Blub: The logging calls will clobber `errno` again. I think returning the value is fine in this function btw., as long as we do it consistently (see below), given that the callers want to handle `ECONNRESET` specifically, so the style guide would currently prefer returning a negative error code.
Also I think the error message is detailed enough and since we know the exact text the `errno` will produce (since we're assigning it), I'm not sure it's worth switching to the `SYS*` macro.

src/lxc/commands.c
```diff
@@ -167,7 +169,7 @@ static int lxc_cmd_rsp_recv(int sock, struct lxc_cmd_rr *cmd)
 		errno = EFBIG;
 		SYSERROR("Response data for command \"%s\" is too long: %d bytes > %d",
```
Blub: Specifically this is an example where the explicit `EFBIG` doesn't really provide any additional info.

src/lxc/commands.c
```diff
@@ -180,15 +182,14 @@ static int lxc_cmd_rsp_recv(int sock, struct lxc_cmd_rr *cmd)
...
 	}
 
-	ret = recv(sock, rsp->data, rsp->datalen, 0);
+	ret = lxc_recv_nointr(sock, rsp->data, rsp->datalen, 0);
 	if (ret != rsp->datalen) {
 		SYSERROR("Failed to receive response data for command \"%s\"",
 		         lxc_cmd_str(cmd->req.cmd));
-		if (ret >= 0)
-			ret = -1;
+		return -1;
```
Blub: The callers seem to want to handle `ECONNRESET`, which AFAIK can occur here as well. Previously we only had the case above where we explicitly returned `-ECONNRESET`, so this looks like it was an actual issue here.

src/lxc/commands.c
brauner: The answer is to not globber errno in the log. I'll push a series of patches on top of this one that will take care of this.

src/lxc/commands.c
```diff
@@ -227,51 +227,33 @@ static int lxc_cmd_rsp_send(int fd, struct lxc_cmd_rsp *rsp)
...
-		if (errno == EPIPE) {
-			close(client_fd);
-			return -EPIPE;
-		}
-		close(client_fd);
-
-		return -1;
-	}
-
-	if ((size_t)ret != sizeof(cmd->req)) {
```
Blub: This would be different in the new code where `errno` would be `0` AFAICT, although I'm not sure this can actually happen when sending. If the message is too long `sendmsg()` (used by `lxc_abstract_unix_send_credential()`) would return `-1` with `errno=EMSGSIZE` anyway. Can a `sendmsg()` actually succeed with less than requested the length?
Anyway, the new code would have to set `errno` explicitly for this case as well.

The `send()` case below is similar.

src/lxc/log.h
```diff
@@ -287,7 +288,9 @@ ATTR_UNUSED static inline void LXC_##LEVEL(struct lxc_log_locinfo* locinfo,	\
 		 * without restrictions. So let's use it for our		\
 		 * logging stamps.						\
 		 */								\
+		saved_errno = errno;						\
 		(void)clock_gettime(CLOCK_REALTIME, &evt.timestamp);		\
+		errno = saved_errno;						\
```
Blub: any reason not to restore after the `va_end`?
I think that might make half the other related commits superfluous? Maybe?

src/lxc/commands.c
brauner: Something I wouldn't rely on.

src/lxc/log.h
brauner: Good call! Done!

src/lxc/commands.c
brauner: Done!

PR #2585, opened 2018-09-03
caps: fix illegal access to array bound

2xsec:
Hello,

fix illegal access to array bound

Thanks.

### DIFF
--- a/src/lxc/caps.c
+++ b/src/lxc/caps.c
@@ -296,15 +296,13 @@ static long int _real_caps_last_cap(void)
 	if (fd >= 0) {
 		ssize_t n;
 		char *ptr;
-		char buf[INTTYPE_TO_STRLEN(int)];
+		char buf[INTTYPE_TO_STRLEN(int)] = {0};
 
 	again:
-		n = read(fd, buf, sizeof(buf));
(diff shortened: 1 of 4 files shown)


src/lxc/caps.c
```diff
@@ -303,7 +303,7 @@ static long int _real_caps_last_cap(void)
 		if (n < 0 && errno == EINTR) {
 			goto again;
 		} else if (n >= 0) {
-			buf[n] = '\0';
+			buf[n-1] = '\0';
```
brauner: That won't work. Assume that `n == 0`. Now you're doing `n - 1` causing `buf[-1] = \0`. Which ironically is an out-of-bounds access. :)
The better way would be to either set the buffer to `0` at declaration time or special case when `n == 0`. That's really up to you. The only case we need to worry about is `n == sizeof(buf)`. An even simpler trick is to cause the read to be truncated, i.e. either add `+1` to the buffer or do `sizeof(buf) - 1` during read. The latter two options are my preference.

2xsec: @brauner oh.. You're right. Thanks for your review. I'll check it again.

brauner:
Thanks!

PR #2589, opened 2018-09-06
lxccontainer: fix dereferenced pointer

2xsec:
Hello,

dereferenced pointer issues are fixed.

Thanks.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -583,6 +583,9 @@ static int do_lxcapi_console_log(struct lxc_container *c, struct lxc_console_log
 {
 	int ret;
 
+	if (!c)
+		return -EINVAL;
+
 	ret = lxc_cmd_console_log(c->name, do_lxcapi_get_config_path(c), log);
 	if (ret < 0) {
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

src/lxc/lxccontainer.c
```diff
@@ -583,6 +583,9 @@ static int do_lxcapi_console_log(struct lxc_container *c, struct lxc_console_log
 {
 	int ret;
 
+	if (!c)
+		return -1;
```
Blub: Thanks for the PR.
Seeing how the rest of this function is returning negative errno codes this should probably return `-EINVAL`.

2xsec: @Blub It's better. Thanks for your review.:)

brauner: Good point. Set errno or negative return in API calls?

brauner:
Thanks! Sorry for the delay I'm currently travelling.

2xsec:
@brauner It's OK. Have a nice trip :)

Issue #2591, opened 2018-09-07
lxc-attach doesn't allow to set uid / gid


Disassembler0:
# Required information

 * Distribution: Alpine
 * Distribution version: 3.8
 * `lxc-start --version`: 3.0.0+, compiled from git master, commit a21ed5555dfc9bbe8bf1fe3004b3878b6e668f94
* `lxc-checkconfig`: N/A
* `uname -a`: Linux vm 4.14.52-0-virt #1-Alpine SMP Tue Jun 26 07:21:56 UTC 2018 x86_64 Linux
* `cat /proc/self/cgroup`:
```
13:pids:/
12:hugetlb:/
11:net_prio:/
10:perf_event:/
9:net_cls:/
8:freezer:/
7:devices:/
6:memory:/
5:blkio:/
4:cpuacct:/
3:cpu:/
2:cpuset:/
1:name=openrc:/
0::/
```

# Issue description
This is probably more a feature request since the functionality was never there, but here goes.

I'd like lxc-attach to be able to switch UID / GID in the entered namespace. Basically a rough equivalent of `docker exec`, which sets UID to the one given in Dockerfile (unless overridden by `-u`). I assume that `lxc-attach` does pretty much the same as `nsenter`, which allows for options `-S` to set UID and `-G` to set GID for the program subsequently executed in the entered namespace.

My use case is that I have bunch of application containers. One container is running something what needs to be accessible for the others at all times (e.g. postgres). When I'm setting up another container with an application which is supposed to use that postgres, I'm initializing a new database for it from the LXC host. However my postgres setup requires that all actions are done by postgres user, so when I `lxc-attach`, I keep being root and I either need to switch inside the container (via `su`, `sudo`, `runas`, `setuidgid`, `s6-setuidgid` etc.) or I simply can't use `lxc-attach` and have to use `nsenter` instead which allows me to set UID. Also `lxc-execute`, which would normally set UID given by `lxc.init.uid` config value is out of the question, because the container is already running and needs to keep running.

**tl;dr:** Please add set UID / GID switches to `lxc-attach` in similar fashion like `nsenter` or `lxc-execute` have.  
(additionally, `lxc-attach` could honor `lxc.init.uid`, but I'd be happy just with the switches.)

brauner:
Yeah, that should not be difficult to add. Just need to see what flags to use.

Disassembler0:
[...]

PR #2595, opened 2018-09-08
utils: allow lxc-attach to set uid / gid

Disassembler0:
- Allow lxc-attach to set UID / GID as requested in #2591
 - Rearrange order of actions on attach
 - Add documentation (also for lxc-execute which has the same params)

Signed-off-by: Disassembler <disassembler@dasm.cz>

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -852,13 +852,6 @@ static int attach_child_main(struct attach_clone_payload *payload)
 	if (options->gid != (gid_t)-1)
 		new_gid = options->gid;
 
-	/* Try to set the {u,g}id combination. */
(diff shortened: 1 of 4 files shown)


src/lxc/attach.c
```diff
@@ -897,6 +890,13 @@ static int attach_child_main(struct attach_clone_payload *payload)
 		TRACE("Loaded seccomp profile");
 	}
 
+	/* Try to set the {u,g}id combination. */
+	if (new_uid != 0 || new_gid != 0 || options->namespaces & CLONE_NEWUSER) {
+		ret = lxc_switch_uid_gid(new_uid, new_gid);
+		if (ret < 0)
+			goto on_error;
+	}
```
brauner: Ok, that needs to be done differently but I can do it. I'll just put on top of your branch.

PR #2601, opened 2018-09-11
log: support dlog

2xsec:
Hello,

This patch is for supporting dlog to print logs.

It can be used on all tizen devices because tizen is using dlog for printing logs as default.

The default option is disable. In order to enable dlog, --enable-dlog option is needed.

Thanks.


brauner:
I have nothing against this patch in principle but can you point me to the dlog source code, please? I need to see the license.

2xsec:
@brauner The license is apache version 2.0. You can see the source codes of the dlog. :)
https://review.tizen.org/git/?p=platform/core/system/dlog.git;a=summary

brauner:
Ok, then seems legit to me. If that's useful to you and you have it behind a compile flag, so be it. :)

brauner:
jenkins: test this please

2xsec:
@brauner Thanks for understanding. :)

PR #2606, opened 2018-09-11
cgroups: scoping for cgroup v2

brauner:
This adds a new scoping mechanism to even better integrate with cgroup v2 but will also be used with cgroup v1.
//fyi @poettering

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -819,7 +819,7 @@ static struct hierarchy *add_hierarchy(struct hierarchy ***h, char **clist, char
 	new->controllers = clist;
 	new->mountpoint = mountpoint;
 	new->base_cgroup = base_cgroup;
-	new->fullcgpath = NULL;
+	new->container_full_path = NULL;
 	new->version = type;
 
 	newentry = append_null_to_list((void ***)h);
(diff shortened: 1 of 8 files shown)


src/lxc/cgroups/cgfsng.c
```diff
@@ -817,7 +817,7 @@ static struct hierarchy *add_hierarchy(struct hierarchy ***h, char **clist, char
 	new->controllers = clist;
 	new->mountpoint = mountpoint;
 	new->base_cgroup = base_cgroup;
-	new->fullcgpath = NULL;
+	new->container_full_path = NULL;
```
hallyn: won't nack, but the new name is less precise.

hallyn:
I'd merge, but there are conflicts to be resolved.

brauner:
@hallyn, resolved :)

poettering:
(thanks for doing this. Don't know LXC too well, to be able to review this, but if this does what @brauner showed me then this is excellent, thank you!)

brauner:
@poettering, np. The only corner-case I have to handle in another PR is that the monitor now needs to delete it's own cgroup. So we have the layout:
```
/sys/fs/cgroup/<delegated>/lxc.monitor/c1
/sys/fs/cgroup/<delegated>/lxc.payload/c1
```
The monitor can now delete `/sys/fs/cgroup/<delegated>/lxc.payload/c1` but it can't delete `/sys/fs/cgroup/<delegated>/lxc.monitor/c1` since it lives in there. I also don't want to spam the hierarchy with useless empty cgroups especially on systems where the init system doesn't clean up. So I'm creating another cgroup
```
/sys/fs/cgroup/<delegated>/lxc.pivot
```
Which only exists so that on container stop the monitor can pivot in there and then delete `/sys/fs/cgroup/<delegated>/lxc.monitor/c1` and then exit. The `lxc.pivot` cgroup is shared by all monitors since it only exits for the sole purpose of pivoting into it during container stop.

Issue #2615, opened 2018-09-15
lxc-3.0.2 fails to build on CentOS 7 with "error: 'for' loop initial declarations are only allowed in C99 mode"


ganto:
# Required information

 * Distribution: CentOS
 * Distribution version: 7

# Issue description

Building `lxc-3.0.2` on CentOS 7 (`gcc-4.8.5-28.el7_5.1.x86_64`) fails with:
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../src -fPIC -DPIC -DLXCROOTFSMOUNT=\"/usr/lib64/lxc/rootfs\" -DLXCPATH=\"/var/lib/lxc\" -DLXC_GLOBAL_CONF=\"/etc/lxc/lxc.conf\" -DLXCINITDIR=\"/usr/libexec\" -DLIBEXECDIR=\"/usr/libexec\" -DLXCTEMPLATEDIR=\"/usr/share/lxc/templates\" -DLXCTEMPLATECONFIG=\"/usr/share/lxc/config\" -DLOGPATH=\"/var/log/lxc\" -DLXC_DEFAULT_CONFIG=\"/etc/lxc/default.conf\" -DLXC_USERNIC_DB=\"/run/lxc/nics\" -DLXC_USERNIC_CONF=\"/etc/lxc/lxc-usernet\" -DDEFAULT_CGROUP_PATTERN=\"lxc/%n\" -DRUNTIME_PATH=\"/run\" -DSBINDIR=\"/usr/sbin\" -I ../../src -I ../../src/lxc -I ../../src/lxc/storage -I ../../src/lxc/cgroups -DHAVE_SECCOMP -DHAVE_SELINUX -pthread -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -m64 -mtune=generic -c confile.c  -fPIC -DPIC -o .libs/liblxc_la-confile.o
BUILDSTDERR: In file included from ./utils.h:48:0,
BUILDSTDERR:                  from lsm/lsm.h:31,
BUILDSTDERR:                  from conf.c:100:
BUILDSTDERR: conf.c: In function 'parse_mntopts':
BUILDSTDERR: ./macro.h:135:2: error: 'for' loop initial declarations are only allowed in C99 mode
[...]

PR #2616, opened 2018-09-15
Declare loop variable outside of 'for' statement

ganto:
This fixes a build failure on CentOS 7 (#2615):
```
  ./macro.h:135:2: error: 'for' loop initial declarations are only allowed in C99 mode
    for (char *__p = NULL, *__it = strtok_r(__splitme, __separators, &__p); \
    ^
```

### DIFF
--- a/src/lxc/macro.h
+++ b/src/lxc/macro.h
@@ -228,7 +228,8 @@ extern int __build_bug_on_failed;
 #endif
 
 #define lxc_iterate_parts(__iterator, __splitme, __separators)                  \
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

brauner:
The problem with declaring those variables before the loop is that they can conflict with variables that users declare that have the same name. I'd like to avoid that very much.

ganto:
My issue already is fixed by #2594. Therefore no need for this patch.

PR #2618, opened 2018-09-18
apparmor: account for specified rootfs path (closes #2617)

### LINKED ISSUE #2617: apparmor abstraction 'start-container' hard codes rootfs mount directory
The configure script allows specifying the directory that lxc pivots to using the `--with-rootfs-path` option. However, the apparmor profile hardcodes the default libdir matches. This results in the following apparmor denial:

    apparmor="DENIED" operation="mount" info="failed flags match" error=-13 profile="/usr/bin/lxc-start" name="/var/lxc/containers/" pid=24775 comm="lxc-start" srcname="/var/lib/lxc/dropkick/rootfs/" flags="rw, rbind"

To fix it, something like the following should be added to the start-container abstraction:

    mount -> @LXCROOTFSMOUNT@/{,**},
    pivot_root @LXCROOTFSMOUNT@/{,**},

And the abstraction should be processed with make somehow.

CameronNemo:
Signed-off-by: Cameron Nemo <camerontnorman@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,6 +111,7 @@ config/ltmain.sh
 config/missing
 config/libtool.m4
 config/lt*.m4
+config/apparmor/abstractions/start-container
 config/bash/lxc
 config/init/common/lxc-containers
 config/init/common/lxc-net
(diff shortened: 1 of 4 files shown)


brauner:
This doesn't compile.

brauner:
Or rather install.

CameronNemo:
I've been meaning to investigate, but it works locally so it may take me some effort. You can see on line 931 of the build log that the supposedly missing file is being generated.

PR #2619, opened 2018-09-18
attach: don't shutdown ipc socket in child

smibarber:
shutdown() affects sockets even across forked processes. The
attached child process doesn't have any interest in using the
IPC socket, so just close it in the child process and let the
intermediate process handle shutting it down.

This fixes a bug seen with lxc exec in crbug.com/884244

Signed-off-by: Stephen Barber <smbarber@chromium.org>


brauner:
jenkins: test this please

PR #2622, opened 2018-09-19
attach: report standard shell exit codes

brauner:
POSIX mandates that on ENOEXEC 126 and on ENOENT 127 is supposed to be
reported.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1510,14 +1510,23 @@ int lxc_attach(const char *name, const char *lxcpath,
 	_exit(0);
 }
 
-int lxc_attach_run_command(void* payload)
+int lxc_attach_run_command(void *payload)
 {
-	lxc_attach_command_t* cmd = (lxc_attach_command_t*)payload;
+	int ret = -1;
+	lxc_attach_command_t *cmd = payload;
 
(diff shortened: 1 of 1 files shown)


src/lxc/attach.c
```diff
@@ -1512,9 +1512,18 @@ int lxc_attach(const char *name, const char *lxcpath,
...
+	if (ret < 0) {
+		switch (errno) {
+		case ENOEXEC:
+			return 126;
+		case ENOENT:
+			return 127;
+		}
+	}
 
 	SYSERROR("Failed to exec \"%s\"", cmd->program);
```
candlerb: Shouldn't SYSERROR be reported *before* the return case above? Otherwise nothing gets logged any more for ENOEXEC and ENOENT!

brauner: Yeah, very obvious oversight on my part. Thanks!

brauner: fixed

candlerb:
Thanks.  Note that assigning and testing the value of `ret` is unnecessary - from the manpage:

> The  exec() functions return only if an error has occurred.  The return value is -1, and errno is set to indicate the error.

brauner:
Yeah, but it looks cleaner to me that way.

PR #2627, opened 2018-09-21
conf: realpath() uses null as second parameter to prevent buffer overflow

2xsec:
Hello,

realpath() does not use null as second parameter. It can cause buffer overflow.

So, the second parameter is modified to null to prevent buffer overflow.

Thanks.


2xsec:
@brauner OK. Thanks for your review. :)

PR #2639, opened 2018-09-23
compiler: compiler based hardening

brauner:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -685,6 +685,7 @@ LXC_CHECK_TLS
 
 AX_CHECK_COMPILE_FLAG([-fdiagnostics-color], [CFLAGS="$CFLAGS -fdiagnostics-color"],,[-Werror])
 AX_CHECK_COMPILE_FLAG([-implicit-fallthrough], [CFLAGS="$CFLAGS -Wimplicit-fallthrough"],,[-Werror])
+AX_CHECK_COMPILE_FLAG([-Wcast-align], [CFLAGS="$CFLAGS -Wcast-align"],,[-Werror])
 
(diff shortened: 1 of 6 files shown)


Blub:
I always wondered why the members of `struct ifaddrs` are defined as `struct sockaddr*` and not `struct sockaddr_storage*` (which AFAIK is supposed to be able to contain all variants of it and should prevent this kind of warning here).
Is this the point where we start using `#pragma clang diagnostic push`+`...ignored "-Wcast-align"`+`... pop` sections? Since with the af_family checks these casts are legal after all.

brauner:
> I always wondered why the members of struct ifaddrs are defined as struct sockaddr* and not struct sockaddr_storage* (which AFAIK is supposed to be able to contain all variants of it and should prevent this kind of warning here).
Is this the point where we start using #pragma clang diagnostic push+...ignored "-Wcast-align"+... pop sections? Since with the af_family checks these casts are legal after all.

The short and generic answer is: anything that hardens the codebase is good. So anything that you want in that direction will be received with generous support. :)
The only things I would like to add:
- The hardening must be reasonable wrt to performance. An example is `fstack-protector-all` probably too much and has performance impact better use `fstack-protector-strong` since it has a lot of benefit but does not impact peformance.
- The hardening must defend against something that is a plausible attach on the library. For example, we likely don't need encryption-based hardening options since we don't do encryption.

stgraber:
jenkins: test this please

brauner:
jenkins: test this please

stgraber:
@brauner try rebasing on master + force-push that may fix it

brauner:
yup, will do

PR #2643, opened 2018-09-23
cgroups: implement monitor cgroup deletion

brauner:
Since we switched to the new cgroup scoping scheme that places the
container payload into `lxc.payload/<container-name>` and
`lxc.monitor/<container-name>` deletion becomes slightly more complicated.
The monitor will be able to `rm_rf(lxc.payload/<container-name>)` but will
not be able to `rm_rf(lxc.monitor/<container-name>)` since it will be
located in that cgroup and it will thus be populated.
My current solution to this is to create a lxc.pivot cgroup that only
exists so that the monitor process on container stop can pivot into it,
call `rm_rf(lxc.monitor/<container-name>)` and can then exit. This group
has not function whatsoever apart from this and can thus be shared by
all monitor processes.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1326,7 +1326,9 @@ static void remove_path_for_hierarchy(struct hierarchy *h, char *cgname, bool mo
 __cgfsng_ops static inline bool cgfsng_monitor_create(struct cgroup_ops *ops,
 							struct lxc_handler *handler)
 {
-	char *monitor_cgroup;
+	char *monitor_cgroup, *offset, *tmp;
+	int idx = 0;
+	size_t len;
 	bool bret = false;
 	struct lxc_conf *conf = handler->conf;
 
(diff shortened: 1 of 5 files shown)


brauner:
I'm sorry, @hallyn. :)

src/lxc/cgroups/cgfsng.c
```diff
@@ -1124,6 +1124,60 @@ __cgfsng_ops__ static void cgfsng_payload_destroy(struct cgroup_ops *ops,
...
+{
+	int len;
+	char *pivot_path;
+	struct lxc_conf *conf = handler->conf;
+	char pidstr[25];
+
+	if (!ops->hierarchies)
+		return;
+
+	len = snprintf(pidstr, 25, "%d", handler->monitor_pid);
```
Blub: I'd prefer `sizeof(pidstr)` in all the code paths below. Magic numbers always makes me search for context.

src/lxc/cgroups/cgfsng.c
```diff
@@ -1317,32 +1317,56 @@ static void remove_path_for_hierarchy(struct hierarchy *h, char *cgname, bool mo
...
+				ops->hierarchies[i]->container_full_path = NULL;
+
+				for (int j = 0; j < i; j++)
+					remove_path_for_hierarchy(ops->hierarchies[j], monitor_cgroup, true);
+
+				idx++;
+				break;
+			}
+		}
+	} while (idx > 0 && idx < 1000);
```
Blub: The loop will end if `idx==0`, but if we hit the error case where `idx` has been incremented, there seems to be a missing success-condition? Maybe pull `int i` out of the `for` loop and break on `!ops->hierarchies[i]`?

PR #2645, opened 2018-09-24
stop: Only freeze if freezer is available

### LINKED ISSUE #2644: liblxc appears to attempt to unfreeze containers on stop when freezer cgroup doesn't exist
On a CGroupV2-only system, attempting to stop a container forcefully (not sending signal) gets me:
```
root@djanet:~# cat /var/log/lxd/c1/lxc.log
lxc c1 20180924182230.487 ERROR    lxc_commands - commands.c:lxc_cmd_stop_callback:656 - Failed to unfreeze container "c1"
lxc 20180924182230.488 ERROR    lxc_commands - commands.c:lxc_cmd_stop:628 - Operation not permitted - Failed to stop container "c1"
```

Which suggests that liblxc is attempting to un-conditionally unfreeze containers even when the freezer cgroup doesn't exist.

stgraber:
Closes #2644

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -654,6 +654,9 @@ static int lxc_cmd_stop_callback(int fd, struct lxc_cmd_req *req,
 		 * lxc_unfreeze() would do another cmd (GET_CGROUP) which would
 		 * deadlock us.
 		 */
+		if (cgroup_ops->get_cgroup(cgroup_ops, "freezer") == NULL)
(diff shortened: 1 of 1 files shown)


stgraber:
@brauner so Jenkins seems happy with it, not sure if there's a cleaner way you'd want this done though :)

brauner:
Let's merge this to unblock things and I can come up with something more performant later.

PR #2652, opened 2018-09-26
tree-wide: fix includes to fix bionic builds

brauner:


brauner:
@stgraber, it cleans up the includes to avoid double definitions and so on to fix Android and cleanup other builds for good. No functional changes.

PR #2657, opened 2018-09-28
oci-template: Add logic for no /etc/passwd, group

ssup2:
OCI image spec dosen't specify an action when there is
no /etc/passwd or /etc/group. So if there is no
/etc/passwd with string user info, set uid to 0. If there
is no /etc/group with string group info, set gid to 0.

Signed-off-by: Jungsub Shin jungsub_shin@tmax.co.kr


stgraber:
Would be nice if we had someone with a clue about OCI stuff who could help review those... the shell seems correct but I have no idea if the action is :)

PR #2661, opened 2018-09-28
autotools: support -z relro and -z now

brauner:


brauner:
Technically, we always did this through our hardening wrapper but let's make it the default in autotools too.

PR #2687, opened 2018-10-10
conf: verify_start_hooks() after lxc.mount.entry

brauner:
Fixes: https://discuss.linuxcontainers.org/t/are-lxc-mount-entry-available-when-lxc-hook-start-is-validated/2906/3

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3514,7 +3514,6 @@ static bool verify_start_hooks(struct lxc_conf *conf)
 
 	lxc_list_for_each (it, &conf->hooks[LXCHOOK_START]) {
 		int ret;
-		struct stat st;
 		char *hookname = it->elem;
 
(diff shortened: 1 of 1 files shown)


joaocfernandes:
Thank you, for your fast reply.

src/lxc/conf.c
```diff
@@ -3521,9 +3520,9 @@ static bool verify_start_hooks(struct lxc_conf *conf)
 		if (ret < 0 || ret >= PATH_MAX)
 			return false;
 
-		ret = stat(path, &st);
+		ret = access(path, F_OK);
```
stgraber: Wouldn't X_OK actually be better since we'll be execing it?

brauner: Hm, good point.

brauner: updated

PR #2691, opened 2018-10-12
Some redundancy codes of abstract unix socket are removed with log cleanups.

2xsec:
Hello,

Some redundancy codes are removed with log cleanups.

Code logic is not changed.

Thanks.


brauner:
Thanks! :)

PR #2694, opened 2018-10-14
lxc-init: log to /dev/console

brauner:


stgraber:
Am I missing something or is that going to seriously break CLI backward compatibility?

brauner:
Not sure I understand. The log and logfile options are still going to be passed. They're not going to be used since we will log to stderr which is duped to /dev/console. So that's a definite change on behavior but logging has just really worked one version back or something when this was fixed so nobody complained about this for a very long time. I also would like to get rid of having the inside of the container have an actual file descriptor for the logfile of the monitor.

stgraber:
@brauner nevermind, misread

PR #2702, opened 2018-10-19
parse: fix uninitialized value

2xsec:
Hello,

uninitialized value is fixed.

Thanks.


brauner:
Good catch, thanks! :)

2xsec:
@brauner You're welcome. A compiler always help us reduce issues. :)

PR #2726, opened 2018-11-16
conf: fix chroot jail issue

2xsec:
Hello,

chroot jail issue is fixed.

Thanks.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1463,6 +1463,12 @@ int lxc_chroot(const struct lxc_rootfs *rootfs)
 		return -1;
 	}
 
+	ret = chdir("/");
+	if (ret < 0) {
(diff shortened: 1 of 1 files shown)


2xsec:
@brauner It is possible to access file system outside of the chroot directory if the current directory is not contained in it. It is just possibility related security issue. But, we also need your opinion. Please review it. :)

Blub:
But we use `chroot(".")`, iow, chroot into "the current directory". Is it still possible for `/` not to be the same as `./` in this case?

2xsec:
@Blub Do you know why a working directory of this is changed to new root by this code?

	/* It is weird, but chdir("..") moves us in a new root */
	ret = chdir("..");
	if (ret < 0) {
		SYSERROR("Failed to chdir(\"..\")");
		return -1;
	}

brauner:
Ok :)

PR #2740, opened 2018-11-29
lxc-usernsexec: fix default map functionality

CameronNemo:
* Place NULL bytes at the end of strings so that
  lxc_safe_ulong() can parse them correctly

* Only free the newly created id_map on error,
  to avoid passing garbage to lxc_map_ids()

Signed-off-by: Cameron Nemo <camerontnorman@gmail.com>

### DIFF
--- a/src/lxc/cmd/lxc_usernsexec.c
+++ b/src/lxc/cmd/lxc_usernsexec.c
@@ -200,6 +200,7 @@ static int read_default_map(char *fnam, int which, char *user)
 {
 	size_t len;
 	char *p1, *p2;
+	unsigned long ul1, ul2;
 	FILE *fin;
 	int ret = -1;
 	size_t sz = 0;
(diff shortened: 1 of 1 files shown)


src/lxc/cmd/lxc_usernsexec.c
```diff
@@ -224,37 +225,42 @@ static int read_default_map(char *fnam, int which, char *user)
...
+		newmap->range = ul2;
 
-		ret = -1;
 		tmp = malloc(sizeof(*tmp));
-		if (!tmp)
-			goto on_error;
+		if (!tmp) {
+			free(newmap);
+			break;
+		}
```
hallyn: This is ok, but another fix would have been to set

```
newmap = NULL;
```

here.

But actually, what you have here changes behavior.  And I think fixes another bug.

If there are no matching entries, you now return -1.  Previously it returned 0 in that case, which the caller took to mean "there was a matching entry".

hallyn:
jenkins: test this please

PR #2749, opened 2018-12-11
storage: src cannot be truncated

brauner:
s/strlcpy/strncpy/g


brauner:
@Blub, you're right. I removed this function now. :)

PR #2750, opened 2018-12-11
CMD log macros: append a newline

tych0:
The log appenders for the regular log macros append a newline as necessary,
and the usage of these is written so it does to. So, let's add a newline.


tych0:
Acutally this isn't quite right, let me send a better version.

PR #2758, opened 2018-12-17
apparmor: allow various remount,bind options

Blub:
RW bind mounts need to be restricted for some paths in
order to avoid MAC restriction bypasses, but read-only bind
mounts shouldn't have that problem.

Additionally, combinations of 'nosuid', 'nodev' and
'noexec' flags shouldn't be a problem either and are
required with newer systemd versions, so let's allow those
as long as they're combined with 'ro,remount,bind'.

(cherry picked from commit e6ec0a9e71aa68c9fd67c691a62aaae87e356cef)

### DIFF
--- a/config/apparmor/abstractions/container-base
+++ b/config/apparmor/abstractions/container-base
@@ -120,6 +120,16 @@
   mount options=(rw,bind) /sy[^s]*{,/**},
   mount options=(rw,bind) /sys?*{,/**},
 
+  # allow various ro-bind-*re*-mounts
+  mount options=(ro,remount,bind),
+  mount options=(ro,remount,bind,nosuid),
+  mount options=(ro,remount,bind,noexec),
+  mount options=(ro,remount,bind,nodev),
+  mount options=(ro,remount,bind,nosuid,noexec),
+  mount options=(ro,remount,bind,noexec,nodev),
+  mount options=(ro,remount,bind,nodev,nosuid),
+  mount options=(ro,remount,bind,nosuid,noexec,nodev),
+
   # allow moving mounts except for /proc, /sys and /dev
   mount options=(rw,move) /[^spd]*{,/**},
   mount options=(rw,move) /d[^e]*{,/**},
--- a/config/apparmor/abstractions/container-base.in
+++ b/config/apparmor/abstractions/container-base.in
@@ -119,6 +119,16 @@
   mount options=(rw,bind) /sy[^s]*{,/**},
   mount options=(rw,bind) /sys?*{,/**},
 
+  # allow various ro-bind-*re*-mounts
+  mount options=(ro,remount,bind),
+  mount options=(ro,remount,bind,nosuid),
+  mount options=(ro,remount,bind,noexec),
+  mount options=(ro,remount,bind,nodev),
+  mount options=(ro,remount,bind,nosuid,noexec),
+  mount options=(ro,remount,bind,noexec,nodev),
+  mount options=(ro,remount,bind,nodev,nosuid),
+  mount options=(ro,remount,bind,nosuid,noexec,nodev),
+
   # allow moving mounts except for /proc, /sys and /dev
   mount options=(rw,move) /[^spd]*{,/**},
   mount options=(rw,move) /d[^e]*{,/**},
@@ -136,4 +146,3 @@
   mount options=(rw,move) /s[^y]*{,/**},
   mount options=(rw,move) /sy[^s]*{,/**},
   mount options=(rw,move) /sys?*{,/**},
-


brauner:
Thanks!

Blub:
`flags=rw,slave`, that's #2778

Zjemm:
Ah so im not the only one. in #2778 the guy is talking about arch, but this issue is beyond arch, as im having this on debian (testing) with debian (testing) lxc containers.

hopes this lands soon

Zjemm:
found anotherone today:

Jan 22 14:02:45 buster-template systemd[89]: systemd-resolved.service: Failed to set up mount namespacing: Permission denied
Jan 22 14:02:45 buster-template systemd[89]: systemd-resolved.service: Failed at step NAMESPACE spawning /lib/systemd/systemd-resolved: Permission denied

Zjemm:
is it a problem in lxc or apparmor or systemd?

r1de:
@Zjemm I believe the issue is with lxc or apparmor, that is what lead me here.
At least it seems like I am not alone... :)

nyc-tinker:
nginx also won't start inside an lxc arch container (4.20.4) on systemd 240.34-3.  Don't have access to host logs so below is client only.  I can confirm mariadb also doesn't work with the same log msg as https://github.com/lxc/lxc/pull/2758#issuecomment-457946063.  Systemd-networkd also didn't get up, but at least that I was able to fix (temporarily, until this gets fixed) removing namespace protection per [here](https://unix.stackexchange.com/questions/291270/systemd-networkd-fails-to-start-when-network-booting-via-pxelinux): 

```Jan 28 13:00:56 sl-cloud systemd[1]: Starting A high performance web server and a reverse proxy server...
Jan 28 13:00:56 srver systemd[832412]: nginx.service: Failed to set up mount namespacing: Permission denied
Jan 28 13:00:56 srver systemd[832412]: nginx.service: Failed at step NAMESPACE spawning /usr/bin/nginx: Permission denied
Jan 28 13:00:56 srver systemd[1]: nginx.service: Control process exited, code=exited, status=226/NAMESPACE

**Comment by fauust** [comment]
Hi,
linking this to @matthewdarwin report:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=920643

**Comment by matthewdarwin** [comment]
The following instructions (step 2) might resolve the situation:

```
lxc (1:3.1.0+really3.0.3-6) unstable; urgency=medium

  LXC 3 got some significant changes from LXC 2.

   1. The configuration files use different variables. A userland script
      lxc-update-config is available to update automatically your
      configuration files. An automatic update is possible and offered by
      debconf during the upgrade of lxc version < 3.0.2 to lxc version >=
      3.0.2. Mind that this update will only work for priviledged containers
      with configurations present in /var/lib/lxc/*/config and any other
      container will not be updated.
   2. AppArmor support in Debian has increased, thus preventing some systemd
      isolation features to work in LXC 3.0.X. Debian has backported some
      patches from LXC 3.1 that, along with some configurations in a
      container, will allow systemd isolation features to work.

      The required configuration parameters are the ones which follow:
        lxc.apparmor.profile = generated
        lxc.apparmor.allow_nesting = 1

      These parameters are provided in the `/etc/lxc/default.conf` file
      shipped with LXC 3. Hence, any newly created container will have these
      parameters set properly, execpt if you alter the forementionned file.
   3. lxc-templates is deprecated by upstream. The new way of building
      containers is via their distrobuilder software. This software isn't in
      Debian Buster, and thus, we still provide lxc-templates. If you relied
      on it (eg, with lxc.include parameter in some configuration file), you
      should install lxc-templates in case it doesn't come by itself (via
      recommends). Otherwise you may experience issues after the upgrade.

 -- Pierre-Elliott Bécue <peb@debian.org>  Sat, 09 Mar 2019 13:09:05 +0100
```

fauust:
Thanks @matthewdarwin !

Zjemm:
i can confirm that for example mariadb, whcih would not start in the container. DOES start with the settings:
lxc.apparmor.profile = generated
lxc.apparmor.allow_nesting = 1

how do those differ from :
lxc.apparmor.profile = unconfined

?

Blub:
`generated` is pretty much what you get from lxd, it builds a profile (starting from the base profile, with some parts controlled by what the kernel can do), and the `allow_nesting` flag allows some additional parts (mostly mounting and, if available, stacking with apparmor namespaces).
For the nesting profile content in particular, see https://github.com/lxc/lxc/blob/master/src/lxc/lsm/apparmor.c#L331

Zjemm:
Ok so will this be the default then?
Or do I have to set these settings Everytime I setup a new Debian container.

Because the problem with not being able to start services  is pretty common.

I noticed it with postfix, mariadb, dovecot. There must be more. 
So wouldn't it make sense to make it default in the Debian containers?

Blub:
I do think it would make sense as a default, but I would feel more comfortable if the remaining quirks on the apparmor side were fixed first.

TrogWarZ:
> i can confirm that for example mariadb, whcih would not start in the container. DOES start with the settings:
> lxc.apparmor.profile = generated
> lxc.apparmor.allow_nesting = 1

Anybody know what else can be tested if with these settings mariadb still does NOT starts with same situation?

AdamN1:
> > i can confirm that for example mariadb, whcih would not start in the container. DOES start with the settings:
> > lxc.apparmor.profile = generated
> > lxc.apparmor.allow_nesting = 1

Adding apparmor entries solves problems running mariadb / apache / etc.

Pastebin error that you receive.

PR #2763, opened 2018-12-26
Set c to NULL after freeing it

caglar10ur:
Signed-off-by: S.Çağlar Onur <caglar@10ur.org>

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -299,6 +299,7 @@ static void lxc_container_free(struct lxc_container *c)
 	c->config_path = NULL;
 
 	free(c);
+	c = NULL;
 }
 
 /* Consider the following case:


brauner:
Thanks!

2xsec:
@caglar10ur 'c' is local variable in lxc_container_free function. Is there any reason for setting variable 'c'?

caglar10ur:
@2xsec 'c' is a reference to the container struct

Blub:
Since this isn't the first commit of the sort: `c` is in fact a local variable here. The data it pointed to was freed in the line above and it goes out of scope afterwards, which means it will not be used anymore. The added line is unnecessary.

brauner:
> Since this isn't the first commit of the sort: `c` is in fact a local variable here. The data it pointed to was freed in the line above and it goes out of scope afterwards, which means it will not be used anymore. The added line is unnecessary.

Feel free to send a revert. :)

PR #2768, opened 2019-01-05
Revert "Set c to NULL after freeing it"

caglar10ur:
Reverts lxc/lxc#2763 as it is unnecessary to set c to NULL after free in this case.

See; https://github.com/lxc/lxc/pull/2763#issuecomment-450723571


brauner:
You need to sign-off your commit. :)

brauner:
Thanks. :)

PR #2771, opened 2019-01-07
confile: add lxc.seccomp.allow_nesting

brauner:
This adds the lxc.seccomp.allow_nesting api extension. If
lxc.seccomp.allow_nesting is set to 1 then seccomp profiles will be
stacked. This way nested containers can load their own seccomp policy on
top of the policy that the outer container might have applied.

Cc: Simon Fels <simon.fels@canonical.com>


morphis:
LGTM, thanks for putting this together so quickly!

PR #2774, opened 2019-01-07
trivial fix: unprivileged veth devices (e.g. vethFWABHX) never contain 'Z' char

hn:
unprivileged veth devices (e.g. vethFWABHX) never contain 'Z' character in the randomly generated device name part because for modulo one does not need to substract 1 from strlen().

### DIFF
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -1982,9 +1982,9 @@ char *lxc_mkifname(char *template)
 		for (i = 0; i < strlen(name); i++) {
 			if (name[i] == 'X') {
 #ifdef HAVE_RAND_R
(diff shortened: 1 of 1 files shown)


brauner:
Hey, thanks for the patch. :) Can you please sign of your commit via `git commit -s`? :)

Blub:
Looks like you pushed a merge. Try force-pushing over the original PR branch instead.

hn:
wuahhh .. reading lots of git documentation for just 2 lines of changed code ... hopefully everything ist fine now

brauner:
Thank you for taking the time!

Blub:
There's a learning curve, but once you get the hang of it you can't live without it ;-)

PR #2776, opened 2019-01-08
wishlist: prefix veth interface name with uid info

hn:
For unprivileged containers, LXC auto-generates a random veth device name (the lxc.network.veth.pair option has no effect). With this patch, the auto-generated name will be prefixed with the reverse-ordered base36-encoded UID of the calling user. This makes it possible to narrow down your firewall rules.

Example: For LXC containers started by uid 713 the veth device name will always start with 'vethTJ' (decimal 713 = 'JT' base36). Now one can insert an iptables rule like 'iptables -i vethTJ+ ...'.

For uids up to 1679615 this will decrease the randomness of the device name to 1295, so make sure to limit the total number of veth bridges per unprivileged user (/etc/lxc/lxc-usernet) to a number well below this (side note: anyone using an LXC setup with 1000+ veth bridges per user?)

(actually, I do not expect this patch to be merged, maybe this is helpful as a starting point for further discussion)

Signed-off-by: Hajo Noerenberg <hajo-github@noerenberg.de>

### DIFF
--- a/src/lxc/cmd/lxc_user_nic.c
+++ b/src/lxc/cmd/lxc_user_nic.c
@@ -711,6 +711,7 @@ static char *get_nic_if_avail(int fd, struct alloted_s *names, int pid,
 	struct stat sb;
 	struct alloted_s *n;
 	char *buf = NULL;
+	uid_t uid;
 
 	for (n = names; n != NULL; n = n->next)
 		cull_entries(fd, n->name, intype, br, NULL, NULL);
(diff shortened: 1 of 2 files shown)


stgraber:
Hmm, so how about trying to cover POSIX?

Say, if geteuid() != 0 && geteuid() <= 65536, then encode in base36 and use vethUID_RAND, so for uid 65536 you'd get something like: veth1EKG_JH0272

Which still fits inside an interface name max length (15 chars).

hn:
@stgraber If you go for POSIX integer uids only, I would vote for having clear text (not base36) uids, e.g. veth1234_JH0272, veth65535_JH027. This would cost one byte of randomness for uids>9999, but increase readability for humans a lot.

stgraber:
Yeah, that'd be fine with me

hn:
Sorry guys ... git is doing things I don't understand ... I'm giving up :)

brauner:
Closing in favor of https://github.com/lxc/lxc/pull/2781.

PR #2786, opened 2019-01-16
seccomp: add rules for specified architecture only 

lifeng68:
If the architecture is specified in the seccomp configuration, like:
```
2
whitelist errno 1
[x86_64]
accept allow
accept4 allow
```
We shoud add rules only for amd64 instead of add rules for
x32/i386/amd64.

1. If the [arch] was not specified in seccomp config, add seccomp rules for all compatible architectures.
2. If the [arch] specified in seccomp config irrelevant to native host arch, the rules will be ignored.
3. If specified [all] in seccomp config, add seccomp rules for all compatible architectures.
4. If specified [arch] as same as native host arch, add seccomp rules for the native host arch.
5. If specified [arch] was not native host arch, but compatible to host arch, add seccomp rules for the specified arch only, NOT add seccomp rules for native arch.

### DIFF
--- a/src/lxc/seccomp.c
+++ b/src/lxc/seccomp.c
@@ -291,7 +291,7 @@ on_error:
 #endif
 
 #if HAVE_DECL_SECCOMP_SYSCALL_RESOLVE_NAME_ARCH
-enum lxc_hostarch_t {
+enum lxc_arch_t {
 	lxc_seccomp_arch_all = 0,
 	lxc_seccomp_arch_native,
 	lxc_seccomp_arch_i386,
@@ -345,8 +345,8 @@ int get_hostarch(void)
 	return lxc_seccomp_arch_unknown;
 }
 
-scmp_filter_ctx get_new_ctx(enum lxc_hostarch_t n_arch,
-			    uint32_t default_policy_action, bool *needs_merge)
+scmp_filter_ctx get_new_ctx(enum lxc_arch_t n_arch,
+			    uint32_t default_policy_action)
 {
 	int ret;
 	uint32_t arch;
@@ -464,10 +464,7 @@ scmp_filter_ctx get_new_ctx(enum lxc_hostarch_t n_arch,
 			return NULL;
 		}
 		TRACE("Removed native arch from main seccomp context");
-
-		*needs_merge = true;
 	} else {
-		*needs_merge = false;
 		TRACE("Arch %d already present in main seccomp context", (int)n_arch);
 	}
 
@@ -550,6 +547,27 @@ bool do_resolve_add_rule(uint32_t arch, char *line, scmp_filter_ctx ctx,
 	return true;
 }
 
+#define SCMP_ARCH_INDEX_MAX 3
+
+struct scmp_ctx_info {
+	uint32_t architectures[SCMP_ARCH_INDEX_MAX];
+	enum lxc_arch_t lxc_arch[SCMP_ARCH_INDEX_MAX];
+	scmp_filter_ctx contexts[SCMP_ARCH_INDEX_MAX];
+	bool needs_merge[SCMP_ARCH_INDEX_MAX];
+};
+
+static int get_arch_index(enum lxc_arch_t arch, struct scmp_ctx_info *ctx)
+{
+	int i;
+
+	for (i = 0; i < SCMP_ARCH_INDEX_MAX; i++) {
+		if (ctx->lxc_arch[i] == arch)
+			return i;
+	}
+
+	return -1;
+}
+
 /*
  * v2 consists of
  * [x86]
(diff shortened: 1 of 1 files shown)


brauner:
So the new `[<arch>]` field is optional, i.e. when it is not specified then all compatibility architectures will load the rule?

lifeng68:
Yes, if [<arch>] field empty, cur_rule_arch will be lxc_seccomp_arch_all, all compatibility architectures will load the rule.

brauner:
//cc @stgraber, @Blub

brauner:
jenkins: test this please

brauner:
This looks like a useful feature to me. Is there any chance that you can add a test for this to verify that this still works:
- when no specific arch is passed
- when a specific arch is passed
?

Blub:
Side note: I think the parser needs some refactoring & restructuring.
For example, we still have a `do_resolve_add_rule(SCMP_ARCH_NATIVE,...)` call above this. While "unknown" architectures (eg. an `[arm]` section on an `x86_64` host) will skip this section entirely, a "known" architecture (eg. an `[x32]` section on `x86_64` hosts) still always also adds a "native" rule. I'm not sure if this is intentional? (Not sure if there are any such arch combinations where you'd even need to distinctly control syscalls for different personalities, but still, it seems wrong).

In other words, with this patch the behavior the code tells me is this:
If the current section is a known architecture, add a "native" rule.
If the current section is `[all]`, add a rule for all compat architectures as well.

src/lxc/seccomp.c
```diff
@@ -935,87 +935,89 @@ static int parse_config_v2(FILE *f, char *line, size_t *line_bufsz, struct lxc_c
...
-						ctx.contexts[2], &rule))
-				goto bad_rule;
-
-			INFO("Added native rule for arch %d for %s action %d(%s)",
-			     ctx.architectures[2], line, rule.action,
-			     get_action_name(rule.action));
-		}
 	}
+	if (cur_rule_arch == lxc_seccomp_arch_all) {
+		INFO("Merging compat seccomp contexts into main context");
```
Blub: This section AFAIK should not depend on `cur_rule_arch`. That variable changes the behavior of the parser, whil this section AFAICT is there to - after finishing parsin - combine all the existing architecture contexts into the one we later pass to `seccomp_load()`?

brauner:
If @Blub's concerns are addresses I'm happy to merge. :)

Blub:
> When [x86], why we check whether the native arch is lxc_seccomp_arch_amd64. Was it intentional？

`lxc_seccomp_arch_unknown` is supposed to mean that we're in an environment where this section does not matter because we cannot be running binaries of that architecture. On an `amd64` host we can be running 32-bit binaries, so we need to create rules for `x86` as well as `amd64`.

That said, I do wonder why we don't check for `i386` in the `[x86_64]` case - shouldn't `get_hostarch()` see `"i686"` on 32-bit containers due to `personality(2)` being called with `PER_LINUX32`?

lifeng68:
I updated the code, please discuss the following strategy whether correct：
1. If the [arch] was not specified in seccomp config, add seccomp rules for all all compat architectures.
2. If the [arch] specified in seccomp config irrelevant to native host arch, the rules will be ignored.
3. If specified [all] in seccomp config, add seccomp rules for all compat architectures.
4. If specified [arch] as same as native host arch, add seccomp rules for the native host arch.
5. If specified [arch] was not native host arch, but compat to host arch, add seccomp rules for the specified arch only, NOT add seccomp rules for native arch.

Blub:
> * If the [arch] was not specified in seccomp config, add seccomp rules for all  compatible architectures.
> * If the [arch] specified in seccomp config irrelevant to native host arch, the rules will be ignored.
> * If specified [all] in seccomp config, add seccomp rules for all compatible architectures.
> * If specified [arch] as same as native host arch, add seccomp rules for the native host arch.
> * If specified [arch] was not native host arch, but compatible to host arch, add seccomp rules for the specified arch only, NOT add seccomp rules for native arch.

This sounds correct to me. I haven't had time to read the updated patch yet, though.

brauner:
If @Blub's happy with this I am too.
@stgraber, care to take a quick look at what @lifeng68 lined out?

src/lxc/seccomp.c
```diff
@@ -550,6 +547,25 @@ bool do_resolve_add_rule(uint32_t arch, char *line, scmp_filter_ctx ctx,
...
+};
+
+static int get_arch_index(enum lxc_arch_t arch, struct scmp_ctx_info *ctx)
+{
+	int i;
+
+	for (i = 0; i < 3; i++) {
+		if (ctx->lxc_arch[i] == arch)
+			return i;
+	}
```
brauner: Hm, this could probably be a define:
```c
#define SCMP_ARCH_INDEX_MAX 3
```
and then
```c
for (i = 0; i < SCMP_ARCH_INDEX_MAX; i++)
        if (ctx->lxc_arch[i] == arch)
                return i;
```

brauner:
jenkins: test this please

src/lxc/seccomp.c
lifeng68: Fixed

brauner:
@lifeng68, can you please send a patch to update the seccomp documentation in `lxc.container.conf.sgml.in` to describe the new format and rules?

brauner:
This branch breaks all seccomp confined workloads. Even the most basic ones indicating that no testing has been done. We need to revert this until this has been sorted out and proper tests added.
//cc @lifeng68 https://github.com/lxc/lxc/pull/2794

lifeng68:
```
If the architecture is specified in the seccomp configuration, like:

2
whitelist errno 1
[x86_64]
accept allow
accept4 allow
We shoud add rules only for amd64 instead of add rules for
x32/i386/amd64.
```
The "seccomp confined workloads" is your internal test case ? This patch changes the behavior of the seccomp parser. Your original configuration file may need to be updated.

brauner:
That's a full-on user breakage. We absolutely can't do that. If the parser needs changing in how it interprets the config file then the obvious thing to do is to require to bump the config file version.
So if it's a seccomp config version 3 you get the new behavior.

Issue #2602, opened 2018-09-11
LXC mounts /sys twice when running privileged container


tommyknows:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 16.04
 * The output of
   * `lxc-start --version` seen in 3.3 and 2.21
   * `uname -a` 4.4.0-134

# Issue description
When starting a privileged container, `/sys` gets mounted twice. Once read-write, and once read-only.
```
root@test:~# mount | grep sysfs
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
sysfs on /sys type sysfs (ro,nosuid,nodev,noexec,relatime)
```

# Steps to reproduce

 1. Create a profile with `security.privileged`:
```
root@-lxc-1:~# lxc profile show test
config:
  security.privileged: "true"
description: Test Profile
devices:
  root:
    path: /
    pool: default
    type: disk
name: test
used_by:
- /1.0/containers/test
```

 2. Launch a container with that profile.
```
root@new-lxc-1:~# lxc launch images:ubuntu/16.04 test -p test
Creating test

The container you are starting doesn't have any network attached to it.
  To create a new network, use: lxc network create
  To attach a network to a container, use: lxc network attach

Starting test
```

 3. Exec into the container and check the mounts:
```
root@lxc-1:~# lxc exec test /bin/bash
root@test:~# mount | grep sysfs
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
sysfs on /sys type sysfs (ro,nosuid,nodev,noexec,relatime)
sysfs on /sys/devices/virtual/net type sysfs (rw,relatime)
sysfs on /sys/devices/virtual/net type sysfs (rw,nosuid,nodev,noexec,relatime)
```

Launching with Debug options:
[lxc.log](https://github.com/lxc/lxc/files/2369975/lxc.log)

PR #2787, opened 2019-01-17
Revert "conf: remove extra MS_BIND with sysfs:mixed"

Blub:
The above commit reverted here confuses the mountall unit of privileged
Ubuntu 14.04 containers at startup so that they cannot finish booting.

This effectively reverts the commit fixing #2602.
@tommyknows, was there any actual breakage causing you to open #2602 or was it just about mount cleanliness? Because apparently removing that mount confuses 14.04 containers, and while that's certainly a bug on the container's side reverting this is the simplest solution if there's no harm in it. Otherwise if there was any breakage from the mount entry I'll instead make a patch to make this configurable.


brauner:
I think we need something that allows us to configure this?

stgraber:
If you want /sys to be rw instead of partly ro, you'd want to use sys:rw in config instead of sys:mixed

brauner:
@stgraber, if you're happy with this, please merge it. :)

tommyknows:
What is the motivation behind mounting `/sys` twice? 
I'm not able to test this, but what happens when I specify `sys:rw` _and_ privileged container?
Does `/sys` get mounted three times? (rw, ro and rw again?)

PR #2789, opened 2019-01-17
Fix memory leak in cgroup_exit

lifeng68:
Add free memory pointed by struct cgroup_ops *ops


brauner:
Thanks!

PR #2802, opened 2019-01-26
Avoid double lxc-freeze/unfreeze

Rachid-Koucha:
If we call lxc-freeze multiple times for an already frozen container, LXC
triggers useless freezing by writing into the "freezer.state" cgroup file.
This is the same when we call lxc-unfreeze multiple times.
Checking the current state with a LXC_CMD_GET_STATE
(calling c->state) would permit to check if the container is FROZEN
or not.


brauner:
jenkins: test this please

PR #2803, opened 2019-01-26
Avoid risk of "too far memory read"

Rachid-Koucha:
As we call "lxc_add_state_client(fd, handler, (lxc_state_t *)req->data)"
which supposes that the last parameter is a table of MAX_STATE
entries when calling memcpy():
memcpy(newclient->states, states, sizeof(newclient->states))


brauner:
Thanks!

PR #2805, opened 2019-01-27
Raise STOPPING state during lxc-stop

Rachid-Koucha:
Before killing the processes in the container, raise the STOPPING state
as the action may take a long time.


brauner:
I don't think that's a great idea. Mostly because the shutdown signal doesn't really need to shutdown the container but also because we do set `STOPPING` already:
https://github.com/lxc/lxc/blob/3309e10c31b04499ceec22353162b09cbf5aa109/src/lxc/start.c#L943

PR #2806, opened 2019-01-27
freezer: non-functional changes

brauner:
Fix the coding style in a few files.

Fixes: db1228b35f3e ("Avoid hardcoded string length")
Fixes: 71fc9c046816 ("Avoid risk of "too far memory read"")
Fixes: 2341916a0367 ("Avoid double lxc-freeze/unfreeze")
Fixes: 9eb9ce3e4778 ("Update freezer.c")

### DIFF
--- a/src/lxc/freezer.c
+++ b/src/lxc/freezer.c
@@ -51,11 +51,11 @@ static int do_freeze_thaw(bool freeze, struct lxc_conf *conf, const char *name,
 	int ret;
 	char v[100];
 	struct cgroup_ops *cgroup_ops;
-        const char *state;
+	const char *state;
 	size_t state_len;
 	lxc_state_t new_state = freeze ? FROZEN : THAWED;
 
-        state = lxc_state2str(new_state);
+	state = lxc_state2str(new_state);
 	state_len = strlen(state);
 
 	cgroup_ops = cgroup_init(conf);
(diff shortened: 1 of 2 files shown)


src/lxc/freezer.c
```diff
@@ -65,19 +65,21 @@ static int do_freeze_thaw(bool freeze, struct lxc_conf *conf, const char *name,
 	ret = cgroup_ops->set(cgroup_ops, "freezer.state", state, name, lxcpath);
 	if (ret < 0) {
 		cgroup_exit(cgroup_ops);
-		ERROR("Failed to %s %s", (new_state == FROZEN ? "freeze" : "unfreeze"), name);
+		ERROR("Failed to %s %s",
+		      (new_state == FROZEN ? "freeze" : "unfreeze"), name);
```
hallyn: TBH if you wanted to actually improve this, it would be better to use a temporary variable with a meaningful name, assigned in the previous line.

src/lxc/lxccontainer.c
```diff
@@ -547,15 +544,12 @@ static bool do_lxcapi_unfreeze(struct lxc_container *c)
...
 		return false;
 
 	s = lxc_getstate(c->name, c->config_path);
-	if (s == FROZEN) {
-	  ret = lxc_unfreeze(c->lxc_conf, c->name, c->config_path);
-	  if (ret < 0)
-	    return false;
-	}
+	if (s == FROZEN)
+		return !lxc_unfreeze(c->lxc_conf, c->name, c->config_path);
```
hallyn: What do you think about using 'return lxc_unfreeze() == 0" here instead?  It's one less conversion in the reader's head.

src/lxc/freezer.c
```diff
@@ -51,11 +51,11 @@ static int do_freeze_thaw(bool freeze, struct lxc_conf *conf, const char *name,
 	int ret;
 	char v[100];
 	struct cgroup_ops *cgroup_ops;
-        const char *state;
+	const char *state;
 	size_t state_len;
 	lxc_state_t new_state = freeze ? FROZEN : THAWED;
 
-        state = lxc_state2str(new_state);
+	state = lxc_state2str(new_state);
```
hallyn: jenkins should really have caught these :)

src/lxc/lxccontainer.c
brauner: yep, will do :)

PR #2809, opened 2019-01-27
Installation of default.script for udhcpc

Rachid-Koucha:
The busybox template installs default.script in /usr/share/udhcpc/.
But the pathname of "default.script" may vary from one busybox
build to another. As the pathname is displayed in udhcpc's help,
grab it from it.


brauner:
Thanks!

PR #2816, opened 2019-01-29
More accurate error msg for template file

Rachid-Koucha:
When calling lxc-create, if the template exists but is not executable, we end with the following error messages which make believe that the template file does not exist when it is merely a execute access problem:

lxc-create: ctn00: utils.c: get_template_path: 918 No such file or directory - bad template: /.../lxc-busybox
lxc-create: ctn00: lxccontainer.c: do_lxcapi_create: 1786 Unknown template "/.../lxc-busybox"
lxc-create: ctn00: tools/lxc_create.c: main: 327 Failed to create container ctn00

Moreover, internally this triggers a useless access to (strace output):

access("/.../lxc-busybox", X_OK) = -1 ENOENT (No such file or directory)

With the above fix, we get a more explicit error message when the template file is missing the "execute" bit:

lxc-create: ctn: utils.c: get_template_path: 926 Permission denied - cannot execute template /h.../lxc-busybox
lxc-create: ctn: lxccontainer.c: do_lxcapi_create: 1816 Unknown template "/.../lxc-busybox"
lxc-create: ctn: tools/lxc_create.c: main: 331 Failed to create container ctn

With the above fix, we get a more explicit error message when the pathname of the template file is incorrect:

lxc-create: ctn: utils.c: get_template_path: 931 No such file or directory - template /.../lxc-busybox does not exist
lxc-create: ctn: lxccontainer.c: do_lxcapi_create: 1816 Unknown template "/.../lxc-busybox"
lxc-create: ctn: tools/lxc_create.c: main: 331 Failed to create container ctn

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -910,9 +910,24 @@ char *get_template_path(const char *t)
 	int ret, len;
 	char *tpath;
 
-	if (t[0] == '/' && access(t, X_OK) == 0) {
-		tpath = strdup(t);
-		return tpath;
+	if (t[0] == '/') {
+        	if (access(t, X_OK) == 0) {
+			tpath = strdup(t);
+			return tpath;
+		} else {
+			if (errno == EACCES) {
+				SYSERROR("cannot execute template %s", t);
+				return NULL;
+			}
+
+			if (errno == ENOENT) {
+				SYSERROR("template %s does not exist", t);
+				return NULL;
+			}
+
+			SYSERROR("Bad template pathname: %s", t);
+			return NULL;
+		}
 	}
 
 	len = strlen(LXCTEMPLATEDIR) + strlen(t) + strlen("/lxc-") + 1;
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -912,19 +912,8 @@ char *get_template_path(const char *t)
 
 	if (t[0] == '/') {
         	if (access(t, X_OK) == 0) {
-			tpath = strdup(t);
-			return tpath;
+			return strdup(t);
 		} else {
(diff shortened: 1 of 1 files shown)


src/lxc/utils.c
```diff
@@ -910,9 +910,24 @@ char *get_template_path(const char *t)
 	int ret, len;
 	char *tpath;
 
-	if (t[0] == '/' && access(t, X_OK) == 0) {
-		tpath = strdup(t);
-		return tpath;
+	if (t[0] == '/') {
+        	if (access(t, X_OK) == 0) {
+			tpath = strdup(t);
+			return tpath;
```
brauner: Just do `return strdup(tpath)` :)

src/lxc/utils.c
```diff
@@ -910,9 +910,24 @@ char *get_template_path(const char *t)
...
+
+			if (errno == ENOENT) {
+				SYSERROR("template %s does not exist", t);
+				return NULL;
+			}
+
+			SYSERROR("Bad template pathname: %s", t);
+			return NULL;
+		}
 	}
```
brauner: Why doesn't a single `SYSERROR` suffice? The errno string will be printed and hence does give you the root cause. So I'd say just do:
```
SYSERROR("Failed to execute template");
```

src/lxc/utils.c
Rachid-Koucha: As I mentionned in the commit comment, the original code does not return iommediately upon this error and tries to open ".../lxc-/pathname_passed". So, the original errno is lost.
But you are right, it is not necessary to dicriminate the errnos in the "else" case.
Concerning "tpath =", I kept the original :-)

PR #2827, opened 2019-02-07
cgroups: partially switch to cleanup macros

brauner:

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -232,10 +232,11 @@ static void append_line(char **dest, size_t oldlen, char *new, size_t newlen)
 /* Slurp in a whole file */
 static char *read_file(const char *fnam)
 {
-	FILE *f;
-	char *line = NULL, *buf = NULL;
-	size_t len = 0, fulllen = 0;
+	__do_free char *line = NULL;
+	__do_fclose FILE *f = NULL;
 	int linelen;
+	char *buf = NULL;
+	size_t len = 0, fulllen = 0;
 
 	f = fopen(fnam, "r");
 	if (!f)
@@ -244,8 +245,6 @@ static char *read_file(const char *fnam)
 		append_line(&buf, fulllen, line, linelen);
 		fulllen += linelen;
 	}
-	fclose(f);
-	free(line);
 	return buf;
 }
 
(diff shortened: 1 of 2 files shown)


src/lxc/cgroups/cgfsng.c
```diff
@@ -232,10 +232,11 @@ static void append_line(char **dest, size_t oldlen, char *new, size_t newlen)
 /* Slurp in a whole file */
 static char *read_file(const char *fnam)
 {
-	FILE *f;
-	char *line = NULL, *buf = NULL;
-	size_t len = 0, fulllen = 0;
+	__do_fclose FILE *f;
```
hallyn: I see that you don't need it here right now, but in general I think adding = NULL; here would be good, in case someone ends up adding an if (solstice) return NULL;  before the f = fopen().

hallyn:
Loving this direction.

hallyn:
Can you point to a doc that clarifies the syntax?  I've usually seen __attribute__  used after declarations, and have not seen anything (including at https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes) that clarifies that you can use it for >1 variables, i.e. __do_free int *a = NULL, *b = NULL; ?

src/lxc/cgroups/cgfsng.c
```diff
@@ -411,7 +412,6 @@ static bool cg_legacy_filter_and_set_cpus(char *path, bool am_initialized)
 	if (!file_exists(__ISOL_CPUS)) {
 		/* This system doesn't expose isolated cpus. */
 		DEBUG("The path \""__ISOL_CPUS"\" to read isolated cpus from does not exist");
-		cpulist = posscpus;
```
Blub: There's a `goto copy_parent;` a few lines below which copies from `cpulist`, so this is a behavior change as well, no?
So unless this is intentional, to deal with the `cpulist != posscpus` comparison the cleanup had before it would probably be the most clear if we add a non-autofreed `char *cpulist_to_copy;` set here, below, and finally before the `copy_parent` label?

src/lxc/cgroups/cgfsng.c
```diff
@@ -503,26 +501,17 @@ static bool cg_legacy_filter_and_set_cpus(char *path, bool am_initialized)
 	bret = true;
 
 on_error:
```
Blub: I think all users of this label could just `return false` at this point, similar for the `on_success` label.

src/lxc/cgroups/cgfsng.c
```diff
@@ -540,21 +529,16 @@ static bool copy_parent_file(char *path, char *file)
 	ret = lxc_read_from_file(fpath, value, len);
 	if (ret != len)
 		goto on_error;
-	free(fpath);
```
Blub: I think this one's still necessary actually as `fpath` is overwritten below.

src/lxc/cgroups/cgfsng.c
brauner: Agreed.

src/lxc/cgroups/cgfsng.c
brauner: Agreed.

src/lxc/cgroups/cgfsng.c
brauner: Good catch!

brauner: I think we can do:
```c
	if (!isdigit(isolcpus[0])) {
		TRACE("No isolated cpus detected");
		/* No isolated cpus but we weren't already initialized by
		 * someone. We should simply copy the parents cpuset.cpus
		 * values.
		 */
		if (!am_initialized) {
			DEBUG("Copying cpu settings of parent cgroup");
			cpulist = posscpus;
			goto copy_parent;
		}
		/* No isolated cpus but we were already initialized by someone.
		 * Nothing more to do for us.
		 */
		return true;
	}
.
.
.	ret = lxc_write_to_file(fpath, cpulist, strlen(cpulist), false, 0666);
	if (cpulist == posscpus)
		cpulist = NULL;
	if (ret < 0) {
		SYSERROR("Failed to write cpu list to \"%s\"", fpath);
		return false;
	}
```

PR #2829, opened 2019-02-08
caps: check uid and euid

brauner:
When we are running inside of a user namespace getuid() will return a
non-zero uid. So let's check euid as well to make sure we correctly drop
capabilities


stgraber:
Any reason why we're not only checking euid?

brauner:
We don't want to play with capabilities when we are an unprivileged uid on the host. :)

PR #2832, opened 2019-02-08
coccinelle support

brauner:


brauner:
Based on systemd's version.

PR #2841, opened 2019-02-11
include: add fexecve() for Android's Bionic

brauner:


stgraber:
Hmm, where did that copyright header come from?

This is adding GPLv2 code to the liblxc library which is released as LGPLv2+.

If it's code you wrote, then you need to make sure it's licensed as LGPLv2+ to be included in liblxc, if it's external code, then this isn't suitable for inclusion in LXC. Code that goes in liblxc itself must be LGPLV2+ or less strict.

Our binaries are GPLv2, that's fine as they consume liblxc and don't provide code to it.

brauner:
I wrote that function myself.

PR #2862, opened 2019-02-18
cleanup macros and rexec fixes

brauner:

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -698,8 +698,8 @@ static int lxc_cmd_terminal_winch_callback(int fd, struct lxc_cmd_req *req,
  */
 int lxc_cmd_console(const char *name, int *ttynum, int *fd, const char *lxcpath)
 {
+	__do_free struct lxc_cmd_console_rsp_data *rspdata = NULL;
 	int ret, stopped;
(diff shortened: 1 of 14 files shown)


brauner:
jenkins: test this please

brauner:
jenkins: test this please

src/lxc/commands.c
```diff
@@ -711,29 +711,25 @@ int lxc_cmd_console(const char *name, int *ttynum, int *fd, const char *lxcpath)
 	if (cmd.rsp.ret < 0) {
 		errno = -cmd.rsp.ret;
 		SYSERROR("Denied access to tty");
-		ret = -1;
-		goto out;
+		return -1;
```
Blub: Previously this case freed `cmd.rsp.data`. The `rspdata` assignment now happens only below though. I do think there are cases where this can make a difference, though I haven't checked all the code paths. At least `lxc_cmd_rsp_recv` allocates it only based on the returned `datalen` independent of `rsp.ret` except for the `LXC_CMD_CONSOLE` case.

brauner: We can just move the `rspdata` assignment up. I did this now. Will force-push now.

PR #2870, opened 2019-02-20
autotools: handle getgrgid_r on bionic

brauner:


stgraber:
```
configure: error: conditional "HAVE_GETGRGID_R" was never defined.
```

So this is making it fail earlier but isn't fixing the issue.

brauner:
Patience, I'll tell you when it's ready.

PR #2871, opened 2019-02-20
[RFC, untested] reject lxc.mount.entries which fall outside the container rootfs

hallyn:
Closes 2766

Add a lxc.container.conf knob to allow them.

### DIFF
--- a/doc/lxc.container.conf.sgml.in
+++ b/doc/lxc.container.conf.sgml.in
@@ -1096,6 +1096,21 @@ dev/null proc/kcore none bind,relative 0 0
 
         <varlistentry>
           <term>
+            <option>lxc.allow_abs_mountentries</option>
+          </term>
+          <listitem>
+            <para>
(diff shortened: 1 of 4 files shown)


stgraber:
@hallyn everything looks sad, including DCO check :)

hallyn:
Yeah I only pushed so far so I could fetch on a different host to test :)

hallyn:
Hah!

We do not need this.

We already ignore all such mounts, i.e.:

```
shallyn@stp:~$ sudo grep tmp /var/lib/lxc/c1/config
lxc.mount.entry = /tmp /tmptmp none bind,create=dir 0 0
shallyn@stp:~$ sudo grep Ignor stdout
lxc-start c1 20190220223457.199 WARN     conf - conf.c:mount_entry_on_absolute_rootfs:2282 - Ignoring mount point "/tmptmp"
```

PR #2882, opened 2019-02-27
gpg: use proxy, if http_proxy is set

misterunknown:
Hi everyone,

I found that `gpg` ignores the `http_proxy` environment variable. As the lxc-download script already adjusts the keyserver when `http_proxy` is set, it also should set the necessary `gpg` option. Currently there is no way to use lxc-download behind a proxy, because of this problem.

I had to put the `gpg` command in a subshell, because if not, `gpg` would not work properly, as it does not treat `${DOWNLOAD_GPG_PROXY}` as a valid argument.

If there is a better solution, let me know about it. This is my first pull request for this project, I hope I got everything right.

Cheers,
Marco

### DIFF
--- a/templates/lxc-download.in
+++ b/templates/lxc-download.in
@@ -60,6 +60,7 @@ if [ -z "${DOWNLOAD_KEYSERVER:-}" ]; then
   # Deal with GPG over http proxy
   if [ -n "${http_proxy:-}" ]; then
     DOWNLOAD_KEYSERVER="hkp://p80.pool.sks-keyservers.net:80"
+    DOWNLOAD_GPG_PROXY="--keyserver-options http-proxy=\"${http_proxy}\""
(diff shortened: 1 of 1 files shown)


brauner:
Hey @misterunknown thanks for patch. Can you please sign-off your commits. Should be as easy as `git commit -s`. :)

misterunknown:
@brauner Thanks for the hint – is it somehow possible to sign-off the commits in the chain, or should I better create a new pull-request?

brauner:
You could do:
```
git rebase --signoff HEAD~<nr-of-commits>
```

misterunknown:
Hm, I somehow messed up my repository, but figured out how to fix it, so now everything should be ok. It's just one single commit and is signed off. Thanks for your patience.

PR #2884, opened 2019-03-01
tree-wide: kill fgets()

brauner:

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1364,8 +1364,6 @@ int lxc_chroot(const struct lxc_rootfs *rootfs)
 {
 	__do_free char *nroot = NULL;
 	int i, ret;
-	char *p, *p2;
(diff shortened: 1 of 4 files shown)


hallyn:
So... what's the rationale for this one?  CODING_STYLE.md doesn't explain why.  Doesn't seem important for run_buffer, for instance.  Is it just a "why not", or is there a security issue I'm not aware of?

brauner:
There are a bunch of places where it is fine like with `run_buffer()` where you don't care about truncation or where you deal with reading partial lines because of truncation in a while loop. But these things can be more easy taking care of with `getline()` which always reads until `\n` or `EOF`. So it's just easier to get right. But I'm open to not enforce this and/or reword to "prefer other methods" instead of saying "never". :)

hallyn:
Ok, I'll take a close look through the day.

PR #2888, opened 2019-03-01
Fix android compilation

kubiko:
Fixing build error when building for android target
Signed-off-by: Ondrej Kubik <ondrej.kubik@canonical.com>


brauner:
Thanks. :)

PR #2889, opened 2019-03-02
gcc -Wall flag compat

tomponline:
Builds on PR from @brauner https://github.com/lxc/lxc/pull/2886

I'm still working through these (the build still fails), but thought I would open in draft form early.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -669,7 +669,7 @@ static bool no_new_privs(struct lxc_container *c, lxc_attach_options_t *options)
 
 static signed long get_personality(const char *name, const char *lxcpath)
 {
-	__do_free char *p;
+	__do_free char *p = NULL;
 
 	p = lxc_cmd_get_config_item(name, "lxc.arch", lxcpath);
 	if (!p)
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -159,7 +159,6 @@ static int in_same_namespace(pid_t pid1, pid_t pid2, const char *ns)
 {
 	__do_close_prot_errno int ns_fd1 = -1, ns_fd2 = -1;
 	int ret = -1;
-	int saved_errno;
 	struct stat ns_st1, ns_st2;
 
 	ns_fd1 = lxc_preserve_ns(pid1, ns);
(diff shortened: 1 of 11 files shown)


tomponline:
@brauner @stgraber that should be all of them now, is now building with `-Wall` flag.

Is it possible to add this flag to the unit test compile?

brauner:
jenkins: test this please

tomponline:
thanks @brauner 

Would there be any problem with adding an `export CFLAGS=-Wall` to https://github.com/lxc/lxc/blob/master/.travis.yml#L15

brauner:
Should be already present by default:
```
https://github.com/lxc/lxc/blob/6400238d08cdf1ca20d49bafb85f4e224348bf9d/configure.ac#L43
```

tomponline:
hrm interesting, I wonder why travis didn't fail the build, and it only failed on centos 7.

I wonder if it is to do with the interplay between `-Wall` and the other CFLAGS centos adds:

```
CFLAGS='-O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches   -m64 -mtune=generic'
```

In my tests the other day, removing `-Wall` allowed it to compile with the other flags present. But perhaps using `-Wall` without the other flags also allows it to build.

I'll do a test an let you know.

tomponline:
@brauner found it, AM_CFLAGS is defined in:

https://github.com/lxc/lxc/blob/1edf8fa48d82a1fe51d231e1d83e1c4bfea7e286/src/tests/Makefile.am#L44-L54

and

https://github.com/lxc/lxc/blob/1edf8fa48d82a1fe51d231e1d83e1c4bfea7e286/src/tests/Makefile.am#L44-L54

If I add the following arguments `-Wall -Werror -Wno-portability -Wall` to those lines, then the build fails when building on the commit https://github.com/lxc/lxc/commit/e613d684fa1757abe72d14ca5e39a08de3364285

So it doesn't look like the options in `AM_INIT_AUTOMAKE` are actually being used.

PR #2890, opened 2019-03-03
fix lxcbasename in lxc/lxccontainer.c

knknkn1162:
After the change, `lxcbasename()` would avoid slash character at the head of the string.
The return value of the `lxcbasename()` is put in the `newargv` variable, and this variable is used in `execvp`.

https://github.com/lxc/lxc/blob/9fc7a50f8a7f1066f658bcb5bab5f3d724318827/src/lxc/lxccontainer.c#L1448

https://github.com/lxc/lxc/blob/9fc7a50f8a7f1066f658bcb5bab5f3d724318827/src/lxc/lxccontainer.c#L1629.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -1322,7 +1322,7 @@ static char *lxcbasename(char *path)
 	while (*p != '/' && p > path)
 		p--;
 
-	return p;
+	return (p == path) ? p : ++p;
 }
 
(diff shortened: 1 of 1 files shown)


brauner:
Thanks for the patch. Can you please sign-off your commit? Should be as easy as `git commit -s`. :)

knknkn1162:
> Can you please sign-off your commit?
OK. So close this PR and push another PR again.

PR #2897, opened 2019-03-06
Fix lxc.cgroup2.<controller> on cgroup2-only systems

TheSeven:
In the pure-unified cgroup case (`cg_unified_init` instead of `cg_hybrid_init`), `ops->unified` was left uninitialized, making any attempts to set cgroup controllers using `lxc.cgroup2.<controller>` fail on systems with only cgroup2 mounted.


brauner:
jenkins: test this please

Issue #2902, opened 2019-03-11
lxc-start doesn't work with --share-pid


flx42:
Hi @brauner! Using latest [`master`](bbebc423bf267a084faec73ee5df72d32ea86751) I see the following:
```
# lxc-create -t download -n bionic -- --dist ubuntu --release bionic --arch amd64
# lxc-start --logfile=/dev/stdout --logpriority=DEBUG --foreground --name=bionic --define 'lxc.net=' --share-pid=${PID} -- bash
[...]
lxc-start bionic 20190311230011.767 ERROR    sync - sync.c:__sync_wait:48 - Connection reset by peer - Sync wait failure
```

This message is coming from the parent, in `lxc_sync_wait_child`.  I believe this is because the expected parent PID is hardcoded to `0` here:
https://github.com/lxc/lxc/blob/bbebc423bf267a084faec73ee5df72d32ea86751/src/lxc/start.c#L1115

So, the child will send itself a SIGKILL here:
https://github.com/lxc/lxc/blob/bbebc423bf267a084faec73ee5df72d32ea86751/src/lxc/utils.c#L1700-L1714

I guess the fix is as simple as:
```patch
-       ret = lxc_set_death_signal(SIGKILL, 0);
+       if (handler->ns_clone_flags & CLONE_NEWPID)
+               ret = lxc_set_death_signal(SIGKILL, 0);
+       else
+               ret = lxc_set_death_signal(SIGKILL, handler->monitor_pid);
```

brauner:
Yep, that's correct. Can you send a patch, please?

flx42:
Wait, is this path valid if doing `--share-pid=othercontainer` where `othercontainer` unshared the PID namespace?

flx42:
Looks like it's not from a quick test, so it's not that easy :)

brauner:
Well, you could check whether the getppid returns 0 and if so you know that you're in a pid namespace. It can't return 0 in all other cases.

brauner:
Actually that check doesn't make a lot of sense rn since getppid() will always return 0 even if the parent died. So this needs to be:
```c
if (ppid && ppid != parent)
```

PR #2903, opened 2019-03-12
Fix PID namespace pdeathsig handling

flx42:
@brauner given what you said in #2902, I believe we can unconditionally pass `monitor_pid` to `lxc_set_death_signal` if we fix the parent check in `lxc_set_death_signal`.


stgraber:
jenkins: test this please

brauner:
Thanks! :)

PR #2904, opened 2019-03-12
lxccontainer: improve shutdown signal detection

brauner:
According to Lennart the official way of detecting whether systemd is runnig is
to check for the existence of /run/systemd/system.


src/lxc/lxccontainer.c
```diff
@@ -2084,7 +2084,9 @@ static bool do_lxcapi_shutdown(struct lxc_container *c, int timeout)
 	/* Detect whether we should send SIGRTMIN + 3 (e.g. systemd). */
 	if (c->lxc_conf && c->lxc_conf->haltsignal)
 		haltsignal = c->lxc_conf->haltsignal;
-	else if (task_blocks_signal(pid, (SIGRTMIN + 3)))
+	else if ((faccessat(AT_FDCWD, "/run/systemd/system/", F_OK,
```
stgraber: Am I missing something or is this looking for /run/systemd/system on the host rather than inside the container?

PR #2894, opened 2019-03-05
Makefile: Adds compile flags -Wall -Werror

tomponline:
Makefile: Adds compile flags -Wall -Werror which were not applying.
These are defined already here https://github.com/lxc/lxc/blob/6400238d08cdf1ca20d49bafb85f4e224348bf9d/configure.ac#L43 but are not applying as CFLAGS are overridden in the Makefile.am files

Closes #2885.

Signed-off-by: tomponline <tomp@tomp.uk>

PR #2910, opened 2019-03-19
Travis: Adds -Wall and -Werror cflags flags to automatic build.

tomponline:
Signed-off-by: tomponline <tomp@tomp.uk>

As per https://github.com/lxc/lxc/pull/2894#issuecomment-469746460


tomponline:
@brauner was this more the approach were thinking about?

PR #2911, opened 2019-03-19
lvm: Updates lvcreate to wipe signatures if supported

tomponline:
lvm: Updates lvcreate to wipe signatures if supported, fallbacks to old command if not.

Signed-off-by: tomponline <tomp@tomp.uk>

Related to https://github.com/lxc/lxd/issues/5482#issuecomment-464810627

I found it quite tricky to test an older version of lvcreate, as LXC won't compile on CentOS 5, which has an older version of lvcreate without the `-Wy` flag.

Instead I created a dummy lvcreate command in `/usr/local/sbin/lvcreate` as follows, which outputs the same error as the old version of lvcreate does when passed the `-Wy` flag.

```
ARGS=$@
logger "lvcreate $ARGS"

# Parse command flags
while [ ! $# -eq 0 ]
do
  	case "$1" in
                -Wy )
echo "lvcreate: invalid option -- W
  Error during parsing of command line."
                        exit 3
                        ;;
        esac
	shift
done

/usr/sbin/lvcreate $ARGS
```

Here is the syslog output from it, showing the detection of the error and falling back to old command:

```
Mar 19 18:10:29 centos7 root: lvcreate -Wy --yes -L 1073741824b VolGroup00 -n test
Mar 19 18:10:29 centos7 root: lvcreate -qq -L 1073741824b VolGroup00 -n test
```

Additionally, because `run_command` doesn't provide the exit code, I instead inspected the error output to detect the unsupported lvcreate command.

### DIFF
--- a/src/lxc/storage/lvm.c
+++ b/src/lxc/storage/lvm.c
@@ -55,6 +55,7 @@ struct lvcreate_args {
 	const char *lv;
 	const char *thinpool;
 	const char *fstype;
+	bool sigwipe;
 
 	/* snapshot specific arguments */
 	const char *source_lv;
@@ -76,12 +77,19 @@ static int lvm_create_exec_wrapper(void *data)
 
 	(void)setenv("LVM_SUPPRESS_FD_WARNINGS", "1", 1);
 	if (args->thinpool)
-		execlp("lvcreate", "lvcreate", "-qq", "--thinpool", args->thinpool,
-		       "-V", args->size, args->vg, "-n", args->lv,
-		       (char *)NULL);
+		if(args->sigwipe)
+			execlp("lvcreate", "lvcreate", "-Wy", "--yes", "--thinpool", args->thinpool,
+			       "-V", args->size, args->vg, "-n", args->lv, (char *)NULL);
+		else
+			execlp("lvcreate", "lvcreate", "-qq", "--thinpool", args->thinpool,
+			       "-V", args->size, args->vg, "-n", args->lv, (char *)NULL);
 	else
-		execlp("lvcreate", "lvcreate", "-qq", "-L", args->size, args->vg, "-n",
-		       args->lv, (char *)NULL);
+		if(args->sigwipe)
+			execlp("lvcreate", "lvcreate", "-Wy", "--yes", "-L", args->size, args->vg, "-n",
+			       args->lv, (char *)NULL);
+		else
+			execlp("lvcreate", "lvcreate", "-qq", "-L", args->size, args->vg, "-n",
+			       args->lv, (char *)NULL);
 
 	return -1;
 }
@@ -177,11 +185,22 @@ static int do_lvm_create(const char *path, uint64_t size, const char *thinpool)
 	cmd_args.vg = vg;
 	cmd_args.lv = lv;
 	cmd_args.size = sz;
+	cmd_args.sigwipe = true;
 	TRACE("Creating new lvm storage volume \"%s\" on volume group \"%s\" "
 	      "of size \"%s\"", lv, vg, sz);
-	ret = run_command(cmd_output, sizeof(cmd_output),
+	ret = run_command_status(cmd_output, sizeof(cmd_output),
 			  lvm_create_exec_wrapper, (void *)&cmd_args);
-	if (ret < 0) {
+
+	/* If lvcreate is old and doesn't support signature wiping, try again without it.
+	 * Test for exit code EINVALID_CMD_LINE(3) of lvcreate command.
+	 */
+	if (WIFEXITED(ret) && WEXITSTATUS(ret) == 3) {
+		cmd_args.sigwipe = false;
+		ret = run_command(cmd_output, sizeof(cmd_output),
+			      lvm_create_exec_wrapper, (void *)&cmd_args);
+	}
+
+	if (ret != 0) {
 		ERROR("Failed to create logical volume \"%s\": %s", lv,
 		      cmd_output);
 		free(pathdup);
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -1594,7 +1594,7 @@ pop_stack:
 	return umounts;
 }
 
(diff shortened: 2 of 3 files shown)


brauner:
jenkins: test this please

src/lxc/storage/lvm.c
```diff
@@ -177,10 +185,18 @@ static int do_lvm_create(const char *path, uint64_t size, const char *thinpool)
 	cmd_args.vg = vg;
 	cmd_args.lv = lv;
 	cmd_args.size = sz;
+	cmd_args.sigwipe = true;
 	TRACE("Creating new lvm storage volume \"%s\" on volume group \"%s\" "
 	      "of size \"%s\"", lv, vg, sz);
 	ret = run_command(cmd_output, sizeof(cmd_output),
 			  lvm_create_exec_wrapper, (void *)&cmd_args);
+
+	//If lvcreate is old and doesn't support signature wiping, try again without it.
```
brauner: We only use `/* */` style comments in LXC. :) Would be nice if you could change it. :)

src/lxc/storage/lvm.c
```diff
@@ -177,10 +185,18 @@ static int do_lvm_create(const char *path, uint64_t size, const char *thinpool)
...
+	if (strstr(cmd_output, "Error during parsing of command line") != NULL) {
```
brauner: Ugh, is that really the only way? That seems like a terrible hack.

src/lxc/storage/lvm.c
tomponline: @brauner got it, will do :)

src/lxc/storage/lvm.c
tomponline: @brauner I agree its far from ideal. 

My original intention with this was to detect the exit code of 3 (EINVALID_CMD_LINE) which hasn't changed for 15 years:

https://github.com/lvmteam/lvm2/blame/6462e8dffc7ed53973ce796cec18fecc0fb89d77/tools/lvmcmdline.c#L2803
https://github.com/lvmteam/lvm2/blame/0ecf232194986b12c3ae95b3f93e7fdd56158dc1/tools/errors.h#L21

However when I came to implement it, I realised that the `run_command` function's return value was not the exit code, but instead a parsed value where -1 was failure and 0 was OK, because it uses `wait_for_pid`.

https://github.com/lxc/lxc/blob/45265bfdfa170a5aab8c655f665b55c425d2abd6/src/lxc/utils.c#L1654
https://github.com/lxc/lxc/blob/45265bfdfa170a5aab8c655f665b55c425d2abd6/src/lxc/utils.c#L308-L309

What I needed was for an equivalent function, but using the `wait_for_pid_status` function which seems to return the exit code: https://github.com/lxc/lxc/blob/45265bfdfa170a5aab8c655f665b55c425d2abd6/src/lxc/utils.c#L314-L331

Perhaps called `run_command_status`?

In the meantime, I researched whether that return response text had changed much over time, and whether it overlapped with introduction of the wipe signatures feature.

The error response output appears to have changed once in 15 years:

15 years ago it was what it is today: "Error during parsing of command line."
https://github.com/lvmteam/lvm2/blame/0dabe7237c531459792de953879a5c4c6de41c65/tools/lvmcmdline.c#L2677

2 years ago it changed briefly for less than a day to:

https://github.com/lvmteam/lvm2/commit/6471bb2c41cf0e692a8fe7ea00033327359d42fc

Before changing back to:

https://github.com/lvmteam/lvm2/commit/2eaca7ab63345594ccfd7133ddf92729b7e7691c#diff-3e7d4e73e6b87bd5050608a3a4bf2285R2677

The `--wipesignatures` and `-Wy` flags were added five years ago (https://github.com/lvmteam/lvm2/commit/169b4c15864330e3fe47efee7b0def06581b8f08).

So that change overlaps the brief fluctuation in error output.

Additionally, comparing to that string `Error during parsing of command line` rather than the more specific output above it `invalid option -- W` also catches versions that don't support the `--yes` flag which outputs: 

```lvcreate: unrecognized option `--yes`
    Error during parsing of command line.
```

I agree that comparing strings in command output is not a great approach, however I wanted to balance that with a more invasive modification to LXC. I also was trying to avoid having to run lvcreate twice on modern systems (in order to introspect the version).

If you're happy with me adding a new `run_command_status` or doing version introspection on lvcreate then I'm happy to change tack on this. Or maybe I've missed a way to get the exit code from `run_command` in an easier way.

Thanks
Tom

tomponline:
@brauner I have taken your feedback onboard and updated this PR to implement my original intention of using the exit code of lvcreate. The change is a bit more invasive than the earlier implementation, but it avoids the output text comparison, which is much preferable. Let me know what you think. Thanks

brauner:
jenkins: test this please

src/lxc/utils.c
```diff
@@ -1651,13 +1651,29 @@ int run_command(char *buf, size_t buf_size, int (*child_fn)(void *), void *args)
 			buf[bytes - 1] = '\0';
 	}
 
-	fret = wait_for_pid(child);
+	if (wait_status) {
+		fret = lxc_wait_for_pid_status(child);
+	}
+	else {
```
brauner: Nit:
- don't put a `\n' after a `}` and the `else`
- don't use `{}` around single-line branches

tomponline:
@brauner thanks! I've fixed that code style change now.

brauner:
jenkins: test this please

tomponline:
@brauner the jenkins test failed, but looking at the console output in jenkins it says:

```
 ==> Executing: "git checkout f85b74ea404e183ff3560c9742d732a141097d8d" in /build
fatal: reference is not a tree: f85b74ea404e183ff3560c9742d732a141097d8d
```

So not sure if that is something I've done wrong, or whether there is something wrong with Jenkins?

stgraber:
jenkins: test this please

stgraber:
This can happen when pushing just as jenkins gets triggered or similar cases, anyway, the re-run looks much better now.

PR #2917, opened 2019-03-25
lxccontainer: check do_lxcapi_init_pid() for failure

tych0:
This function can fail, because it uses the command API. If it does fail,
we get weird errors about not being able to open strange proc paths:

xc authyldapservice-c8020e20-e203-e852-90ef-4d378e8d1444 20190323163231.386 ERROR    lxc_utils - utils.c:switch_to_ns:1184 - No such file or directory - failed to open /proc/-104/ns/net

So let's check for errors before then.


brauner:
Thanks! :)

PR #2920, opened 2019-03-27
Update docs

tenforward:
This is the patch that adds configuration items added in LXC 3.1 to the man pages. Each description is based on a release announcement. If there is an incorrect part, please correct it. (And I'm not good at English, so if incorrect, please touch up.  :sweat_smile: )


brauner:
Perfect as always!

PR #2921, opened 2019-03-28
Adds veth router mode

tomponline:
This PR adds a new "router" mode for veth networking.

The idea behind this is to provide an ipvlan-like networking mode that has the following properties:

- Works on older kernels.
- Uses the host's routing table (and netfilter rules) to route packets (potentially out of different interfaces or between containers), unlike ipvlan.
- Prevents containers from altering their IP.
- Prevents broadcast/multicast traffic to/from containers.
- Provides same MAC externally for all containers.
- No bridge interface to manage.
- Supports layer 3 only mode for setups where BGP (or other routing protocols) are running on the host to distribute container's IPs in the local routing table to the wider network.
- Containers can optionally have IPs published on local LAN at layer 2 using the existing `l2proxy` and `link` settings.

To implementation adds a new `mode` config property for veth type networks, which can support 2 values; "bridge" and "router".

If `mode` is "bridge" then all existing behaviour is maintained.
If `mode` is not defined in the config file, it defaults to "bridge". 

This "router" mode will configure the host machine as a router for the container by adding static routes for the container's IPs on the host pointing to the host-side veth interface. It will also add static IP proxy entries on the host-side veth interface for the container's gateway IPs.

Example usage:

```
lxc.net.0.type = veth
lxc.net.0.link = eth0
lxc.net.0.flags = up
lxc.net.0.veth.mode = router
lxc.net.0.ipv4.address = 192.168.3.3/32
lxc.net.0.ipv6.address = 2a02:xxx:xxx:3::3/128
lxc.net.0.ipv4.gateway = 169.254.0.1
lxc.net.0.ipv6.gateway = auto
lxc.net.0.l2proxy = 1
lxc.net.0.link = eth0
```

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -139,6 +139,7 @@ lxc_config_define(net_nic);
 lxc_config_define(net_script_down);
 lxc_config_define(net_script_up);
 lxc_config_define(net_type);
+lxc_config_define(net_veth_mode);
 lxc_config_define(net_veth_pair);
 lxc_config_define(net_veth_ipv4_route);
 lxc_config_define(net_veth_ipv6_route);
@@ -234,6 +235,7 @@ static struct lxc_config_t config_jump_table[] = {
 	{ "lxc.net.script.up",             set_config_net_script_up,               get_config_net_script_up,               clr_config_net_script_up,             },
 	{ "lxc.net.type",                  set_config_net_type,                    get_config_net_type,                    clr_config_net_type,                  },
 	{ "lxc.net.vlan.id",               set_config_net_vlan_id,                 get_config_net_vlan_id,                 clr_config_net_vlan_id,               },
+	{ "lxc.net.veth.mode",             set_config_net_veth_mode,               get_config_net_veth_mode,               clr_config_net_veth_mode,             },
 	{ "lxc.net.veth.pair",             set_config_net_veth_pair,               get_config_net_veth_pair,               clr_config_net_veth_pair,             },
 	{ "lxc.net.veth.ipv4.route",       set_config_net_veth_ipv4_route,         get_config_net_veth_ipv4_route,         clr_config_net_veth_ipv4_route,       },
 	{ "lxc.net.veth.ipv6.route",       set_config_net_veth_ipv6_route,         get_config_net_veth_ipv6_route,         clr_config_net_veth_ipv6_route,       },
@@ -303,6 +305,7 @@ static int set_config_net_type(const char *key, const char *value,
 		netdev->type = LXC_NET_VETH;
 		lxc_list_init(&netdev->priv.veth_attr.ipv4_routes);
 		lxc_list_init(&netdev->priv.veth_attr.ipv6_routes);
+		lxc_veth_mode_to_flag(&netdev->priv.veth_attr.mode, "bridge");
 	} else if (strcmp(value, "macvlan") == 0) {
 		netdev->type = LXC_NET_MACVLAN;
 		lxc_macvlan_mode_to_flag(&netdev->priv.macvlan_attr.mode, "private");
@@ -448,6 +451,21 @@ static int set_config_net_name(const char *key, const char *value,
 		return -1;
 
 	return network_ifname(netdev->name, value, sizeof(netdev->name));
+}
+
+
+static int set_config_net_veth_mode(const char *key, const char *value,
+				       struct lxc_conf *lxc_conf, void *data)
+{
+	struct lxc_netdev *netdev = data;
+
+	if (lxc_config_value_empty(value))
+		return clr_config_net_veth_mode(key, lxc_conf, data);
+
+	if (!netdev)
+		return -1;
+
+	return lxc_veth_mode_to_flag(&netdev->priv.veth_attr.mode, value);
 }
 
 static int set_config_net_veth_pair(const char *key, const char *value,
@@ -5094,6 +5112,22 @@ static int clr_config_net_ipvlan_isolation(const char *key,
 	return 0;
 }
 
+static int clr_config_net_veth_mode(const char *key,
+				       struct lxc_conf *lxc_conf, void *data)
+{
+	struct lxc_netdev *netdev = data;
+
+	if (!netdev)
+		return minus_one_set_errno(EINVAL);
+
+	if (netdev->type != LXC_NET_VETH)
+		return 0;
+
+	netdev->priv.veth_attr.mode = -1;
+
+	return 0;
+}
+
 static int clr_config_net_veth_pair(const char *key, struct lxc_conf *lxc_conf,
 				    void *data)
 {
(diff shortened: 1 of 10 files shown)


tomponline:
@brauner my two main questions around the approach I've used here are:

1. Does this need to be supported by `lxc-user-nic`, I thought it might, but it appears to be relevant only for bridging https://linuxcontainers.org/lxc/manpages//man1/lxc-user-nic.1.html
2. Before this PR `instantiate_veth` function was at the top of the file, but due to the use of static functions added to this PR used inside `instantiate_veth` I needed to add new functions above it. All other functions used inside `instantiate_veth` are defined extern and added to network.h or not defined as static. Is there any guidance around when to use static vs extern and ordering of functions inside a file?
3. I've made the default of the `mode` property "bridge" so that the old behaviour is the default, however I am concerned that there may be another way to call `instantiate_veth` that may not pass the config initialisation here https://github.com/lxc/lxc/pull/2921/files#diff-2f3344d3cb9700d11140decc560aaacaR291, would it be worth checking for "bridge" and empty string here https://github.com/lxc/lxc/pull/2921/files#diff-49260cced2f6f9c5617962928313c73aR242 to ensure the old behaviour id `mode` is empty?

Thanks
Tom

tomponline:
@brauner please can you take a look at this when you get chance, I'd like to try and get it ready in time for LXC 3.2 release. It also includes error_log_errno so we can close https://github.com/lxc/lxc/pull/3012  too.

src/lxc/confile.c
```diff
@@ -5060,6 +5078,22 @@ static int clr_config_net_ipvlan_isolation(const char *key,
 	return 0;
 }
 
+static int clr_config_net_veth_mode(const char *key,
+				       struct lxc_conf *lxc_conf, void *data)
+{
+	struct lxc_netdev *netdev = data;
+
+	if (!netdev)
+		return -1;
```
brauner: Please use `minus_one_set_errno(EINVAL)`

src/lxc/confile.c
```diff
@@ -5482,6 +5516,42 @@ static int get_config_net_ipvlan_isolation(const char *key, char *retv, int inle
...
+	const char *mode;
+	struct lxc_netdev *netdev = data;
+
+	if (!retv)
+		inlen = 0;
+	else
+		memset(retv, 0, inlen);
+
+	if (!netdev)
+		return -1;
```
brauner: Please use `minus_one_set_errno(EINVAL)`

src/lxc/confile_utils.c
```diff
@@ -501,6 +501,30 @@ void lxc_free_networks(struct lxc_list *networks)
...
+} veth_mode[] = {
+    { "bridge", VETH_MODE_BRIDGE },
+    { "router", VETH_MODE_ROUTER },
+};
+
+int lxc_veth_mode_to_flag(int *mode, const char *value)
+{
+	size_t i;
+
+	for (i = 0; i < sizeof(veth_mode) / sizeof(veth_mode[0]); i++) {
```
brauner:
```
for (size_t i = 0; ...)
```

src/lxc/confile_utils.c
```diff
@@ -501,6 +501,30 @@ void lxc_free_networks(struct lxc_list *networks)
...
+		if (strcmp(veth_mode[i].name, value))
```
brauner:
```
if (strcmp(...) != 0))
```

src/lxc/confile_utils.c
```diff
@@ -501,6 +501,30 @@ void lxc_free_networks(struct lxc_list *networks)
...
+			continue;
+
+		*mode = veth_mode[i].mode;
+		return 0;
+	}
+
+	return -1;
```
brauner: Please use `minus_one_set_errno(EINVAL)`

src/lxc/network.c
```diff
@@ -178,6 +178,102 @@ static int lxc_setup_ipv6_routes(struct lxc_list *ip, int ifindex)
...
+	lxc_list_for_each(iterator, ip) {
+		struct lxc_inetdev *inetdev = iterator->elem;
+
+		err = lxc_ipv4_dest_add(ifindex, &inetdev->addr, 32);
+
+		if (err) {
+			return error_log_errno(err,
+				"Failed to setup ipv4 address route for network device with eifindex %d",
+				ifindex);
+		}
```
brauner: No `{}` around single-line if statements.

src/lxc/network.c
```diff
@@ -178,6 +178,102 @@ static int lxc_setup_ipv6_routes(struct lxc_list *ip, int ifindex)
...
+static int setup_ipv6_addr_routes(struct lxc_list *ip, int ifindex)
+{
+	struct lxc_list *iterator;
+	int err;
+
+	lxc_list_for_each(iterator, ip) {
+		struct lxc_inet6dev *inet6dev = iterator->elem;
+
+		err = lxc_ipv6_dest_add(ifindex, &inet6dev->addr, 128);
+		if (err) {
```
brauner: No `{}` around single-line if statements

src/lxc/network.c
```diff
@@ -178,6 +178,102 @@ static int lxc_setup_ipv6_routes(struct lxc_list *ip, int ifindex)
...
+	struct ip_proxy_args *args = data;
+
+	execlp("ip", "ip", "neigh", "flush", "proxy", args->ip, "dev", args->dev, (char *)NULL);
+	return -1;
+}
+
+static int lxc_add_ip_neigh_proxy(const char *ip, const char *dev)
+{
+	int ret;
+	char cmd_output[PATH_MAX];
```
brauner: Should likely initialize first byte to 0 to not print garbage below

src/lxc/network.c
```diff
@@ -178,6 +178,102 @@ static int lxc_setup_ipv6_routes(struct lxc_list *ip, int ifindex)
...
+		return -1;
+	}
+
+	return 0;
+}
+
+static int lxc_del_ip_neigh_proxy(const char *ip, const char *dev)
+{
+	int ret;
+	char cmd_output[PATH_MAX];
```
brauner: Should likely initialize first byte to 0 to not print garbage below

src/lxc/network.c
```diff
@@ -306,6 +402,62 @@ static int instantiate_veth(struct lxc_handler *handler, struct lxc_netdev *netd
 		goto out_delete;
 	}
 
+	if (netdev->priv.veth_attr.mode == VETH_MODE_ROUTER) {
+		if (netdev->ipv4_gateway) {
+			char bufinet4[INET_ADDRSTRLEN];
+			inet_ntop(AF_INET, netdev->ipv4_gateway, bufinet4, sizeof(bufinet4));
```
brauner: Shouldn't you check for error here?

src/lxc/network.c
```diff
@@ -306,6 +402,62 @@ static int instantiate_veth(struct lxc_handler *handler, struct lxc_netdev *netd
...
+			err = lxc_add_ip_neigh_proxy(bufinet4, veth1);
+			if (err) {
+				error_log_errno(err, "Failed to add gateway ipv4 proxy on \"%s\"", veth1);
+				goto out_delete;
+			}
+		}
+
+		if (netdev->ipv6_gateway) {
+			char bufinet6[INET6_ADDRSTRLEN];
+			inet_ntop(AF_INET6, netdev->ipv6_gateway, bufinet6, sizeof(bufinet6));
```
brauner: Shouldn't you check for error here?

tomponline:
> Mostly nits otherwise good to go.
> One question though: does this patchset take care to set bridge mode by default?

@brauner thanks, yes it sets it to bridge mode by default here https://github.com/lxc/lxc/pull/2921/files#diff-2f3344d3cb9700d11140decc560aaacaR306

brauner:
@tomponline, did you get around to fixing this up? :)

tomponline:
@brauner not yet, will get it done today though :)

tomponline:
@brauner I've made those modifications, and also added an additional sysctl check for `net.ipv6.conf.all.forwarding=1` as this is needed for any sort of IPv6 forwarding (unlike IPv4).

brauner:
Sweet, thanks!

Issue #2927, opened 2019-04-06
API doc generation fails under out of source build.


yosukesan:
# Required information 

* Distibution: Debian
* Distribution Version: 9
* commit 4f34c6f95dfd94a1736d1bf73783ab97aedb7fd6.

# Issue description

If --enable-api-docs is enabled under out of source directory build, api doc generation by doxygen failes due to absence of input file.

'''
make[3]: Nothing to be done for 'all-am'.
make[3]: Leaving directory '/home/yosuke/Data/github/work_lxc/build_org/doc/ko'
make[2]: Leaving directory '/home/yosuke/Data/github/work_lxc/build_org/doc/ko'
Making all in api
make[2]: Entering directory '/home/yosuke/Data/github/work_lxc/build_org/doc/api'
doxygen
error: Doxyfile not found and no input file specified!
'''

# Steps to reproduce

1. Install doxygen
2. Used commit 4f34c6f95dfd94a1736d1bf73783ab97aedb7fd6.
3. Generated configure file in the source dir.
4. Prepare src (lxc below) and build directory as described. Assume we are at ${work_root}.

```
${work_root}
      |-- lxc
      |-- build
```

5. cd build; ../lxc/configure; make

The step 1. might be omitted if --enable-api-docs is manually enabled. However, I did not test it.

# Information to attach

If you recognise this is a problem, I am happy to send a pull request :).

stgraber:
Not really a setup that we or so far, much anyone appears to be running, but if API doc is the only thing failing in such a setup, then sure, a pull request to fix it would be fine :)

PR #2928, opened 2019-04-06
fix: #2927 api doc generation fails under out of source build.

yosukesan:
Signed-off-by: yosukesan <y.otsuki30@gmail.com>

# Description of the patch

This patch fix api doc generation failure by doxgen.

# Trigger of the problem

Absence of argument when doxygen is called was the cause.
If no argument is specified, doxygen searches a Doxyfile in current dir. Hence, in source build worked.
However, in out of source build this did not.

# Test

I tested both out of source and in source build both successfully terminated.

* In source build was just simple.
```
bash autogen.sh && ./configure && make
```

* out of source build
Identical to "Step to reproduce" in #2972

Do I need test more ? If so let me know.

# Note

For CMake users, out of source build is standard procedure :).

### DIFF
--- a/doc/api/Makefile.am
+++ b/doc/api/Makefile.am
@@ -2,7 +2,7 @@ EXTRA_DIST = Doxyfile
 
 if ENABLE_API_DOCS
 html: Doxyfile
-	doxygen
+	doxygen $<
 endif
 
 .PHONY: html


brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
Thanks!

PR #2933, opened 2019-04-13
start: handle setting pdeath signal in new pidns

brauner:
In the usual case the child runs in a separate pid namespace. So far we haven't
been able to reliably set the pdeath signal. When we set the pdeath signal we
need to verify that we haven't lost a race whereby we have been orphaned and
though we have set a pdeath signal it won't help us since, well, the parent is
dead.
We were able to correctly handle this case when we were in the same pidns since
getppid() will return a valid pid. When we are in a separate pidns 0 will be
returned since the parent doesn't exist in our pidns.
A while back, while Jann and I were discussing other things he came up
with a nifty idea: simply pass an fd for the parent's status file and check the
"State:" field. This is the implementation of that idea.

Suggested-by: Jann Horn <jann@thejh.net>

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -709,6 +709,9 @@ void lxc_free_handler(struct lxc_handler *handler)
 		if (handler->conf->maincmd_fd >= 0)
 			lxc_abstract_unix_close(handler->conf->maincmd_fd);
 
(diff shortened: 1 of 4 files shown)


brauner:
//Cc @flx42

brauner:
jenkins: test this please

PR #2934, opened 2019-04-13
attach: use tighter scope for fd variable

brauner:


brauner:
This is really trivial and I'm abusing this improvement to test Jenkins and see whether the error I'm seeing in my other branch is actually triggered by my changes. It seems really unlikely that this is caused by my changes though.

PR #2939, opened 2019-04-15
execute: add a way to wait in lxc-init before spawning

tych0:
We're working on an implementation of crio-lxc. The crio spec thinks about
create and start slightly differently than liblxc does: create actually
creates the namespaces that the container will finally be run in to allow
them to be manipulated. It also allows passing the fds that the container
will use as stdin/stdout, etc. Alternatively, create() in lxc parlance is
just an on-disk operation, where start() creates the namespaces and spawns
the "real" init.

To support the crio semantics in liblxc, we introduce an option to have
lxc-init (blocking) write to a fifo, so that the crio implementation can
wait to read from the fifo until crio's start method is called. This means
that all the namespaces are around and can be manipulated, and the fds
passed at create() time are kept around and correct.

This abstraction is slightly leaky, since it requires the user to both set
up the fifo (e.g. bind mount it in, by setting a mount on liblxc) and then
say again where it is. A less leaky approach might be to just handle the
fifo setup inside liblxc entirely, but then we'd need a new
->execute_really_start() api call instead of just a new config option.

Signed-off-by: Michael McCracken <mikmccra@cisco.com>

### DIFF
--- a/src/lxc/cmd/lxc_init.c
+++ b/src/lxc/cmd/lxc_init.c
@@ -74,9 +74,10 @@ static struct option long_options[] = {
 	    { "version",     no_argument,       0, OPT_VERSION },
 	    { "quiet",       no_argument,       0, 'q'         },
 	    { "lxcpath",     required_argument, 0, 'P'         },
+	    { "exec-wait",   no_argument,       0, 'w'         },
 	    { 0,             0,                 0, 0           }
 	};
-static const char short_options[] = "n:hqo:l:P:";
+static const char short_options[] = "n:hqo:l:P:wk";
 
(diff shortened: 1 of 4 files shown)


brauner:
You think you will only ever need this for containers started through `lxc-init` and never for full system containers?

brauner:
Basically, if you think this is generally useful and not just for this special case, you should probably do:
```
setup()
launch()
```
where `setup()` sets up the container and `launch()` actually execs...

tych0:
Per some discussion on IRC, we're going to drop this entirely and do things a different way.

Issue #2932, opened 2019-04-12
Document mount order


flx42:
Mostly a nice to have, but I got confused for a while on why our `mount` hook ("a hook to be run in the container's namespace after mounting has been done, but before the pivot_root") was called before the `lxc.mount.entry` but after the `lxc.mount.auto`. 

From the code:
https://github.com/lxc/lxc/blob/9810d1916eacea64c838681d3045cac60e7f00c2/src/lxc/conf.c#L3537-L3614
The order is:
* `lxc.mount.auto = proc` and `lxc.mount.auto = sys`
* `lxc.mount.fstab`
* `lxc.mount.auto = cgroup`
* `lxc.hook.mount`
* `lxc.autodev`
* `lxc.mount.entry`

By the way, was there a reason to **not** have `lxc.mount.entry` immediately after `lxc.mount.fstab`?

brauner:
> By the way, was there a reason to not have lxc.mount.entry immediately after lxc.mount.fstab?

Not that I know of.

You want the mount order documented in `lxc.container.conf` or in the code?

flx42:
> By the way, was there a reason to not have lxc.mount.entry immediately after lxc.mount.fstab?

Should we try to fix it or is too late and part of the implicit API?

> You want the mount order documented in lxc.container.conf or in the code?

In `lxc.container.conf`, the code is easy to follow once you find the right place.

brauner:
Yeah, I think we can send a patch, clean this up and then document it. :)

brauner:
Care to send one, @flx42?

flx42:
Was out of the office for a few days, I'll try to do it this week.

flx42:
Let's see if changing the mount order breaks anything before tackling documentation.

tardyp:
> By the way, was there a reason to not have lxc.mount.entry immediately after lxc.mount.fstab?

A use case is to bind mount additional stuff like /dev/bus/usb

This is documented to be working here:
https://medium.com/@konpat/usb-passthrough-to-an-lxc-proxmox-15482674f11d

does not work for me, but I am in lxc 3.03 so before this patch..

PR #2941, opened 2019-04-16
conf: do lxc.mount.entry mounts right after lxc.mount.fstab

flx42:
These configuration options use the same syntax and therefore it seems
more intuitive to have the same behavior for both of them, which is
not the case today since mount hooks and autodev mounts are called
between the two.

See: https://github.com/lxc/lxc/issues/2932

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -3546,6 +3546,15 @@ int lxc_setup(struct lxc_handler *handler)
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

brauner:
If there are any regressions the LXD edge snap should catch them.

brauner:
Thanks! :)

PR #2942, opened 2019-04-16
hooks/nvidia: handle spaces in NVIDIA_REQUIRE variables

flx42:
Previously, environment variables with a space where splitted.


brauner:
Thanks!

PR #2948, opened 2019-04-24
Fix 'zfs get' command order

pgauret:
Another case of calling 'zfs get' which requires reordering arguments to work with latest zfs.

Signed-off-by: Paul Gauret <pgauret@yahoo.com>


brauner:
Thanks!

PR #2963, opened 2019-04-30
hooks: Adds LXC_NET_ID enviroment variable

tomponline:
Adds LXC_NET_ID environment variable to network up and down hooks.

This represents the netdev index ID of the network device in the config file.

E.g. lxc.net.6.type=ipvlan would result in LXC_NET_ID = 6

This allows hooks to parse the relevant config file section related to the network being processed.

### DIFF
--- a/doc/lxc.container.conf.sgml.in
+++ b/doc/lxc.container.conf.sgml.in
@@ -699,6 +699,14 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
                  to 1.
                   </para>
                 </listitem>
+
+                <listitem>
+                 <para>
+                 LXC_NET_ID: the network device index ID in the config file.
+                 Note that this information is only available when
+                 <option>lxc.hook.version</option> is set to 1.
+                  </para>
+                </listitem>
               </itemizedlist>
 
               Whether this information is provided in the form of environment
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -360,7 +360,7 @@ static int run_buffer(char *buffer)
 
 int run_script_argv(const char *name, unsigned int hook_version,
 		    const char *section, const char *script,
-		    const char *hookname, char **argv)
+		    const char *hookname, char **argv, int netdev_idx)
 {
 	__do_free char *buffer = NULL;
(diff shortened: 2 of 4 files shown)


brauner:
What do we need that for exactly?

tomponline:
@brauner this is something I've thought about adding for a while. When writing network hook scripts that need to interrogate the container's config file (either directly or via a binding like go-lxc) in order to read config settings to use in the hook, it is currently quite tricky to figure out which network device the hook is being executed for (if you are not using static interface names in the config file).

This provides an indication to the hook script as to which network index the hook is being executed for, which then allows the config to be access, e.g.:

```go
package main

import "gopkg.in/lxc/go-lxc.v2"
import "log"
import "fmt"

func main() {
        ctName := "test1" //LXC_NAME env var
        fmt.Println(lxc.DefaultConfigPath())
        c, err := lxc.NewContainer(ctName, lxc.DefaultConfigPath())
        if err != nil {
                log.Fatalf("Cannot load config for: %s, %s\n", ctName, err.Error())
        }

        if !c.Defined() {
                log.Fatalf("Container %s not defined", ctName)
        }

        i := 4 //LXC_NET_ID env var
        fmt.Println(c.ConfigKeys(fmt.Sprintf("%s.%d", "lxc.net", i)))
        fmt.Println(c.ConfigItem(fmt.Sprintf("%s.%d.type", "lxc.net", i)))
}
```

brauner:
I want to see people who actually need before exposing information like this in an environment variable. We can leave the PR open for a while.

tomponline:
@brauner fair enough :)

This originally presented itself as a problem for me when writing the "routed veth" mode as a network hook. This required retrieving IP and gateway address info about the container from the config file in order to setup the relevant IP neigh proxy and routes.

In my original implementation I have naively iterated through each network and assumed the ID would be the same as the order it was defined in the file and starting at 0. https://github.com/tomponline/lxc-routednet/blob/master/lxc-routednet.go#L71-L78

However now I realise that is not the case, and you can end up retrieving the wrong network device's config using that approach.

I thought this would be useful to any future network hook writers who need to access the network device config.

There may be another way of achieving this that I have not thought of though.

tomponline:
This is not needed anymore.

PR #2964, opened 2019-04-30
network: Adds layer 2 (ARP/NDP) proxy mode

tomponline:
Adds the `lxc.net.[i].l2proxy` flag that can be either 0 or 1.

Defaults to 0.

This, when used with `lxc.net.[i].link`, will add IP neighbour proxy entries on the linked device
for any IPv4 and IPv6 addresses on the container's network device.

E.g. where enp3s0 is the linked parent interface.

```
ip -4 route add local ${IF_IP4}/32 dev enp3s0
ip neigh add proxy ${IF_IP4} dev enp3s0
ip -6 route add local ${IF_IP6}/128 dev enp3s0
ip neigh add proxy ${IF_IP6} dev enp3s0
```

Additionally, for IPv4 addresses it will check the following sysctls are enabled and fail with an error if not:

```
sysctl net.ipv4.conf.enp3s0.forwarding=1
```

Additionally, for IPv6 addresses it will check the following sysctls are enabled and fail with an error if not:

```
sysctl net.ipv6.conf.enp3s0.proxy_ndp=1
sysctl net.ipv6.conf.enp3s0.forwarding=1
```

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -129,6 +129,7 @@ lxc_config_define(net_ipv4_gateway);
 lxc_config_define(net_ipv6_address);
 lxc_config_define(net_ipv6_gateway);
 lxc_config_define(net_link);
+lxc_config_define(net_l2proxy);
 lxc_config_define(net_macvlan_mode);
 lxc_config_define(net_ipvlan_mode);
 lxc_config_define(net_ipvlan_isolation);
@@ -222,6 +223,7 @@ static struct lxc_config_t config_jump_table[] = {
 	{ "lxc.net.ipv6.address",          set_config_net_ipv6_address,            get_config_net_ipv6_address,            clr_config_net_ipv6_address,          },
 	{ "lxc.net.ipv6.gateway",          set_config_net_ipv6_gateway,            get_config_net_ipv6_gateway,            clr_config_net_ipv6_gateway,          },
 	{ "lxc.net.link",                  set_config_net_link,                    get_config_net_link,                    clr_config_net_link,                  },
+	{ "lxc.net.l2proxy",               set_config_net_l2proxy,                 get_config_net_l2proxy,                 clr_config_net_l2proxy,               },
 	{ "lxc.net.macvlan.mode",          set_config_net_macvlan_mode,            get_config_net_macvlan_mode,            clr_config_net_macvlan_mode,          },
 	{ "lxc.net.ipvlan.mode",           set_config_net_ipvlan_mode,             get_config_net_ipvlan_mode,             clr_config_net_ipvlan_mode,           },
 	{ "lxc.net.ipvlan.isolation",      set_config_net_ipvlan_isolation,        get_config_net_ipvlan_isolation,        clr_config_net_ipvlan_isolation,      },
@@ -401,6 +403,35 @@ static int set_config_net_link(const char *key, const char *value,
 		ret = network_ifname(netdev->link, value, sizeof(netdev->link));
 
 	return ret;
+}
+
+static int set_config_net_l2proxy(const char *key, const char *value,
+				     struct lxc_conf *lxc_conf, void *data)
+{
+	struct lxc_netdev *netdev = data;
+	unsigned int val = 0;
+	int ret;
+
+	if (lxc_config_value_empty(value))
+		return clr_config_net_l2proxy(key, lxc_conf, data);
+
+	if (!netdev)
+		return minus_one_set_errno(EINVAL);
+
+	ret = lxc_safe_uint(value, &val);
+	if (ret < 0)
+		return minus_one_set_errno(-ret);
+
+	switch (val) {
+	case 0:
+		netdev->l2proxy = false;
+		return 0;
+	case 1:
+		netdev->l2proxy = true;
+		return 0;
+	}
+
+	return minus_one_set_errno(EINVAL);
 }
 
 static int set_config_net_name(const char *key, const char *value,
@@ -4960,6 +4991,19 @@ static int clr_config_net_link(const char *key, struct lxc_conf *lxc_conf,
 	return 0;
 }
 
+static int clr_config_net_l2proxy(const char *key, struct lxc_conf *lxc_conf,
+			       void *data)
+{
+	struct lxc_netdev *netdev = data;
+
+	if (!netdev)
+		return minus_one_set_errno(EINVAL);
+
+	netdev->l2proxy = false;
+
+	return 0;
+}
+
 static int clr_config_net_macvlan_mode(const char *key,
 				       struct lxc_conf *lxc_conf, void *data)
 {
@@ -5280,6 +5324,13 @@ static int get_config_net_link(const char *key, char *retv, int inlen,
 		strprint(retv, inlen, "%s", netdev->link);
 
 	return fulllen;
+}
+
+static int get_config_net_l2proxy(const char *key, char *retv, int inlen,
+			       struct lxc_conf *c, void *data)
+{
+	struct lxc_netdev *netdev = data;
+	return lxc_get_conf_bool(c, retv, inlen, netdev->l2proxy);
 }
 
 static int get_config_net_name(const char *key, char *retv, int inlen,
--- a/src/lxc/file_utils.c
+++ b/src/lxc/file_utils.c
@@ -147,7 +147,7 @@ ssize_t lxc_read_nointr_expect(int fd, void *buf, size_t count, const void *expe
 	ssize_t ret;
 
 	ret = lxc_read_nointr(fd, buf, count);
-	if (ret <= 0)
+	if (ret < 0)
 		return ret;
 
 	if ((size_t)ret != count)
@@ -158,7 +158,18 @@ ssize_t lxc_read_nointr_expect(int fd, void *buf, size_t count, const void *expe
 		return -1;
 	}
 
-	return ret;
+	return 0;
+}
+
+ssize_t lxc_read_file_expect(const char *path, void *buf, size_t count, const void *expected_buf)
+{
+	__do_close_prot_errno int fd = -EBADF;
+
+	fd = open(path, O_RDONLY | O_CLOEXEC);
+	if (fd < 0)
+		return -1;
+
+	return lxc_read_nointr_expect(fd, buf, count, expected_buf);
 }
 
 bool file_exists(const char *f)
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -1660,6 +1660,24 @@ static int proc_sys_net_write(const char *path, const char *value)
 	return err;
 }
 
+static int lxc_is_ip_forwarding_enabled(const char *ifname, int family)
+{
+	int ret;
+	char path[PATH_MAX];
+	char buf[1] = "";
+
+	if (family != AF_INET && family != AF_INET6)
+		return minus_one_set_errno(EINVAL);
+
+	ret = snprintf(path, PATH_MAX, "/proc/sys/net/%s/conf/%s/%s",
+		       family == AF_INET ? "ipv4" : "ipv6", ifname,
+		       "forwarding");
+	if (ret < 0 || (size_t)ret >= PATH_MAX)
+		return minus_one_set_errno(E2BIG);
+
+	return lxc_read_file_expect(path, buf, 1, "1");
+}
+
 static int neigh_proxy_set(const char *ifname, int family, int flag)
 {
 	int ret;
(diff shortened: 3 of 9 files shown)


tomponline:
@stgraber here is the implementation of the `l2proxy` feature.

On IPv6 systems, in order for the proxy entry to take effect (aside from needing a static route which applies equally to IPv4 entries too) we also need to enable `sysctl net.ipv6.conf.[link].proxy_ndp=1` as it works differently to IPv4.

However I wanted to check this is OK as previous comments on this thread suggest that this would be better left as a documentation point:
https://github.com/lxc/lxd/pull/5182#discussion_r229118305

tomponline:
@stgraber as discussed I'll update this to check for the specific sysctl entries being configured and if not then fail to start with an error explaining what is wrong,

tomponline:
@stgraber @brauner this is ready for review now. Thanks

src/lxc/confile.c
```diff
@@ -403,6 +405,33 @@ static int set_config_net_link(const char *key, const char *value,
...
+				     struct lxc_conf *lxc_conf, void *data)
+{
+	struct lxc_netdev *netdev = data;
+	unsigned int val = 0;
+
+	if (lxc_config_value_empty(value))
+		return clr_config_net_l2proxy(key, lxc_conf, data);
+
+	if (!netdev)
+		return -1;
```
brauner:
```c
return minus_one_set_errno(EINVAL);
```

src/lxc/confile.c
```diff
@@ -403,6 +405,33 @@ static int set_config_net_link(const char *key, const char *value,
...
+
+	if (lxc_safe_uint(value, &val) < 0)
+		return minus_one_set_errno(EINVAL);
```
brauner: I think the `lxc_safe_uint()` functions will set errno correctly so better would be:
```c
ret = lxc_safe_uint();
if (ret < 0)
        return minus_...
```

src/lxc/confile.c
```diff
@@ -4960,6 +4989,19 @@ static int clr_config_net_link(const char *key, struct lxc_conf *lxc_conf,
 	return 0;
 }
 
+static int clr_config_net_l2proxy(const char *key, struct lxc_conf *lxc_conf,
+			       void *data)
+{
+	struct lxc_netdev *netdev = data;
+
+	if (!netdev)
+		return -1;
```
brauner:
```c
return minus_one_set_errno(EINVAL);
```

src/lxc/file_utils.c
```diff
@@ -147,7 +147,7 @@ ssize_t lxc_read_nointr_expect(int fd, void *buf, size_t count, const void *expe
 	ssize_t ret;
 
 	ret = lxc_read_nointr(fd, buf, count);
-	if (ret <= 0)
+	if (ret < 0)
```
brauner: What, why?

src/lxc/file_utils.c
```diff
@@ -158,6 +158,21 @@ ssize_t lxc_read_nointr_expect(int fd, void *buf, size_t count, const void *expe
 		return -1;
 	}
 
+	return 0;
+}
+
+ssize_t lxc_read_file_expect(const char *path, void *buf, size_t count, const void *expected_buf)
+{
+	int fd;
```
brauner:
```c
__do_close_prot_errno int fd = -EBADF;
```

src/lxc/file_utils.c
```diff
@@ -158,6 +158,21 @@ ssize_t lxc_read_nointr_expect(int fd, void *buf, size_t count, const void *expe
...
+	ssize_t ret;
+
+	fd = open(path, O_RDONLY | O_CLOEXEC);
+	if (fd < 0)
+		return -1;
+
+	ret = lxc_read_nointr_expect(fd, buf, count, expected_buf);
+	close(fd);
```
brauner: With the cleanup macro above this line needs to be removed.

brauner: At which point you can remove remove the `ssize_t ret` declaration and return with:
```c
return lxc_read_nointr_expect(...
```

src/lxc/network.c
```diff
@@ -1660,6 +1660,25 @@ static int proc_sys_net_write(const char *path, const char *value)
...
+	char buf[1] = "";
+
+	if (family != AF_INET && family != AF_INET6)
+		return minus_one_set_errno(EINVAL);
+
+	ret = snprintf(path, PATH_MAX, "/proc/sys/net/%s/conf/%s/%s",
+		       family == AF_INET ? "ipv4" : "ipv6", ifname,
+		       "forwarding");
+
+	if (ret < 0 || (size_t)ret >= PATH_MAX)
```
brauner: Please remove the additional '\n'.

src/lxc/network.c
```diff
@@ -1677,6 +1696,25 @@ static int neigh_proxy_set(const char *ifname, int family, int flag)
...
+	char buf[1] = "";
+
+	if (family != AF_INET && family != AF_INET6)
+		return minus_one_set_errno(EINVAL);
+
+	ret = snprintf(path, PATH_MAX, "/proc/sys/net/%s/conf/%s/%s",
+		       family == AF_INET ? "ipv4" : "ipv6", ifname,
+		       family == AF_INET ? "proxy_arp" : "proxy_ndp");
+
+	if (ret < 0 || (size_t)ret >= PATH_MAX)
```
brauner: Please remove the additional `\n`.

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+static int lxc_add_ip_proxy(const char *ip, const char *dev)
+{
+	int ret;
+	char cmd_output[PATH_MAX];
+	struct ip_proxy_args args;
+	args.ip = ip;
+	args.dev = dev;
+
+	ret = run_command(cmd_output, sizeof(cmd_output),
+			  lxc_add_ip_proxy_exec_wrapper, (void *)&args);
```
brauner: Why the explicit cast to void?

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+	return 0;
+}
+
+static int lxc_del_ip_proxy(const char *ip, const char *dev)
+{
+	int ret;
+	char cmd_output[PATH_MAX];
+	struct ip_proxy_args args;
+	args.ip = ip;
+	args.dev = dev;
```
brauner: Should be:
```c
struct ip_proxy_args args = {
        .ip = ip,
        .dev = dev,
};

File: src/lxc/network.c [hunk 11]
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+
+	ret = run_command(cmd_output, sizeof(cmd_output),
+			  lxc_del_ip_proxy_exec_wrapper, (void *)&args);
```
brauner: The cast to void is not needed.

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+			ERROR("Requires sysctl net.ipv6.conf.%s.forwarding be set to 1", netdev->link);
+			return minus_one_set_errno(EINVAL);
+		}
+	}
+
+	lxc_list_for_each_safe(cur, &netdev->ipv4, next) {
+		inet4dev = cur->elem;
+		if (!inet_ntop(AF_INET, &inet4dev->addr, bufinet4, sizeof(bufinet4))) {
+			return minus_one_set_errno(EINVAL);
+		}
```
brauner: No {} around single-line if.

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+
+		if (lxc_add_ip_proxy(bufinet4, netdev->link)) {
+			return minus_one_set_errno(EINVAL);
+		}
```
brauner: No {} around single-line if.

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+	}
+
+	lxc_list_for_each_safe(cur, &netdev->ipv6, next) {
+		inet6dev = cur->elem;
+		if (!inet_ntop(AF_INET6, &inet6dev->addr, bufinet6, sizeof(bufinet6))) {
+			return minus_one_set_errno(EINVAL);
+		}
```
brauner: No {} around single-line if.

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+
+		if (lxc_add_ip_proxy(bufinet6, netdev->link)) {
+			return minus_one_set_errno(EINVAL);
+		}
```
brauner: No {} around single-line if.

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+
+		if (lxc_del_ip_proxy(bufinet6, netdev->link)) {
+			err = -1;
+			continue; /* Try to remove any other l2proxy entries */
+		}
+	}
+
+	if (err < 0) {
+		return minus_one_set_errno(EINVAL);
+	}
```
brauner: No {} around single-line if.

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+	struct lxc_inet6dev *inet6dev;
+	char bufinet4[INET_ADDRSTRLEN], bufinet6[INET6_ADDRSTRLEN];
+	int err = 0;
+
+	lxc_list_for_each_safe(cur, &netdev->ipv4, next) {
+		inet4dev = cur->elem;
+		if (!inet_ntop(AF_INET, &inet4dev->addr, bufinet4,sizeof(bufinet4))) {
+			err = -1;
+			continue; /* Try to remove any other l2proxy entries */
+		}
```
brauner: Should probably log which ip proxy failed.

src/lxc/network.c
```diff
@@ -2679,6 +2717,151 @@ bool lxc_delete_network_unpriv(struct lxc_handler *handler)
...
+		if (lxc_del_ip_proxy(bufinet4, netdev->link)) {
+			err = -1;
+			continue; /* Try to remove any other l2proxy entries */
+		}
+	}
+
+	lxc_list_for_each_safe(cur, &netdev->ipv6, next) {
+		inet6dev = cur->elem;
+		if (!inet_ntop(AF_INET6, &inet6dev->addr, bufinet6, sizeof(bufinet6))) {
+			err = -1;
```
brauner: Should probably log which ip proxy failed.

src/lxc/network.c
```diff
@@ -2795,6 +2985,14 @@ bool lxc_delete_network_priv(struct lxc_handler *handler)
 		if (!netdev->ifindex)
 			continue;
 
+		/* Delete l2proxy entries if enabled and used with a link property */
+		if (netdev->l2proxy && netdev->link[0] != '\0') {
+			if (lxc_delete_l2proxy(netdev)) {
+				WARN("Failed to delete l2proxy");
+				/* Don't return, let the network be cleaned up as normal. */
+			}
```
brauner: No {} around single-line if.

src/lxc/file_utils.c
tomponline: @brauner this function isn't used anywhere currently, so I've re-purposed it for my sysctl comparisons needs. Whilst doing so I noticed that this appears to have an edge case bug in it where if the result of `lxc_read_nointr` was 0 (and this would happen if the file being read is zero length) then it would return 0 without doing the buffer comparison - thus indicating that the the 2 buffers were equal even when they are potentially not.

src/lxc/file_utils.c
tomponline: @brauner thanks will do

src/lxc/network.c
tomponline: @brauner this is how it is used elsewhere in the code base. I think its because the function signature for `run_command` requires that way https://github.com/lxc/lxc/blob/master/src/lxc/utils.c#L1597 so that the same function can be used with commands that take different sets of arguments.

tomponline: @brauner I've tried removing the cast and it compiles, so I do wonder why it was added in the past:

https://github.com/lxc/lxc/blob/ac7148050126cc3cd9872007c5ea69cad44baa91/src/lxc/lsm/apparmor.c#L811
https://github.com/lxc/lxc/blob/ea84ddf9e2e4933bf8110366be3c3f6dd3c4b6a6/src/lxc/network.c#L2127-L2128

There's quite a few like that.

src/lxc/network.c
tomponline: @brauner will do,  see https://github.com/lxc/lxc/pull/2964/files#r280346212

src/lxc/network.c
tomponline: @brauner that makes sense for the `lxc_del_ip_proxy` below, but here and will add, but I don't have a string representation of the IP yet to log at this point (this is what failed). I'll fix the missing space in the function params though.

tomponline: @brauner actually `lxc_del_ip_proxy` already logs an error.

tomponline: @brauner will add some basic logging here without the IP, but with the dev instead

src/lxc/network.c
tomponline: See https://github.com/lxc/lxc/pull/2964/files#r280351891

tomponline:
@brauner I have re-worked this based on your review, thanks.

PR #2971, opened 2019-05-02
namespaces: allow a pathname to a nsfd for namespace to share

hallyn:
This is something that might be needed per the crio/runc bundle spec documentation, i.e.

https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#namespaces

### DIFF
--- a/src/lxc/confile_utils.c
+++ b/src/lxc/confile_utils.c
@@ -859,23 +859,27 @@ static int lxc_container_name_to_pid(const char *lxcname_or_pid,
 	return pid;
 }
 
-int lxc_inherit_namespace(const char *lxcname_or_pid, const char *lxcpath,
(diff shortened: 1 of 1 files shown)


hallyn:
For some reason, PR 2970 won't let me re-open it.

src/lxc/confile_utils.c
```diff
@@ -865,6 +865,10 @@ int lxc_inherit_namespace(const char *lxcname_or_pid, const char *lxcpath,
 	int fd, pid;
 	char *dup, *lastslash;
 
+	if (lxcname_or_pid[0] == '/') {
+		return open(lxcname_or_pid, O_RDONLY | O_CLOEXEC);
```
tych0: lxcname_or_pid is probably not an accurate name for this variable any more.

PR #2973, opened 2019-05-03
network: Adds gateway device route mode

tomponline:
Adds ability to specify "dev" as the gateway value, which will cause a device route to be set as default gateway.

This is primarily intended for use with IPVLAN where due to the nature of it operating at layer 3, we do not need to specify a gateway IP address.

### DIFF
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -2063,8 +2063,12 @@ static int ip_gateway_add(int family, int ifindex, void *gw)
 	rt->rtm_dst_len = 0;
 
 	err = -EINVAL;
-	if (nla_put_buffer(nlmsg, RTA_GATEWAY, gw, addrlen))
-		goto out;
+
+	/* If gateway address not supplied, then a device route will be created instead */
+	if (gw != NULL) {
+		if (nla_put_buffer(nlmsg, RTA_GATEWAY, gw, addrlen))
+			goto out;
+	}
 
 	/* Adding the interface index enables the use of link-local
 	 * addresses for the gateway.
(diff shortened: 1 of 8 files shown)


src/lxc/network.c
```diff
@@ -3438,29 +3452,39 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
...
-					      "\"%s\"", buf);
+				err = lxc_ipv6_gateway_add(netdev->ifindex, netdev->ipv6_gateway);
+				if (err) {
+					errno = -err;
+					SYSERROR("Failed to setup ipv6 gateway for network device \"%s\"",
+						ifname);
+
+					if (netdev->ipv6_gateway_auto) {
+						char buf[INET6_ADDRSTRLEN];
+						inet_ntop(AF_INET6, netdev->ipv6_gateway, buf, sizeof(buf));
```
brauner: Is it safe to ignore errors here?

tomponline: @brauner this isn't actually new code, its the existing gateway setup code just showing as new as I moved it into an `else` statement on line https://github.com/lxc/lxc/pull/2973/files#diff-49260cced2f6f9c5617962928313c73aR3463.

To answer your actual question, I think it is safe, because the gateway setting has failed and the container start will fail anyway on line 3486. It is just used to print out the detected gateway IP for more information in the errors.

However it is not ideal to not be checking the error from this function as presumably if it does fail the error message would contain an uninitialised char array, although thats very unlikely to happen.

More of concern for me is that there are several error messages that dont contain the problem IP above here, and only in this specific scenario does it actually log the gateway IP being used.

Further more on line https://github.com/lxc/lxc/blob/a2f9a6706d6363faa1fb6a091679799a1d41f374/src/lxc/network.c#L3466 if that fails it really shouldnt proceed any further as it has already tried to add the gateway and failed.

I can a look at re-structuring this code, would you prefer that done in a separate PR or as part of this? 

Thanks

brauner: Feel free to restructure the code! But yes, please do it in a separate pr.

PR #2977, opened 2019-05-06
raw_syscalls: add initial support for pidfd_send_signal()

brauner:
Well, I added this syscall so we better use it. :)


stgraber:
jenkins: test this please

PR #2979, opened 2019-05-07
network: Makes vlan network interfaces set mtu before upscript called

tomponline:
This is consistent with veth and ipvlan types.

Also makes the debug message for success occur after up script has run.

Also makes device clean up on error more thorough and consistent.


brauner:
jenkins: test this please

PR #2981, opened 2019-05-07
network: Re-works veth gateway logic

tomponline:
Handles more errors and gives better error messages.

### DIFF
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -3377,6 +3377,7 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
 	int err;
 	const char *net_type_name;
 	char *current_ifname = ifname;
+	char bufinet4[INET_ADDRSTRLEN], bufinet6[INET6_ADDRSTRLEN];
 
 	/* empty network namespace */
 	if (!netdev->ifindex) {
@@ -3501,11 +3502,6 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
 		}
 	}
 
-	/* We can only set up the default routes after bringing
-	 * up the interface, since bringing up the interface adds
-	 * the link-local routes and we can't add a default
-	 * route if the gateway is not reachable. */
-
 	/* setup ipv4 gateway on the interface */
 	if (netdev->ipv4_gateway || netdev->ipv4_gateway_dev) {
 		if (!(netdev->flags & IFF_UP)) {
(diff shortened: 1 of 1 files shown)


src/lxc/network.c
```diff
@@ -3529,26 +3525,31 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
 				return minus_one_set_errno(-err);
 			}
 		} else {
+			/* Check the gateway address is valid */
+			if (!inet_ntop(AF_INET, netdev->ipv4_gateway, bufinet4, sizeof(bufinet4)))
+				return minus_one_set_errno(-errno);
```
brauner: Why `-errno` and not `errno`?

src/lxc/network.c
```diff
@@ -3529,26 +3525,31 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
...
+				 */
 				err = lxc_ipv4_dest_add(netdev->ifindex, netdev->ipv4_gateway, 32);
-				if (err) {
+				if (err < 0) {
 					errno = -err;
-					SYSERROR("Failed to add ipv4 dest for network device \"%s\"",
-						ifname);
+					SYSERROR("Failed to add ipv4 dest \"%s\" for network device \"%s\"",
+						bufinet4, ifname);
+					return -1;
```
brauner: Should we do `minus_one_set_errno` here too?

src/lxc/network.c
```diff
@@ -3578,28 +3579,31 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
...
+				 */
 				err = lxc_ipv6_dest_add(netdev->ifindex, netdev->ipv6_gateway, 128);
-				if (err) {
+				if (err < 0) {
 					errno = -err;
-					SYSERROR("Failed to add ipv6 dest for network device \"%s\"",
-						ifname);
+					SYSERROR("Failed to add ipv6 dest \"%s\" for network device \"%s\"",
+						bufinet6, ifname);
+					return -1;
```
brauner: Should we do `minus_one_set_errno` here too?

src/lxc/network.c
```diff
@@ -3529,26 +3525,31 @@ static int lxc_setup_netdev_in_child_namespaces(struct lxc_netdev *netdev)
...
+			if (err < 0) {
+				/* If adding the default route fails, this could be because the
+				 * gateway address is in a different subnet to the container's address.
+				 * To work around this, we try adding a static device route to the
+				 * gateway address first, and then try again.
+				 */
 				err = lxc_ipv4_dest_add(netdev->ifindex, netdev->ipv4_gateway, 32);
-				if (err) {
+				if (err < 0) {
 					errno = -err;
```
tomponline: @brauner in this scenario, I could have remove these `errno = -err` lines and then replaced the `return -1` 2 lines down with `return minus_one_set_errno(-err);`. However that would have resulted in the call to the `SYSERROR` function not having the most recent `errno` set (assuming that the previously called function `lxc_ipv4_dest_add` did not set errno, which is does it some situations and not others).

Alternatively I could have just used `ERROR` so that it wouldnt matter.

src/lxc/network.c
tomponline: @brauner thats a mistake, will fix. Thanks

src/lxc/network.c
brauner: Meh, your call. No strong opinion from my side.

PR #2987, opened 2019-05-09
Pass zero to clone

tych0:
Here's some things I was thinking about on a plane while working on something else :)

### DIFF
--- a/doc/lxc.container.conf.sgml.in
+++ b/doc/lxc.container.conf.sgml.in
@@ -1722,6 +1722,12 @@ dev/null proc/kcore none bind,relative 0 0
             process wants to inherit the other's network namespace it usually
             needs to inherit the user namespace as well.
             </para>
+
+            <para>
+            Note that without careful additional configuration of an LSM,
+            sharing user+pid namespaces with a task may allow that task to
+            escalate privileges to that of the task calling liblxc.
+            </para>
           </listitem>
         </varlistentry>
       </variablelist>
--- a/src/lxc/namespace.c
+++ b/src/lxc/namespace.c
@@ -42,7 +42,7 @@
 
 lxc_log_define(namespace, lxc);
 
-#define __LXC_STACK_SIZE 4096
+#define __LXC_STACK_SIZE (8 * 1024 * 1024)
 pid_t lxc_clone(int (*fn)(void *), void *arg, int flags, int *pidfd)
 {
 	pid_t ret;
--- a/src/lxc/namespace.c
+++ b/src/lxc/namespace.c
@@ -42,25 +42,10 @@
 
 lxc_log_define(namespace, lxc);
 
-struct clone_arg {
-	int (*fn)(void *);
-	void *arg;
-};
-
-static int do_clone(void *arg)
-{
-	struct clone_arg *clone_arg = arg;
-	return clone_arg->fn(clone_arg->arg);
-}
-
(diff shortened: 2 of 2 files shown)


brauner:
That's what I added lxc_raw_clone for and we already use that in a bunch of places. lxc_clone is used for CLONE_VM and we want the stack.

tych0:
On Thu, May 09, 2019 at 11:27:27AM -0700, Christian Brauner wrote:
> That's what I added lxc_raw_clone for and we already use that in a bunch of places. lxc_clone is used for CLONE_VM and we want the stack.

Huh? I don't think you do.

brauner:
> On Thu, May 09, 2019 at 11:27:27AM -0700, Christian Brauner wrote: That's what I added lxc_raw_clone for and we already use that in a bunch of places. lxc_clone is used for CLONE_VM and we want the stack.
> Huh? I don't think you do.

So there's
[`lxc_raw_clone()`](https://github.com/lxc/lxc/blob/master/src/lxc/raw_syscalls.c#L36)
and
[`lxc_raw_clone_cb()`](https://github.com/lxc/lxc/blob/master/src/lxc/raw_syscalls.c#L105)
both rely on COW aka don't pass stack as `NULL`.

The default in `start` is to use `lxc_raw_clone_cb()`:
https://github.com/lxc/lxc/blob/master/src/lxc/start.c#L1766
and when you share namespaces we use `CLONE_VM` and `lxc_clone()` with a separate stack. `lxc_clone()` is used on purpose here
https://github.com/lxc/lxc/blob/master/src/lxc/start.c#L1753
as a `child_fn` it calls [`lxc_raw_clone_cb()`](https://github.com/lxc/lxc/blob/master/src/lxc/start.c#L1619) after it as attached to all namespaces.

In fact `lxc_clone()` is only used in a few places. In all other places it has been replaced by `lxc_raw_clone{_cb}()`. Even `attach()` calls `lxc_raw_clone()`. It's been like that for a long time. :)

tych0:
On Thu, May 09, 2019 at 02:38:55PM -0700, Christian Brauner wrote:
> > On Thu, May 09, 2019 at 11:27:27AM -0700, Christian Brauner wrote: That's what I added lxc_raw_clone for and we already use that in a bunch of places. lxc_clone is used for CLONE_VM and we want the stack.
> > Huh? I don't think you do.
> 
> So there's
> [`lxc_raw_clone()`](https://github.com/lxc/lxc/blob/master/src/lxc/raw_syscalls.c#L36)
> and
> [`lxc_raw_clone_cb()`](https://github.com/lxc/lxc/blob/master/src/lxc/raw_syscalls.c#L105)
> both rely on COW aka don't pass stack as `NULL`.

Right, sorry, so you really do want a heap allocated stack in this
case. But you don't want it on your same stack.

I went ahead and updated the PR and the commit message to reflect
this.

brauner:
Hm, you need to rebase. :) Why are you switching to using heap memory?

tych0:
On Wed, May 15, 2019 at 06:41:50AM -0700, Christian Brauner wrote:
> Hm, you need to rebase. :) Why are you switching to using heap memory?

Sure, just rebased. See the commit message, there's a reasonble
writeup there.

brauner:
Sweet! :)

brauner:
jenkins: test this please

brauner:
jenkins: test this please

PR #2997, opened 2019-05-10
criu: Use -v4 instead of -vvvvvv

rst0git:
CRIU has only 4 [levels of verbosity](https://criu.org/Logging) (errors, warnings, info, debug). Thus, using `-v4` is more appropriate.


brauner:
Thanks!

PR #2998, opened 2019-05-12
Fix returning -1 in functions with return type bool

rikardfalkeborn:
This fixes a couple of issues in error paths when returning -1 when false was intended. Detected by the help of cppcheck.


brauner:
Thanks!

PR #2999, opened 2019-05-12
initutils: Fix memleak on realloc failure

rikardfalkeborn:


rikardfalkeborn:
Oh, this is not right. I'll fix..

brauner:
Thanks!

PR #3001, opened 2019-05-13
Use %m instead of strerror() when available

Rachid-Koucha:
Use %m under HAVE_M_FORMAT instead of strerror()

### DIFF
--- a/src/lxc/log.h
+++ b/src/lxc/log.h
@@ -416,53 +416,94 @@ ATTR_UNUSED static inline void LXC_##LEVEL(struct lxc_log_locinfo* locinfo,	\
 	LXC_FATAL(&locinfo, format, ##__VA_ARGS__);			\
 } while (0)
 
+#if HAVE_M_FORMAT
+#define SYSTRACE(format, ...)                              \
+		TRACE("%m - " format, ##__VA_ARGS__);
+#else
 #define SYSTRACE(format, ...)                              \
 	do {                                               \
 		lxc_log_strerror_r;                        \
 		TRACE("%s - " format, ptr, ##__VA_ARGS__); \
 	} while (0)
+#endif
 
+#if HAVE_M_FORMAT
+#define SYSDEBUG(format, ...)                              \
+                DEBUG("%m - " format, ##__VA_ARGS__)
+#else
 #define SYSDEBUG(format, ...)                              \
 	do {                                               \
 		lxc_log_strerror_r;                        \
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

brauner:
Does `%m` use a thread-safe way of getting at the `errno`?

Rachid-Koucha:
> Does `%m` use a thread-safe way of getting at the `errno`?

This is a good question !

It is done inside GLIBC. I guess it hides a call to strerror() internally. The Linux manual of printf() merely says the following :

m      (Glibc extension; supported by uClibc and musl.)  Print output of strerror(errno).  No argument is required.

And strerror() source code is not thread-safe as it uses a static global malloced internal buffer into which it stores the error string. Here is the source code of strerror in GLIBC 2.29:

/* Return a string describing the errno code in ERRNUM.
   The storage is good only until the next call to strerror.
   Writing to the storage causes undefined behavior.  */
libc_freeres_ptr (static char *buf);

char *
strerror (int errnum)
{
  char *ret = __strerror_r (errnum, NULL, 0);
  int saved_errno;

  if (__glibc_likely (ret != NULL))
    return ret;
  saved_errno = errno;
  if (buf == NULL)
    buf = malloc (1024);
  __set_errno (saved_errno);
  if (buf == NULL)
    return _("Unknown error");
  return __strerror_r (errnum, buf, 1024);
}

So, I would say no : %m is not thread safe :-(

Rachid-Koucha:
There is also a discussion here saying that %m would be thread-safe. So, I am confused...

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=621057

Rachid-Koucha:
I dived into the source code og GLIBC/printf(). And it appears that '%m' is thread-safe as it hides a call to strerror_r() to write into the user buffer (in case of snprintf() or the the internal buffer of GLIBC in case of printf(). Here is the source code (from GLIBC2.29/stdio-common/vfprintf-internal.c):

    LABEL (form_strerror):						      \
      /* Print description of error ERRNO.  */				      \
      string =								      \
	(CHAR_T *) __strerror_r (save_errno, (char *) work_buffer,	      \
				 WORK_BUFFER_SIZE * sizeof (CHAR_T));	      \
      is_long = 0;		/* This is no wide-char string.  */	      \
      goto LABEL (print_string)

So, I would say that %m is thread-safe.

PR #3005, opened 2019-05-15
network: move phys netdevs back to monitor's net ns rather than pid 1's

tomponline:
Updates lxc_restore_phys_nics_to_netns() to move phys netdevs back to the monitor's network namespace rather than the previously hardcoded PID 1 net ns.

This is to fix instances where LXC is started inside a net ns different from PID 1 and physical devices are moved back to a different net ns when the container is shutdown than the net ns than where the container was started from.

### DIFF
--- a/src/lxc/macro.h
+++ b/src/lxc/macro.h
@@ -261,6 +261,10 @@ extern int __build_bug_on_failed;
 #define IFLA_NET_NS_PID 19
 #endif
 
+#ifndef IFLA_NET_NS_FD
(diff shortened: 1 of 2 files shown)


tomponline:
@brauner this should fix the specific issue I was having with the MTU tests over at https://github.com/lxc/lxd/pull/5745

I'll check other uses of that function and see if any more are appropriate to be changed too.

tomponline:
@brauner no other instances of `lxc_netdev_move_by_*` supply a pid of 1, so this should suffice as is.

PR #3025, opened 2019-06-04
lxc_usernsexec: continuing after unshare fails leads to confusing and…

xmcqueen:
… misleading error messages


xmcqueen:
That was a ridiculous effort for a three line commit! wow.  I'm glad its setup now. I think its right.  Getting the gpg key thing going right on this osx was not easy to figure out.

brauner:
Ah, I think you misunderstood me. :) All we need is a DCO line at the end of your commit message like this:

Signed-off-by: Firstname Lastname <valid.email@address.com>

brauner:
You can achieve this by doing:

got commit -s

brauner:
The gpg part is useful too though so thanks for going through that hazzle. :)

xmcqueen:
this is merged from a different PR

PR #3034, opened 2019-06-05
start: generate new boot id on container start

### LINKED ISSUE #3027: RFE: provide /proc/sys/kernel/random/boot_id
As originally filed in ubuntu at [bug 1831258](https://bugs.launchpad.net/bugs/1831258) `journalctl --list-boots` identifies boots by reading `/proc/sys/kernel/random/boot_id`.  Without doing this, the journal can only detect a reboot if the *host* has rebooted.

Without this ability to detect boots, functionality like `journalctl --boot=X` will also not work.

nspawn seems to do something similar in [nspawn/nspawn.c:setup_boot_id](https://github.com/systemd/systemd/blob/master/src/nspawn/nspawn.c#L1841)

brauner:
Closes #3027.

BugLink: https://bugs.launchpad.net/bugs/1831258
Cc: Dimitri John Ledkov <xnox@ubuntu.com>
Cc: Scott Moser <smoser@ubuntu.com>

### DIFF
--- a/src/lxc/Makefile.am
+++ b/src/lxc/Makefile.am
@@ -45,7 +45,8 @@ noinst_HEADERS = api_extensions.h \
 		 ../tests/lxctest.h \
 		 tools/arguments.h \
 		 storage/storage_utils.h \
-		 utils.h
+		 utils.h \
+		 uuid.h
 
 if IS_BIONIC
 noinst_HEADERS += ../include/fexecve.h \
@@ -143,6 +144,7 @@ liblxc_la_SOURCES = af_unix.c af_unix.h \
 		    syscall_wrappers.h \
 		    terminal.c \
 		    utils.c utils.h \
+		    uuid.c uuid.h \
 		    version.h \
 		    $(LSM_SOURCES)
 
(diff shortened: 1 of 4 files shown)


xnox:
Why implement uuid from scratch, instead of using libuuid1 ?

xnox:
I noticed that in a comple of places, upon attach, one can remount /proc. Will this id continue to be accessible in those cases?

brauner:
> Why implement uuid from scratch, instead of using libuuid1 ?

We don't link against stuff whenever we can avoid it especially if its something like this.

brauner:
> I noticed that in a comple of places, upon attach, one can remount /proc. Will this id continue to be accessible in those cases?

With an explicit request only and LXD doesn't allow you to do that. A lot of things would not be available in this case including all lxcfs-based mocking.

xnox:
By libuuid1 I mean to use util-linux's uuid.h header/lib and call `uuid_generate_random`  https://github.com/karelzak/util-linux/blob/master/libuuid/src/uuid.h#L91

xnox:
> > Why implement uuid from scratch, instead of using libuuid1 ?
> 
> We don't link against stuff whenever we can avoid it especially if its something like this.

ok.

PR #3045, opened 2019-06-14
[RFC\ Switch from gnutls to openssl for sha1

hallyn:
The reason for this is because openssl can be statically linked
against, gnutls cannot.

[ commit msg addendum ] I don't know that openssl is what we want.  Maybe we just want to hardcode our own.  But at least openssl is widely avaiable, and available as .a's.  Also, I guess lxc.pc will also need to be updated if we do this.

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -330,17 +330,30 @@ again:
 	return status;
 }
 
-#if HAVE_LIBGNUTLS
-#include <gnutls/gnutls.h>
-#include <gnutls/crypto.h>
+#ifdef HAVE_OPENSSL
+#include <openssl/evp.h>
 
-__attribute__((constructor))
-static void gnutls_lxc_init(void)
+static int do_sha1_hash(const char *buf, int buflen, unsigned char *md_value, int *md_len)
 {
-	gnutls_global_init();
+	EVP_MD_CTX *mdctx;
+	const EVP_MD *md;
+
+	md = EVP_get_digestbyname("sha1");
+	if(!md) {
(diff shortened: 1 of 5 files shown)


src/lxc/utils.c
```diff
@@ -330,17 +330,30 @@ int lxc_wait_for_pid_status(pid_t pid)
...
 
-__attribute__((constructor))
-static void gnutls_lxc_init(void)
+static int do_sha1_hash(const char *buf, int buflen, unsigned char *md_value, int *md_len)
 {
-	gnutls_global_init();
+	EVP_MD_CTX *mdctx;
+	const EVP_MD *md;
+
+	md = EVP_get_digestbyname("sha1");
```
Blub: What about using `EVP_sha1()` directly? Its code is AFAIK literally just `return &sha1_md;` so we'd get rid of the error case (which is the main thing which bothers me in gnutls, btw. - many function can't actually really fail, except gnutls has a global "the world has ended" error state...)

hallyn: Hi,

I was going by https://www.openssl.org/docs/man1.1.0/man3/EVP_get_digestbyname.html  which recommended this route.  I don't mind either way.  We might I suppose one day want to make the algorithm configurable (if we ever start using this function for something more security related).

I don't mind either way.

hallyn:
Ok I've only done rudementary testing so far, so probably shouldn't merge it just yet.

hallyn:
> Ok I've only done rudementary testing so far, so probably shouldn't merge it just yet.

Well, seems all right here.

Issue #3066, opened 2019-06-29
Can LXC_NET_NS be exported when net hook being executed?


ss1h2a3tw:
# Issue description

Since the net hook script may need to modify the network configs in the lxc's network namespace. Exporting this variable will be helpful.

stgraber:
Seems reasonable to me, can you send a PR that adds the environment variable?

ss1h2a3tw:
I found that the  `LXC_[NAMESPACE IDENTIFIER]_NS` are not set even in the start-host hook. Is it intentional?

ss1h2a3tw:
But if `LXC_PID` are set, maybe this variable won't be needed. Because during starting the container, the script can get the namespace by `/proc/PID/ns`. Maybe moving the setenv of the `LXC_PID` forward will be a easier approach.

stgraber:
@brauner @tomponline I'm guessing you two will have a stronger opinion than I do

tomponline:
@brauner has said in the past that LXC_PID can.be used for this as @ss1h2a3tw says.

ss1h2a3tw:
I found that the time that the hook being executed will differ by whether the lxc is running as root. If it is running as root, the hook will be executed before clone(). If not running as root, the hook will be executed after clone(). So `LXC_PID` or `LXC_NET_NS` will be only available for the hook if the user is not root.

brauner:
For containers started as root neither LXC_PID nor LXC_NET_NS can be made available since it is done before the container is started. But I'm sure we could redesign this.

PR #3069, opened 2019-07-01
start: unify network creation

brauner:
Make sure that network creation happens at the same time for containers started
by privileged and unprivileged users. The only reason we didn't do this so far
was to avoid sending network device ifindices around in the privileged case.

Link: https://github.com/lxc/lxc/issues/3066


stgraber:
jenkins not happy

brauner:
> jenkins not happy

That's why it's labeled with "incomplete". :)

brauner:
@stgraber

PR #3080, opened 2019-07-05
Seccomp notify api update

Blub:
General description (from the `seccomp: update notify api` commit:)
```
The previous API doesn't reflect the fact that
`seccomp_notif` and `seccomp_notif_resp` are allocatd
dynamically with sizes figured out at runtime.

We now query the sizes via the seccomp(2) syscall and change
`struct seccomp_notify_proxy_msg` to contain the sizes
instead of the data, with the data following afterwards.

Additionally it did not provide a convenient way to identify
the container the message originated from, for which we now
include a cookie configured via `lxc.seccomp.notify.cookie`.

Also verify the `id` in the client's response.
```

Another big change is the switch from `SOCK_STREAM` to `SOCK_SEQPACKET`. I think it makes sense: the seccomp notify kernel api is basically packet based as well.


Blub:
^ Changes: Added a compile fix (`HAVE_STRUCT_SECCOMP_NOTIF_SIZES`) and the `MSG_TRUNC` commit.

PR #3081, opened 2019-07-05
start: call lxc_find_gateway_addresses early

### LINKED ISSUE #3078: lxc.net.x.ipv4.gateway = auto no longer creates route inside container
The template below is mostly useful for bug reports and support questions.
Feel free to remove anything which doesn't apply to you and add more information where it makes sense.

# Required information

 * Distribution: Ubuntu
 * Distribution version: 18.04

# Issue description

Since commit e389f2afd8509b0996bf83bf99303eea88a590f7, and still occurring in master, the `lxc.net.x.ipv4.gateway=auto` setting no longer works.

Before that, this setting, when coupled with the `lxc.net.x.ipv4.link` setting would take the IP of the link device and add a default route inside the container to that address.

# Steps to reproduce

 1. Step one

Example config:
In this case lxdbr0 interface has IP 10.111.199.1
```
lxc.net.0.type = veth
lxc.net.0.flags = up
lxc.net.0.ipv4.address = 192.168.1.200/32
lxc.net.0.ipv4.gateway = auto
lxc.net.0.link = lxdbr0
```

 2. Step two
```
lxc-start ct1
lxc-attach ct1
```
 3. Step three
```
ip r
```
You would expect to see a route to 192.168.1.1 as default gateway inside container.

brauner:
This restores the lxc.net.x.ipv4.gateway = auto and
lxc.net.x.ipv6.gateway = auto functionality.

When the child is created the parent and child have different views of
struct lxc_handler since - obviously - virtual memory is duplicated. So any
changes to done by the parent that the child should see need to be IPCed to it.
For any non-actual device creation stuff this does not make much sense. This
includes finding gateway addresses. Move it back prior to clone().

Fixes #3078

[christian.brauner@ubuntu.com: non-functional changes and update commit message]

### DIFF
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -2473,8 +2473,7 @@ int lxc_find_gateway_addresses(struct lxc_handler *handler)
 			continue;
 
 		if (netdev->type != LXC_NET_VETH && netdev->type != LXC_NET_MACVLAN) {
-			ERROR("Automatic gateway detection is only supported "
-			      "for veth and macvlan");
+			ERROR("Automatic gateway detection is only supported for veth and macvlan");
 			return -1;
 		}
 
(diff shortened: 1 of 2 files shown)


tomponline:
@brauner gcc tests showing as failed but not sure if thats just some keyserver issue (there are some errors about that way down).

brauner:
Yeah, flaky gpg.

Issue #3082, opened 2019-05-22
Bad error hanling in do_lxcapi_mount can result do dataloss


tchwpkgorg:
Ubuntu 19.04, lxd installed from snap and initialized as follows:
```
root@tomasz:~# snap install lxd
lxd 3.13 from Canonical✓ installed
root@tomasz:~# lxd init
Would you like to use LXD clustering? (yes/no) [default=no]: 
Do you want to configure a new storage pool? (yes/no) [default=yes]: 
Name of the new storage pool [default=default]: 
Name of the storage backend to use (btrfs, ceph, dir, lvm) [default=btrfs]: dir
Would you like to connect to a MAAS server? (yes/no) [default=no]: 
Would you like to create a new local network bridge? (yes/no) [default=yes]: 
What should the new bridge be called? [default=lxdbr0]: 
What IPv4 address should be used? (CIDR subnet notation, “auto” or “none”) [default=auto]: 
What IPv6 address should be used? (CIDR subnet notation, “auto” or “none”) [default=auto]: 
Would you like LXD to be available over the network? (yes/no) [default=no]: 
Would you like stale cached images to be updated automatically? (yes/no) [default=yes] 
Would you like a YAML "lxd init" preseed to be printed? (yes/no) [default=no]: 

```

It's now possible to start / delete containers as follows, which is expected:

```
# lxc launch images:ubuntu/bionic/amd64 testct
To start your first container, try: lxc launch ubuntu:18.04

Creating testct
Starting testct                             

# lxc delete testct --force

# lxc launch images:ubuntu/bionic/amd64 testct
Creating testct
Starting testct

# lxc delete testct --force
```

Now, let's launch a new container and try to add a disk to it - this fails, which is unexpected:

```
root@tomasz:~# lxc launch images:ubuntu/bionic/amd64 testct
Creating testct
Starting testct

# lxc config device add testct TDv2 disk source=/srv path=/srv
Error: Add disk devices: Failed to add mount for device: Failed to run: /snap/lxd/current/bin/lxd forkmount lxc-mount testct /var/snap/lxd/common/lxd/containers /var/snap/lxd/common/lxd/logs/testct/lxc.conf /var/snap/lxd/common/lxd/devices/testct/disk.TDv2.srv /srv none 4096: 
```

Let's delete the container and launch it again - this fails, which is again unexpected:

```
# lxc delete testct --force

# lxc launch images:ubuntu/bionic/amd64 testct
Creating testct
Starting testct
[...]

PR #3083, opened 2019-07-07
lxccontainer: properly cleanup on mount injection failure

brauner:
Closes: #3082
Reported-by: Stéphane Graber <stgraber@ubuntu.com>

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -5117,33 +5117,37 @@ static int do_lxcapi_mount(struct lxc_container *c, const char *source,
 
 		suff = strrchr(template, '/');
 		if (!suff)
-			_exit(EXIT_FAILURE);
+			goto cleanup_target_in_child;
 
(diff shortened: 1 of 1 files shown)


src/lxc/lxccontainer.c
```diff
@@ -5117,33 +5117,39 @@ static int do_lxcapi_mount(struct lxc_container *c, const char *source,
...
 		TRACE("Moved mount from \"%s\" to \"%s\"", path, target);
 
 		_exit(EXIT_SUCCESS);
+
+	cleanup_target_in_child:
+		(void)remove(target);
+		_exit(EXIT_FAILURE);
 	}
 
 	ret = wait_for_pid(pid);
```
stgraber: Doesn't the child returning a failure cause ret here to < 0 making us jump to out without going through umount?

In the issue I tracked down, it looked like the umount2() line was never called and I don't see what part of your commit makes this change.

brauner: I forgot to rmove the `out`.

brauner: done

PR #3090, opened 2019-07-11
Suppress hardcoded table sizes

Rachid-Koucha:
. Use sizeof() instead of hardcoded values
. snprintf(..., size, ""...) is in error if the return code is >= size (not sufficient to set only ">")


brauner:
jenkins: test this please

brauner:
Thanks.

PR #3092, opened 2019-07-16
seccomp: open memfd read-write

Blub:
Makes it easier to implement syscalls which need to write to
a buffer passed by user space as a pointer.


brauner:
Thanks!

PR #3097, opened 2019-07-22
suppress false-negative error in templates and nvidia hook

lpirl:
When creating unprivileged containers I came across the following error:

> $ lxc-create … -t download … -- --dist debian …
> /proc/1/uid_map: No such file or directory

The cause seems to be that ``/proc`` is mounted with ``hidepid=2`` which makes ``/proc/1/…`` appear absent for non-root users.

Since the script works correctly despite the error, this error message might be confusing for users.

I am not into LXC, so please check carefully and let me know if I can help further in fixing this.

Thanks for LXC and considering this PR.

### DIFF
--- a/hooks/nvidia
+++ b/hooks/nvidia
@@ -58,8 +58,12 @@ in_userns() {
         echo $fields | grep -q " 0 1$" && { echo userns-root; return; } || true
     done < /proc/self/uid_map
 
-    [ "$(cat /proc/self/uid_map)" = "$(cat /proc/1/uid_map)" ] && \
-        { echo userns-root; return; }
+    if [ -e /proc/1/uid_map ]; then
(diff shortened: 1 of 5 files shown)


brauner:
//Cc @flx42, @3XX0 :)

lpirl:
We could also add a comment to the outermost ``if`` statements mentioning the ``hidepid`` mount option of ``proc``. Wasn't sure if this would be too much bloat.

hallyn:
Hm, I suppose in that case we might instead want to do the less precise check of seeing whether /proc/self/uid_map has the full uid range or not.  It's *possible* that you'd be in a non-init userns with full uid range (to drop capabilities against the init_user_ns , i.e. devices, etc), but I don't know of anyone doing that.

lpirl:
Thanks for your response @hallyn.
Unfortunately, I know too little to adapt the changes.
If you tell me what to check exactly, I'd be willing to give it a shot.

PR #3106, opened 2019-07-26
network: restore ability to move nl80211 devices

### LINKED ISSUE #3105: After upgrade: Failed to move network device “wlan0” to network namespace: Operation not permitted
# Required information

 * Distribution: Ubuntu
 * Distribution version: 18.04
   * Kernel version: 4.18.0-25-generic
   * LXD version: 3.15
   * Storage backend in use: btrfs

# Issue description

After lxd snap auto-refresh, the problem occurred.

```
lxc firewall 20190725200021.783 ERROR    start - start.c:proc_pidfd_open:1644 - Function not implemented - Failed to send signal through pidfd
lxc firewall 20190725200021.900 ERROR    network - network.c:lxc_network_move_created_netdev_priv:3292 - Operation not permitted - Failed to move network device "wlp3s0" with ifindex 3 to network namespace 25846
lxc firewall 20190725200021.900 ERROR    start - start.c:lxc_spawn:1846 - Failed to create the network
lxc firewall 20190725200021.900 WARN     network - network.c:lxc_delete_network_priv:3374 - Failed to rename interface with index 2 from "enp2s0" to its initial name "enp2s0"
lxc firewall 20190725200021.960 ERROR    lxccontainer - lxccontainer.c:wait_on_daemonized_start:873 - Received container state "ABORTING" instead of "RUNNING"
lxc firewall 20190725200021.961 ERROR    start - start.c:__lxc_start:2036 - Failed to spawn container "firewall"
lxc 20190725200022.408 WARN     commands - commands.c:lxc_cmd_rsp_recv:135 - Connection reset by peer - Failed to receive response for command "get_state"
```

The only solution was to revert.

```
# snap changes
ID   Status  Spawn               Ready               Summary
18   Done    today at 14:45 -03  today at 14:45 -03  Auto-refresh snap "lxd"
19   Done    today at 17:15 -03  today at 17:15 -03  Revert "lxd" snap
#

### PR DESCRIPTION
**Pull request by brauner**
Closes #3105.

### DIFF
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -1248,22 +1248,21 @@ static int lxc_netdev_rename_by_name_in_netns(pid_t pid, const char *old,
 static int lxc_netdev_move_wlan(char *physname, const char *ifname, pid_t pid,
 				const char *newname)
 {
-	char *cmd;
+	__do_free char *cmd = NULL;
 	pid_t fpid;
-	int err = -1;
 
 	/* Move phyN into the container.  TODO - do this using netlink.
 	 * However, IIUC this involves a bit more complicated work to talk to
 	 * the 80211 module, so for now just call out to iw.
 	 */
 	cmd = on_path("iw", NULL);
-	if (!cmd)
-		goto out1;
-	free(cmd);
+	if (!cmd) {
+		return -1;
+	}
 
 	fpid = fork();
 	if (fpid < 0)
-		goto out1;
+		return -1;
 
 	if (fpid == 0) {
 		char pidstr[30];
@@ -1274,21 +1273,18 @@ static int lxc_netdev_move_wlan(char *physname, const char *ifname, pid_t pid,
 	}
 
 	if (wait_for_pid(fpid))
-		goto out1;
+		return -1;
 
-	err = 0;
 	if (newname)
-		err = lxc_netdev_rename_by_name_in_netns(pid, ifname, newname);
+		return lxc_netdev_rename_by_name_in_netns(pid, ifname, newname);
 
-out1:
-	free(physname);
-	return err;
+	return 0;
 }
 
 int lxc_netdev_move_by_name(const char *ifname, pid_t pid, const char* newname)
 {
+	__do_free char *physname = NULL;
 	int index;
-	char *physname;
 
 	if (!ifname)
 		return -EINVAL;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by tomponline** [comment]
@brauner looks good. Once this is merged I can add a test to LXD using this module to test for it working https://wireless.wiki.kernel.org/en/users/Drivers/mac80211_hwsim.

I've used that module this morning on ubuntu 18.04 to re-create the issue and check this PR fixes it.

I've also noticed that LXD has a bug related to detaching it (both when a container shutsdown and when its live unplugged), which I'll take a look at fixing in the device PR https://github.com/lxc/lxd/pull/5900

**Comment by tomponline** [comment]
@brauner ah, actually I've found a bug in this (that was also present in LXC 3.1.0).

If you use a config such as:

```
lxc.net.0.type = phys
lxc.net.0.link = wlan0
lxc.net.0.flags = up
lxc.net.0.name = eth0
```

When the container starts, it successfully gets moved into the container and renamed to eth0.

However when the container is stopped it gets removed from the container but re-appears back on the host as eth0, rather than the name it was when it was attached.

brauner:
Yeah, I observed something similar too.

tomponline:
@brauner I think on the LXD side, I need to add something similar to forkdetach that detects a wifi device and removes its physical device back instead using iw command. This should fix the live detach not working.

tomponline:
@brauner oh actually, that is not run until container stop, its ``cc.DetachInterface()`` that is failing, which is go-lxc calling into liblxc, so maybe this PR could address that issue too?

marcialvieira:
> @brauner ah, actually I've found a bug in this (that was also present in LXC 3.1.0).
> 
> If you use a config such as:
> 
> ```
> lxc.net.0.type = phys
> lxc.net.0.link = wlan0
> lxc.net.0.flags = up
> lxc.net.0.name = eth0
> ```
> 
> When the container starts, it successfully gets moved into the container and renamed to eth0.
> 
> However when the container is stopped it gets removed from the container but re-appears back on the host as eth0, rather than the name it was when it was attached.

I also had this problem, so I'm using the same name on both of them to avoid this, thanks for fixing both problems, I really appreciate the great work you guys are doing.

tomponline:
@stgraber this fixes the immediate issue, there are other issues left to fix, but they were already present in lxc 3.1.0.

PR #3107, opened 2019-07-26
lxccontainer: do_lxcapi_detach_interface to support detaching wlan devs

tomponline:
lxccontainer: do_lxcapi_detach_interface to support detaching wlan devices.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -4793,6 +4793,7 @@ static bool do_lxcapi_detach_interface(struct lxc_container *c,
 {
 	int ret;
 	pid_t pid, pid_outside;
+	__do_free char *physname = NULL;
 
 	/*
 	 * TODO - if this is a physical device, then we need am_host_unpriv.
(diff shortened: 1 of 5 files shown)


src/lxc/lxccontainer.c
```diff
@@ -4843,7 +4844,14 @@ static bool do_lxcapi_detach_interface(struct lxc_container *c,
 			}
 		}
 
-		ret = lxc_netdev_move_by_name(ifname, pid_outside, dst_ifname);
+		/* remount /sys so is_wlan() can check if this device is a wlan device. */
+		lxc_attach_remount_sys_proc();
```
tomponline: @brauner is this OK to do or is it a bad idea?

brauner: You can't do this. Not without a new mount namespace. What you're dong right now is re-mounting proc and sys on the host.

tomponline: @brauner I've updated this to make a call to `unshare(CLONE_NEWNS);` before remounting /sys but after the fork. Thanks

brauner: You also need to do:
```c
	if (mount(NULL, "/", NULL, MS_REC | MS_PRIVATE, 0) < 0) {
```
to prevent mount propagation from getting in your way.

tomponline: @brauner thanks for the advice, I've added now so that the / path inside the mount namespace is recursively set as a private mount before /sys is remounted, so that any existing mount propagation rules can't take effect inside the new mount namespace.

brauner: Excellent!

PR #3117, opened 2019-08-10
[aa-profile] Deny access to /proc/acpi/**

### LINKED ISSUE #3115: [Debian][aa profiles] Also deny /proc/acpi to the containers
Hi,

It occurred to me via bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=906805 on Debian that LXC might be vulnerable to CVE-2018-10892.

I gave a look at the apparmor profiles and the source code, and nothing seems to mask /proc/acpi from the container when it's not an unprivileged one.

Did I miss something? Otherwise maybe it'd be a good idea to prevent the access to /proc/acpi via apparmor, or, better, via the core code of LXC with, potentially, a config parameter to allow the access to /proc/acpi?

With best regards <3

P-EB:
Would close #3115


stgraber:
Hi,

Per CONTRIBUTING, all commits need to include a Signed-off-by line.

Can you update your commit then we'll review it and land it.

Thanks!

P-EB:
@stgraber ok, I'll do that.

PR #3119, opened 2019-08-13
attach: don't close stdout of getent

Blub:
See the first context line, we want to replace stderr, as stdout is the pipe from which we want to read the result.


brauner:
Grazie mille.

Issue #781, opened 2016-01-29
why /dev is limited to 100k, and how to increase or change it?


mingewang:
see the man page:
       If  lxc.autodev is set to 1, then after mounting the container's rootfs LXC
       will mount a fresh tmpfs under /dev (limited to 100k)  and  fill  in  a
       minimal  set  of  initial  devices.   

the question is: why /dev is limited to 100K, and is there anyway to increase it? where is the code to set this limit.

I have a box which has /dev 100%. ( due to 100K limit)

thanks.

stgraber:
On Fri, Jan 29, 2016 at 06:57:40AM -0800, Min Wang wrote:
> see the man page:
>        If  lxcautodev is set to 1, then after mounting the container's rootfs LXC
>        will mount a fresh tmpfs under /dev (limited to 100k)  and  fill  in  a
>        minimal  set  of  initial  devices   
> 
> the question is: why /dev is limited to 100K, and is there anyway to increase it? where is the code to set this limit
> 
> I have a box which has /dev 100%
> 
> 
> thanks

./lxc/conf.c:   if (safe_mount("none", path, "tmpfs", 0, "size=100000,mode=755",

A patch to bump it to say 500k would be welcome.

Stéphane

mingewang:
@stgraber , thanks. I submitted a patch and created a pull request, please review. thanks.

stgraber:
Merged, closing this issue.

PR #3127, opened 2019-08-27
Add devfs_size config parameter

Napsty:
This PR wants to solve an issue with a limited size of the `/dev` tmpfs, which already happened a while ago in https://github.com/lxc/lxc/issues/781. Back then the size of the `/dev` tmpfs was increased but with a fixed size. 
Under some rare but real circumstances, the `/dev` tmpfs can run out of space. Here from a real container in production:

```
root@irnsrvc16 ~ # df -h /dev
Filesystem     Type   Size  Used Avail Use% Mounted on
none           tmpfs  492K  492K     0 100% /dev
```

The reason in this case is that this specific container is used as central syslog server (using syslog-ng) for a couple dozen other LXC containers. syslog-ng uses its console logging to /dev/tty10, which just uses up the whole space

```
root@irnsrvc16 ~ # ll /dev/tt*
crw-rw-rw- 1 root root   5, 0 Aug 21  2018 /dev/tty
crw--w---- 1 root tty  136, 0 Aug 21  2018 /dev/tty1
-rw-r----- 1 root adm  503808 Aug 23 09:21 /dev/tty10
crw--w---- 1 root tty  136, 1 Aug 21  2018 /dev/tty2
crw--w---- 1 root tty  136, 2 Aug 21  2018 /dev/tty3
crw--w---- 1 root tty  136, 3 Aug 21  2018 /dev/tty4
```

This is just one practical example. There may be other uses cases where `/dev` may run out of space.

Anyways, making the size configurable is a win for all. 

Let me know if the PR is fine or if the parameter name needs to change.

### DIFF
--- a/doc/lxc.container.conf.sgml.in
+++ b/doc/lxc.container.conf.sgml.in
@@ -1064,9 +1064,10 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
         <filename>/dev</filename> to be set up as needed in the container
         rootfs.  If lxc.autodev is set to 1, then after mounting the container's
         rootfs LXC will mount a fresh tmpfs under <filename>/dev</filename>
-        (limited to 500k) and fill in a minimal set of initial devices.
+        (limited to 500k by default, unless defined in lxc.devfs_size) and
+        fill in a minimal set of initial devices.
         This is generally required when starting a container containing
-        a "systemd" based "init" but may be optional at other times.  Additional
+        a "systemd" based "init" but may be optional at other times. Additional
         devices in the containers /dev directory may be created through the
         use of the <option>lxc.hook.autodev</option> hook.
       </para>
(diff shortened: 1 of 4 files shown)


Napsty:
> All commits need to be signed-off (see CONTRIBUTING) before we can review them.

sorry, I misunderstood the signed-off part. Will add `git commit --signoff ...` to the PR. That should be it or do I need to make additional changes? 

> That limit was put in place, specifically for preventing a full system denial of service should such a piece of broken software exist :)

I fully agree with you. But the default limit of 500K still stays for security and if one admin really must increase the `/dev`size (for whatever reason really, https://github.com/lxc/lxc/issues/781 didn't mention a reason) a new limit is set. Giving the admin at least the possibility to make the size configurable. 

A temporary workaround is to remount the `/dev` tmpfs with a new size value inside the container, but of course after a restart of the container, the original size kicks in again.

Napsty:
Closed for cleanup and favour of #3128

PR #3128, opened 2019-08-28
Add config parameter for /dev tmpfs size

Napsty:
Second try's a charm. Sorry. Closed the other PR in favour for this PR. 

This PR wants to solve an issue with a limited size of the `/dev` tmpfs, which already happened a while ago in https://github.com/lxc/lxc/issues/781. Back then the size of the `/dev` tmpfs was increased but with a fixed size. 
Under some rare but real circumstances, the `/dev` tmpfs can run out of space. Here from a real container in production:

```
root@irnsrvc16 ~ # df -h /dev
Filesystem     Type   Size  Used Avail Use% Mounted on
none           tmpfs  492K  492K     0 100% /dev
```

The reason in this case is that this specific container is used as central syslog server (using syslog-ng) for a couple dozen other LXC containers. syslog-ng uses its console logging to /dev/tty10, which just uses up the whole space

```
root@irnsrvc16 ~ # ll /dev/tt*
crw-rw-rw- 1 root root   5, 0 Aug 21  2018 /dev/tty
crw--w---- 1 root tty  136, 0 Aug 21  2018 /dev/tty1
-rw-r----- 1 root adm  503808 Aug 23 09:21 /dev/tty10
crw--w---- 1 root tty  136, 1 Aug 21  2018 /dev/tty2
crw--w---- 1 root tty  136, 2 Aug 21  2018 /dev/tty3
crw--w---- 1 root tty  136, 3 Aug 21  2018 /dev/tty4
```

This is just one practical example. There may be other uses cases where `/dev` may run out of space.

Important: The current default of 500K as tmpfs size will stay unless `lxc.devfs_size` is used to overwrite that value. This way security to protect against a misbehaving software spamming `/dev` is in place, yet the admin has the freedom to increase (or decrease) the default size for his own reasons.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1135,18 +1135,21 @@ on_error:
  * error, log it but don't fail yet.
  */
 static int mount_autodev(const char *name, const struct lxc_rootfs *rootfs,
-			 const char *lxcpath)
+			 int autodevtmpfssize, const char *lxcpath)
 {
 	__do_free char *path = NULL;
 	int ret;
 	size_t clen;
 	mode_t cur_mask;
+        char mount_options[128];
 
 	INFO("Preparing \"/dev\"");
 
 	/* $(rootfs->mount) + "/dev/pts" + '\0' */
 	clen = (rootfs->path ? strlen(rootfs->mount) : 0) + 9;
 	path = must_realloc(NULL, clen);
+	sprintf(mount_options, "size=%d,mode=755", (autodevtmpfssize != 0) ? autodevtmpfssize : 500000);
+	DEBUG("Using mount options: %s", mount_options);
 
 	ret = snprintf(path, clen, "%s/dev", rootfs->path ? rootfs->mount : "");
 	if (ret < 0 || (size_t)ret >= clen)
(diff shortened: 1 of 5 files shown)


Napsty:
Yep, the name is not fixed in stone (@brauner you understand this ;-) ), we can vote on the parameter name.

src/lxc/conf.c
```diff
@@ -1135,18 +1135,21 @@ static int lxc_create_ttys(struct lxc_handler *handler)
...
 	mode_t cur_mask;
+        char mount_options[128];
 
 	INFO("Preparing \"/dev\"");
 
 	/* $(rootfs->mount) + "/dev/pts" + '\0' */
 	clen = (rootfs->path ? strlen(rootfs->mount) : 0) + 9;
 	path = must_realloc(NULL, clen);
+	sprintf(mount_options, "size=%d,mode=755", (devfs_size != 0) ? devfs_size : 500000);
+	INFO("Using mount options: %s", mount_options);
```
Napsty: INFO or TRACE  here?

brauner: `DEBUG`, I think.

hallyn:
lgtm, but please squash the two commits.

Issue #3139, opened 2019-09-14
Build failure due to OpenSSL


h1z1:
* Distribution: CentOS
 * Distribution version: 7

# Issue description

Current git head fails to build as a result of fa2bb6ba532c5e7f92df8cbae50a68af519f9997 

# Steps to reproduce

 1. Step one
Clone lxc

 2. Step two
autogen.sh
./configure --enable-openssl

 3. Step three
make

# Information to attach

```
utils.c: In function 'do_sha1_hash':
utils.c:347:2: error: implicit declaration of function 'EVP_MD_CTX_new' [-Werror=implicit-function-declaration]
  mdctx = EVP_MD_CTX_new();
  ^
utils.c:347:2: error: nested extern declaration of 'EVP_MD_CTX_new' [-Werror=nested-externs]
utils.c:347:8: error: assignment makes pointer from integer without a cast [-Werror]
  mdctx = EVP_MD_CTX_new();
        ^
utils.c:351:2: error: implicit declaration of function 'EVP_MD_CTX_free' [-Werror=implicit-function-declaration]
  EVP_MD_CTX_free(mdctx);
  ^
utils.c:351:2: error: nested extern declaration of 'EVP_MD_CTX_free' [-Werror=nested-externs]
cc1: all warnings being treated as errors
```

EVP_MD_CTX_new and EVP_MD_CTX_free were introduced in OpenSSL_1_1_0-pre1 ~2015.  Given the number of people who got bit by this I can only imagine the thought process behind it.

CentOS 7 ships with 1.0.2k-16

PR #3140, opened 2019-09-16
Ensures OpenSSL compatibility with older versions of EVP API.

caioboffo:
Signed-off-by: Caio B. Silva <caioboffo@gmail.com>

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -344,11 +344,11 @@ static int do_sha1_hash(const char *buf, int buflen, unsigned char *md_value, in
 		return -1;
 	}
 
-	mdctx = EVP_MD_CTX_new();
+	mdctx = EVP_MD_CTX_create();
 	EVP_DigestInit_ex(mdctx, md, NULL);
 	EVP_DigestUpdate(mdctx, buf, buflen);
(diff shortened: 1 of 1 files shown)


brauner:
Can you explain this change in a little more detail, please. :)

brauner:
jenkins: test this please

caioboffo:
> Can you explain this change in a little more detail, please. :)

Sure, sorry about that . This is fix for Issue #3139. OpenSSL Version <= 1_1_0 Does not have the EVP_MD_CTX_(new|free) functions. In order to maintain compatibility with OpenSSL 1_0_2k, for example, the do_sha1_hash function in util.c needs to call the EVP_MD_CTX_(create|delete) functions.

brauner:
Ah, ok. Can we detect this through autoconf which functions are supported?

caioboffo:
If I'm not mistaken we can detect lib version and also functions presence with Autoconf. But I don't see the point in doing this. Unless we need to ensure that OpenSSL versions greater or equal than 1.1.0. Public APIs aren't supposed to change drastically from version to version.

brauner:
jenkins: test this please

Issue #3035, opened 2019-06-06
pam_cgfs should have man page


AkihiroSuda:

brauner:
Yes, agreed. Someone care to come up with one?

adamRogerson:
@brauner I will give this a go.

CameronNemo:
Is there any documentation currently?

avhvr:
Hi,

I'd like to contribute to this issue. Is anyone working on it currently?

brauner:
@avhr feel free to give it a try

AkihiroSuda:
Thanks @avhvr

PR #3141, opened 2019-09-17
doc: add man page for pam_cgfs

avhvr:
#3035
Signed-off-by: Venkata Harshavardhan Reddy Allu <venkataharshavardhan_ven@srmuniv.edu.in>


stgraber:
I believe you need to add this to configure.ac at least

PR #3142, opened 2019-09-18
cmd: fix usernet quota handling

Piscolero:
When comparing the usernet quota definitions with the usernic database
only the owner is evaluated and the nicname is ignored.

The following example leads to 'Quota reached' error although it shouldn't.

  lxc-usernet:
    user veth bridge1 1
    user veth bridge2 1

  container1:
    lxc.net.0.type = veth
    lxc.net.0.link = bridge1
    lxc.net.0.flags = up

  container2:
    lxc.net.0.type = veth
    lxc.net.0.link = bridge2
    lxc.net.0.flags = up

This fix counts only lines with matching user and nicname when checking
the quota.

### DIFF
--- a/src/lxc/cmd/lxc_user_nic.c
+++ b/src/lxc/cmd/lxc_user_nic.c
@@ -412,7 +412,8 @@ static char *get_eow(char *s, char *e)
 
 static char *find_line(char *buf_start, char *buf_end, char *name,
 		       char *net_type, char *net_link, char *net_dev,
-		       bool *owner, bool *found, bool *keep)
+		       bool *owner, bool *found_link, bool *found_dev,
+		       bool *keep)
 {
 	char *end_of_line, *end_of_word, *line;
 
@@ -420,7 +421,8 @@ static char *find_line(char *buf_start, char *buf_end, char *name,
 		size_t len;
 		char netdev_name[IFNAMSIZ];
 
-		*found = false;
+		*found_link = false;
+		*found_dev = false;
 		*keep = true;
 		*owner = false;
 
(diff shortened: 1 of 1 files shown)


brauner:
Hey, can you please sign-off your commit as per our contributing rules? :) I can't look at the patch before that. :)

Piscolero:
Sorry, my mistake. Signed-off now.

brauner:
> Sorry, my mistake. Signed-off now.

No worries at all. :)

brauner:
jenkins: test this please

brauner:
@stgraber, @hallyn, @Blub

brauner:
jenkins: test this please

brauner:
jenkins: test this please

src/lxc/cmd/lxc_user_nic.c
```diff
@@ -457,9 +457,6 @@ static char *find_line(char *buf_start, char *buf_end, char *name,
 		if (!end_of_word)
 			return NULL;
 
-		if (strncmp(buf_start, net_type, strlen(net_type)))
-			*found = false;
```
Blub: If I see this right, `net_type` is now entirely unused, which doesn't feel right. I suppose a mismatching type could set `found_dev` to `false`?

Piscolero: It really doesn't feel right, but is no change in behavior since before `*found` was set to `false` which is the same as its initial value and there is no setting to `true` before this line.

Piscolero: Since I didn't dive into the `cull_entries` function, I wasn't sure about changing the old behavior. But `find_line` probably really should return only one `found` value if all parameters (`!= NULL`) are matching.

brauner: Did you ever get around to updating the patch to do that?

Piscolero: Sorry, I didn't have any time to do this.
Will try to look into it next weekend, but can't promise anything.

brauner:
Closing due to age. Please re-open if there's sill interest.

desultory:
I think this is still an issue.

PR #3149, opened 2019-10-03
Fix lxc-update-config in network.address

AlbanVidal:
Dear developpers,

When we use lxc-update-config to translate a LXCv2 to LXCv3 configuration file, we have a issue with the parameters lxc.network.X.ipv4.address.

The output of the translate is:
lxc.net.0.ipv4.address.address

Instead of:
lxc.net.0.ipv4.address

This introduce a bug: the container want to use the gateway address as a secondary network address.

Regards,

Signed-off-by: Alban VIDAL <zordhak@debian.org>


brauner:
Thanks! :)

PR #3154, opened 2019-10-09
doc: Add more info about 'lxc.start.order'

ljelinek-cznic:
This trivial patch adds more information about 'lxc.start.order'. Currently there is nothing about how the "order" is treated (whether a lower value means an earlier start or opposite). This patch adds a sentence which explains it.


brauner:
Thanks for the patch! Can you please sign-off your commit. :)

PR #3165, opened 2019-10-18
Update Makefiles to uninstall paths owned by LXC on installation. 

caioboffo:
This PR updates Makefiles.am to remove directories owned by LXC that are created on 'make install'. It also removes some paths that LXC does not own and are being removed on 'make uninstall'. Although it could not remove the ${libexecdir}/lxc/hooks or ${libdir}/lxc/rootfs (the former could be specified by the user when running the ./configure script).


brauner:
jenkins: test this please

PR #3168, opened 2019-10-22
syscall_wrappers: rename internal memfd_create to memfd_create_lxc

havmind:
In case the internal memfd_create has to be used, make sure we don't
clash with the already existing memfd_create function from glibc.

This can happen if this glibc function is a stub. In this case, at
./configure time, the test for this function will return false, however
the declaration of that function is still available. This leads to
compilation errors.

Signed-off-by: Patrick Havelange <patrick.havelange@essensium.com>


brauner:
Good idea now that glibc finally added a wrapper.

PR #3169, opened 2019-10-23
apparmor: Prevent writes to /proc/acpi/**

Blub:
Same as #3117.


brauner:
Thanks!

PR #3178, opened 2019-11-02
conf: fix memory leak for set config rootfs options

xinhua9569:
fix memory leak for  lxc_grow_array

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -4096,6 +4096,7 @@ void lxc_conf_free(struct lxc_conf *conf)
 	free(conf->rootfs.bdev_type);
 	free(conf->rootfs.options);
 	free(conf->rootfs.path);
+	free(conf->rootfs.data);
 	free(conf->logfile);
 	if (conf->logfd != -1)
 		close(conf->logfd);


brauner:
Thanks for the patch. Can you please sign-off your commit?

src/lxc/string_utils.c
```diff
@@ -519,6 +519,7 @@ int lxc_grow_array(void ***array, size_t *capacity, size_t new_size, size_t capa
 	/* first time around, catch some trivial mistakes of the user
 	 * only initializing one of these */
 	if (!*array || !*capacity) {
+		free(*array);
```
brauner: That whole block confuses me. This makes it look like we expect the caller to have an uninitialized pointer and then initialize it? That whole block doesn't make much sense to me.
Wouldn't it be better if we'd report an error here instead of freeing the caller's memory behind their back?

xinhua9569: if we have for loop to call lxc_grow_array many times,we initialize it every time,the old don't free, I think it will memory leek

brauner:
jenkins: test this please

brauner:
Thanks!

PR #3187, opened 2019-11-18
tests: use /dev/null instead of /dev/network_latency

brauner:
BugLink: https://bugs.launchpad.net/bugs/1848587

The latter device has been removed apparently.

### DIFF
--- a/src/tests/device_add_remove.c
+++ b/src/tests/device_add_remove.c
@@ -21,7 +21,7 @@
 #include <lxc/lxccontainer.h>
 
 #define NAME "device_add_remove_test"
-#define DEVICE "/dev/network_latency"
+#define DEVICE "/dev/loop-control"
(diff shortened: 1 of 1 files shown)


src/tests/device_add_remove.c
```diff
@@ -21,7 +21,7 @@
 #include <lxc/lxccontainer.h>
 
 #define NAME "device_add_remove_test"
-#define DEVICE "/dev/network_latency"
+#define DEVICE "/dev/null"
```
stgraber: That doesn't work, we need to add/remove a device which isn't pre-created by liblxc.

brauner: hmkay

brauner: switched to /dev/loop-control

PR #3190, opened 2019-11-20
fix wrong order of bridge/nic in error message

pokab:
Seems to be a simple mixup.
Found in LXC 3.0.3 shipped with Ubuntu 18.04.


stgraber:
Thanks!

PR #3202, opened 2019-12-03
cgroups/freezer: rework cgroup2 freezer feature usage

brauner:

### DIFF
--- a/src/lxc/cgroups/cgroup.h
+++ b/src/lxc/cgroups/cgroup.h
@@ -28,6 +28,7 @@
 #include <stddef.h>
 #include <sys/types.h>
 
+#define DEFAULT_CGROUP_MOUNTPOINT "/sys/fs/cgroup"
 #define PAYLOAD_CGROUP "lxc.payload"
 #define MONITOR_CGROUP "lxc.monitor"
 #define PIVOT_CGROUP "lxc.pivot"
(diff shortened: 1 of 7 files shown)


src/lxc/cgroups/cgroup.h
```diff
@@ -28,6 +28,7 @@
 #include <stddef.h>
 #include <sys/types.h>
 
+#define DEFAULT_MOUNTPOINT "/sys/fs/cgroup"
```
stgraber: Shouldn't that have CGROUP in the name to avoid potential later clashes with other things that may have a default mount point?

brauner: myeah

stgraber:
@brauner so if I'm reading this right, on upgrade, we won't be able to freeze/unfreeze any existing running containers, are we fine with that?

brauner:
> @brauner so if I'm reading this right, on upgrade, we won't be able to freeze/unfreeze any existing running containers, are we fine with that?

Hm, not really, as this only affects the v2 freezer. The legacy freezer hierarchy works exactly the way it did before.
The new freezer feature of cgroup has not worked correctly before this patch with lxc anyway.

stgraber:
Fair enough :)

PR #3204, opened 2019-12-03
lxc: switch to SPDX

brauner:


src/lxc/tools/lxc_snapshot.c
```diff
@@ -1,21 +1,4 @@
...
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program; if not, write to the Free Software Foundation, Inc.,
- * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
- */
+/* SPDX-License-Identifier: GPL-2.0+ */
```
stgraber: The GPL-2 licensing looks incorrect in this PR, those aren't GPL-2.0+, they are GPL-2.0 only.

brauner: Is this the only file you spotted where this is the case?

stgraber: No, all the GPL v2 ones are affected, as far as I know, we don't have any GPL v2 "or later" anywhere in the LXC codebase.

Conan-Kudo:
Was this approved by the original authors mentioned in the copyright headers?

brauner:
Yes, we wrote mails and got approval. This has been taking over a year.

PR #3213, opened 2019-12-06
config: Fix parsing of mount options

blenk92:
When parsing mount options e.g. from lxc.mount.entry the specified options are mapped to the flags constants. To do so, the strings are compared to the options contained in mount_opt. However, when comparing the strings, the length of the string is not checked. That entails that the option "rootcontext=selinux-context" is mapped to the mount option read-only (ro). This commit fixes
this issue by checking if a '=' is contained in the specified option and additionally comparing the length of the strings.


brauner:
jenkins: test this please

brauner:
Thanks for catching and fixing that!

PR #3214, opened 2019-12-06
"busy" field init to -1 instead of 0

Rachid-Koucha:
"busy" field is assigned with the command socket descriptor when the terminal is in use. So, use "-1" to disable it.


brauner:
jenkins: test this please

brauner:
Thanks!

PR #3223, opened 2019-12-09
cgroups: flatten hierarchy

brauner:


chobostar:
Is that part should be fixed too ?
https://github.com/lxc/lxc/blob/master/configure.ac#L487-L490

brauner:
> @brauner Is that part should be fixed too ?
> https://github.com/lxc/lxc/blob/master/configure.ac#L487-L490

Yep, looks like it. You want to give it a shot?

brauner:
Otherwise I can send a pr once I'm back.

chobostar:
https://github.com/lxc/lxc/pull/3259

PR #3235, opened 2019-12-21
fd: only add valid fd to mainloop

xinhua9569:
Signed-off-by: dongxinhua <dongxinhua@huawei.com>

### DIFF
--- a/src/lxc/mainloop.c
+++ b/src/lxc/mainloop.c
@@ -84,6 +84,8 @@ int lxc_mainloop_add_handler(struct lxc_epoll_descr *descr, int fd,
 	struct epoll_event ev;
 	struct mainloop_handler *handler;
 	struct lxc_list *item;
+	if (fd < 0)
+		return -1;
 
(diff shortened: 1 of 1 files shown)


src/lxc/mainloop.c
```diff
@@ -84,6 +84,8 @@ int lxc_mainloop_add_handler(struct lxc_epoll_descr *descr, int fd,
 	struct epoll_event ev;
 	struct mainloop_handler *handler;
 	struct lxc_list *item;
+	if (fd < 0)
+		return 0;
```
brauner: This should probably log at least a warning.

Why do you want to return success and not an error?

xinhua9569: accept your suggestion for return -1, but don't add error print,because when every call lxc_mainloop_add_handler function, it had checked the return value

brauner:
jenkins: test this please

brauner:
Thanks!

PR #3248, opened 2020-01-08
network: improve veth device creation

brauner:
This allows us to avoid having to move the network device. It also allows us to
work around a kernel bug that in combination with a recent change in systemd
244 causes uses of systemd-networkd to not get an ip address.


brauner:
jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

PR #3251, opened 2020-01-14
test: increase timeout for api reboot tests

brauner:


stgraber:
Why is this needed? That's a very long time to wait for.

brauner:
We used to wait with -1 before. I just recently experimented with shorter wait times. Sometimes when jenkins is under heavy load this test takes a long time.

stgraber:
ok

PR #3261, opened 2020-02-05
container.conf: Fix parsing of config options:

blenk92:
When parsing config options, the order in which the option has been
added to config_jump_table is important. If a more specific option,
such as lxc.selinux.context.keyring, is added after a less specific
(i.e. lxc.selinux.context), the less specific option is taken. This
is unexpected and lead to the mistake that lxc.selinux.context.keyring
has been added at the wrong place. This patch proposes a different
strategy of option parsing: always take the longest match.

In addition, this commit fixes a off by 2 in config option parsing
(due to missing parenthesis). The error occurs if for instance
lxc.net.0.type is parsed. In that case, the .0 is removed from the
string. However, due to the missing parenthesis, the null terminating
character is off by two which means the modified config option would
be lxc.net.typepe instead of lxc.net.type.


src/lxc/confile.c
```diff
@@ -4929,7 +4938,7 @@ static struct lxc_config_t *get_network_config_ops(const char *key,
 		}
 
 		memmove(copy + 8, idx_end + 1, strlen(idx_end + 1));
-		copy[strlen(key) - numstrlen + 1] = '\0';
+		copy[strlen(key) - (numstrlen + 1)] = '\0';
```
brauner: This should be a separate bugfix commit.

src/lxc/confile.c
```diff
@@ -255,12 +255,21 @@ static const size_t config_jump_table_size = sizeof(config_jump_table) / sizeof(
 struct lxc_config_t *lxc_get_config(const char *key)
 {
 	size_t i;
+	struct lxc_config_t *best_match = NULL;
+	size_t best_match_name_len = 0;
 
-	for (i = 0; i < config_jump_table_size; i++)
-		if (!strncmp(config_jump_table[i].name, key, strlen(config_jump_table[i].name)))
-			return &config_jump_table[i];
+	for (i = 0; i < config_jump_table_size; i++) {
```
brauner: Me, that means we always run through the whole table, no? That sounds suboptimal.
Wouldn't it just suffice if you placed the `lxc.selinux.context.keyring` key at a different location? This already works for other keys, no?

blenk92: Yes, that would be sufficient. If this is the preferred way, I can simply do this. But I guess somebody will run into that issue again...

brauner: Hm, I thought we had documented that behavior... Weird. If you want, you can add a separate commit here to this branch that adds a comment above the jump table explaining this behavior.

blenk92: hmm, at least i didn't stumble across that... Nevertheless, I added a comment ;-)

brauner:
jenkins: test this please

src/lxc/confile.c
brauner: Thanks! :)

brauner:
Thanks! :)

blenk92:
Thanks for the fast review ;-)

PR #3277, opened 2020-03-08
network: Uses netlink for IP neighbour proxy management

tomponline:
Removes need for ip command when managing IP neighbour proxies.

Signed-off-by: Thomas Parrott <thomas.parrott@canonical.com>

PR #3278, opened 2020-03-08
network: Adds short sleep between veth setup and neighbour proxy add

tomponline:
There is an intermittent issue, experienced on at least Ubuntu 18.04 (5.3.0-40-generic) and Alpine 3.11 (5.4.12-1-virt) when using the router network interface type that causes the IP proxy neighbour entries on the host side of the veth pair to not be created (or to disappear very shortly after being added).

The `ip neigh add proxy` command returns without an error, however by the time the network up hook has started a short time later, the IP neighbour proxy entries are no longer there (if they ever were).

I've also tested this using netlink rather than the ip command to add and both are equally affected.

Adding a short sleep between setting up the veth pair and adding the proxy entries appears to fix it.

### DIFF
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -382,6 +382,13 @@ static int instantiate_veth(struct lxc_handler *handler, struct lxc_netdev *netd
 	}
 
 	if (netdev->priv.veth_attr.mode == VETH_MODE_ROUTER) {
+		/* sleep for a short period of time to work around a bug that intermittently prevents IP neighbour
+		   proxy entries from being added using lxc_ip_neigh_proxy below. When the issue occurs the entries
+		   appear to be added successfully but then do not appear in the proxy list. The length of time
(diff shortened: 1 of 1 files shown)


tomponline:
@stgraber this was the router proxy issue we discussed in the past (its what caused the jenkins tests to fail when adding `routed` mode to LXD). The work around for LXD was to add IP addresses to the host-side interface, however that is not always possible/wanted. 

I'm not very happy with adding a "nanosleep" (even though it is very short), but its the only thing that seems to help (and type of sleep works, even `sleep(0)` or `usleep(0)`).

I've also tried with netlink rather than the `ip` command (see PR https://github.com/lxc/lxc/pull/3277) and that doesn't help the situation.

CC @brauner any insight into what could be going on here would be appreciated.

brauner:
Sounds like a kernel bug tbh. But let's merge it for now.

brauner:
Thanks!

PR #3280, opened 2020-03-09
commands_utils: fix socket leak when adding state client

Piscolero:
If lxc_add_state_client() is called with the container already being in
the desired state the client fd will never be closed and is leaking.
This due to setting stay_connected in lxc_cmd for
LXC_CMD_ADD_STATE_CLIENT. If the desired state isn't already achieved
the client fd will later be closed by calling lxc_cmd_fd_cleanup() but
in the other case the client configuration isn't added to the handlers
state clients. So the client fd has to be closed explicitely.

This is simply tested by starting container A and calling lxc-wait -n A
-s RUNNING.

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -863,7 +863,7 @@ int lxc_cmd_add_state_client(const char *name, const char *lxcpath,
 	return MAX_STATE;
 }
 
-static int lxc_cmd_add_state_client_callback(int fd, struct lxc_cmd_req *req,
+static int lxc_cmd_add_state_client_callback(__owns int fd, struct lxc_cmd_req *req,
 					     struct lxc_handler *handler,
 					     struct lxc_epoll_descr *descr)
 {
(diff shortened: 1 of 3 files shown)


brauner:
Excellent catch! Would you mind adding a define to `compiler.h` like
```c
#define __owns
```
and then redefining `lxc_add_state_client`() as:
```c
int lxc_add_state_client(__owns int state_client_fd, struct lxc_handler *handler,
			 lxc_state_t states[MAX_STATE])

```

You can also add a short comment on top of `__owns` that it indicates that the function takes ownership of the fd and might close it.

brauner:
jenkins: test this please

brauner:
I think Jenkins is offline at the moment because @stgraber is reshuffling his storage a bit but it should be back in a little bit.

Piscolero:
Alright, this is not the correct location to close the filedescriptor since it's used for the response afterwards of course.

Piscolero:
Moved the closing of the filedescriptor after sending the response. Didn't check the exit code of lxc-wait when testing before.

brauner:
jenkins: test this please

brauner:
Looks good, thank you!

PR #3290, opened 2020-03-11
pidfds: switch infrastructure to rely on pidfds whenever possible

brauner:

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -117,9 +117,7 @@ cgroup of the container when the controller is available.
 
 # cgroup2
 
-This enables `LXC` to make complete use of the unified cgroup hierarchy. With
-this extension it is possible to run `LXC` containers on systems that use
-a pure unified cgroup layout.
+This enables `LXC` to make complete use of the unified cgroup hierarchy. With this extension it is possible to run `LXC` containers on systems that use a pure unified cgroup layout.
 
 # init\_pidfd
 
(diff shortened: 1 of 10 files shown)


doc/api-extensions.md
```diff
@@ -46,6 +46,8 @@ the kernel will generate a poll event and send a message over the file descripto
 
 The caller can read this message, inspect the syscalls including its arguments. Based on this information the caller is expected to send back a message informing the kernel which action to take. Until that message is sent the kernel will block the calling process. The format of the messages to read and sent is documented in seccomp itself.
 
+A new API function `seccomp_notify_fd()` has been which allows callers to retrieve the notifier fd for the container's seccomp filter.
```
stgraber: `has been added`?

brauner: oh yeah

stgraber:
What's the upgrade handling for the pidfd logic. That is, we're running on a kernel which supports it but are running existing containers which were started with a liblxc version which doesn't support it.

Do we properly fallback to the old logic in that case?

brauner:
> What's the upgrade handling for the pidfd logic. That is, we're running on a kernel which supports it but are running existing containers which were started with a liblxc version which doesn't support it.
> 
> Do we properly fallback to the old logic in that case?

Yeah, the callback required won't be available so we get a -1 and fallback to pids.

PR #3296, opened 2020-03-16
Deactivate autodev in busybox template

Rachid-Koucha:
Add lxc.autodev = 0 to use the dev directory built by the busybox template

### DIFF
--- a/templates/lxc-busybox.in
+++ b/templates/lxc-busybox.in
@@ -235,6 +235,7 @@ lxc.cap.drop = sys_module mac_admin mac_override sys_time
 # When using LXC with apparmor, uncomment the next line to run unconfined:
 #lxc.apparmor.profile = unconfined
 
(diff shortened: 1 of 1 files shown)


stgraber:
Why is this needed?

Rachid-Koucha:
Without this, LXC builds a default /dev directory without shm.

stgraber:
@brauner any reason why autodev would run after the mounts?
If the mounts are applied after autodev, then this isn't needed.

Rachid-Koucha:
For your information, I tried this with LXC 2.1.1 as it is the version used in my company.

brauner:
> @brauner any reason why autodev would run after the mounts?
> If the mounts are applied after autodev, then this isn't needed.

That's how it is today.

stgraber:
Ok, then this isn't needed.

PR #3323, opened 2020-03-24
conf: fix read-only bind mounts

tych0:
Here we would always set MS_RDONLY in required_flags if it was set in
mountflags, so the expression:

!(required_flags & ~mountflags)

would always be true, and we would always skip the remount.

Instead, let's treat readonly as special: always do the remount if
MS_RDONLY is present. Unfortunately it doesn't seem to show up in
sb.f_flag, so we can't use the same path as everything else.

This only inadvertently worked before because of a bug fixed in
f75917858023 ("conf: don't accidently double-mount").


brauner:
Thanks for that!

PR #3329, opened 2020-03-26
cgroups: fix attaching to the unified cgroup

brauner:

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -2057,12 +2057,11 @@ static inline char *build_full_cgpath_from_monitorpath(struct hierarchy *h,
 	return must_make_path(h->mountpoint, inpath, filename, NULL);
 }
 
-static int cgroup_attach_leaf(int unified_fd, int64_t pid)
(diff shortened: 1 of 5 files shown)


src/lxc/cgroups/cgfsng.c
```diff
@@ -2082,6 +2081,8 @@ static int cgroup_attach_leaf(int unified_fd, int64_t pid)
 		return log_error_errno(-1, errno, "Failed to attach to unified cgroup");
 
 	do {
+		bool rm = false;
+		char attach_cgroup[STRLITERALLEN("lxc-1000/cgroup.procs") + 1];
```
stgraber: Hmm, what's that?

stgraber: As in, what happens if my container name is significantly longer than `lxc-1000`?

brauner: That's irrelevant. That's the maximum name of the sub-cgroup we create to attach to the container in the unified hierarchy and it is created via `mkdirat()` and has nothing to do with the cgroup that container itself uses.

PR #3339, opened 2020-03-30
verify cgroup controller name

Blub:
validate that a cgroup controller name is a valid
zero-terminated string before passing it to
`cgroup_ops->get_cgroup()`.


brauner:
Thanks!

PR #3340, opened 2020-03-30
cgroups: handle older kernels (e.g. v4.9)

brauner:
On olders kernels the restrictions to move processes between cgroups are
different than they are on newer kernels. Specifically, we're running into the
following check:
```c
if (!uid_eq(cred->euid, GLOBAL_ROOT_UID) &&
    !uid_eq(cred->euid, tcred->uid) &&
    !uid_eq(cred->euid, tcred->suid))
        ret = -EACCES;
```
which dictates that in order to move a process into a cgroup one either needs
to be global root (no restrictions apply) or the effective uid of the process
trying to move the process and the {saved}uid of the process that is supposed
to be mvoed need to be identical. The new attaching logic we did didn't
fulfill this criterion for because it's not present on new kernels.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -30,6 +30,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include "af_unix.h"
 #include "caps.h"
 #include "cgroup.h"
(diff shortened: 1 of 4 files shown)


stgraber:
jenkins: test this please

PR #3341, opened 2020-03-31
init: add ExecReload to lxc.service to only reload profiles

Blub:
I wanna say it seems nicer for packaging, but then again debhelpers cannot reload anyway ;-)
Still, I think it makes sense.


brauner:
jenkins: test this please

brauner:
Thank you!

PR #3342, opened 2020-03-31
allow running lxc-monitord as a system daemon

Blub:
lxc-monitord instances are spawned on demand and, if this
happens from a service, the daemon is considered part of
it by systemd, as it is running in the same cgroups. This
can be avoided by leaving it running permanently.


brauner:
jenkins: test this please

brauner:
Thank you!

PR #3344, opened 2020-04-01
fix non-root user cannot write /dev/stdout

gaohuatao-1:
Signed-off-by: gaohuatao <gaohuatao@huawei.com>


brauner:
Oh, I didn't get notified about this patch. So sorry!

brauner:
jenkins: test this please

PR #3347, opened 2020-04-02
cgroups: fix "uninitialized transient_len" warning

cyphar:
Without this change, a build error is triggered if you compile with
`-Werror=maybe-uninitialized`.

```
cgroups/cgfsng.c: In function 'cgfsng_monitor_enter':
groups/cgfsng.c:1387:9: error: 'transient_len' may be used uninitialized in this function
   ret = lxc_writeat(h->cgfd_mon, "cgroup.procs", transient, transient_len);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The issue is that if `handler->transient_pid` is 0, then `transient_len` is
uninitialised but `lxc_writeat(..., transient_len)` still gets called.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1388,7 +1388,7 @@ __cgfsng_ops static bool cgfsng_monitor_enter(struct cgroup_ops *ops,
 		if (ret)
 			return log_error_errno(false, errno, "Failed to enter cgroup \"%s\"", h->monitor_full_path);
 
-                if (handler->transient_pid < 0)
+		if (handler->transient_pid <= 0)
 			return true;
 
 		ret = lxc_writeat(h->cgfd_mon, "cgroup.procs", transient, transient_len);


cyphar:
Another solution would be to just initialise `transient_len` but this seemed more like a logic error.

brauner:
jenkins: test this please

brauner:
> Another solution would be to just initialise `transient_len` but this seemed more like a logic error.

Odd, I wonder why this isn't initialized, i.e. where would this become 0...

cyphar:
The compiler is just being too clever but it's preventing LXC from building on Leap, so I just decided to take C3PO's advice and "let the compiler win" on this one. Let me just double-check it does actually fix the compiler error...

brauner:
> The compiler is just being too clever but it's preventing LXC from building on Leap, so I just decided to take C3PO's advice and "let the compiler win" on this one. Let me just double-check it does actually fix the compiler error...

Yeah, I think `<= 0` checking against pid is the sane thing to do anyway. Especially in the context of cgroups since writing 0 means move the current process which isn't a great idea.

brauner:
jenkins: test this please

brauner:
Thank you very much, good sir.

cyphar:
Okay, so the compiler actually isn't smart enough -- even with the patch the warning is still present. I'll send another patch to just `0`-initialise the variable...

brauner:
> Okay, so the compiler actually isn't smart enough -- even with the patch the warning is still present. I'll send another patch to just `0`-initialise the variable...

Which compiler version is that? I wonder if we can get travis to also compile with a specific compiler version...

cyphar:
Looks like Leap 15.1 uses GCC 7.5.0 -- presumably that's why it's smart but not smart enough.

PR #3352, opened 2020-04-03
Revert "start: remove unnecessary check for valid cgroup_ops"

Blub:
This reverts commit 52520e4f793f73e5956c2d9de9c83f074622ce1d.

This can be NULL when there's a pre-start hook which fails.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -933,8 +933,10 @@ void lxc_end(struct lxc_handler *handler)
 
 	lsm_process_cleanup(handler->conf, handler->lxcpath);
 
-	cgroup_ops->payload_destroy(cgroup_ops, handler);
-	cgroup_ops->monitor_destroy(cgroup_ops, handler);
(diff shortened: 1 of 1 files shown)


src/lxc/start.c
```diff
@@ -933,8 +933,10 @@ void lxc_end(struct lxc_handler *handler)
 
 	lsm_process_cleanup(handler->conf, handler->lxcpath);
 
-	cgroup_ops->payload_destroy(cgroup_ops, handler);
-	cgroup_ops->monitor_destroy(cgroup_ops, handler);
+	if (cgroup_ops) {
```
brauner, line 936: Oh, this was me having a brainfart. I was thinking that we have check in the callbacks for whether or not cgroup_ops is initialized but that obviously doesn't help when cgroup_ops is NULL...

brauner: You're missing a SOB line, though!

PR #3353, opened 2020-04-03
 introduce lxc.cgroup.dir.{monitor,container,container.inner}

Blub:
This is a new approach to #1302 with a container-side
configuration instead of a global boolean flag.

Contrary to the previous PR using an optional additional
parameter for the get-cgroup command, this introduces two
new additional commands to get the limiting cgroup path and
cgroup2 file descriptor. If the limiting option is not in
use, these behave identical to their full-path counterparts.

If these variables are used the payload will end up in the
concatenation of lxc.cgroup.dir.payload and
lxc.cgroup.dir.namespace (namespace may be empty), and the
monitor will end up in lxc.cgruop.dir.monitor. The
directories are fixed, no retry count logic is applied,
failing to create these directories will simply be a hard
error.

Note that this is designed to be opt in, so if there are still some setup
issues with eg. the devices cgroup in v1 or with cgroupv2 (which I'll be
testing shortly), it still shouldn't affect existing configurations.
Posting this now for initial review.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1238,7 +1238,6 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
 	const char *monitor_dir = conf->cgroup_meta.monitor_dir,
 		   *container_dir = conf->cgroup_meta.container_dir,
 		   *namespace_dir = conf->cgroup_meta.namespace_dir;
-	size_t prefix_len;
 
 	/* none of the new options are set, all is fine */
 	if (!monitor_dir && !container_dir && !namespace_dir)
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -725,6 +725,7 @@ static struct hierarchy *add_hierarchy(struct hierarchy ***h, char **clist, char
 	new->container_base_path = container_base_path;
 	new->version = type;
 	new->cgfd_con = -EBADF;
+	new->cgfd_limit = -EBADF;
 	new->cgfd_mon = -EBADF;
 
 	newentry = append_null_to_list((void ***)h);
@@ -956,13 +957,15 @@ static int cgroup_tree_remove(struct hierarchy **hierarchies,
 		struct hierarchy *h = hierarchies[i];
 		int ret;
 
-		if (!h->container_full_path)
+		if (!h->container_limit_path)
 			continue;
 
-		ret = lxc_rm_rf(h->container_full_path);
+		ret = lxc_rm_rf(h->container_limit_path);
 		if (ret < 0)
-			WARN("Failed to destroy \"%s\"", h->container_full_path);
+			WARN("Failed to destroy \"%s\"", h->container_limit_path);
 
+		if (h->container_limit_path != h->container_full_path)
+			free_disarm(h->container_limit_path);
 		free_disarm(h->container_full_path);
 	}
 
@@ -1089,7 +1092,12 @@ __cgfsng_ops static void cgfsng_monitor_destroy(struct cgroup_ops *ops,
 			goto try_lxc_rm_rf;
 		}
 
-		if (conf && conf->cgroup_meta.dir)
+		if (conf && conf->cgroup_meta.monitor_dir)
+			pivot_path = must_make_path(h->mountpoint,
+						    h->container_base_path,
+						    conf->cgroup_meta.monitor_dir,
+						    CGROUP_PIVOT, NULL);
+		else if (conf && conf->cgroup_meta.dir)
 			pivot_path = must_make_path(h->mountpoint,
 						    h->container_base_path,
 						    conf->cgroup_meta.dir,
@@ -1147,7 +1155,8 @@ static int mkdir_eexist_on_last(const char *dir, mode_t mode)
 }
 
 static bool cgroup_tree_create(struct hierarchy *h, const char *cgroup_tree,
-			       const char *cgroup_leaf, bool payload)
+			       const char *cgroup_leaf, bool payload,
+			       const char *cgroup_limit_dir)
 {
 	__do_free char *path = NULL;
 	int ret, ret_cpuset;
@@ -1176,6 +1185,16 @@ static bool cgroup_tree_create(struct hierarchy *h, const char *cgroup_tree,
 		if (h->cgfd_con < 0)
 			return log_error_errno(false, errno, "Failed to open %s", path);
 		h->container_full_path = move_ptr(path);
+		if (cgroup_limit_dir) {
+			path = must_make_path(h->mountpoint, h->container_base_path, cgroup_limit_dir, NULL);
+			h->cgfd_limit = lxc_open_dirfd(path);
+			if (h->cgfd_limit < 0)
+				return log_error_errno(false, errno, "Failed to open %s", path);
+			h->container_limit_path = move_ptr(path);
+		} else {
+			h->container_limit_path = h->container_full_path;
+			h->cgfd_limit = h->cgfd_con;
+		}
 	} else {
 		h->cgfd_mon = lxc_open_dirfd(path);
 		if (h->cgfd_mon < 0)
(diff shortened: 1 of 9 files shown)


src/lxc/cgroups/cgfsng.c
```diff
@@ -1188,18 +1207,71 @@ static bool cgroup_tree_create(struct hierarchy *h, const char *cgroup_tree,
...
+ * Returns the prefix length if it is set, otherwise zero on success.
+ */
+static bool check_cgroup_dir_config(struct lxc_conf *conf)
+{
+	const char *payload_dir, *limit_prefix, *monitor_dir;
+	size_t prefix_len;
+
+	monitor_dir = conf->cgroup_meta.monitor_dir;
+	payload_dir = conf->cgroup_meta.payload_dir;
+	limit_prefix = conf->cgroup_meta.limit_prefix_dir;
```
brauner, line 1243: Hm, you could just:
```c
const char *monitor_dir = conf->cgroup_meta.monitor_dir,
           *payload_dir = conf->cgroup_meta.payload_dir,
           *limit_prefix = conf->cgroup_meta.limit_prefix_dir;
```

src/lxc/cgroups/cgfsng.c
```diff
@@ -1188,18 +1207,71 @@ static bool cgroup_tree_create(struct hierarchy *h, const char *cgroup_tree,
...
+				       " set");
+
+	if (!limit_prefix)
+		return true;
+
+	/* we have a limit prefix directory, check it: */
+
+	prefix_len = strlen(limit_prefix);
+	if (strncmp(payload_dir, limit_prefix, prefix_len) == 0 &&
+	    payload_dir[prefix_len] == '/')
```
brauner, line 1269: Any concerns about having `../../../` or sm in the path?

brauner: Maybe we just just make this a hard requirement for all `lxc.cgroup.{dir,monitor,payload,limit_prefix}`?

src/lxc/cgroups/cgfsng.c
```diff
@@ -1268,9 +1348,9 @@ __cgfsng_ops static inline bool cgfsng_monitor_create(struct cgroup_ops *ops,
 			idx++;
 			break;
 		}
-	} while (ops->hierarchies[i] && idx > 0 && idx < 1000);
+	} while (ops->hierarchies[i] && idx > 0 && idx < 1000 && suffix);
 
-	if (idx == 1000)
+	if (idx == 1000 || (!suffix && idx != 0))
```
brauner, line 1353: We really need to rethink how we handle this too at some point (the retry logic that is).

src/lxc/cgroups/cgroup.h
```diff
@@ -54,7 +54,11 @@ typedef enum {
  *   init's cgroup (if root).
  *
  * @container_full_path
- * - The full path to the containers cgroup.
+ * - The full path to the container's cgroup.
+ *
+ * @container_limit_path
+ * - The full path to the container's limiting cgroup. May simply point to
+ *   container_full_path.
```
brauner, line 61: The case where they both point to the same memory is handled safely, i.e. we can't end-up double-freeing?

src/lxc/cgroups/cgroup.h
```diff
@@ -146,7 +153,7 @@ struct cgroup_ops {
 	bool (*monitor_enter)(struct cgroup_ops *ops, struct lxc_handler *handler);
 	bool (*payload_create)(struct cgroup_ops *ops, struct lxc_handler *handler);
 	bool (*payload_enter)(struct cgroup_ops *ops, struct lxc_handler *handler);
-	const char *(*get_cgroup)(struct cgroup_ops *ops, const char *controller);
+	const char *(*get_cgroup)(struct cgroup_ops *ops, const char *controller, bool limiting);
```
brauner, line 156: Hm, wouldn't a new `get_limiting_cgroup()` be nicer and move that abstraction with the `bool` argument into `cgfsng.c`?

brauner: I think that would make the codepath in `commands.c` a little nicer, too.

src/lxc/confile.c
```diff
@@ -3644,6 +3682,56 @@ static int get_config_cgroup_dir(const char *key, char *retv, int inlen,
...
+{
+	int len;
+	int fulllen = 0;
+
+	if (!retv)
+		inlen = 0;
+	else
+		memset(retv, 0, inlen);
+
+	strprint(retv, inlen, "%s", lxc_conf->cgroup_meta.payload_dir);
```
brauner, line 3712: Fwiw, I really hate this macro. :)

src/lxc/confile.c
```diff
@@ -4458,6 +4546,42 @@ static int clr_config_cgroup_dir(const char *key, struct lxc_conf *lxc_conf,
...
 }
 
+static int clr_config_cgroup_monitor_dir(const char *key,
+					 struct lxc_conf *lxc_conf,
+					 void *data)
+{
+	if (lxc_conf->cgroup_meta.monitor_dir) {
+		free(lxc_conf->cgroup_meta.monitor_dir);
+		lxc_conf->cgroup_meta.monitor_dir = NULL;
+	}
```
brauner, line 4556: I think this should just be:
```c
free_disarm(lxc_conf->cgroup_meta.monitor_dir);
```
which does all that. :)

src/lxc/confile.c
```diff
@@ -4458,6 +4546,42 @@ static int clr_config_cgroup_dir(const char *key, struct lxc_conf *lxc_conf,
...
+}
+
+static int clr_config_cgroup_payload_dir(const char *key,
+					 struct lxc_conf *lxc_conf,
+					 void *data)
+{
+	if (lxc_conf->cgroup_meta.payload_dir) {
+		free(lxc_conf->cgroup_meta.payload_dir);
+		lxc_conf->cgroup_meta.payload_dir = NULL;
+	}
```
brauner, line 4568: I think this should just be:
```c
free_disarm(lxc_conf->cgroup_meta.payload_dir);
```
which does all that. :)

src/lxc/confile.c
```diff
@@ -4458,6 +4546,42 @@ static int clr_config_cgroup_dir(const char *key, struct lxc_conf *lxc_conf,
...
+}
+
+static int clr_config_cgroup_limit_prefix_dir(const char *key,
+					      struct lxc_conf *lxc_conf,
+					      void *data)
+{
+	if (lxc_conf->cgroup_meta.limit_prefix_dir) {
+		free(lxc_conf->cgroup_meta.limit_prefix_dir);
+		lxc_conf->cgroup_meta.limit_prefix_dir = NULL;
+	}
```
brauner, line 4580: I think this should just be:
```c
free_disarm(lxc_conf->cgroup_meta.limit_prefix_dir);
```
which does all that. :)

src/lxc/cgroups/cgfsng.c
Blub: Not sure, if so, `lxc.cgroup.dir` should be restricted as well. I'm not sure it's our job to check this here though. An unprivileged user shouldn't have the permissions for "bad paths" anyway.

Blub: Alternatively we could go with `payload_dir` and `namespace_dir` (closer to the old patch), that might also be easier in some places (eg. I already need to take the special care of the cpuset cgroup which I just realized wouldn't work with deeper paths than 1 additional dir).

src/lxc/cgroups/cgroup.h
Blub: I went through all the cases I found, but if you feel safer with a simple `strdup()` that's fine with me.

src/lxc/cgroups/cgroup.h
Blub: True, will change it. (At least this prevented me from forgetting to handle all the call sites :) )

src/lxc/confile.c
Blub: Right. I see a cleanup commit for existing cases in the future. ;-)

src/lxc/cgroups/cgroup.h
brauner: Nah, it's fine.

PR #3356, opened 2020-04-05
doc: Add lxc.cgroup.dir.{monitor,container,container.inner} to Japanese man

tenforward:
Update for commit a900cba


brauner:
Thank you very much! :)

PR #3359, opened 2020-04-07
cgroup isolation: handle devices cgroup early

Blub:
Otherwise we cannot use an 'a' entry in devices.deny/allow
as these are not permitted once a subdirectory was created.

Without isolation we initialize the devices cgroup
particularly late, so there are probably cases which cannot
work with isolation.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1154,11 +1154,12 @@ static int mkdir_eexist_on_last(const char *dir, mode_t mode)
 	return 0;
 }
 
-static bool cgroup_tree_create(struct hierarchy *h, const char *cgroup_tree,
+static bool cgroup_tree_create(struct cgroup_ops *ops, struct lxc_conf *conf,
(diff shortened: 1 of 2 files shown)


src/lxc/cgroups/cgfsng.c
```diff
@@ -1169,6 +1170,37 @@ static bool cgroup_tree_create(struct hierarchy *h, const char *cgroup_tree,
...
+		h->container_limit_path = move_ptr(limit_path);
+
+		/*
+		 * With isolation the devices legacy cgroup needs to be
+		 * iinitialized early, as it typically contains an 'a' (all)
+		 * line, which is not possible once a subdirectory has been
+		 * created.
+		 */
+		if (string_in_list(h->controllers, "devices")) {
+			ret = ops->setup_limits_legacy(ops, conf, true);
```
brauner, line 1198: Any clue as to whether you'd need something for the cgroup2 devices controller too or is this cgroup v1 specific?

PR #3360, opened 2020-04-07
start: ensure all file descriptors are closed during exec

brauner:
Closes https://github.com/checkpoint-restore/criu/issues/1011.


necessarily-equal:
This fixes https://github.com/anbox/anbox/issues/1397, thanks for that! Do you have an idea of when will LXC4.0.2 will be released and if it will bundle the fix?

PR #3376, opened 2020-04-12
Fix lxc-oci template with loop backingstore

toddnni:
Move the content of rootfs inside OCI package to rootfs instead of replacing it, as the directory is used as the mountpoint.

Tested with directory and loop backingstore.

Noticed this on Debian, where the error message was `rmdir: failed to remove '/usr/lib/x86_64-linux-gnu/lxc/rootfs': Device or resource busy` when running 

    lxc-create -n test -t oci -B loop -- -u docker://centos:latest


brauner:
jenkins: test this please

brauner:
@stgraber, can you double-check?

brauner:
Thanks everyone!

PR #3410, opened 2020-05-13
reboot fixes

brauner:

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -942,7 +942,7 @@ static void do_restore(struct lxc_container *c, int status_pipe, struct migrate_
 		close(fd);
 	}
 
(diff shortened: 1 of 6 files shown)


tomponline:
@brauner thanks! Do you know why this only happened with syslog level 2 or below?

brauner:
On Wed, May 13, 2020 at 05:07:07AM -0700, TomP wrote:
> @brauner thanks! Do you know why this only happened with syslog level 2 or below?

Yeah. When you request LXC logs to syslog then LXC will open a syslog
connection when it starts the container and enable logging to it with a
global flag. This means the monitor will now have an additional syslog
fd open. When you reboot LXC will still hold that connection open. Now
lxc_check_inherited() will close the syslog connection and log that it
closed the syslog connection but since syslog logging is enabled it will
re-open the syslog connection causing a new unexpected fd to show and
the cycle repeats.

PR #3411, opened 2020-05-14
console: only create detached mount when a console is requested

brauner:
otherwise weird things might happen.


stgraber:
jenkins not happy

brauner:
what the?...

brauner:
Ok, fixed. This was interesting.

PR #3414, opened 2020-05-15
improve LXC_CMD_GET_CGROUP compatibility

Blub:
When a newer lxc library communicates with an older one
(such as running an lxc 4.0 lxc-freeze on a longer running
container which was started while lxc was still at version
3), the LXC_CMD_GET_LIMITING_CGROUP command is not
available, causing the remote to just close the socket.
Catch this and try the previous command instead.


brauner:
Keyserver failures.

PR #3428, opened 2020-05-29
Add test of lxc-usernsexec

smoser:
The test executes lxc-usernsexec to create some files and chmod them.
Then makes assertions on the uid and gid of those files from outside.


stgraber:
jenkins: test this please

stgraber:
I think you're missing some autotools/automake changes for this to actually be installed as a test binary, that's why our testsuite isn't running it right now.

smoser:
adjusted to add it to bin_scripts if ENABLE_TOOLS.

stgraber:
jenkins: test this please

stgraber:
You also need EXTRA_DIST or it won't make it to the release tarball.

smoser:
and updated one more time.  thank you, @stgraber.

stgraber:
jenkins: test this please

PR #3432, opened 2020-06-01
lxc-test-usernsexec: If user is root, then create and use non-root user.

smoser:
Previously if the user was root, then the test would just skip
running (and exit 0).  The lxc test environment is run as root.
So, instead of never doing anything there, we create a user,
make sure it is in /etc/sub{ug}id and then execute the test as that
user.

If user is already non-root, then just execute the tests as before.


stgraber:
DCO please :)

brauner:
jenkins: test this please

PR #3442, opened 2020-06-09
Coverity fixes for veth vlan

tomponline:


src/lxc/network.c
```diff
@@ -470,9 +473,9 @@ static int setup_veth_ovs_bridge_vlan(char *veth1, struct lxc_netdev *netdev)
 	int taggedLength = lxc_list_len(&netdev->priv.veth_attr.vlan_tagged_ids);
 	struct ovs_veth_vlan_args args;
 	args.nic = veth1;
-	args.vlan_mode = "";
-	args.vlan_id = -1;
-	args.trunks = "";
+	args.vlan_mode = NULL;
```
tomponline, line 476: @brauner do you think coverity will also complain that these char pointers are not being freed (`args.vlan_mode` and `args.trunks`)?

brauner: Yeah, you should free this memory if it is separate from `trunks`.

src/lxc/confile.c
```diff
@@ -5906,7 +5906,7 @@ static int get_config_net_veth_vlan_tagged_id(const char *key, char *retv, int i
 	struct lxc_netdev *netdev = data;
 
 	if (!netdev)
-		ret_errno(EINVAL);
+		return ret_errno(EINVAL);
```
brauner, line 5909: Good spot.

src/lxc/network.c
```diff
@@ -426,38 +421,45 @@ struct ovs_veth_vlan_args {
...
 	const char *vlan_mode;	/* Port VLAN mode. */
 	short vlan_id;		/* PVID VLAN ID. */
-	const char *trunks;	/* Comma delimited list of tagged VLAN IDs. */
+	char *trunks;		/* Comma delimited list of tagged VLAN IDs. */
 };
 
+static void free_ovs_veth_vlan_args(struct ovs_veth_vlan_args *args)
+{
+	if (args->trunks)
+		free_disarm(args->trunks);
```
brauner, line 430: This van just be:
```c
static inline void free_ovs_veth_vlan_args(struct ovs_veth_vlan_args *args)
{
		free_disarm(args->trunks);
}
```

PR #3450, opened 2020-06-13
containertests: fix null pointer defereference

gaurav1086:
Signed-off-by: Gaurav Singh <gaurav1086@gmail.com>

In the if condition : if !str is true i.e. str is NULL, it does further operations on str (fprintf) which can cause segmentation fault. Fix this by by changing the if condition to : 
if (str && strcmp(str, CONFIGFNAM))


brauner:
jenkins: test this please

brauner:
Thanks for fixing this.

PR #3452, opened 2020-06-18
Revert "nl: fix memory leak"

brauner:
This was an obvious braino on my part to merge this. Out automatic cleanup
macros will handle this case just fine so this introduces a double-free. Rever
this commit.

This reverts commit 9d05339487f4e9c4e7f700f963c161a4d9977ae4.


stgraber:
Missing sign off

PR #3468, opened 2020-06-30
templates/lxc-download.in: fix wrong if-condition

johanneskastl:
Use the result of the gpg command, not the result when executing the result of the gpg command.

In my understanding, we want to execute the gpg command and use its return code. We do not want to execute the gpg command, then collect its output, execute that and then return the exit code from that command.


johanneskastl:
Apparently `${DOWNLOAD_GPG_PROXY:-}` only works if unquoted, so I removed that one again from my commit.

PR #3469, opened 2020-06-30
templates/lxc-download.in: use GPG option "--receive-keys"

johanneskastl:
Using the option --recv-keys I get an error on openSUSE Tumbleweed which has `gpg2-2.2.20-1.2.x86_64`.

Not sure if this needs to be set differently for older systems, that do not have a recent gpg2 and thus might require the old  behaviour.


stgraber:
Can you rebase on master? I think recv-keys was always a shortcut for receive-keys, so we should be fine with this.

johanneskastl:
> Can you rebase on master? I think recv-keys was always a shortcut for receive-keys, so we should be fine with this.

Done.

PR #3487, opened 2020-07-17
Improve efficiency of lxc_ifname_alnum_case_sensitive

samboyles1:
To detect if a newly generated interface name is a duplicate of an existing interface lxc_ifname_alnum_case_sensitive() currently gets a list of all interfaces using netns_getifaddrs(). When the system has a small number of interfaces this works fine, however when there are thousands or tens of thousands of interfaces this quickly becomes less than optimal.

As we only need to check if an interface name exists, and do not need the detailed information about the interfaces provided by netns_getifaddrs(), we can instead use the if_nametoindex() function, which is much more efficient.

Signed-off-by: Sam Boyles <sam.boyles@alliedtelesis.co.nz>

### DIFF
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -2754,9 +2754,7 @@ static const char padchar[] = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLM
 char *lxc_ifname_alnum_case_sensitive(char *template)
 {
 	int ret;
-	struct netns_ifaddrs *ifa, *ifaddr;
 	char name[IFNAMSIZ];
-	bool exists = false;
 	size_t i = 0;
 #ifdef HAVE_RAND_R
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

src/lxc/network.c
```diff
@@ -2792,18 +2783,11 @@ char *lxc_ifname_alnum_case_sensitive(char *template)
...
-			if (!strcmp(ifa->ifa_name, name)) {
-				exists = true;
-				break;
-			}
-		}
-
-		if (!exists)
+		if (if_nametoindex(name) == 0) {
 			break;
+		}
```
brauner, line 2788: The only tiny thing I have is: We don't do `{}` around single-line if statements. :)

brauner:
Thanks!

PR #3491, opened 2020-07-18
fix the return value check of snprintf

JingWoo:
Signed-off-by: wujing <wujing50@huawei.com>


src/lxc/commands_utils.c
```diff
@@ -117,7 +117,7 @@ int lxc_make_abstract_socket_name(char *path, size_t pathlen,
 	}
 
 	ret = snprintf(offset, len, "%s/%s/%s", lxcpath, name, suffix);
-	if (ret < 0)
+	if (ret < 0 || (size_t)ret >= len)
```
brauner, line 120: So this is intentional. If you look below you'll see that overflow is handled further below by falling back to hashing.

JingWoo: thanks...

PR #3532, opened 2020-09-02
avoid a NULL pointer dereference in lxc-attach

sparlane:
Seems to appear when stderr is a terminal and not stdin or stdout.

Signed-off-by: Scott Parlane <scott.parlane@alliedtelesis.co.nz>

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1384,14 +1384,10 @@ int lxc_attach(struct lxc_container *container, lxc_attach_exec_t exec_function,
 
 	if (pid == 0) {
 		if (options->attach_flags & LXC_ATTACH_TERMINAL) {
-			if (terminal.tty_state)
-			{
-				ret = pthread_sigmask(SIG_SETMASK,
-							  &terminal.tty_state->oldmask, NULL);
-				if (ret < 0) {
-					SYSERROR("Failed to reset signal mask");
-					_exit(EXIT_FAILURE);
-				}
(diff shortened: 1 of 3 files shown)


src/lxc/attach.c
```diff
@@ -1384,11 +1384,14 @@ int lxc_attach(struct lxc_container *container, lxc_attach_exec_t exec_function,
...
-				SYSERROR("Failed to reset signal mask");
-				_exit(EXIT_FAILURE);
+			if (terminal.tty_state)
+			{
+				ret = pthread_sigmask(SIG_SETMASK,
+							  &terminal.tty_state->oldmask, NULL);
+				if (ret < 0) {
+					SYSERROR("Failed to reset signal mask");
+					_exit(EXIT_FAILURE);
+				}
```
brauner, line 1394: I think we should not open-code this. Can you add sm like the following on top of your patch, please?

file:///home/brauner/src/git/lxc/src/lxc/tmp/0001-terminal-introduce-lxc_terminal_signal_sigmask_safe_.patch
file:///home/brauner/src/git/lxc/src/lxc/tmp/0002-attach-use-lxc_terminal_signal_sigmask_safe_blocked.patch

brauner (changes requested):
Thank you for your patch! Good catch!

Just one request. :)

src/lxc/attach.c
sparlane: I gave these patches a test with in my setup, no issues seen.

brauner:
jenkins: test this please

brauner:
Thanks!

PR #3536, opened 2020-09-04
configure: enable flto

brauner:
https://llvm.org/docs/LinkTimeOptimization.html


brauner:
Hm, `-flto` support is detected by llvmGOLD is not found. Not sure currently if this is a autotools bug or we're missing an install target on our clang builds.

brauner:
Already enabled.

Issue #3471, opened 2020-07-02
lxc-stop in some Docker won't work


P-EB:
Hi,

Since we upgraded LXC to 4.0.2 in Debian unstable, the Debian Salsa CI seems to break out.

This post: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=961584#22 contains a Dockerfile that allows to reproduce the issue. 

The trace obtained is:

```
Starting LXC network bridge: :Starting LXC autoboot containers: :
lxc-stop autopkgtest-stable-amd64 20200603085059.718 TRACE    commands - commands.c:lxc_cmd_rsp_recv:123 - Command "get_init_pid" received response
lxc-stop autopkgtest-stable-amd64 20200603085059.718 DEBUG    commands - commands.c:lxc_cmd_rsp_recv:156 - Response data length for command "get_init_pid" is 0
lxc-stop autopkgtest-stable-amd64 20200603085059.718 TRACE    commands - commands.c:lxc_cmd:293 - Opened new command socket connection fd 4 for command "get_init_pid"
lxc-stop autopkgtest-stable-amd64 20200603085059.719 TRACE    commands - commands.c:lxc_cmd_rsp_recv:123 - Command "get_state" received response
lxc-stop autopkgtest-stable-amd64 20200603085059.719 DEBUG    commands - commands.c:lxc_cmd_rsp_recv:156 - Response data length for command "get_state" is 0
lxc-stop autopkgtest-stable-amd64 20200603085059.719 TRACE    commands - commands.c:lxc_cmd:293 - Opened new command socket connection fd 4 for command "get_state"
lxc-stop autopkgtest-stable-amd64 20200603085059.719 DEBUG    commands - commands.c:lxc_cmd_get_state:656 - Container "autopkgtest-stable-amd64" is in "RUNNING" state
lxc-stop autopkgtest-stable-amd64 20200603085059.719 TRACE    commands - commands.c:lxc_cmd_rsp_recv:123 - Command "stop" received response
lxc-stop autopkgtest-stable-amd64 20200603085059.719 DEBUG    commands - commands.c:lxc_cmd_rsp_recv:156 - Response data length for command "stop" is 0
lxc-stop autopkgtest-stable-amd64 20200603085059.719 TRACE    commands - commands.c:lxc_cmd:293 - Opened new command socket connection fd 4 for command "stop"
[...]

PR #3539, opened 2020-09-17
commands: don't fail if unfreeze fails

brauner:
We can e.g. fail the unfreeze because the freezer cgroup is not available and
then we erronously report that stopping the container failed.

Closes: #3471.


stgraber:
Jenkins not looking happy

brauner:
Oh odd...

PR #3567, opened 2020-10-27
lxc-attach: Enable setting the SELinux context

blenk92:
Enable lxc-attach to set the SELinux context that the user will end up
in when attaching to a container (This can be used to overwrite the
context set in the config file). If the option is not used, behavior
will be as before


brauner:
jenkins: test this please

brauner (approved):
Seems reasonable, thank you!

brauner:
Thanks!

PR #3579, opened 2020-11-13
cgfsng: adjust log level to warn instead of error

lifeng68:
These errors do not affect the normal operation. In order to avoid user misunderstanding,  adjust the level  to warning instead of error.

Signed-off-by: lifeng68 <lifeng68@huawei.com>


brauner:
jenkins: test this please

PR #3582, opened 2020-11-17
file_utils: fix config file parsing

brauner:
We accidently used the "bytes_to_write" variable after we've written all the
bytes at which point it is guaranteed to be 0. Let's use the "bytes_read"
variable instead.


brauner:
//Cc @stgraber, @tomponline

PR #3588, opened 2020-12-01
Added fuzz targets for string_util functions.

jesusluna99:
Also added Dockerfile and build.sh for oss-fuzz integration

To build and run fuzz targets:

make CC=clang
clang -fsanitize=fuzzer-no-link,address -Isrc -c <path/to/fuzz_target.c>
clang -fsanitize=fuzzer,address -o fuzzMe fuzz_target.o liblxc.a 
./fuzzMe

Signed-off-by: Jesus Luna <jesusluna99@gmail.com>

### DIFF
--- a/src/lxc/fuzz/Dockerfile
+++ b/src/lxc/fuzz/Dockerfile
@@ -1,0 +1,5 @@
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y make autoconf automake libtool pkgconf
+RUN git clone --depth 1 https://github.com/lxc/lxc.git lxc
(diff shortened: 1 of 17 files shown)


brauner (changes requested):
Thanks for working on this!

I don't think we need to have the Dockerfile and the custom build script. From what it looks like we can just integrate this with the rest of the test files under `/src/test/` and add a new target to the Makefile. @stgraber ?

stgraber:
Yeah, we can't exactly run Docker within our CI environment so having a way to just run the fuzzing code from normal tests would be better.

stgraber:
Closing for inactivity.

PR #3591, opened 2020-12-05
disable ipv6 upon creating a bridge in lxc-net.in and enable it on user specifications

### LINKED ISSUE #1414: Disabling link-local IPv6 on veth interfaces
I am using containers via lxd, see https://github.com/lxc/lxd/issues/2845.  However I am told that this change requires work on the lxc side.

Platform: ubuntu 16.04 with these standard packages:

* liblxc1 2.0.6-0ubuntu1~ubuntu16.04.2
* lxc-common 2.0.6-0ubuntu1~ubuntu16.04.2
* lxcfs 2.0.5-0ubuntu1~ubuntu16.04.1

# Issue description

When lxc attaches containers to bridges, there is a vethXXXXXX interface on the host side and an ethX interface at the container side.

The vethXXXXXX interfaces all get IPv6 link local addresses (which are different from the link-local addresses of the ethX interfaces inside the containers).

With many containers, the host gains many link-local addresses:

~~~
        "ansible_all_ipv6_addresses": [
            "fe80::fc7d:f7ff:fed7:d4f5",
            "fe80::fc2f:1eff:fe51:1633",
            "fe80::fc69:31ff:fe23:2990",
            "fe80::fcf8:a8ff:fed7:7e7",
            "fe80::fc66:8eff:fe07:e10a",
            "fe80::fca2:26ff:feb6:c370",
            "fe80::fc0c:90ff:fe3c:31ef",
            "fe80::fc03:c9ff:fe2d:fcc1",
... etc
~~~

A veth interface is never a source or target for traffic, so should never have its own IPv4 or IPv6 address.

I'd like to disable these link-local addresses: for cleanliness, and to reduce risk of traffic being able to reach bridges that I don't want it to.

This could probably be done by:

~~~
echo 0 > /proc/sys/net/ipv6/conf/vethXXXXXX/autoconf
~~~

and it definitely can be removed by:

~~~
echo 1 > /proc/sys/net/ipv6/conf/vethXXXXXX/disable_ipv6
~~~

However this depends on the actual assigned name, so has to be done when the interface is created. I don't want to set `net.ipv6.conf.default.autoconf=0` as that would have global effect, only for veth interfaces.

# Steps to reproduce

* Start containers
* 'ifconfig -a' on host
* Look at all those link local addresses.

AndrewElvisDeng:
Fixes #1414. This is the proposed solution; please let me know if there's an issue with this solution.

### DIFF
--- a/config/init/common/lxc-net.in
+++ b/config/init/common/lxc-net.in
@@ -78,6 +78,7 @@ start() {
     [ ! -d /sys/class/net/${LXC_BRIDGE} ] && ip link add dev ${LXC_BRIDGE} type bridge
     echo 1 > /proc/sys/net/ipv4/ip_forward
     echo 0 > /proc/sys/net/ipv6/conf/${LXC_BRIDGE}/accept_dad || true
+    echo 1 > /proc/sys/net/ipv6/conf/${LXC_BRIDGE}/disable_ipv6
 
     # if we are run from systemd on a system with selinux enabled,
     # the mkdir will create /run/lxc as init_var_run_t which dnsmasq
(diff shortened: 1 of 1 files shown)


stgraber:
Nope, that's wrong, we definitely do want IPv6 support on the bridge!

What we don't want is IPv6 enabled on the host side veth interface as that causes the host (rather than the container) to incorrectly get itself an IPv6 address from that bridge.

PR #3597, opened 2020-12-12
unmounted proc/sys/net if dropping CAP_NET_ADMIN

zhenr667:


src/lxc/conf.c
```diff
@@ -616,6 +616,7 @@ static int lxc_mount_auto_mounts(struct lxc_conf *conf, int flags, struct lxc_ha
 		const char *fstype;
 		unsigned long flags;
 		const char *options;
+		int net_cap_dropped;
```
brauner, line 619: Probably more obvious to make it:
```c
bool requires_cap_net_admin
```

src/lxc/conf.c
```diff
@@ -666,6 +667,10 @@ static int lxc_mount_auto_mounts(struct lxc_conf *conf, int flags, struct lxc_ha
 		if (!default_mounts[i].destination)
 			return log_error(-1, "BUG: auto mounts destination %d was NULL", i);
 
+		if(!in_caplist(CAP_NET_ADMIN, &conf->caps))
+			if(default_mounts[i].net_cap_dropped)
+				continue;
```
brauner, line 672: Add a bool at the top of the this function:
```c
bool has_cap_net_admin
```

and then before you enter the loop do:

```c
has_cap_net_admin = in_caplist(CAP_NET_ADMIN, &conf->caps);
for (i = 0...

        if (!has_cap_net_admin && default_mounts[i].requires_cap_net_admin)
                continue;
```

This should be more obvious to the reader. You could also add:
```c
if (!has_cap_net_admin && default_mounts[i].requires_cap_net_admin) {
        TRACE("Container does not have CAP_NET_ADMIN. Skipping \"%s\" mount", default_mounts[i].source ?: "(null)");
        continue;
}
```

brauner (changes requested):
Thanks for taking this!
This looks mostly sane. I've left a few comments.

Can you please also make sure that you squash the two commits into a single one, please?

PR #3599, opened 2020-12-13
unmounted proc/sys/net if dropping CAP_NET_ADMIN

zhenr667:


brauner (changes requested):
Fails to compile apparently:

```
../../../src/lxc/conf.c: In function ‘lxc_mount_auto_mounts’:
../../../src/lxc/conf.c:653:2: error: ‘has_cap_net_admin’ undeclared (first use in this function)
  has_cap_net_admin = in_caplist(CAP_NET_ADMIN, &conf->caps);
  ^~~~~~~~~~~~~~~~~
../../../src/lxc/conf.c:653:2: note: each undeclared identifier is reported only once for each function it appears in
```

Issue #3598, opened 2020-12-13
Changes to `src/lxc/conf.c:mount_autodev` from version 4.0.4 to 4.0.5 break containers with ZFS backing stores


ahesford:
# Issue description

After upgrading to version 4.0.5 as packaged in Void Linux (the build options were identical in both versions), containers created with a ZFS backing store cannot be started. The problem appears to be related to changes in `src/lxc/conf.c:mount_autodev` in commit ee8eeba881274bc4f4cd028255d019a4a41f73f3 (cc: @brauner), as evidenced by the debug log. However, I have *not* bisected the issue to this specific commit.

This is a regression with respect to version 4.0.4 and does not affect directory-backed containers.

I only use unprivileged containers, started as root, so I cannot say if the same issue affects privileged containers.

# Required information

- Distribution: Void Linux
- Distribution version: rolling release, up to date
- `lxc-start --version`: 4.0.5
- `lxc-checkconfig`: All items "enabled" EXCEPT: cgroup v1 systemd controller ("missing"); CONFIG_NF_NAT_IPV4 ("missing"); CONFIG_NF_NAT_IPV6 ("missing"). I am *not* using systemd-based containers and the output is identical between the functional 4.0.4 and the affected 4.0.5.
- `uname -a`: `Linux host.domain 5.9.13_1 #1 SMP 1607463776 x86_64 GNU/Linux`
- `cat /proc/self/cgroup`:
```
12:pids:/
11:hugetlb:/
10:net_prio:/
9:perf_event:/
8:net_cls:/
7:freezer:/
6:devices:/
5:memory:/
4:blkio:/
3:cpuacct:/
2:cpu:/
1:cpuset:/
0::/
```
   * `cat /proc/1/mounts`:
```
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
devtmpfs /dev devtmpfs rw,nosuid,noexec,size=8108656k,nr_inodes=2027164,mode=755,inode64 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000 0 0
tmpfs /dev/shm tmpfs rw,nosuid,nodev,noexec,inode64 0 0
tmpfs /run tmpfs rw,nosuid,nodev,noexec,mode=755,inode64 0 0
system/OS/void / zfs rw,relatime,xattr,posixacl 0 0
[...]

PR #3601, opened 2020-12-14
conf: fix block-device based rootfs mounting

brauner:
Fixes: #3598
Cc: stable-4.0


ahesford:
Thanks for your fast turnaround. Applying the patch against 4.0.5 does indeed fix #3598.

PR #3602, opened 2020-12-15
Support nftables using symbolic links to run nftables commands in legacy mode

comannnnndooooo:
This implementation uses the update-alternatives command to set nftables to be running in legacy mode when iptables commands need to be run. Then unsets the symbolic link after finishing the ensure there's no unintended behavior afterwards.

### DIFF
--- a/config/init/common/lxc-net.in
+++ b/config/init/common/lxc-net.in
@@ -92,32 +92,13 @@ start() {
     _ifup
 
     nftables_ver_output=$(nft --version)
-    if [$nftables_ver_output != *"not found"*]; then
-        update-alternatives --set iptables /usr/sbin/iptables-legacy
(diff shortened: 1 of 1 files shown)


stgraber:
This is extremely racy not to mention will potentially mess with kernel state.

What we are looking for is:
 - Move existing iptables calls to internal functions named after what they're doing
 - For each of those functions add support for performing the same action through the nft command
 - At the beginning of the script perform detection to see if the system relies on traditional iptables or nft and use the appropriate code path from that point on

comannnnndooooo:
Ohh okay I understand, I'm on it.

comannnnndooooo:
I reworked the change to move the iptables calls and the new nft calls within their own functions. The script checks at the beginning to see which command set to use.

PR #3604, opened 2020-12-17
fix: gcc8.3.0 and memset overflow

fingera:


stgraber:
You're sending this against the LXC 2.1 branch, LXC 2.1 is unsupported, please retarget this to the master branch if applicable.

src/lxc/network.c
```diff
@@ -2151,8 +2151,8 @@ static int lxc_create_network_unpriv_exec(const char *lxcpath, const char *lxcna
 		return -1;
 	}
 
-	memset(netdev->name, 0, IFNAMSIZ + 1);
```
fingera, line 2154: char name[IFNAMSIZ];

fingera:
> You're sending this against the LXC 2.1 branch, LXC 2.1 is unsupported, please retarget this to the master branch if applicable.

I plan to use version 2.1
Isn't that a good idea?
Should I use 4.0?

fingera:
4.0 is okay

stgraber:
It's not a good idea to use a release which we don't support anymore, no :)
Non LTS releases like 2.1 (as opposed to 2.0) are only supported until the next release comes out, so that particular release was only supported for a few months.

LXC 2.0, 3.0 and 4.0 are all LTS releases that are still supported, though both 2.0 and 3.0 are now in the security-only part of their life, 4.0 is the latest LTS release with full bugfix and security support.

src/lxc/network.c
fingera: @stgraber There seems to be a small security problem here. on 2.x

PR #3634, opened 2021-01-27
mkdir -p /proc /sys on container startup

motiejus:
Some container images don't have it, and strange things happen when they don't. Since these directories are required for normal functioning, it's wise to pre-create them while starting the container.

Signed-off-by: Motiejus Jakštys <motiejus@jakstys.lt>


brauner:
Agreed, that's a good idea. Thanks for the patch!

brauner:
jenkins: test this please

PR #3647, opened 2021-02-02
cgroup2: only rely on command socket when getting cgroup values

brauner:

### DIFF
--- a/src/tests/cgpath.c
+++ b/src/tests/cgpath.c
@@ -104,33 +104,33 @@ static int test_running_container(const char *lxcpath, const char *name)
 		goto err3;
 
 	/* test get/set value using memory.soft_limit_in_bytes file */
-	ret = cgroup_ops->get(cgroup_ops, "memory.soft_limit_in_bytes", value,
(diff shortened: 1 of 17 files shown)


src/tests/cgpath.c
```diff
@@ -104,33 +104,33 @@ static int test_running_container(const char *lxcpath, const char *name)
 		goto err3;
 
 	/* test get/set value using memory.soft_limit_in_bytes file */
-	ret = cgroup_ops->get(cgroup_ops, "memory.soft_limit_in_bytes", value,
-			      sizeof(value), c->name, c->config_path);
-	if (ret < 0) {
+	ret = cgroup_ops->get(cgroup_ops, "pids.max", value, sizeof(value),
```
stgraber, line 107: Won't those changes break on non-cgroup2?

brauner: No, non-cgroup2 has `pids.max` too and the Jenkins runners run on non-cgroup2. :)

stgraber: doh, read cpu.max :)

PR #3661, opened 2021-02-06
network: Add error message if iw couldn't be found

blenk92:
When iw is not installed on the host, lxc fails to move wireless devices
into a container. Unfortunately there is no real error message yet
(other than that it didn't work), so its quite unobvious what causes
this. This commit adds a error message that clearly states the
application iw is missing.


brauner:
Thanks for the pr.
Can you please sign-offs your commit? :)

blenk92:
done :-)

brauner:
jenkins: test this please

PR #3670, opened 2021-02-13
tree-wide: convert to strequal() and strnequal()

brauner:

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -191,7 +191,7 @@ static int exec_criu(struct cgroup_ops *cgroup_ops, struct lxc_conf *conf,
 	 * --enable-fs hugetlbfs --enable-fs tracefs --ext-mount-map console:/dev/pts/n
 	 * +1 for final NULL */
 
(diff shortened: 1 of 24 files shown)


src/lxc/criu.c
```diff
@@ -407,7 +407,7 @@ static int exec_criu(struct cgroup_ops *cgroup_ops, struct lxc_conf *conf,
 			return log_error_errno(-EIO, EIO, "Failed to freezer cgroup entry");
 
 		if (!opts->user->disable_skip_in_flight &&
-				strcmp(opts->criu_version, CRIU_IN_FLIGHT_SUPPORT) >= 0)
+		    strcmp(opts->criu_version, CRIU_IN_FLIGHT_SUPPORT) >= 0)
```
stgraber, line 410: Not converting that one?

brauner: Ah no, because we want to check for greater than not just 0. :)

stgraber: Oh right, missed that.

PR #3686, opened 2021-02-19
apparmor: prefer /proc/.../attr/apparmor/current over legacy interface

### LINKED ISSUE #3685: apparmor: lxc doesn't support new /proc/self/attr/apparmor/... setup
This is related to opencontainers/runc#2801 and docker/for-linux#1199, and is a heads-up since it looks like you have the same bug we did.

It turns out that with certain kernel configurations (such as the one enabled by Arch recently -- where `CONFIG_LSM` has `bpf` as the last entry), the `/proc/self/attr` files no longer refer to AppArmor and instead operate on the BPF LSM. It turns out that some time ago, "major" LSMs got their own subdirectories (`/proc/self/attr/apparmor/...`) which is what we should be using when writing process labels. I believe the same thing goes for SELinux but I haven't tried. [It turns out that `/proc/self/attr/...` files are "legacy"](https://www.kernel.org/doc/html/latest/admin-guide/LSM/index.html):

> Process attributes associated with “major” security modules should be accessed and maintained using the special files in `/proc/.../attr`. A security module may maintain a module specific subdirectory there, named after the module. `/proc/.../attr/smack` is provided by the Smack security module and contains all its special files. The files directly in `/proc/.../attr` remain as legacy interfaces for modules that provide subdirectories.

As far as I can tell, `src/lxc/lsm/apparmor.c` only references `/proc/self/attr/` files so it should have the same issue as us. (You use `/proc/self/attr/current` while we use `/proc/self/attr/exec`, but I believe it's the same issue either way).

I can send a patch for this if you'd like -- it'll look something like opencontainers/runc#2803.

cyphar:
It turns out that since Linux 5.1 there are now per-LSM subdirectories
for major LSMs, [which users are recommended to use over the "legacy"
top-level /proc/$pid/attr/... files][1]:

> Process attributes associated with “major” security modules should be
> accessed and maintained using the special files in /proc/.../attr. A
> security module may maintain a module specific subdirectory there,
> named after the module. /proc/.../attr/smack is provided by the Smack
> security module and contains all its special files. The files directly
> in /proc/.../attr remain as legacy interfaces for modules that provide
> subdirectories.

[AppArmor has had such a directory since Linux 5.8][2], and it turns out
that with certain CONFIG_LSM configurations you can end up with AppArmor
files not being accessible from the legacy interface. Arch Linux
recently added BPF as one of the enabled LSM in their configuration, [and
this broke runc][3] and LXC.

The solution is to first try to use /proc/$pid/attr/apparmor/current and
fall back to /proc/$pid/attr/current if the former is not available.

[1]: https://www.kernel.org/doc/html/latest/admin-guide/LSM/index.html
[2]: https://github.com/torvalds/linux/commit/6413f852ce086c0f95817012c08d481ce24d8b1a
[3]: https://github.com/opencontainers/runc/issues/2801

**NOTE**: I have not tested this thoroughly, I've only compile-tested it.

Fixes #3685
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/src/lxc/lsm/apparmor.c
+++ b/src/lxc/lsm/apparmor.c
@@ -395,52 +395,43 @@ static int apparmor_enabled(struct lsm_ops *ops)
 	return 0;
 }
 
+static int __apparmor_process_label_open(struct lsm_ops *ops, pid_t pid, int o_flags, bool on_exec)
+{
+	int ret = -1;
+	int labelfd;
+	char path[LXC_LSMATTRLEN];
+
+	if (on_exec)
+		TRACE("On-exec not supported with AppArmor");
+
+	ret = snprintf(path, LXC_LSMATTRLEN, "/proc/%d/attr/current", pid);
+	if (ret < 0 || ret >= LXC_LSMATTRLEN)
+		return -1;
+
+	labelfd = open(path, o_flags);
+	if (labelfd < 0)
+		return log_error_errno(-errno, errno, "Unable to open AppArmor LSM label file descriptor");
+
+	return labelfd;
+}
+
 static char *apparmor_process_label_get(struct lsm_ops *ops, pid_t pid)
 {
-	char path[100], *space;
-	int ret;
-	char *buf = NULL, *newbuf;
-	int sz = 0;
-	FILE *f;
+	int label_fd;
+	__do_free char *label = NULL;
+	size_t len;
 
-	ret = snprintf(path, 100, "/proc/%d/attr/current", pid);
-	if (ret < 0 || ret >= 100) {
-		ERROR("path name too long");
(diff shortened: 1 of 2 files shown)


cyphar:
**NOTE**: I have not tested this thoroughly, I've only compile-tested it.

brauner (approved):
How this isn't a giant kernel uapi regression is absolutely beyond me.

brauner:
jenkins: test this please

brauner:
Thank you very much!

cyphar:
Oh you merged it before I fixed the coverity failures. :cry:

cyphar:
I'll make a new PR.

cyphar:
> How this isn't a giant kernel uapi regression is absolutely beyond me.

Yeah this escapes me too. :eyes:

PR #3689, opened 2021-02-23
cgroups: introduce fd-only cgroup attach via LXC_CMD_GET_CGROUP_CTX

brauner:


src/lxc/af_unix.c
```diff
@@ -168,60 +164,168 @@ int lxc_unix_send_fds(int fd, int *sendfds, int num_sendfds, void *data,
...
+		return 0;
+
+	/* If SO_PASSCRED is set we will always get a ucred message. */
+	for (cmsg = CMSG_FIRSTHDR(&msg); cmsg; cmsg = CMSG_NXTHDR(&msg, cmsg)) {
+                if (cmsg->cmsg_level == SOL_SOCKET && cmsg->cmsg_type == SCM_RIGHTS) {
+			__u32 idx;
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-align"
+			int *fds_raw = (int *)CMSG_DATA(cmsg);
+#pragma GCC diagnostic pop
```
stgraber, line 207: What's that?

brauner:
We always compile with -Wcast-align and clang-13 will complain if we cast
like this. But we have no choice because netlink... systemd does it too. I
haven't found a better way yet...

src/lxc/af_unix.c
brauner: I've added a comment in the code now too.

PR #3709, opened 2021-03-17
Initial support for idmapped mounts

brauner:
```
lxc.rootfs.options = idmap=/path/to/user/ns/to/idmap/to
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -652,8 +652,9 @@ AC_CHECK_HEADER([ifaddrs.h],
 AC_HEADER_MAJOR
 
 # Check for some syscalls functions
-AC_CHECK_FUNCS([setns pivot_root sethostname unshare rand_r confstr faccessat gettid memfd_create move_mount open_tree execveat clone3 fsopen fspick fsconfig fsmount, openat2, close_range, statvfs])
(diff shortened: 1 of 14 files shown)


stgraber:
jenkins: test this please

brauner:
> jenkins: test this please

Currently we require the caller to pass us a path to a userns via:
```
lxc.rootfs.option = idmap=</path/to/userns>
```

I think in the future we also want to accept:
```
lxc.rootfs.option = idmap
```
where `idmap` without a path argument means that the caller's user namespace is supposed to be used.
But for that to work we need to rework how LXC starts the container a bit. I think it's a bit of work but not too much.

stgraber:
Looks like the new options aren't documented.
Also does the same apply to additional mounts through `lxc.mount.entry`?

brauner:
> Looks like the new options aren't documented.
> Also does the same apply to additional mounts through `lxc.mount.entry`?

It could but it doesn't just yet. It would require a larger rework how we setup those mounts.

stgraber:
@brauner want to do a rebase and land this one?

brauner:
Yep, rebased now!

stgraber:
jenkins: test this please

PR #3711, opened 2021-03-17
Switch to Github actions

stgraber:


stgraber:
Note that this will fail on gcc+ubuntu-16.04, we need a fix from @brauner so that this can go green.

brauner (approved):

PR #3716, opened 2021-03-19
string_utils: provide a version of strchrnul() in case it's not avail…

brauner:
…able

This should only happen on Android.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -640,7 +640,7 @@ AC_CHECK_HEADER([ifaddrs.h],
 AC_HEADER_MAJOR
 
 # Check for some syscalls functions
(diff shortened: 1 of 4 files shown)


src/lxc/string_utils.c
```diff
@@ -1028,3 +1028,135 @@ void remove_trailing_slashes(char *p)
 	while (--l >= 0 && (p[l] == '/' || p[l] == '\n'))
 		p[l] = '\0';
 }
+
+#ifndef HAVE_STRCHRNUL
+/* Taken from glibc. */
```
stgraber, line 1033: For this kind of stuff we have `lxc/include/` where we track where things come from and their license.

brauner: Ugh, I was trying to avoid that shenanigans...

stgraber: Yeah but `taken from glibc` while our licenses are compatible is a bit light on details for proper tracking of code origin :)

brauner: Done.

PR #3726, opened 2021-03-25
ci: turn on CIFuzz

evverx:
Now that lxc has been integrated into OSS-Fuzz it should be
possible to start using https://google.github.io/oss-fuzz/getting-started/continuous-integration/
(mostly to make sure that the project is buildable there).

It should help to keep the integration in more or less good shape.


brauner:
You are definitely me hero for the forseeable future. :)

brauner:
Thanks!

Issue #3727, opened 2021-03-25
lxc seems to be failing to compile with ASan/UBsan/MSan


evverx:
It seems `liblxc` is built with `-shared -Wl,-no-undefined` by default but unfortunately those flags are incompatible with the sanitizers: https://github.com/google/sanitizers/issues/380. I "fixed" it in https://github.com/google/oss-fuzz/blob/589afcfbd1daafd96dfa178bfc3c3656bc6df81d/projects/lxc/build.sh#L18-L20 with `sed` but it would be great to avoid relying on kludges like that :-)

```
./autogen.sh && CFLAGS=-fsanitize=address ./configure && make -j$(nproc)
...
[...]

PR #3728, opened 2021-03-25
configure: add --enable-fuzz

brauner:
Fixes: #3727

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -746,6 +746,12 @@ AC_SEARCH_LIBS(clock_gettime, [rt])
 # See if we support thread-local storage.
 LXC_CHECK_TLS
 
+# Build with ASAN commands
+AC_ARG_ENABLE([fuzz],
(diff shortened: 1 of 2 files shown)


evverx:
@brauner thanks!

I wonder if it would be better to turn it off with something like `--disable-no-undefined` to make it possible to build lxc with sanitizers regarless of whether the fuzz targets are built or not? This way `--enable-asan` and `--enable-ubsan` could turn  that flag on automatically as well.

brauner:
Oh yeah, that's a good point.

evverx:
Looking at src/lxc/Makefile.am, it seems `liblxc_la_LDFLAGS` takes precedence over `AM_LDFLAGS`:
https://github.com/lxc/lxc/blob/65c5bc571b94345decd779092056c11981f6fea4/src/lxc/Makefile.am#L272-L275

The sed script removes -Wl,-no-undefined from there to compile liblxc on OSS-Fuzz as far as I can tell. Let me double check

brauner:
Closing until we have a better way of dealing with this.

PR #3696, opened 2021-02-25
Rename and move physical interfaces at the same time

samboyles1:
We have come across an problem where containers fail to start because of an issue
when renaming physical interfaces to names of a similar pattern.

With a network config of:
    lxc.net.0.type = veth
    lxc.net.0.name = eth0
    lxc.net.0.flags = up
    lxc.net.0.link = br0
    lxc.net.0.vlan.id = 1
    lxc.net.0.hwaddr = ...
    lxc.net.1.type = phys
    lxc.net.1.name = eth1
    lxc.net.1.flags = up
    lxc.net.1.link = eth2
    lxc.net.2.type = phys
    lxc.net.2.name = eth2
    lxc.net.2.flags = up
    lxc.net.2.link = eth1

lxc will first rename vethXXXXX -> eth0, which is fine, but when it comes to
renaming eth2 -> eth1, it fails with:

ERROR  network - network.c:__instantiate_ns_common:883 - File exists - Failed to rename network device "eth2" to "eth1"
ERROR  network - network.c:lxc_setup_network_in_child_namespaces:3603 - File exists - Failed to setup netdev

It seems to be that lxc is confusing interface names in different network
namespaces when interfaces are being renamed.

This change makes it so that physical interfaces will be renamed at
the same time as they are moved into the containers network namespace.

Signed-off-by: Sam Boyles <sam.boyles@alliedtelesis.co.nz>
Signed-off-by: Blair Steven <blair.steven@alliedtelesis.co.nz>

PR #3731, opened 2021-03-26
network: handle name collisions when returning interfaces to host

samboyles1:
Reviewed-by: Blair Steven <blair.steven@alliedtelesis.co.nz>
Signed-off-by: Sam Boyles <sam.boyles@alliedtelesis.co.nz>

### DIFF
--- a/src/lxc/network.c
+++ b/src/lxc/network.c
@@ -3612,6 +3612,10 @@ static bool lxc_delete_network_priv(struct lxc_handler *handler)
 		}
 
 		if (netdev->type == LXC_NET_PHYS) {
+			/* Physical interfaces are initially returned to the parent namespace
+			 * with their transient name to avoid collisions
+			 */
+			netdev->ifindex = if_nametoindex(netdev->transient_name);
 			ret = lxc_netdev_rename_by_index(netdev->ifindex, netdev->link);
 			if (ret < 0)
 				WARN("Failed to rename interface with index %d "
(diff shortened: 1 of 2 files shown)


samboyles1:
Hi @brauner 
I ran into an issue related to changes made in https://github.com/lxc/lxc/pull/3696

When rebooting our container instances from within the container, the physical network attachments are being returned to the host's network namespace with the same name as they had inside the container. The subsequent rename inside the host namespace may then fail if there is a collision in the same way as it was failing in the initial issue I raised.

I have also cleared netdev->transient_name after the interface has been fully returned to the host because following the reboot process, the interfaces are once again sent to the container, but a check in create_transient_name was returning an error if the transient name was already set, causing lxc to exit every time I tried to reboot my container.

Let me know if this is not the correct way to handle this, but it seems likely that we need to do the same temporary interface naming when returning network interfaces to the host as we do when passing them into the container.

brauner:
jenkins: test this please

brauner:
> Hi @brauner
> I ran into an issue related to changes made in #3696
> 
> When rebooting our container instances from within the container, the physical network attachments are being returned to the host's network namespace with the same name as they had inside the container. The subsequent rename inside the host namespace may then fail if there is a collision in the same way as it was failing in the initial issue I raised.
> 
> I have also cleared netdev->transient_name after the interface has been fully returned to the host because following the reboot process, the interfaces are once again sent to the container, but a check in create_transient_name was returning an error if the transient name was already set, causing lxc to exit every time I tried to reboot my container.
> 
> Let me know if this is not the correct way to handle this, but it seems likely that we need to do the same temporary interface naming when returning network interfaces to the host as we do when passing them into the container.

This is actually a pretty clean solution, so no objections from me! :)

PR #3733, opened 2021-03-26
oss-fuzz: make it possible to build the fuzzer without docker

evverx:
With this patch applied the fuzz target can be built (with ASan)
and run with
```
./src/tests/oss-fuzz.sh
./out/fuzz-lxc-config-read doc/examples/
```

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32475 can be
reproduced by running
```
$ echo "lxc.console.buffer.size=d" >oss-fuzz-32475
$ ./out/fuzz-lxc-config-read ./oss-fuzz-32475
INFO: Seed: 1044753468
INFO: Loaded 1 modules   (18770 inline 8-bit counters): 18770 [0x883cc0, 0x888612),
INFO: Loaded 1 PC tables (18770 PCs): 18770 [0x888618,0x8d1b38),
./out/fuzz-lxc-config-read: Running 1 inputs 1 time(s) each.
Running: oss-fuzz-32475
=================================================================
==2052097==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffcca063e7f at pc 0x000000659e0d bp 0x7ffcca063e30 sp 0x7ffcca063e28
READ of size 1 at 0x7ffcca063e7f thread T0
...
```

I'll point OSS-Fuzz to the build script once this patch is merged.

### DIFF
--- a/src/tests/fuzz-lxc-config-read.c
+++ b/src/tests/fuzz-lxc-config-read.c
@@ -1,0 +1,28 @@
+/* SPDX-License-Identifier: LGPL-2.1+ */
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include "conf.h"
+#include "confile.h"
+#include "lxctest.h"
+#include "utils.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	int fd = -1;
+	char tmpf[] = "fuzz-lxc-config-read-XXXXXX";
+	struct lxc_conf *conf = NULL;
+
+	fd = lxc_make_tmpfile(tmpf, false);
+	lxc_test_assert_abort(fd >= 0);
+	lxc_write_nointr(fd, data, size);
+	close(fd);
+
+	conf = lxc_conf_init();
+	lxc_test_assert_abort(conf);
+	lxc_config_read(tmpf, conf, false);
+	lxc_conf_free(conf);
+
+	(void) unlink(tmpf);
+	return 0;
+}
--- a/src/tests/oss-fuzz.sh
+++ b/src/tests/oss-fuzz.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+set -ex
+
+export SANITIZER=${SANITIZER:-address}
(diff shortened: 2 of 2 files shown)


src/tests/fuzz-lxc-config-read.c
```diff
@@ -0,0 +1,25 @@
...
+#include "utils.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	int fd = -1;
+	char tmpf[] = "fuzz-lxc-config-read-XXXXXX";
+	struct lxc_conf *conf = NULL;
+
+	fd = lxc_make_tmpfile(tmpf, false);
+	lxc_write_nointr(fd, data, size);
+	close(fd);
```
brauner, line 17: This looks like it's missing some error checking. This should check whether `lxc_make_tmpfile()` failed and error out. It should also check for `lxc_conf_init()` failure and not call `lxc_config_read()` in that case and error out instead.

brauner (changes requested):
Just a small change required, I think.

src/tests/fuzz-lxc-config-read.c
evverx: > This looks like it's missing some error checking

Agreed. Given that it isn't expected that it should fail I wonder if something like `lxc_test_assert_abort(fd >= 0);` would suffice. That's what we usually do in the `systemd` fuzz targets :-)

brauner: Yeah, that sounds like a good idea!

evverx: Done

brauner:
Thank you very much!

evverx:
@brauner I'm planning to integrate the fuzz target into the build system eventually (to make it possible to build and run it with something like `./configure --with-fuzzers && make check`) but I can't seem to figure out how to run the tests. As far as I can see, Jenkins runs `./build/lxc-exercise` which in turn runs all the tests but it looks like the script isn't here

brauner:
@stgraber, can you quickly run @evverx through the `lxc-ci` here, please?

stgraber:
The script should be in the deps folder of the lxc-ci repo.

PR #3737, opened 2021-03-26
oss-fuzz: a few follow-up commits

evverx:
* fuzz: generate all the config keys and add them to the seed corpus
* README: add OSS-Fuzz/CIFuzz badges
* fuzz: create tmpfiles in /tmp


brauner:
Looks good!

PR #3741, opened 2021-03-27
confile_utils: free list during lxc_remove_nic_by_idx()

brauner:
Reported-by: Evgeny Vereshchagin <evvers@ya.ru>
Link: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32484


brauner:
//Cc @evverx

evverx:
Thanks! Looks like the memory leak is gone.

To judge from https://github.com/lxc/lxc/pull/3741/checks?check_run_id=2209302973, the fuzz target ran for about 3 minutes without crashing so it's safe to say that it's now capable of catching regressions (with ASan)

PR #3742, opened 2021-03-27
test: introduce a memory leak

evverx:
to make sure CIFuzz can catch it


evverx:
It seems to be working. Closing

PR #3743, opened 2021-03-27
oss-fuzz: fixes

brauner:

### DIFF
--- a/src/lxc/compiler.h
+++ b/src/lxc/compiler.h
@@ -7,6 +7,9 @@
 #define _GNU_SOURCE 1
 #endif
 
+#include <stdbool.h>
+#include <linux/types.h>
+
 #include "config.h"
 
 #ifndef thread_local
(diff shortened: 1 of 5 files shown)


evverx:
With the integer overflow gone it should be possible to unblock UBSan on CIFuzz:
```diff
diff --git a/.github/workflows/cifuzz.yml b/.github/workflows/cifuzz.yml
index 30d4e0d64..074269040 100644
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -29,7 +29,7 @@ jobs:
         with:
           oss-fuzz-project-name: 'lxc'
           fuzz-seconds: 180
-          dry-run: ${{ matrix.sanitizer != 'address' }}
+          dry-run: ${{ matrix.sanitizer == 'memory' }}
           sanitizer: ${{ matrix.sanitizer }}
       - name: Upload Crash
         uses: actions/upload-artifact@v1
```

evverx:
Oh, to judge from https://bugs.webkit.org/show_bug.cgi?id=190208, `__builtin_*` doesn't work with clang on 32-bit machines without compiler-rt :-(
```
Since r183319, __builtin_mul_overflow is used with gcc or clang in WTF/wtf/CheckedArithmetic.h

This leads to a link failure when WebKit is built on Linux with clang and the libgcc runtime:
 BINXX  WPELauncher
output/staging/usr/lib/libWPEWebKit-0.1.so: undefined reference to `__mulodi4'
collect2: error: ld returned 1 exit status 

This is because clang generates code using the __mulodi4 symbol for __builtin_mul_overflow.
But this symbol is available only in compiler-rt, and not in the libgcc runtime used by most Linux distributions of clang.

See also this upstream clang bug: https://bugs.llvm.org/show_bug.cgi?id=28629
```

stgraber:
clang i386 failed with what looks like a regression

brauner:
Android and i386 with clang should now be fixed. The rest are two download failures.

PR #3749, opened 2021-03-29
compiler: fix thread_local detection

brauner:


brauner:
//Cc @evverx

brauner:
This means we always have TLS and remove the ifdef shenanigans.

evverx:
Thanks! I've just compiled lxc on Fedora 32 successfully.

PR #3754, opened 2021-03-30
confile: don't jump into the global table twice

brauner:
instead move networking keys into a subtable. This avoids even just the
remote danger of recursion and also speeds up config parsing.

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -229,29 +229,6 @@ static struct lxc_config_t config_jump_table[] = {
 	{ "lxc.namespace.share",            true,  set_config_namespace_share,            get_config_namespace_share,            clr_config_namespace_share,            },
 	{ "lxc.time.offset.boot",           true,  set_config_time_offset_boot,	   get_config_time_offset_boot,           clr_config_time_offset_boot,           },
 	{ "lxc.time.offset.monotonic",      true,  set_config_time_offset_monotonic,      get_config_time_offset_monotonic,      clr_config_time_offset_monotonic,      },
-	{ "lxc.net.flags",                  true,  set_config_net_flags,                  get_config_net_flags,                  clr_config_net_flags,                  },
-	{ "lxc.net.hwaddr",                 true,  set_config_net_hwaddr,                 get_config_net_hwaddr,                 clr_config_net_hwaddr,                 },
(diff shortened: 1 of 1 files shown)


src/lxc/confile.c
```diff
@@ -229,29 +229,6 @@ static struct lxc_config_t config_jump_table[] = {
...
-	{ "lxc.net.vlan.id",                true,  set_config_net_vlan_id,                get_config_net_vlan_id,                clr_config_net_vlan_id,                },
-	{ "lxc.net.veth.mode",              true,  set_config_net_veth_mode,              get_config_net_veth_mode,              clr_config_net_veth_mode,              },
-	{ "lxc.net.veth.pair",              true,  set_config_net_veth_pair,              get_config_net_veth_pair,              clr_config_net_veth_pair,              },
-	{ "lxc.net.veth.ipv4.route",        true,  set_config_net_veth_ipv4_route,        get_config_net_veth_ipv4_route,        clr_config_net_veth_ipv4_route,        },
-	{ "lxc.net.veth.ipv6.route",        true,  set_config_net_veth_ipv6_route,        get_config_net_veth_ipv6_route,        clr_config_net_veth_ipv6_route,        },
-	{ "lxc.net.veth.vlan.id",           true,  set_config_net_veth_vlan_id,           get_config_net_veth_vlan_id,           clr_config_net_veth_vlan_id,           },
-	{ "lxc.net.veth.vlan.tagged.id",    true,  set_config_net_veth_vlan_tagged_id,    get_config_net_veth_vlan_tagged_id,    clr_config_net_veth_vlan_tagged_id,    },
```
evverx, line 254: `src/tests/oss-fuzz.sh` uses `config_jump_table` to put all the keys into the seed corpus (to make it easier for the fuzz target to cover the parser). With this part of the table gone it seems it should start using `config_jump_table_net` as well. It shouldn't block this PR though. I'll open a follow-up PR once this is merged.

brauner: Yeah. (I once experimented with generating a perfect hash table (Probably still have a branch somewhere.) instead of using a jump table. That actually shouldn't be too difficult to implement by itself. I think the main thing was to make sure that devs have an easy way to regenerate the table via a script or sm.)

PR #3755, opened 2021-03-30
oss-fuzz.sh: put the "lxc.net" keys to the seed corpus as well

evverx:
It's just a follow-up to 0abcc213e2291d71 (where the "lxc.net" keys
were moved from config_jump_table to config_jump_table_net)


brauner:
Excellent, thanks. You're only missing the Signed-off-by line which we require. :) I know, Lennart doesn't like them. :)

evverx:
Fixed

brauner:
Thanks!

PR #3758, opened 2021-03-31
confile: cap to last bit in set_config_net_ipv4_address()

brauner:
Link: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32708

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -781,22 +781,22 @@ static int set_config_net_ipv4_address(const char *key, const char *value,
 	}
 
 	/* No prefix specified, determine it from the network class. */
-	if (prefix) {
+	ret = 0;
+	if (prefix)
 		ret = lxc_safe_uint(prefix, &inetdev->prefix);
-		if (ret < 0)
-			return ret;
-	} else {
+	else
 		inetdev->prefix = config_ip_prefix(&inetdev->addr);
-	}
-	if (inetdev->prefix > 32)
+	if (ret || inetdev->prefix > 32)
 		return ret_errno(EINVAL);
 
-	/* If no broadcast address, let compute one from the
-	 * prefix and address.
-	 */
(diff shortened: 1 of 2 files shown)


brauner:
jenkins: test this please

brauner:
jenkins: test this please

stgraber:
Hmm, the arm64 failure needs investigation as it's not a network failure but instead we're hitting the veth quota...

brauner:
> Hmm, the arm64 failure needs investigation as it's not a network failure but instead we're hitting the veth quota...

I restarted these runners quite a lot because the download failures are increasing.

stgraber:
Each test is running in a LXD container so there can't be leftover veth.

So errors like this one showing veth quota errors aren't random, it's an actual regression.

stgraber:
FAIL: lxc-tests: /usr/bin/lxc-test-usernic (13s)
---
/usr/sbin/deluser: The user `usernic-user' does not exist.
Using image from local cache
Unpacking the rootfs

---
You just created an Ubuntu trusty arm64 (20210331_07:42) container.

To enable SSH, run: apt install openssh-server
No default root or user password are set by LXC.
cmd/lxc_user_nic.c: 1209: main: Quota reached
cmd/lxc_user_nic.c: 1209: main: Quota reached
xx2:2:veth1001_kODz:126
xx3:3:veth1001_35hO:127
cmd/lxc_user_nic.c: 1209: main: Quota reached
cmd/lxc_user_nic.c: 1209: main: Quota reached
FAIL: unable to create nic after destroying the old
FAIL
FAIL
---

brauner:
> Each test is running in a LXD container so there can't be leftover veth.
> 
> So errors like this one showing veth quota errors aren't random, it's an actual regression.

This change doesn't change any of that code so I'm not sure where it would regress...

stgraber:
jenkins: test this please

PR #3772, opened 2021-04-06
templates/*.in: fixed PATH handling with spaces

merlin1991:
if $PATH already contains a path with a space the append of the
default directories in all template scripts fails with an error
like the following:

/usr/share/lxc/templates/lxc-download: 69: export: (x86)/NVIDIA: bad
variable name


stgraber:
Thanks!

Issue #3782, opened 2021-04-13
lxcapi_createl seems to crash without templates


evverx:
It seems all the tests where `lxcapi_createl` is called are run after the "busybox" template has been installed with `make install` (which is probably why it has never crashed on Jenkins). Without the template `lxcapi_createl` crashes with
```
$ ./src/tests/lxc-test-console
lxccontainer.c:1927:8: runtime error: member access within null pointer of type 'struct lxc_conf'
AddressSanitizer:DEADLYSIGNAL
=================================================================
==370477==ERROR: AddressSanitizer: SEGV on unknown address 0x0000000031f8 (pc 0x7fccf3242a6d bp 0x7ffe9322db00 sp 0x7ffe9322d840 T0)
==370477==The signal is caused by a READ memory access.
    #0 0x7fccf3242a6d in do_lxcapi_create /home/vagrant/lxc/src/lxc/lxccontainer.c:1927
    #1 0x7fccf3246075 in lxcapi_createl /home/vagrant/lxc/src/lxc/lxccontainer.c:2184
    #2 0x557e905a339d in test_console /home/vagrant/lxc/src/tests/console.c:152
    #3 0x557e905a3938 in main /home/vagrant/lxc/src/tests/console.c:186
    #4 0x7fccf21b1081 in __libc_start_main (/lib64/libc.so.6+0x27081)
    #5 0x557e905a1a6d in _start (/home/vagrant/lxc/src/tests/.libs/lt-lxc-test-console+0x3a6d)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /home/vagrant/lxc/src/lxc/lxccontainer.c:1927 in do_lxcapi_create
==370477==ABORTING
```

brauner:
Ah, I see the issue.

PR #3785, opened 2021-04-14
lxccontainer: fix container creation error paths

brauner:
Fixes: #3782

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -1775,21 +1775,19 @@ static bool do_lxcapi_create(struct lxc_container *c, const char *t,
 			     const char *bdevtype, struct bdev_specs *specs,
 			     int flags, char *const argv[])
 {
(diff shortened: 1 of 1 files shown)


brauner:
You've compiled with `flto`?

evverx:
I built it with
```
./autogen.sh &&
CFLAGS=-fsanitize=address,undefined ./configure --enable-tests --prefix=/usr/ --sysconfdir=/etc/ --localstatedir=/var/ --disable-no-undefined &&
make -j$(nproc)
```
and by default `-flto=thin` is passed to the compiler/linker as far as I can see

brauner:
I wonder if it's related to `lto` and I've messed up the Makefile at some point. I do have an item to rework the testsuite on my TODO.

evverx:
> I wonder if it's related to lto

I think it could be. FWIW I can't reproduce the crash with neither clang nor newer versions of gcc.

PR #3787, opened 2021-04-15
ci: an attempt to run the tests under ASan/UBsan

evverx:
The idea is to run the tests under ASan/UBsan on GHActions when PRs are opened to prevent memory leaks, heap-buffer-overflows and the like from making it into the lxc repository.


evverx:
Now that it's green I think it should be more or less ready for review. @brauner could you take a look?

brauner:
jenkins: test this please

.github/workflows/sanitizers.sh
```diff
@@ -0,0 +1,67 @@
...
+printf "Memory leaks are ignored due to https://github.com/lxc/lxc/issues/3788.\n"
+ASAN_OPTIONS=$ASAN_OPTIONS:detect_leaks=0 UBSAN_OPTIONS=$UBSAN_OPTIONS /usr/bin/test-concurrent.orig
+EOF
+chmod +x /usr/bin/lxc-test-concurrent
+
+if [[ "$CC" == "clang" ]]; then
+    mv /usr/bin/{lxc-test-console-log,test-console-log.orig}
+    cat <<EOF >/usr/bin/lxc-test-console-log
+#!/bin/bash
+printf "Memory leaks are ignored due to https://github.com/lxc/lxc/issues/3796.\n"
```
brauner, line 55: This should hopefully be fixed with:
https://github.com/lxc/lxc/pull/3790

brauner:
Looks good to me. I think we want to move all the ASAN options (including compiler dependent ones) into the makefile itself in the future.

.github/workflows/sanitizers.sh
evverx: I tried rebasing this PR on top of the master branch (with those commits included) but the test built with clang somehow triggered the same memory leak again. Weirdly, gcc seems to be fine and I can't seem to reproduce it locally with the same version of clang.

PR #3801, opened 2021-04-20
ci: make use of --enable-sanitizers instead of CFLAGS

evverx:


evverx:
It's just a follow-up to https://github.com/lxc/lxc/pull/3787. I think it should make it a little bit easier to build the code with ASan/UBsan. @brauner could you take a look?

brauner:
Thanks!

PR #3818, opened 2021-04-30
oss-fuzz: always turn off logging on OSS-Fuzz

evverx:
Apparently /proc/self/comm can't be used (reliably) on OSS-Fuzz to figure out
whether the code is run inside the fuzz targets, which causes the
fuzz targets to fill the filesystem with log files.

Related: https://github.com/google/oss-fuzz/issues/5509
Should address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33835


brauner:
Thanks!

Issue #3828, opened 2021-05-06
Error after update to Linux zen 5.12.1-zen2-1-zen on Archlinux


pixrl:
# Required information

 * Distribution: Archlinux
 * Distribution version: rolling
 * `lxc-start --version`
```
4.0.8
```
 * `lxc-checkconfig`
```
LXC version 4.0.8
--- Namespaces ---
Namespaces: enabled
Utsname namespace: enabled
Ipc namespace: enabled
Pid namespace: enabled
User namespace: enabled
newuidmap is not installed
newgidmap is not installed
Network namespace: enabled

--- Control groups ---
Cgroups: enabled

Cgroup v1 mount points: 


Cgroup v2 mount points: 
/sys/fs/cgroup

Cgroup v1 systemd controller: missing
Cgroup v1 freezer controller: missing
Cgroup namespace: required
Cgroup device: enabled
Cgroup sched: enabled
Cgroup cpu account: enabled
Cgroup memory controller: enabled
Cgroup cpuset: enabled

--- Misc ---
Veth pair device: enabled, loaded
Macvlan: enabled, not loaded
Vlan: enabled, not loaded
Bridges: enabled, loaded
Advanced netfilter: enabled, not loaded
CONFIG_NF_NAT_IPV4: missing
CONFIG_NF_NAT_IPV6: missing
CONFIG_IP_NF_TARGET_MASQUERADE: enabled, not loaded
CONFIG_IP6_NF_TARGET_MASQUERADE: enabled, not loaded
CONFIG_NETFILTER_XT_TARGET_CHECKSUM: enabled, loaded
CONFIG_NETFILTER_XT_MATCH_COMMENT: enabled, loaded
FUSE (for use with lxcfs): enabled, loaded

--- Checkpoint/Restore ---
checkpoint restore: enabled
CONFIG_FHANDLE: enabled
CONFIG_EVENTFD: enabled
CONFIG_EPOLL: enabled
CONFIG_UNIX_DIAG: enabled
CONFIG_INET_DIAG: enabled
CONFIG_PACKET_DIAG: enabled
CONFIG_NETLINK_DIAG: enabled
File capabilities: 

Note : Before booting a new kernel, you can check its configuration
usage : CONFIG=/path/to/config /usr/bin/lxc-checkconfig

```
   * `uname -a`
```
[...]

PR #3831, opened 2021-05-09
Skip rootfs pinning for ZFS roots.

sjuxax:
Hello! This is definitely a drive-by contribution, only tested through to "Works For Me" status, but it fixes a predicament for me. The last couple versions of LXC failed to start my containers, so I pinned to 4.0.6, but was bitten by the bug discussed at https://github.com/lxc/lxc/issues/3828 after upgrading the kernel to 5.12.1, so couldn't start any unprivileged containers that had ZFS roots.

It seems that the recently-rewritten code for rootfs pinning in conf.c (see https://github.com/lxc/lxc/commit/79ff643d24593a1b77bb39233219d55d20efa4bc) is not considering the case of ZFS roots. The old version reads like it would return -2, i.e., "nothing needed to be pinned", for ZFS roots since the paths would fail to canonicalize. In the new version, containers with ZFS roots fail to start with an error that reads `Bad file descriptor - Failed to open "pool/dataset"`.

This commit checks if `rootfs->bdev_type` is ZFS before the path is analyzed and simply does `goto out`, following the pattern established a few lines later. I'm not sure whether it's correct, but it lets me start my containers again on 5.12.1. :)

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -543,16 +543,13 @@ int lxc_rootfs_init(struct lxc_conf *conf, bool userns)
 			return syserror_set(-EINVAL, "Idmapped rootfs currently only supports the \"dir\" storage driver");
 	}
 
-	if (rootfs->bdev_type && strequal(rootfs->bdev_type, "zfs")) {
-		TRACE("Not pinning because container uses ZFS");
-		goto out;
-	}
-
 	if (rootfs->path) {
-		if (rootfs->bdev_type &&
-		    (strequal(rootfs->bdev_type, "overlay") ||
-		     strequal(rootfs->bdev_type, "overlayfs")))
-			return log_trace_errno(0, EINVAL, "Not pinning on stacking filesystem");
+		if (rootfs->bdev_type) {
+			if (strequal(rootfs->bdev_type, "overlay") || strequal(rootfs->bdev_type, "overlayfs"))
+				return log_trace_errno(0, EINVAL, "Not pinning on stacking filesystem");
+			if (strequal(rootfs->bdev_type, "zfs"))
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

brauner (approved):
This looks good! Thank you! :)

src/lxc/conf.c
```diff
@@ -543,6 +543,11 @@ int lxc_rootfs_init(struct lxc_conf *conf, bool userns)
 			return syserror_set(-EINVAL, "Idmapped rootfs currently only supports the \"dir\" storage driver");
 	}
 
+	if (rootfs->bdev_type && strequal(rootfs->bdev_type, "zfs")) {
+		TRACE("Not pinning because container uses ZFS");
+		goto out;
+	}
```
brauner, line 549: Hm, actually this should be slighlty different. You should move this into the `if (rootfs->path)` condition below where overlayfs is already checked:
```c
	if (rootfs->path) {
		if (rootfs->bdev_type) {
	                if (strequal(rootfs->bdev_type, "overlay") || strequal(rootfs->bdev_type, "overlayfs")))
			        return log_trace_errno(0, EINVAL, "Not pinning on stacking filesystem");
                        if (strequal(rootfs->bdev_type, "zfs"))
                                return log_trace_errno(0, EINVAL, "Not pinning on ZFS filesystem");
	        }
```

brauner (changes requested):
On second thought one small change. :)

src/lxc/conf.c
sjuxax: Agreed, that reads much better. Pushed the change now.

brauner:
jenkins: test this please

brauner (approved):
Thank you!

Issue #3836, opened 2021-05-10
There seems to be a memory leak in cgroup_init


evverx:
It was discovered at https://oss-fuzz-build-logs.storage.googleapis.com/log-d028cc78-2165-40e7-905e-9712f7e4fa64.txt
```
Step #5: /tmp/not-out/fuzz-lxc-cgroup-init -rss_limit_mb=2560 -timeout=25 -seed=1337 -runs=4 < /dev/null
Step #5: INFO: Running with entropic power schedule (0xFF, 100).
Step #5: INFO: Seed: 1337
Step #5: INFO: Loaded 1 modules   (19239 inline 8-bit counters): 19239 [0xaaab80, 0xaaf6a7), 
Step #5: INFO: Loaded 1 PC tables (19239 PCs): 19239 [0xaaf6a8,0xafa918), 
Step #5: INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
Step #5: INFO: A corpus is not provided, starting from an empty corpus
Step #5: #2	INITED cov: 253 ft: 254 corp: 1/1b exec/s: 0 rss: 32Mb
Step #5: 	NEW_FUNC[1/10]: 0x5a1080 in append_unexp_config_line /src/lxc/src/lxc/confile.c:2654
Step #5: 	NEW_FUNC[2/10]: 0x5a1450 in update_hwaddr /src/lxc/src/lxc/confile.c:2628
Step #5: 
Step #5: =================================================================
Step #5: ==192==ERROR: LeakSanitizer: detected memory leaks
Step #5: 
Step #5: Direct leak of 16 byte(s) in 1 object(s) allocated from:
Step #5:     #0 0x5391f3 in realloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:164:3
Step #5:     #1 0x6b9117 in list_add /src/lxc/src/lxc/cgroups/cgfsng.c:79:6
Step #5:     #2 0x6badad in cgroup_hierarchy_add /src/lxc/src/lxc/cgroups/cgfsng.c:451:8
Step #5:     #3 0x6b797a in __initialize_cgroups /src/lxc/src/lxc/cgroups/cgfsng.c:3242:9
Step #5:     #4 0x6a72b8 in initialize_cgroups /src/lxc/src/lxc/cgroups/cgfsng.c:3316:8
[...]

PR #3837, opened 2021-05-11
cgroups: clean up cgroup_ops on initialization error

brauner:
Fixes: #3836


evverx:
Looks like the memory leak is gone. Thanks!

PR #3864, opened 2021-06-12
string utils: Make sure don't return uninitialized memory.

lifeng68:
The function lxc_string_split_quoted and lxc_string_split_and_trim use
realloc to reduce the memory. But the result may be NULL, then the
returned memory will be uninitialized


lifeng68:
@brauner PTAL

brauner:
jenkins: test this please

brauner (approved):
Thanks for the patch! I think that's a good enough fix. Alternatively we could've just returned NULL in that case instead but this is fine too!

Issue #3845, opened 2021-05-19
Support LISTEN_FDS to keep inherited file descriptors open.


r10r:
Hi there,

while implementing https://github.com/lxc/lxcri/issues/13  for lxcri. I noticed that lxc warns about inherited file descriptors.

```
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 3
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 4
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 8
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 9
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 10
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 11
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 12
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 13
lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 14
```

The `LISTEN_FDS` environment variable is mentioned in the OCI spec and used by systemd to support socket activiation.

>  *`LISTEN_FDS`:* The number of file descriptors passed.
>       For example, `LISTEN_FDS=2` would mean that the runtime MUST pass file descriptors 3 and 4 to the container process (in addition to the standard streams) to support [socket activation][systemd-listen-fds].

see https://github.com/opencontainers/runtime-tools/blame/master/docs/command-line-interface.md#L78

What about supporting `LISTEN_FDS` in lxc instead ?

brauner:
> Hi there,
> 
> while implementing [lxc/lxcri#13](https://github.com/lxc/lxcri/issues/13) for lxcri. I noticed that lxc warns about inherited file descriptors.
> 
> ```
> lxc buildah-buildah975299976 20210519121608.238 WARN     start - start.c:lxc_check_inherited:300 - Inherited fd 3
[...]

PR #3865, opened 2021-06-14
Add support for LISTEN_FDS environment variable.

brauner:
The LISTEN_FDS environment variable defines the number of
file descriptors that should be inherited by the container,
in addition to stdio.
The LISTEN_FDS environment variable is defined in the OCI spec
and used to support socket activation.

Refs #3845

Signed-off-by: Ruben Jenster <r.jenster@drachenfels.de>

### DIFF
--- a/src/lxc/lxccontainer.h
+++ b/src/lxc/lxccontainer.h
@@ -223,7 +223,8 @@ struct lxc_container {
 
 	/*!
 	 * \brief Change whether the container wishes all file descriptors
-	 *  to be closed on startup.
+	 *  to be closed on startup. The LISTEN_FDS environment variable
+	 *  can be set to keep inherited file descriptors open.
 	 *
 	 * \param c Container.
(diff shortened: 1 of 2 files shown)


brauner:
@r10r do you need this in stable for `lxcri`?

r10r:
Yes I do. Setting additional groups is required for proper support of the kubernetes security context.
The file system group (fsGroup) maybe different from the user/group of the container process.
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

There are currently two tests in e2e test suite that set a filesystem group different from the provided user / group  that 
will fail with permission denied if the container user is not a member of the filesystem group.

Unfortunately I did not record the test names and broke my test cluster - so I'll have to rerun the tests this evening and report it back.

r10r:
As far as I know it's primarily used by kubernetes (downward API) to share (secret) information between containers which run as different users.

brauner:
Ok, backported it.

r10r:
Thanks!

Issue #3876, opened 2021-06-23
remove_from_array() causes a crash


tomasz-blaszczak-red:
# Required information
The issue is not related to any distribution. 

# Issue description
- `remove_from_array()` causes a crash, the `free()` should be done on `*result`. An array should be `realloc()`ed when an item is removed.
- there is also a potential memory leak in `add_to_array()`. If `strdup()` fails, then an array is not shrink back.

# Steps to reproduce
Call list_active_containers() inside a container where lxc_container_new() is going to fail and returns NULL (e.g. because of get_rundir() returns NULL: get_euid() != 0, no XDG_RUNTIME_DIR and HOME environment variables). Program will crash most likely on next call to add_to_array() or when items are freed.
The following code demonstrates the crash in simpler manner:
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

static inline int string_cmp(char **first, char **second)
{
    return strcmp(*first, *second);
}

static char** get_from_array(char ***names, char *cname, int size)
{
    return (char **)bsearch(&cname, *names, size, sizeof(char *), (int (*)(const void *, const void *))string_cmp);
}

static bool add_to_array(char ***names, char *cname, int pos)
{
    char **newnames = (char**)realloc(*names, (pos+1) * sizeof(char *));
    if (!newnames) {
        fprintf(stderr, "Out of memory\n");
        return false;
    }

    *names = newnames;
    newnames[pos] = strdup(cname);
    if (!newnames[pos])
        return false;

    // sort the arrray as we will use binary search on it
    qsort(newnames, pos + 1, sizeof(char *), (int (*)(const void *,const void *))string_cmp);

    return true;
}

static bool remove_from_array(char ***names, char *cname, int size)
{
    char **result = get_from_array(names, cname, size);
    if (result != NULL) {
        free(result);
        return true;
    }
    return false;
}

void test_realloc()
{
  char **ct_name = NULL;
  int i, ct_name_cnt = 0;
  add_to_array(&ct_name, "Item#01", ct_name_cnt++);
  remove_from_array(&ct_name, "Item#01", ct_name_cnt--);
  add_to_array(&ct_name, "Item#01", ct_name_cnt++);
  add_to_array(&ct_name, "Item#02", ct_name_cnt++);
  add_to_array(&ct_name, "Item#03", ct_name_cnt++);
  add_to_array(&ct_name, "Item#04", ct_name_cnt++);
  for (i = 0; i < ct_name_cnt; i++) {
[...]

PR #3877, opened 2021-06-23
remove_from_array() causes a crash

tomasz-blaszczak-red:
Fix for #3876.

When an item is added to an array, then the array is realloc()ed (to size+1),
and the item is copied (strdup()) to the array.
Thus, when an item is removed from an array, memory allocated for that item
should be freed, successive items should be left-shifted and the array
realloc()ed again (size-1).

Additional changes:
- If strdup() fails in add_to_array(), then an array should be
  realloc()ed again to original size.
- Initialize an array in list_all_containers().

Valgrind summary on the patched version:
```
==476062==
==476062== HEAP SUMMARY:
==476062==     in use at exit: 0 bytes in 0 blocks
==476062==   total heap usage: 13 allocs, 13 frees, 73,880 bytes allocated
==476062==
==476062== All heap blocks were freed -- no leaks are possible
==476062==
==476062== For lists of detected and suppressed errors, rerun with: -s
==476062== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

static inline int string_cmp(char **first, char **second)
{
    return strcmp(*first, *second);
}

static char** get_from_array(char ***names, char *cname, int size)
{
    return (char **)bsearch(&cname, *names, size, sizeof(char *), (int (*)(const void *, const void *))string_cmp);
}

static bool add_to_array(char ***names, char *cname, int pos)
{
    char **newnames = (char**)realloc(*names, (pos+1) * sizeof(char *));
    if (!newnames) {
        fprintf(stderr, "Out of memory\n");
        return false;
    }

    *names = newnames;
    newnames[pos] = strdup(cname);
    if (!newnames[pos]) {
        *names = (char**)realloc(*names, (pos) * sizeof(char *));
        return false;
    }

    // sort the arrray as we will use binary search on it
    qsort(newnames, pos + 1, sizeof(char *), (int (*)(const void *,const void *))string_cmp);

    return true;
}

static bool remove_from_array(char ***names, char *cname, int size)
{
    char **result = get_from_array(names, cname, size);
    if (result != NULL) {
        int i;
        for (i = 0; (*names)[i] != *result && i < size; i++) {
        }
        free(*result);
        memmove(*names+i, *names+i+1, (size-i-1) * sizeof(char*));
        *names = (char**)realloc(*names, (size-1) * sizeof(char *));
        return true;
    }
    return false;
}

void test_realloc()
{
  char **ct_name = NULL;
  int i, ct_name_cnt = 0;
  add_to_array(&ct_name, "Item#01", ct_name_cnt++);
  remove_from_array(&ct_name, "Item#01", ct_name_cnt--);
  add_to_array(&ct_name, "Item#01", ct_name_cnt++);
  add_to_array(&ct_name, "Item#02", ct_name_cnt++);
  add_to_array(&ct_name, "Item#03", ct_name_cnt++);
  add_to_array(&ct_name, "Item#04", ct_name_cnt++);
  for (i = 0; i < ct_name_cnt; i++) {
      printf("i[%d]='%s'\n", i, ct_name[i]);
  }
  remove_from_array(&ct_name, "Item#03", ct_name_cnt--);
  for (i = 0; i < ct_name_cnt; i++) {
      printf("i[%d]='%s'\n", i, ct_name[i]);
  }
  for (i = 0; i < ct_name_cnt; i++) {
      free(ct_name[i]);
  }
  free(ct_name);
}

int main() {
    test_realloc();
    return 0;
}
```

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -2326,7 +2326,7 @@ static bool remove_from_array(char ***names, char *cname, int size)
 		char **newnames = (char**)realloc(*names, (size-1) * sizeof(char *));
 		if (!newnames) {
 			ERROR("Out of memory");
-			return false;
+			return true;
 		}
 
 		*names = newnames;
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -2262,7 +2262,7 @@ static inline int container_cmp(struct lxc_container **first,
 
 static bool add_to_array(char ***names, char *cname, int pos)
 {
-	char **newnames = realloc(*names, (pos+1) * sizeof(char *));
+	char **newnames = (char**)realloc(*names, (pos+1) * sizeof(char *));
 	if (!newnames) {
 		ERROR("Out of memory");
 		return false;
@@ -2270,10 +2270,8 @@ static bool add_to_array(char ***names, char *cname, int pos)
 
 	*names = newnames;
 	newnames[pos] = strdup(cname);
-	if (!newnames[pos]) {
-		*names = (char**)realloc(*names, (pos) * sizeof(char *));
+	if (!newnames[pos])
 		return false;
-	}
 
 	/* Sort the array as we will use binary search on it. */
 	qsort(newnames, pos + 1, sizeof(char *),
@@ -2322,12 +2320,16 @@ static bool remove_from_array(char ***names, char *cname, int size)
 {
 	char **result = get_from_array(names, cname, size);
 	if (result != NULL) {
-		int i;
-		for (i = 0; (*names)[i] != *result && i < size; i++) {
-		}
+		size_t i = result - *names;
 		free(*result);
 		memmove(*names+i, *names+i+1, (size-i-1) * sizeof(char*));
-		*names = (char**)realloc(*names, (size-1) * sizeof(char *));
+		char **newnames = (char**)realloc(*names, (size-1) * sizeof(char *));
+		if (!newnames) {
+			ERROR("Out of memory");
+			return false;
+		}
+
+		*names = newnames;
 		return true;
 	}
 
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -2270,8 +2270,10 @@ static bool add_to_array(char ***names, char *cname, int pos)
 
 	*names = newnames;
 	newnames[pos] = strdup(cname);
-	if (!newnames[pos])
+	if (!newnames[pos]) {
+		*names = (char**)realloc(*names, (pos) * sizeof(char *));
 		return false;
+	}
 
 	/* Sort the array as we will use binary search on it. */
 	qsort(newnames, pos + 1, sizeof(char *),
@@ -2320,7 +2322,12 @@ static bool remove_from_array(char ***names, char *cname, int size)
 {
 	char **result = get_from_array(names, cname, size);
 	if (result != NULL) {
-		free(result);
+		int i;
+		for (i = 0; (*names)[i] != *result && i < size; i++) {
+		}
+		free(*result);
+		memmove(*names+i, *names+i+1, (size-i-1) * sizeof(char*));
+		*names = (char**)realloc(*names, (size-1) * sizeof(char *));
 		return true;
 	}
 
@@ -5659,7 +5666,7 @@ int list_all_containers(const char *lxcpath, char ***nret,
 {
 	int i, ret, active_cnt, ct_cnt, ct_list_cnt;
 	char **active_name;
-	char **ct_name;
+	char **ct_name = NULL;
 	struct lxc_container **ct_list = NULL;
 
 	ct_cnt = list_defined_containers(lxcpath, &ct_name, NULL);


src/lxc/lxccontainer.c
```diff
@@ -2270,8 +2270,10 @@ static bool add_to_array(char ***names, char *cname, int pos)
 
 	*names = newnames;
 	newnames[pos] = strdup(cname);
-	if (!newnames[pos])
+	if (!newnames[pos]) {
+		*names = (char**)realloc(*names, (pos) * sizeof(char *));
```
Blub, line 2274: This `realloc` can fail in which case `*names` must not be updated.
(I also don't think this is really necessary or would help all that much. For a further next successful `add_to_array()` call the `realloc` on top should end up being a nop anyway, and if the `strdup` fails it's easily possible the shrinking here fails, too.)

src/lxc/lxccontainer.c
```diff
@@ -2320,7 +2322,12 @@ static bool remove_from_array(char ***names, char *cname, int size)
 {
 	char **result = get_from_array(names, cname, size);
 	if (result != NULL) {
-		free(result);
+		int i;
+		for (i = 0; (*names)[i] != *result && i < size; i++) {
```
Blub, line 2326: You should be able to easily get the index with `size_t i = result - *names` instead of searching for it.

src/lxc/lxccontainer.c
```diff
@@ -2320,7 +2322,12 @@ static bool remove_from_array(char ***names, char *cname, int size)
...
+		}
+		free(*result);
+		memmove(*names+i, *names+i+1, (size-i-1) * sizeof(char*));
+		*names = (char**)realloc(*names, (size-1) * sizeof(char *));
```
Blub, line 2330: Same as above in `add_to_array`.

Blub (changes requested):

src/lxc/lxccontainer.c
tomasz-blaszczak-red: Hi,
I took back this change. In the current implementation all places where `add_to_array()` fails break further execution and free an array instantly. However, if a usage of `add_to_array()` changes at some time and a new item is added after adding of another one has failed, it will cause a memory leak. See my explanation for shrinking in 'remove_from_array()'.

src/lxc/lxccontainer.c
tomasz-blaszczak-red: Done

src/lxc/lxccontainer.c
tomasz-blaszczak-red: I would disagree here. If you leave it without shrinking, then size of an array becomes bigger than the number of elements the array has. It brings the following issues:
- When iterating over the array you access to elements that are already freed. Such iteration leads to duoble-free at the end of every function that has called remove_from_array(), and to crash. The simple NULLing the element via `*result = NULL` after `free()` does not help too much because of the second issue.
- When adding a new element after one was freed, the call to realloc() causes memory leak, because the number of elements (`pos+1`) is passed to realloc() instead of the real size of the array:
  ```
  char **newnames = (char**)realloc(*names, (pos+1) * sizeof(char *));
  ```

Please, refer to this output produced by Valgring:
```
$ clang++ -std=c++11 -g -O0 -Wall -Wno-disabled-macro-expansion -Wno-writable-strings -Wno-deprecated lxc-top.test.c -o lxc-top.test.out
$ valgrind ./lxc-top.test.out

==1690443== Memcheck, a memory error detector
==1690443== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==1690443== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==1690443== Command: ./lxc-top.test.out
==1690443==
i[0]='Item#01'
i[1]='Item#02'
i[2]='Item#03'
i[3]='Item#04'
i[0]='Item#01'
i[1]='Item#02'
i[2]='(null)'
==1690443==
==1690443== HEAP SUMMARY:
==1690443==     in use at exit: 8 bytes in 1 blocks
==1690443==   total heap usage: 10 allocs, 9 frees, 73,840 bytes allocated
==1690443==
==1690443== LEAK SUMMARY:
==1690443==    definitely lost: 8 bytes in 1 blocks
==1690443==    indirectly lost: 0 bytes in 0 blocks
==1690443==      possibly lost: 0 bytes in 0 blocks
==1690443==    still reachable: 0 bytes in 0 blocks
==1690443==         suppressed: 0 bytes in 0 blocks
==1690443== Rerun with --leak-check=full to see details of leaked memory
==1690443==
==1690443== For lists of detected and suppressed errors, rerun with: -s
==1690443== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

And the code:
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#define ERROR(x) do { fprintf(stderr, x); fprintf(stderr, "\n"); } while (0)

static inline int string_cmp(char **first, char **second)
{
	return strcmp(*first, *second);
}

static char** get_from_array(char ***names, char *cname, int size)
{
	return (char **)bsearch(&cname, *names, size, sizeof(char *), (int (*)(const void *, const void *))string_cmp);
}

static bool add_to_array(char ***names, char *cname, int pos)
{
	char **newnames = (char**)realloc(*names, (pos+1) * sizeof(char *));
	if (!newnames) {
		ERROR("Out of memory");
		return false;
	}

	*names = newnames;
	newnames[pos] = strdup(cname);
	if (!newnames[pos])
		return false;

	// sort the arrray as we will use binary search on it
	qsort(newnames, pos + 1, sizeof(char *), (int (*)(const void *,const void *))string_cmp);

	return true;
}

static bool remove_from_array(char ***names, char *cname, int size)
{
	char **result = get_from_array(names, cname, size);
	if (result != NULL) {
		free(*result);
		*result = NULL;
		return true;
	}
	return false;
}

void test_realloc()
{
	char **ct_name = NULL;
	int i, ct_name_cnt = 0;
	add_to_array(&ct_name, "Item#01", ct_name_cnt++);
	add_to_array(&ct_name, "Item#02", ct_name_cnt++);
	add_to_array(&ct_name, "Item#03", ct_name_cnt++);
	add_to_array(&ct_name, "Item#04", ct_name_cnt++);
	for (i = 0; i < ct_name_cnt; i++) {
		printf("i[%d]='%s'\n", i, ct_name[i]);
	}
	remove_from_array(&ct_name, "Item#03", ct_name_cnt--);
	for (i = 0; i < ct_name_cnt; i++) {
		printf("i[%d]='%s'\n", i, ct_name[i]);
	}
	for (i = 0; i < ct_name_cnt; i++) {
		free(ct_name[i]);
	}
	free(ct_name);
}

int main()
{
	test_realloc();
	return 0;
}

```

brauner:
jenkins: test this please

brauner:
@Blub I'll wait for you to merge this.

src/lxc/lxccontainer.c
```diff
@@ -2320,7 +2320,16 @@ static bool remove_from_array(char ***names, char *cname, int size)
...
 	char **result = get_from_array(names, cname, size);
 	if (result != NULL) {
-		free(result);
+		size_t i = result - *names;
+		free(*result);
+		memmove(*names+i, *names+i+1, (size-i-1) * sizeof(char*));
+		char **newnames = (char**)realloc(*names, (size-1) * sizeof(char *));
+		if (!newnames) {
+			ERROR("Out of memory");
+			return false;
```
Blub, line 2329: The `size` here is tracked by the *caller*, and we just erased an element (`memmoved` a range of elements over it).
The above will log an error, but we still return `false` to the caller, indicating that the element was *not* actually removed and will assume the array has not changed in length.
This means that the caller will now see the final element twice.
I think it's better to just return `true` in either case. Having an extra element allocated and not used is harmless. Having an element in there twice on the other hand may lead to double free corruptions later down the line.

Blub (changes requested):

src/lxc/lxccontainer.c
tomasz-blaszczak-red: Indeed, well spotted, thanks

Blub:
Thanks!

tomasz-blaszczak-red:
Thanks!

Issue #2477, opened 2018-07-17
shared propagation of "mount.entry" does not work


flx42:
Could probably be bundled in #2475

Let's assume I want to bind-mount `/run` as `shared` inside my container:
```
$ findmnt /run -o TARGET,OPT-FIELDS,PROPAGATION
TARGET OPT-FIELDS PROPAGATION
/run   shared:5   shared
```
Currently, we get the following:
```
$ sudo lxc-execute --define 'lxc.mount.entry = /run mnt/run none bind,shared,create=dir 0 0' -n bionic -- findmnt -o TARGET,OPT-FIELDS,PROPAGATION /mnt/run
TARGET   OPT-FIELDS          PROPAGATION
/mnt/run shared:237 master:5 shared,slave
```
This mount inside the container is a slave of the `/run` group (group 5). It should have been `shared:5`.
Doing a mount inside the container will not make the mount visible on the host.

I think what happens is:
- `remount_all_slave` will remount `/run` as `slave`.
- when bind-mounting `/run` inside the container's filesystem, the resulting mount point will be `slave` per the bind semantics of the [kernel documentation](https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt)
- `make_shared` on a slave mount will make it `shared and slave` per the state diagram in the documentation.

brauner:
Yeah, we need to discuss this. There are multiple problems with all of this. One of the main problems is that `pivot_root()` will `no no` you when your rootfs is `MS_SHARED`. I've been thinking about this and I haven't come up with a clean and clear solution yet.

brauner:
So suggestion welcome!

brauner:
Hm, can you test whether this would work for LXC too?

flx42:
Thinking about it more, it looks like this approach is better (slightly more fine-grained) but still has a major issue. Let's take the following example
```
lxc.mount.entry = /run/nvidia mnt/nvidia none bind,shared,create=dir 0 0
lxc.mount.entry = /home/fabecassis/test mnt/test none bind,shared,create=dir 0 0
```

If the parent mount is set to `private` (or `slave`). The behavior of these two lines above will depend on where the `rootfs` is:
* `/run/rootfs`: `/run` will be made `private`, the first mount can't be `shared`.
* `/home/fabecassis/rootfs`: `/home/ will be made `private`, the second mount can't be `shared`.
* `/tmp/rootfs`: `/tmp` or `/` will be made private, both mounts will be fine.

[...]

PR #3880, opened 2021-06-23
Mount propagation fixes

r10r:
Hi there,

I made some changes to allow shared mount propagation and to fix mount option parsing.
As shared mount propagation is considered dangerous and  I'm quite new to it please
review carefully.

There is a pending issue #2477  that may be closed if this is merged.

###  background (mount propgation fixes):
I tried out several CSI (Container Storage Interface) providers to add persistent 
storage for my lxcri/lxc backed kubernetes cluster and noticed that it did not work.
CSI requires`Bidirectional` (shared) mount propagation to work.

See also:
* https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation
* https://kubernetes-csi.github.io/docs/deploying.html#enabling-mount-propagation

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2217,6 +2217,12 @@ static int parse_mntopt(char *opt, unsigned long *flags, char **data, size_t siz
 {
 	size_t ret;
 
+	/* Propagation opts are parsed separately and must not go into data. */
+	for (struct mount_opt *mo = &propagation_opt[0]; mo->name != NULL; mo++) {
+		if (strnequal(opt, mo->name, strlen(mo->name)))
+			return 0;
+	}
+
 	/* If '=' is contained in opt, the option must go into data. */
 	if (!strchr(opt, '=')) {
 		/*
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -2390,6 +2390,10 @@ int parse_mount_attrs(struct lxc_mount_options *opts, const char *mntopts)
 	if (*mntopts_new)
 		opts->data = move_ptr(mntopts_new);
 
+	// Use MS_SLAVE as default mount propagation.
+	if (opts->attr.propagation == 0)
+		opts->attr.propagation = MS_SLAVE;
+
 	return 0;
 }
 
@@ -3936,29 +3940,60 @@ static void turn_into_dependent_mounts(const struct lxc_rootfs *rootfs)
 	 */
 	move_fd(memfd);
 
+	char p_mount[PATH_MAX], p_path[PATH_MAX];
+	bool p_mount_shared, p_path_shared;
+
+	p_mount[0] = '\0';
+	p_path[0] = '\0';
+
 	while (getline(&line, &len, f) != -1) {
 		char *opts, *target;
+		bool shared;
+
+		TRACE("mountinfo: %s", line);
 
 		target = get_field(line, 4);
 		if (!target)
 			continue;
 
 		opts = get_field(target, 2);
-		if (!opts)
-			continue;
-
 		null_endofword(opts);
-		if (!strstr(opts, "shared"))
-			continue;
+
+		if (opts && strstr(opts, "shared"))
+			shared = true;
 
 		null_endofword(target);
-		ret = mount(NULL, target, NULL, MS_SLAVE, NULL);
-		if (ret < 0) {
-			SYSERROR("Failed to recursively turn old root mount tree into dependent mount. Continuing...");
-			continue;
+
+		// Find parent mount of rootfs->path (longest prefix match)
+		if ((strnequal(rootfs->path, target, strlen(target))) &&
+		    (strlen(target) > strlen(p_path))) {
+			(void)strlcpy(p_path, target, sizeof(p_path));
+			p_path_shared = shared;
+		}
+
+		// Find parent mount of rootfs->mount (longest prefix match)
+		if ((strnequal(rootfs->mount, target, strlen(target))) &&
+		    (strlen(target) > strlen(p_mount))) {
+			(void)strlcpy(p_mount, target, sizeof(p_mount));
+			p_mount_shared = shared;
 		}
 	}
-	TRACE("Turned all mount table entries into dependent mount");
+
+	if (p_path_shared) {
+		DEBUG("Change mount propagation for \"%s\" to private (lxc.rootfs.path parent)", p_path);
+		ret = mount(NULL, p_path, NULL, MS_SLAVE, NULL);
+		if (ret < 0) {
+			SYSERROR("Failed to change mount propagation for \"%s\" to private (lxc.rootfs.path parent)", p_path);
+		}
+	}
+
+	if (p_mount_shared) {
+		DEBUG("Change mount propagation for \"%s\" to private (lxc.rootfs.mount parent)", p_mount);
+		ret = mount(NULL, p_mount, NULL, MS_SLAVE, NULL);
+		if (ret < 0) {
+			SYSERROR("Failed to change mount propagation for \"%s\" to private (lxc.rootfs.mount parent)", p_mount);
+		}
+	}
 }
 
 /* This does the work of remounting / if it is shared, calling the container
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -111,7 +111,7 @@ https://www.kernel.org/doc/html/latest/process/coding-style.html
   punctuation sign.
 - They should be descriptive, without being needlessly long. It is best to just
   use already existing error messages as examples.
-- The commit message itself is not subject to rule 4), i.e. it should not be
+- The error message itself is not subject to rule 4), i.e. it should not be
   wrapped at 80chars. This is to make it easy to grep for it.
 - Examples of acceptable error messages are:
   ```C


r10r:
Some more work seems to be done for automounts ...

brauner:
jenkins: test this please

r10r:
@brauner  Currently my changes do break the auto mounts, since the mount-propagation for the auto mounts will change to 'shared' so they will propagate to the host - wich is bad. I tried to bind mount the rootfs and change the rootfs propagation to slave but this did't help - I definitely need your help here! Can you have a look at this ?

r10r:
I'm looking at it right now - hopefully i'll find a solution.

brauner:
> I'm looking at it right now - hopefully i'll find a solution.

I haven't gotten around looking into this yet but mount propagation is tricky. So the requirement that Kubernetes has is that the _volume_ has mount propagation applied not the rootfs itself, right? Because the rootfs itself won't work since `pivot_root()` doesn't work on shared mounts.

r10r:
Right! I think I found the solution right now - simply making '/' shared before the rootfs mount should do it - i'll test again and push my changes and we'll see.

r10r:
sorry - I meant 'slave'  😄

brauner:
> sorry - I meant 'slave'

Ah, right.

r10r:
Hmm, making / 'slave' will break shared mount propgation for bind mounts to '/' again ...

r10r:
@brauner This should be it. What do you think ? 

NOTE: I'll transform my test scripts into a test case 'lxc-test-shared-propagation' later on (tomorrow)

> So the requirement that Kubernetes has is that the volume has mount propagation applied not the rootfs itself, right? Because the rootfs itself won't work since pivot_root() doesn't work on shared mounts.

I thought about that again, and usually it should not be an issue because the container rootfs parent mountpoint is not '/' (e.g when using crio and the overlayfs driver) - but this may be different with other storage drivers and I described the workaround in the commit. But I think the workaround is more relevant when using (plain) lxc containers with the storage backend dir.
I think it would be good to put that into the manpage of *lxc.container.conf* ?

stgraber:
@r10r sorry for the crazy delay, we've got work to do to improve our responsivenes on liblxc changes...

Would you mind rebasing this as the PR no longer applies cleanly?

Once it does, I'll get a new test run and then nag Christian and Alex to do a review.

r10r:
> @r10r sorry for the crazy delay, we've got work to do to improve our responsivenes on liblxc changes...
> 
> Would you mind rebasing this as the PR no longer applies cleanly?
> 
> Once it does, I'll get a new test run and then nag Christian and Alex to do a review.

Hi @stgraber I've rebased to main. 
Please double check the default mount propagations settings in https://github.com/lxc/lxc/blob/3f85b34bc73de539174c6bcda43150cbf64a3539/src/lxc/conf.c#L2393 which created the merge conflict (I did neither build nor test it after rebasing).  Btw. you have switched the build system. Do you provide a *build* container for building liblxc statically and running the tests locally ?

stgraber:
jenkins: test this please

stgraber:
@r10r nope, we don't have a build container. When I need to build things locally I usually go through https://github.com/lxc/lxc/blob/main/.github/workflows/build.yml.

Currently the testsuite run logic is still on Jenkins which is a bit annoying as it uses a bunch of files from different locations. Our short term plan is to move those tests over to Github Actions too which should make it easier for folks to run on their own forks or even locally.

stgraber:
@brauner @mihalicyn can either of you review this one please?

mihalicyn:
I'm on it.

src/lxc/conf.c
```diff
@@ -2217,6 +2217,12 @@ static int parse_mntopt(char *opt, unsigned long *flags, char **data, size_t siz
 {
 	size_t ret;
 
+	/* Propagation opts are parsed separately and must not go into data. */
```
mihalicyn, line 2220: Good catch!

Just one small question here:
Have you tried to reproduce this issue on a recent LXC versions?
I have tried to add `lxc.mount.entry = tmpfs foobar tmpfs rw,rprivate,noexec,nosuid,nodev,create=dir` to reproduce this and haven't succeed. And as I can see from the code this function is called only from a few codepath like storage drivers (btrfs,zfs,overlayfs).

src/lxc/conf.c
```diff
@@ -2390,6 +2390,10 @@ int parse_mount_attrs(struct lxc_mount_options *opts, const char *mntopts)
 	if (*mntopts_new)
 		opts->data = move_ptr(mntopts_new);
 
+	// Use MS_SLAVE as default mount propagation.
```
mihalicyn, line 2393: nit: please use `/* ... */` comments style.

https://github.com/lxc/lxc/blob/main/CODING_STYLE.md#3-only-use---style-comments

src/lxc/conf.c
```diff
@@ -3936,29 +3940,60 @@ static void turn_into_dependent_mounts(const struct lxc_rootfs *rootfs)
...
-		if (!strstr(opts, "shared"))
-			continue;
+
+		if (opts && strstr(opts, "shared"))
+			shared = true;
 
 		null_endofword(target);
-		ret = mount(NULL, target, NULL, MS_SLAVE, NULL);
+
+		// Find parent mount of rootfs->path (longest prefix match)
```
mihalicyn, line 3967: you don't need to do this string comparisons to find parent mount, because `/proc/self/mountinfo` file provides you with this data ;-)

For example:
```
root@ubuntu:/# cat /proc/self/mountinfo
3541 156 259:4 /lxc/ubuntu/rootfs / rw,relatime master:120 - ext4 /dev/nvme1n1p4 rw
3896 3541 0:138 / /dev rw,relatime - tmpfs none rw,size=492k,mode=755,uid=100000,gid=100000,inode64
3897 3541 0:153 / /proc rw,nosuid,nodev,noexec,relatime - proc proc rw
3963 3964 0:153 /sys/net /proc/sys/net rw,nosuid,nodev,noexec,relatime - proc proc rw
```

A first column is the unique `mnt_id` of a mount (host-wide!). A second column shows parent `mnt_id` of the mount.
`/` (`mnt_id = 3541`) is obviously the parent mount for `/dev` (`mnt_id = 3896`) and `/proc` (`mnt_id = 3897`). As you can see both `/dev` and `/proc` have `3541` in the 2nd column (parent mount `mnt_id`).

And also using path-based way of finding parent mounts is not reliable because somebody can create an overmount, for example.

src/lxc/conf.c
```diff
@@ -3930,29 +3940,60 @@ static void turn_into_dependent_mounts(const struct lxc_rootfs *rootfs)
...
+
+	if (p_mount_shared) {
+		DEBUG("Change mount propagation for \"%s\" to private (lxc.rootfs.mount parent)", p_mount);
+		ret = mount(NULL, p_mount, NULL, MS_SLAVE, NULL);
+		if (ret < 0) {
+			SYSERROR("Failed to change mount propagation for \"%s\" to private (lxc.rootfs.mount parent)", p_mount);
 		}
 	}
-	TRACE("Turned all mount table entries into dependent mount");
 }
```
mihalicyn, line 3997: just to ensure that I understand your idea right:
you are changing the `turn_into_dependent_mounts` function such that it stops setting `MS_SLAVE` on all container mounts, but only for `rootfs->path` and `rootfs->mount` parent mounts, right?

r10r: @mihalicyn Yes that was my intention. I tried to describe it in the commit message of  https://github.com/lxc/lxc/pull/3880/commits/3f85b34bc73de539174c6bcda43150cbf64a3539 as best as I can.
I'll have to look for my mount propagation test scripts - they should be a start point for implementing lxc tests.

I share all of your points in the review. But since I've switched to runc as container runtime and paused development of https://github.com/lxc/lxcri I really wonder if anyone benefits from this PR ?

mihalicyn: To be honest I'm not very familiar with the lxcri project ;-)
But I guess that if this issue is blocking lxcri from functioning properly we have to do something with this.

>Yes that was my intention. I tried to describe it in the commit message of https://github.com/lxc/lxc/commit/3f85b34bc73de539174c6bcda43150cbf64a3539 as best as I can.

Your description was great, but I just wanted to ensure and sync our understanding of the problem and approach, because as you've mentioned all of this is very sensitive stuff and can cause a security issues. That's why it's better to ask a few extra questions than having any doubts :)

>But since I've switched to runc as container runtime and paused development

Ah. You was using the lxcri with Kubernetes, right?

brauner (changes requested):
```
To allow pivot_root and avoid unwanted mount propagation to the host
it should be enough to set the propagation for the (parent)
mountpoints of lxc.rootfs.mount and and lxc.rootfs.path to MS_SLAVE.
This will allow bind mounts with the mount option 'shared' and 'rshared'
to stay in the peer group and share propagation.
The default mount propagation for user defined mounts defined by
lxc.mount.entry and lxc.mount.fstab is set to MS_SLAVE.

Shared propagation will not work if a mount shares the parent mount
with either lxc.rootfs.mount or lxc.rootfs.path.
A workaround is to bind mount lxc.rootfs.mount and/or lxc.rootfs.path
or a parent path of it to itself on the host. E.g

mount --bind /usr/lib/lxc/rootfs /usr/lib/lxc/rootfs
mount --make-private /usr/lib/lxc/rootfs
```

Messing with the mount propagation of the parent mount of the rootfs mount is wrong. LXC has no business changing that mount. You could be changing mount propagation of the host rootfs which is completely unacceptable.

What exactly do you need? The rootfs itself _cannot_ be shared. The only thing that you can do is make the rootfs MS_SHARED | MS_SLAVE. Such that it is a slave to the parent (e.g., /) but also is the master in its own peer group. Please see the comment at the end of `lxc_pivot_root()`.

The other thing is making individual `lxc.mount.entry` entries shared mounts. Why doesn't that work exactly?

r10r:
I decided to close this PR because I don't need it anymore, and no one else seems to need it.
The review reveals that there is a lot of work to do for a secure and decent implementation.
Unfortunately I don't have the required knowledge and time for this.
@mihalicyn Thanks for the review - and sorry for taking your time.
Feel free to use any changes from the PR if there is any need for it.

PR #3885, opened 2021-06-28
execute: don't exec init, call it

tych0:
Instead of having a statically linked init that we put on the host fs
somewhere via packaging, have to either bind mount in or detect fexecve()
functionality, let's just call it as a library function. This way we don't
have to do any of that.

This also fixes up a bunch of conditions from:

if (quiet)
    fprintf(stderr, "log message");

to

if (!quiet)
    fprintf(stderr, "log message");

:)

and it drops all the code for fexecve() detection and bind mounting our
init in, since we no longer need any of that.

A couple other thoughts:

* I left the lxc-init binary in since we ship it, so someone could be using
  it outside of the internal uses.
* There are lots of unused arguments to lxc-init (including presumably
  --quiet, since nobody noticed the above); those may be part of the API
  though and so we don't want to drop them.


brauner:
jenkins: test this please

brauner (commented):
I like the idea.
I just had to put a few fixes on top. We need to make sure that the parent is notified about the child's exec. For regular containers this is done automatically when the child exec since the sync fds will be closed during the exec at which point the parent knows the child has exec'ed. But this doesn't happen for the execute/stub-init case as the stub-init exec's the payload process as a child and sticks around so the sync fd is never closed and the parent never notified. In addition we would leak all file descriptors into the stub init process that the monitor held open. So we need to make sure to close all file descriptors upwards of stderr.

brauner:
Thanks, @tycho!

PR #3888, opened 2021-06-30
Improve read-only /sys with read-write /sys/devices/virtual/net

brauner:


bluca:
To close the loop: root cause was found and fixed by https://github.com/systemd/systemd/pull/20075 - nevertheless, it's a nice change to have in LXC, as the setup will be more aligned with expectations, so thanks!

PR #3896, opened 2021-07-05
RFC: conf: userns.conf: include userns.conf.d

Blub:
With `common.conf` including `common.conf.d/` I find it would make sense to have the "common" `userns.conf` include `userns.conf.d/`.

The only downside is that it some people might already be doing this and possibly end up double-including those files...

### DIFF
--- a/config/templates/userns.conf.in
+++ b/config/templates/userns.conf.in
@@ -19,3 +19,6 @@ lxc.tty.dir =
 
 # Setup the default mounts
 lxc.mount.auto = sys:rw
+
+# Lastly, include all the configs from @LXCTEMPLATECONFIG@/userns.conf.d/
+lxc.include = @LXCTEMPLATECONFIG@/userns.conf.d/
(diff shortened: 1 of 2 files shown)


brauner:
So a simple drop-in directory. Yeah, I think that sounds good.

Blub:
Ah somehow I thought that if the include path ended in a `/` an `ENOENT` would be accepted, but we only do an `is_dir()` call checking it via `stat()`.
So would it be better to force the directory to exist at install time, or change include handling so that so that a trailing `/` in the include line forces it to be treated as a directory and ignore `ENOENT`?

brauner:
> Ah somehow I thought that if the include path ended in a `/` an `ENOENT` would be accepted, but we only do an `is_dir()` call checking it via `stat()`.
> So would it be better to force the directory to exist at install time, or change include handling so that a trailing `/` in the include line forces it to be treated as a directory and ignore `ENOENT`?

I like the latter option.

An alternative is to enforce that the drop-in directory is called `userns.conf.d` or in general any drop in directories must end in `*.d`?

Blub:
Not sure enforcing the `*.d` naming would help much, has more potential to break existing setups I think.

PR #3897, opened 2021-07-05
lxc-unshare: fixes

brauner:

### DIFF
--- a/src/lxc/tools/lxc_unshare.c
+++ b/src/lxc/tools/lxc_unshare.c
@@ -416,7 +416,7 @@ int main(int argc, char *argv[])
 			if (lpid == 0) {
 				char buf[256];
 
-				ret = snprintf(buf, 256, "%d", lpid);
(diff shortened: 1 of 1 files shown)


src/lxc/tools/lxc_unshare.c
```diff
@@ -246,8 +246,13 @@ static int do_start(void *arg)
 		}
 	}
 
-	if ((start_arg->flags & CLONE_NEWNS) && start_arg->want_default_mounts)
-		lxc_setup_fs();
+	if (start_arg->flags & CLONE_NEWNS) {
```
hallyn, line 249: pre-morning-coffee question: does this mean that mounts done in the -attach will not be reflected in the container and vice versa?

brauner: Yes. So right now we seem to have the problem that when you run `lxc-unshare` it can end up unmounting filesystems and then it's reflected on the host. Most notably `proc`. If this is ok then I can drop that change!

PR #3898, opened 2021-07-06
lxc: remove CLONE_PIDFD flag from fallback clone() syscall

salonijain0918:
It doesn't work with kernels older than 5.2.

https://man7.org/linux/man-pages/man2/clone.2.html

"       CLONE_PIDFD (since Linux 5.2)"

Fix lxc container startup error:

clone3({flags=CLONE_PIDFD|CLONE_NEWNS|CLONE_NEWUTS|CLONE_NEWIPC|CLONE_NEWPID|0x200000000, pidfd=0x55904a09a0, exit_signal=SIGCHLD, stack=NULL, stack_size=0, /* bytes 80..87 */ "\x17\x00\x00\x00\x00\x00\x00\x00"}, 88) = -1 ENOSYS (Function not implemented)
clone3({flags=CLONE_PIDFD|CLONE_NEWNS|CLONE_NEWUTS|CLONE_NEWIPC|CLONE_NEWPID, pidfd=0x55904a09a0, exit_signal=SIGCHLD, stack=NULL, stack_size=0}, 64) = -1 ENOSYS (Function not implemented)
clone(child_stack=NULL, flags=CLONE_PIDFD|CLONE_NEWNS|CLONE_NEWUTS|CLONE_NEWIPC|CLONE_NEWPID|SIGCHLD, parent_tid=0x55904a09a0) = -1 EINVAL (Invalid argument)

Signed-off-by: Saloni.Jain <jainsaloni0918@gmail.com>

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -1692,9 +1692,10 @@ static int lxc_spawn(struct lxc_handler *handler)
 			TRACE("Spawned container directly into target cgroup via cgroup2 fd %d", cgroup_fd);
 		}
 
-		/* Kernel might be too old for clone3(). */
+		/* Kernel might be too old for clone3() and CLONE_PIDFD. */
 		if (handler->pid < 0) {
 			SYSTRACE("Failed to spawn container via clone3()");
+			handler->clone_flags &= ~CLONE_PIDFD;
 
 		/*
 		 * In contrast to all other architectures arm64 verifies that


src/lxc/start.c
```diff
@@ -1692,9 +1692,10 @@ static int lxc_spawn(struct lxc_handler *handler)
 			TRACE("Spawned container directly into target cgroup via cgroup2 fd %d", cgroup_fd);
 		}
 
-		/* Kernel might be too old for clone3(). */
+		/* Kernel might be too old for clone3() and CLONE_PIDFD. */
 		if (handler->pid < 0) {
 			SYSTRACE("Failed to spawn container via clone3()");
+			handler->clone_flags &= ~CLONE_PIDFD;
```
brauner, line 1698: Hey, thanks for the patch!

What kernel and architecture are you seeing this on?
The legacy `clone()` system call ignores unknown flags so passing `CLONE_PIDFD` shouldn't cause `EINVAL` to be returned.

salonijain0918: Hello, thanks for the reply!

This was encountered on kernel 4.14.180 for aarch64.

brauner: Yep, that's what I suspected. I tried to fix this a while ago. If you look a few lines below you'll see:
```c
		/*
		 * In contrast to all other architectures arm64 verifies that
		 * the argument we use to retrieve the pidfd with is
		 * initialized to 0. But we need to be able to initialize it to
		 * a negative value such as our customary -EBADF so we can
		 * detect whether this kernel supports pidfds. If the syscall
		 * returns and the pidfd variable is set to something >= 0 then
		 * we know this is a kernel supporting pidfds. But if we can't
		 * set it to -EBADF then this won't work since 0 is a valid
		 * file descriptor too. And since legacy clone silently ignores
		 * unknown flags we are left without any way to detect support
		 * for pidfds. So let's special-case arm64 to not fail starting
		 * containers.
		 */
		#if defined(__aarch64__)
			handler->pid = lxc_raw_legacy_clone(handler->clone_flags & ~CLONE_PIDFD, NULL);
		#else
			handler->pid = lxc_raw_legacy_clone(handler->clone_flags, &handler->pidfd);
		#endif
```
So somehow `aarch64` isn't set?

salonijain0918:
As I can check, the above commit is merged in lxc versions 4.0.7, and not available on previous versions, I am using lxc version 4.0.6 version tag.

brauner:
> As I can check, the above commit is merged in lxc versions 4.0.7, and not available on previous versions, I am using lxc version 4.0.6 version tag.

Ah right, I've backported that commit to our stable branch and newer releases have it available. If you want to stay on 4.0.6 you could cherry-pick that fix?

salonijain0918:
Sure, will do that. Thanks!!

brauner:
If this works for you I'm closing the PR here for now. If you run into any more issues, please tell us. :)

salonijain0918:
Sure, thanks a lot! I'm working on upstreaming your solution for lower lxc versions as well.

PR #3909, opened 2021-07-19
bpf: simplify detection if BPF is supported

petris:
bpf_devices_cgroup_supported() tries to load a simple BPF program to test
if BPF works. This is problematic because the function used to load the
program - bpf_program_load_kernel() - emits an error to the log if BPF is
not enabled in the kernel although device controller is not requested in
the configuration.

To detect if BPF is supported we need to check if the BPF syscall is
available. We can do it by passing a NULL pointer instead of the syscall
argument as the kernel returns either ENOSYS, when the sysall is not
implemented or EFAULT, when it is implemented.

### DIFF
--- a/src/lxc/cgroups/cgroup2_devices.c
+++ b/src/lxc/cgroups/cgroup2_devices.c
@@ -538,6 +538,10 @@ bool bpf_devices_cgroup_supported(void)
 		return log_trace(false,
 				 "The bpf device cgroup requires real root");
 
+	ret = bpf(BPF_PROG_LOAD, NULL, sizeof(union bpf_attr));
+	if (ret < 0 && errno == ENOSYS)
+		return log_trace(false, "The bpf syscall is not available");
+
 	prog = bpf_program_new(BPF_PROG_TYPE_CGROUP_DEVICE);
 	if (!prog)
 		return log_trace(false, "Failed to allocate new bpf device cgroup program");


brauner:
> bpf_devices_cgroup_supported() tries to load a simple BPF program to test
> if BPF works. This is problematic because the function used to load the
> program - bpf_program_load_kernel() - emits an error to the log if BPF is
> not enabled in the kernel although device controller is not requested in
> the configuration.
> 
> To detect if BPF is supported we need to check if the BPF syscall is
> available. We can do it by passing a NULL pointer instead of the syscall
> argument as the kernel returns either ENOSYS, when the sysall is not
> implemented or EFAULT, when it is implemented.

Hey, thanks for the patch!
Hm, we don't want to just test that `bpf()` is available. We do want to test whether `bpf()` is available and whether the `bpf()` subsystem supports the device cgroup. So this change would give the wrong impression that the kernel supports the bpf devices cgroup even though it doesn't.

I can think of two solutions:
1. Split the test in two steps, i.e. first perform a `bpf()` syscall with a `NULL` pointer and if this works perform it again with a dummy bpf devices program.
2. Only perform the `bpf()` test if the devices cgroup for `lxc.cgroup2.devices.*` has been requested.

Probably a combination of the 2 would work.

petris:
You are right, one can enable bpf syscall without cgroup support. I will rework the patch.

brauner:
> You are right, one can enable bpf syscall without cgroup support. I will rework the patch.

Thanks. Please make sure to sign-off your commits with `git commit -s` or `git commit --amend -s`.

petris:
Updated - just simply check if the syscall is there before doing anything else.

brauner:
Thanks!

PR #3910, opened 2021-07-19
lxc_setup_ttys: Handle existing ttyN file without underlying device

petris:
If a device file is opened and there isn't the underlying device,
the open call fails with ENXIO, but the path can be opened with
O_PATH, which is enough for mounting over the device file.

Generalize this idea and use O_PATH for all cases when the file
is there. One still must check for both ENXIO and EEXIST as it's
unspecified what error is reported if multiple error conditions
occur at the same time.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -942,6 +942,19 @@ static bool append_ttyname(char **pp, char *name)
 	return true;
 }
 
+static int open_ttymnt_at(int dfd, const char *path)
+{
+	int fd;
+
+	fd = open_at(dfd, path, PROTECT_OPEN | O_CREAT | O_EXCL,
+		     PROTECT_LOOKUP_BENEATH, 0);
+	if (fd < 0 && (errno == ENXIO || errno == EEXIST))
+		fd = open_at(dfd, path, PROTECT_OPATH_FILE,
+			     PROTECT_LOOKUP_BENEATH, 0);
+
+	return fd;
+}
+
 static int lxc_setup_ttys(struct lxc_conf *conf)
 {
 	int ret;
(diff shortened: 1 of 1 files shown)


brauner:
Can you please sign-off your commit? :)

petris:
> Can you please sign-off your commit? :)

Updated

src/lxc/conf.c
```diff
@@ -942,6 +942,19 @@ static bool append_ttyname(char **pp, char *name)
...
 
+static int open_ttymnt_at(int dfd, const char *path)
+{
+	int fd;
+
+	fd = open_at(dfd, path, PROTECT_OPEN | O_CREAT | O_EXCL,
+		     PROTECT_LOOKUP_BENEATH, 0);
+	if (fd < 0 && (errno == ENXIO || errno == EEXIST))
+		fd = open_at(dfd, path, PROTECT_OPATH_FILE,
+			     PROTECT_LOOKUP_BENEATH, 0);
```
brauner, line 953: Hm, I'd like to understand when this happens. So, if we used `lxc.autodev = 1` then we've mounted a new `tmpfs` instance over `/dev` which means the device files can't possible exist before. Can you outline your usecase?

petris: In my scenario devtmpfs is mounted on /dev, because containers are used to run multiple instances of SW which has been designed to exist only in one instance on the system and creating per container dev would make it more complex. The host distribution creates /dev/tty1 and as there is just one devtmpfs instance in Linux, it exists in the container as well and because the underlying device is not there, LXC fails to open it and stops.

This used to work in the past as this issue popped up after updating to LXC4 without changing anything else.

brauner: Ah, ok thank you.

brauner:
Thank you! Backported this to `stable-4.0` as well.

PR #3924, opened 2021-08-06
mainloop: io_uring support

brauner:
Users can choose to compile liblxc with io_uring support. This will
cause LXC to use io_uring instead of epoll.
We're using both, io_uring's one-shot and multi-shot poll mode depending
on the type of handler.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -671,6 +671,22 @@ AC_CHECK_HEADER([ifaddrs.h],
 	AC_DEFINE(HAVE_IFADDRS_H, 1, [Have ifaddrs.h]),
 	AM_CONDITIONAL(HAVE_IFADDRS_H, false))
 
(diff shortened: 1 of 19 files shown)


stgraber:
@brauner looks like this is break lxc-test-concurrent (timing out in Github Actions)

brauner:
> @brauner looks like this is break lxc-test-concurrent (timing out in Github Actions)

Yeah, this isn't done yet. I converted it from draft to review since I had a working version based on io_uring oneshot mode. But now I added support for multishot and will allow to mix and match oneshot and multishot mode depending on the handler.
However, this will need a revamp of the cleanup logic that we currently use in our event loops. Which is what I'm doing. I also need to make sure to keep our epoll implementation working of course. :)

PR #3933, opened 2021-08-11
lxc-start SIGABRT on systems with defined offline cpus and a total nu…

jimferr:
…mber of cpus divisible by 32

Details: https://discuss.linuxcontainers.org/t/lxc-4-0-9-lxc-start-sigabrt-on-systems-with-defined-offline-cpus-and-a-total-number-of-cpus-divisible-by-32/11783

Signed-off-by: Jim Ferrigno <jim.ferrigno@oracle.com>

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -250,7 +250,7 @@ static char *lxc_cpumask_to_cpulist(uint32_t *bitarr, size_t nbits)
 	char numstr[INTTYPE_TO_STRLEN(size_t)] = {0};
 	int ret;
 
-	for (size_t i = 0; i <= nbits; i++) {
+	for (size_t i = 0; i < nbits; i++) {
 		if (!is_set(i, bitarr))
 			continue;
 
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

brauner (approved):
Thank you!

brauner (changes requested):
Looking at this closely there's still a bug in there. Namely, when neither offline nor isolated cpus are present then the current code using `<=` is correct which is due to the fact that we don't increment `maxposs` after we read it. I'll pull your change and fix it up locally and the push it out.

brauner:
@jimferr I merged your patch in https://github.com/lxc/lxc/pull/3934 in an altered version. Can you verify this fixes things for you, please?

jimferr:
@brauner I reviewed the changes in #3934 and have comments about the change.    To note the changes in #3934 introduce another off by one error that can result in memory corruption.  I will continue the discussion in #3934

brauner:
@jimferr, ok. Thank you. I'll wait for your reply.

PR #3934, opened 2021-08-12
cgroups: cpumask fixes

brauner:

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -596,7 +596,6 @@ static bool cpuset1_cpus_initialize(int dfd_parent, int dfd_child,
 
 		if (maxposs < maxisol)
 			maxposs = maxisol;
-		maxposs++;
 	} else {
 		TRACE("The path \""__ISOL_CPUS"\" to read isolated cpus from does not exist");
 	}
@@ -615,7 +614,6 @@ static bool cpuset1_cpus_initialize(int dfd_parent, int dfd_child,
 
 		if (maxposs < maxoffline)
 			maxposs = maxoffline;
-		maxposs++;
 	} else {
 		TRACE("The path \""__OFFLINE_CPUS"\" to read offline cpus from does not exist");
 	}
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -215,7 +215,7 @@ static uint32_t *lxc_cpumask(char *buf, size_t nbits)
 	size_t arrlen;
 
 	arrlen = BITS_TO_LONGS(nbits);
-	bitarr = calloc(arrlen, sizeof(uint32_t));
+	bitarr = zalloc(arrlen * sizeof(uint32_t));
 	if (!bitarr)
 		return ret_set_errno(NULL, ENOMEM);
 


jimferr:
General Comment:  Array indexing and object counting leads to these type of errors.  I always found it helpful to have comments in the code to avoid these type of mismatches.

@brauner  using file cgfsng.c as the reference for the comments below:

- The removal of maxposs++ at lines 599 and 618 introduces another off by one error which was perhaps previously discovered and perhaps the reason for lines 599 and 618 in the first place.  By way of example to illustrate the issue lets have the OFFLINE CPU range from 15-32 (and assume the ONLINE CPU range was 0-14) so we 33 CPUs in total and maxposs value is 32.  When lxc_cpumask() is called it will allocate an array of sizeof(uint32_t) size of 1 when it should have been 2 to accommodate the 33rd CPU.  This will lead to memory corruption.
- Again the confusion of array indexing and object counting come into play.
- Two things that need to happen / decided upon and applied consistently for the maxposs.  Does maxposs represent the maximum from the array indexing (i.e. starting count from 0) or object counting (i.e. starting count from 1) or object counting perspective.  And it may have to be normalized to account for computers that may be accounting for CPUs based on array indexing or object counting... To Note: I have no idea if there are any systems out there that start cpu counts from 1 instead of 0.
- To aide in the decision on whether or not maxposs needs to be normalized (this is after the first decision has been made for how lxc treats maxposs) another function say get_min_cpu() could get the minimum / staring value for the cpu which would be 0 or 1.

brauner:
Ah you working on another patch @jimferr?

jimferr:
However you think `maxposs` should reflect is fine but keep in mind that some systems may starting counting the CPUs from 1 ... As I said earlier I have no idea if any such do.  I look forward to your finalized solution!

On another topic related to this which was brought up in the original discussion I questioned the logic of the following code snippet in cpuset1_cpus_initialize() and asked if it should be reversed.

`        
        if (!flipped_bit) {
                cpulist = lxc_cpumask_to_cpulist(possmask, maxposs);
                TRACE("No isolated or offline cpus present in cpuset");
        } else {
                cpulist = move_ptr(posscpus);
                TRACE("Removed isolated or offline cpus from cpuset");
        }
`

brauner:
> However you think `maxposs` should reflect is fine but keep in mind that some systems may starting counting the CPUs from 1 ... As I said earlier I have no idea if any such do. I look forward to your finalized solution!

If such systems exist that would break in interesting places, I would assume but maybe we can remove that assumption altogether.

> 
> On another topic related to this which was brought up in the original discussion I questioned the logic of the following code snippet in cpuset1_cpus_initialize() and asked if it should be reversed.
> 
> ```
>         if (!flipped_bit) {
>                 cpulist = lxc_cpumask_to_cpulist(possmask, maxposs);
>                 TRACE("No isolated or offline cpus present in cpuset");
>         } else {
>                 cpulist = move_ptr(posscpus);
>                 TRACE("Removed isolated or offline cpus from cpuset");
>         }
> ```

jimferr:
@brauner Assuming another fix / PR based on what we discussed is forthcoming can I be put on the list to review it.

brauner:
> @brauner Assuming another fix / PR based on what we discussed is forthcoming can I be put on the list to review it.

Sure, I just pinged you on a PR. I'm putting it into draft state for now.

PR #3938, opened 2021-08-16
cgroups: simplify offline and isolated cpumask handling

brauner:

//Cc @jimferr

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -262,6 +262,50 @@ static int lxc_cpumask(char *buf, uint32_t **bitarr, size_t *last_set_bit)
 	return 0;
 }
 
+static int lxc_cpumask_update(char *buf, uint32_t *bitarr, size_t last_set_bit,
+			      bool clear)
+{
+	bool flipped = false;
+	char *token;
+
+	lxc_iterate_parts(token, buf, ",") {
+		unsigned last_bit, first_bit;
+		char *range;
+
+		errno = 0;
+		first_bit = strtoul(token, NULL, 0);
+		last_bit = first_bit;
+		range = strchr(token, '-');
+		if (range)
+			last_bit = strtoul(range + 1, NULL, 0);
+
+		if (!(first_bit <= last_bit)) {
+			WARN("The cup range seems to be inverted: %u-%u", first_bit, last_bit);
+			continue;
+		}
+
+		if (last_bit > last_set_bit)
+			continue;
+
+		while (first_bit <= last_bit) {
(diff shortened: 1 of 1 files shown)


src/lxc/cgroups/cgfsng.c
```diff
@@ -554,12 +598,9 @@ static bool cpuset1_cpus_initialize(int dfd_parent, int dfd_child,
...
-	__do_free uint32_t *isolmask = NULL, *offlinemask = NULL,
-			   *possmask = NULL;
+	__do_free uint32_t *possmask = NULL;
 	int ret;
-	size_t isol_last_set_bit = 0, offline_last_set_bit = 0,
-	       poss_last_set_bit = 0;
-	bool flipped_bit = false;
+	size_t poss_last_set_bit = 0;
 
 	posscpus = read_file_at(dfd_parent, "cpuset.cpus", PROTECT_OPEN, 0);
```
jimferr, line 605: I don't think it is a foregone conclusion that cpuset.cpus file provides all possible cpus.  In a system I am looking at comprised of 448 cpus viewing /sys/fs/cgroup/cpuset/cpuset.cpus it lists only the on-line cpus which happens to be 0-103.  /sys/devices/system/cpu/possible actually provides the entire range of cpus.

brauner: But that shouldn't be relevant. We only care about the cpus that are available according to `cpuset.cpus`. Where would "possible" be relevant?

src/lxc/cgroups/cgfsng.c
```diff
@@ -262,6 +262,50 @@ static int lxc_cpumask(char *buf, uint32_t **bitarr, size_t *last_set_bit)
...
+
+		errno = 0;
+		first_bit = strtoul(token, NULL, 0);
+		last_bit = first_bit;
+		range = strchr(token, '-');
+		if (range)
+			last_bit = strtoul(range + 1, NULL, 0);
+
+		if (!(first_bit <= last_bit)) {
+			WARN("The cup range seems to be inverted: %u-%u", first_bit, last_bit);
```
jimferr, line 283: Minor comment: the word "cup" should be "cpu".

brauner: Yup, thanks.

src/lxc/cgroups/cgfsng.c
jimferr: Ahh.  I did not notice that lxc_cpumask() was changed and added the third argument for last_set_bit so in the case I cited above it would be a NOP as there are no offlines within the range of the onlines.  And the new code handles cases where offlines would be within the range of cpuset.cpus.

PR #3946, opened 2021-08-23
Fix attach c

blenk92:
Currently, the -c command (to set the selinux context) seems to be
broken because lxc-attach expects that also a new mount namespace
is specified via command line. This commit remove the check for the new
mount namespace to fix this issue. Please note that the
--elevated-privileges option is not affected by this issue.

Current situation:
```
~ # lxc-attach -n container --elevated-privileges=LSM
~ # id
uid=0(root) gid=0(root) context=u:r:unconfined_t:s0
# can not enter container
~ # lxc-attach -n container -c u:r:unconfined_t:s0
lxc-attach: node1: attach.c: lxc_attach_run_shell: 1860 Permission denied - Failed to execute shell
```

Expected:
```
~ # lxc-attach -n container --elevated-privileges=LSM
~ # id
uid=0(root) gid=0(root) context=u:r:unconfined_t:s0
# works
~ # lxc-attach -n container -c u:r:unconfined_t:s0
# in container now
~ # id
uid=0(root) gid=0(root) context=u:r:unconfined_t:s0
```

Please also note that when --elevated-privileges=LSM and -c are specified the -c context is silently ignored. This was pretty confusing to me since `-c u:r:unconfined_t:s0` didn't work (due to the bug) but `-c u:r:init_t:s0 -e` did work but transited into the `unconfined_t` context. Thus, i've added a patch to make the -c and the lsm privileges option exclusive.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1430,6 +1430,11 @@ int lxc_attach(struct lxc_container *container, lxc_attach_exec_t exec_function,
 		return log_error(-1, "Failed to get attach context");
 	}
 
+	if (options->lsm_label) {
+		free(ctx->lsm_label);
+		ctx->lsm_label = strdup(options->lsm_label);
+	}
+
 	conf = ctx->container->lxc_conf;
 
 	if (!fetch_seccomp(ctx->container, options))
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -3252,7 +3252,7 @@ int lxc_config_parse_arch(const char *arch, signed long *persona)
 	return ret_errno(EINVAL);
 }
 
-int lxc_fill_elevated_privileges(char *flaglist, int *flags)
+int lxc_fill_elevated_privileges(char *flaglist, int *flags, const char *selinux_context)
 {
 	char *token;
 	int i, aflag;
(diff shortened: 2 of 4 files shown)


brauner:
jenkins: test this please

blenk92:
Hi @brauner saw you spend some effort to create the other PR. Unfortunately i got tricked by another issue in our system (labeling wasn't working properly so i accidentally ended up in the desired context without violations). I spend some more time debugging whats going on and i think i found the root cause now: The context passed via the -c is completely ignored and always the context specified in the config file is taken. This patch should fix the issue by just overwriting the context parsed from the config file if a context has been passed via command line

brauner:
> Hi @brauner saw you spend some effort to create the other PR. Unfortunately i got tricked by another issue in our system (labeling wasn't working properly so i accidentally ended up in the desired context without violations). I spend some more time debugging whats going on and i think i found the root cause now: The context passed via the -c is completely ignored and always the context specified in the config file is taken. This patch should fix the issue by just overwriting the context parsed from the config file if a context has been passed via command line. Sorry for the confusion...

Ah. Should we still merge the other branch  I sent and you sent a separate PR on top of that?

brauner:
Closing this PR as we've merged your patches and a few additional ones in another PR. Please do send a PR in case there's anything to fix still!

PR #3947, opened 2021-08-23
config: enable seccomp profile only when compiled with libseccomp

blenk92:
Make lxc fail if seccomp.profile is specified but lxc is compiled without seccomp support. Currently, seccomp.profile is silently ignored if is specified in such a scenario. This could lead to the false impression that the seccomp filter is applied while it actually isn't.

With this patch make the behavior of lxc.seccomp.profile compliant to the other lxc.seccomp,* options and will thus prevent a situation as described above.

Signed-off-by: Maximilian Blenk <Maximilian.Blenk@bmw.de>

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -1211,7 +1211,11 @@ static int set_config_seccomp_notify_proxy(const char *key, const char *value,
 static int set_config_seccomp_profile(const char *key, const char *value,
 				      struct lxc_conf *lxc_conf, void *data)
 {
+#ifdef HAVE_SECCOMP
 	return set_config_path_item(&lxc_conf->seccomp.seccomp, value);
+#else
+	return ret_set_errno(-1, ENOSYS);
+#endif
 }
 
 static int set_config_execute_cmd(const char *key, const char *value,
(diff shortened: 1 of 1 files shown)


brauner:
jenkins: test this please

brauner:
Thanks! I don't expect this to break anyone but there's a slight chance. :)

blenk92:
Yeah, actually i was thinking if it makes sense to move the seccomp filter from common.conf into a seperate confg file, but then on the other hand this would silently turn the filter off which is probably way worse... Maybe that seccomp filter there should be commented out if seccomp is not enabled. What do you think?

brauner:
> Maybe that seccomp filter there should be commented out if seccomp is not enabled. What do you think?

You mean during build?

blenk92:
> > Maybe that seccomp filter there should be commented out if seccomp is not enabled. What do you think?
> 
> You mean during build?

yes, during build ;-)

brauner:
I tend to think this is a good idea under the assumption that config files and shared library form a unit, i.e. if you update the shared library then you're also expected to update the config files. @stgraber what's your take on this?

stgraber:
Yeah, makes sense I think

brauner:
@blenk92 feel free to send a patch!

PR #3963, opened 2021-09-03
build: add basic meson support

brauner:
Allow the shared library to be compiled with meson.


meson.build
```diff
@@ -0,0 +1,624 @@
...
+threads = dependency('threads')
+libseccomp = cc.find_library('seccomp', required : false)
+if libseccomp.found()
+	conf.set10('HAVE_SECCOMP', libseccomp.found())
+
+	seccomp_headers = '''
+	#include <seccomp.h>
+	'''
+
+	if cc.has_function('seccomp_notify_fd', prefix : seccomp_headers, args : '-D_GNU_SOURCE', dependencies: libseccomp)
```
eli-schwartz, line 247: libseccomp comes with a pkg-config file, so you can just use

```
libseccomp = dependency('libseccomp')
if libseccomp.version().version_compare('>=2.5.0')
    # https://github.com/seccomp/libseccomp/commit/dead12bc788b259b148cc4d93b970ef0bd602b1a
    conf.set10('HAVE_DECL_SECCOMP_NOTIFY_FD', true)
endif
if libseccomp.version().version_compare('>=2.0.0')
    # https://github.com/seccomp/libseccomp/commit/6220c8c0fc479d97b6d3e3166a4e46fbfe25a3c0
    conf.set10('HAVE_DECL_SECCOMP_SYSCALL_RESOLVE_NAME_ARCH', true)
endif
```

meson.build
```diff
@@ -0,0 +1,624 @@
...
+	        # We get -1 if the size cannot be determined
+	        have = cc.sizeof(decl, prefix : seccomp_headers, args : '-D_GNU_SOURCE') > 0
+
+		if have == true
+			conf.set10('HAVE_' + decl.underscorify().to_upper(), have)
+		endif
+	endforeach
+endif
+
+libselinux = cc.find_library('selinux', required : false)
```
eli-schwartz, line 270: Has a pkg-config file.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+if libselinux.found()
+	conf.set10('HAVE_SELINUX', libselinux.found())
+endif
+
+libapparmor = cc.find_library('apparmor', required : false)
```
eli-schwartz, line 275: Has a pkg-config file.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+	conf.set10('HAVE_OPENSSL', libopenssl.found())
+endif
+
+libcap = dependency('libcap', required : false)
+if not libcap.found()
+        # Compat with Ubuntu 14.04 which ships libcap w/o .pc file
+        libcap = cc.find_library('cap')
+endif
+
+if libcap.found()
```
eli-schwartz, line 291: If it isn't found via pkg-config you fall back on checking for it with cc.find_library using the default `required: true`, so this condition can never be false. I think for consistency with the other checks, you meant to use `required: false` there too.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+
+config_h = configure_file(
+        output : 'config.h',
+        configuration : conf)
+
+liblxc = shared_library(
+        'lxc',
+        version : liblxc_version,
+        include_directories : tools_liblxc_includes,
+        link_args : ['-shared', '-fPIC', '-DPIC'],
```
eli-schwartz, line 319: What is the purpose of these arguments? e.g. shared_library() obviously handles -shared for you, and sensibly does -fPIC too. And the define is probably supposed to be in `c_args`, not just during linking.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+        link_whole : [liblxc_static],
+        dependencies : [threads,
+			libseccomp,
+			libcap,
+			libopenssl,
+			libselinux,
+			libapparmor],
+        install : true,
+        install_dir : libdir)
```
eli-schwartz, line 328: install_dir for shared_library and library defaults to `get_option('libdir')`, relative paths for install_dir on any target are rendered relative to the prefix.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+                include_directories : tools_liblxc_includes,
+                link_with : [liblxc],
+		dependencies : [threads,
+				libseccomp,
+				libcap,
+				libopenssl,
+				libselinux,
+				libapparmor],
+                install : true,
+                install_dir : bindir)
```
eli-schwartz, line 344: bindir is the default.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+                'lxc-autostart',
+		tools_lxc_autostart_sources,
+                include_directories : tools_liblxc_includes,
+                link_with : [liblxc],
+		dependencies : [threads,
+				libseccomp,
+				libcap,
+				libopenssl,
+				libselinux,
+				libapparmor],
```
eli-schwartz, line 342: Since you link with liblxc for many executables, and also a list of the same dependencies each time, you could consider using

```
liblxc_dep = declare_dependency(
    link_with: liblxc,
    dependencies: [threads,
                  libseccomp,
                  libcap,
                  libopenssl,
                  libselinux,
                  libapparmor])
```

Then each executable can `dependencies: liblxc_dep` and it expands to all these repetitive arguments.

src/include/meson.build
```diff
@@ -0,0 +1,95 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+include_sources = files(
+	'bpf.h',
+	'bpf_common.h',
+	'netns_ifaddrs.c',
+	'netns_ifaddrs.h')
+
+if cc.has_function('getline', prefix : '#include <stdio.h>', args : '-D_GNU_SOURCE')
+	conf.set10('HAVE_GETLINE', have)
```
eli-schwartz, line 10: Should be `true`, not `have`.

The latter is a variable last set in a foreach loop as
```
have = cc.sizeof(decl, prefix : seccomp_headers, args : '-D_GNU_SOURCE') > 0
```

meson.build
```diff
@@ -0,0 +1,624 @@
...
+	                'scmp_filter_ctx',
+	                'struct seccomp_notif_sizes',
+	                'struct clone_args',
+	               ]
+
+	        # We get -1 if the size cannot be determined
+	        have = cc.sizeof(decl, prefix : seccomp_headers, args : '-D_GNU_SOURCE') > 0
+
+		if have == true
+			conf.set10('HAVE_' + decl.underscorify().to_upper(), have)
```
eli-schwartz, line 265: Couldn't this just be,
```
if cc.sizeof(....) > 0
    conf.set10(..., true)
endif
```

src/include/meson.build
```diff
@@ -0,0 +1,95 @@
...
+else
+	include_sources += files(
+		'getline.c',
+		'getline.h')
+endif
+
+if cc.has_function('fexecve', prefix : '#include <unistd.h>', args : '-D_GNU_SOURCE')
+	conf.set10('HAVE_FEXECVE', have)
```
eli-schwartz, line 18: Lots of leaky reuse of the "have" variable here...

src/lxc/meson.build
```diff
@@ -0,0 +1,133 @@
...
+	'syscall_numbers.h',
+	'syscall_wrappers.h',
+	'terminal.c',
+	'terminal.h',
+	'utils.c',
+	'utils.h',
+	'uuid.c',
+	'uuid.h')
+
+liblxc_c_args = ['-fvisibility=default']
```
eli-schwartz, line 125: >  `gnu_symbol_visibility` (since 0.48.0): specifies how symbols should be exported, see e.g the GCC Wiki for more information. This value can either be an empty string or one of `default`, `internal`, `hidden`, `protected` or `inlineshidden`, which is the same as hidden but also includes things like C++ implicit constructors as specified in the GCC manual. Ignored on compilers that do not support GNU visibility arguments.

You can set this directly on the static_library.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+# AS_AC_EXPAND(LXCINITDIR, "$libexecdir")
+# 
+# # We need the install path so criu knows where to reference the hook scripts.
+# AC_DEFINE_UNQUOTED([DATADIR], "$DATADIR", ["Prefix for shared files."])
+
+cc = meson.get_compiler('c')
+pkgconfig = import('pkgconfig')
+
+possible_cc_flags = [
+	'-fPIE',
```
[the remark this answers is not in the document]
brauner: Now set in `default_options`.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+	'-fexceptions',
+	'-Warray-bounds',
+	'-Wrestrict',
+	'-Wreturn-local-addr',
+	'-fsanitize=cfi',
+	'-Wstringop-overflow',
+]
+
+possible_link_flags = [
+	'-Wl,--as-needed',
```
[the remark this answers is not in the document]
brauner: Set in `default_options` explicitly now.

meson.build
brauner: I've simplified it a bit at least.

meson.build
brauner: Doesn't find it here with `dependency()` but doe find it with `find_library()`.

meson.build
brauner: Doesn't find it here with `dependency()` but doe find it with `find_library()`.

meson.build
brauner:
```
Run-time dependency selinux found: NO (tried pkgconfig and cmake)
Run-time dependency apparmor found: NO (tried pkgconfig and cmake)
Run-time dependency openssl found: YES 1.1.1j
Run-time dependency libcap found: YES 2.44
```

meson.build
brauner:
```
Run-time dependency selinux found: NO (tried pkgconfig and cmake)
Run-time dependency apparmor found: NO (tried pkgconfig and cmake)
Run-time dependency openssl found: YES 1.1.1j
Run-time dependency libcap found: YES 2.44
```

meson.build
brauner: Ah ok, I see. I need to look for `libselinux` instead of `selinux`. Weird, I would've expected `find_library` to require `libselinux` not dependency.

meson.build
brauner: Ah ok, I see. I need to look for `libselinux` instead of `selinux`. Weird, I would've expected `find_library` to require `libselinux` not dependency.

meson.build
brauner: I simply tried to follow other projects of similar size and scope.

brauner: And I find almost nothing obvious when it comes to build systems, especially coming from autotools. :)

meson.build
brauner: done

meson.build
brauner: removed

meson.build
brauner: removed

meson.build
brauner: done

meson.build
brauner: done

src/include/meson.build
brauner: removed

src/include/meson.build
brauner: removed

src/lxc/meson.build
brauner: done

meson.build
```diff
@@ -0,0 +1,624 @@
...
+# # We need the install path so criu knows where to reference the hook scripts.
+# AC_DEFINE_UNQUOTED([DATADIR], "$DATADIR", ["Prefix for shared files."])
+
+cc = meson.get_compiler('c')
+pkgconfig = import('pkgconfig')
+
+possible_cc_flags = [
+	'-fPIE',
+	'-Wvla',
+	'-fdiagnostics-color',
```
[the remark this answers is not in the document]
brauner: Noted, I added it explicitly to default_options anyway though.

meson.build
eli-schwartz: dependency() will look for the exact filename minus .pc, because that is the calling convention of pkg-config, `pkg-config --cflags --libs {name}`..

cc.find_library() will look for the library name minus the platform prefix (lib) and suffix (.so), because that is the calling convention of the compiler, `-l{name}`.

meson.build
```diff
@@ -0,0 +1,624 @@
...
+
+project_source_root = meson.current_source_dir()
+project_build_root = meson.current_build_dir()
+
+# join_paths ignores the preceding arguments if an absolute component is
+# encountered, so this should canonicalize various paths when they are
+# absolute or relative.
+prefixdir = get_option('prefix')
+if not prefixdir.startswith('/')
+        error('Prefix is not absolute: "@0@"'.format(prefixdir))
```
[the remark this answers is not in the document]
eli-schwartz: I see you marked this as resolved, but didn't remove the custom error message. Any particular reason?

Again, if you try to pass a non-absolute prefix then meson itself will immediately error out before even parsing the build file (note the context is line 1 in project declaration) and your custom error message will never be emitted.

You can be absolutely guaranteed that prefixdir will start with a "/", you do not need to check. :)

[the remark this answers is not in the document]
brauner: Ah, I might have missed this. Let me push a fix for this!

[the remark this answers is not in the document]
brauner: Ok, fixed now!

src/lxc/meson.build
eli-schwartz: I have just realized that in 1531f8d53df57dc8d1a6604fef593b4d65ec90ec you moved -fvisibility to the static library, when I actually meant you can set the special gnu_symbol_visibility attribute there.

PR #3966, opened 2021-09-08
Improve bash completion - part 2.

denisfa:
Hi,

this is a follow-up to #3899. It includes some ideas discussed there, as well as a couple of fixes for stuff my tests did not catch at the time.

However, I would like to have feedback about some changes I did, mostly on design:

1. **Names with spaces**: when you create a container like `lxc-create -t download -n "arch linux"`, you can autocomplete the name as `'arch linux'`, `"arch linux"`or `arch\ linux`. I choose the one with backslash because it felt natural to me. However, you may be able to tell this is a bad idea if there is an edge case that I am unaware of.
2. **cgroup completion**: I am using the entry `lxc.payload.${name}`, where `name` is the container name. Not sure if this is correct. I try first the `user.slice` entry and the payload is my fallback for `cgroups_v2`. I tried to not use external tools to check if I have v1 or v2 by checking dir availability.
3. **Bash version 4**: I use associative arrays and some variable operations that  are available for `bash>=4`. I did not disclose that on my previous pr, but I am always assuming `bash>=4`.
4. **Missing completion for `lxc-info --config` and `lxc-{start,execute} --define`**: I did not find a way to reliable list these values unless I hard-code them into completion. Therefore, they will remain out of this work scope.

In case the commit history looks all over the place, I can later squash/fix-up/force-push for a cleaner history. All the changes I wanted to do are here.

I will await for your word before rewriting anything.

### DIFF
--- a/config/bash/lxc.in
+++ b/config/bash/lxc.in
@@ -39,6 +39,7 @@ __lxc_check_name_present() {
     mapfile -t names < <(command lxc-ls -1)
     local -r shortoptnamexp="^-[0-9A-Za-mo-z]*n[0-9A-Za-mo-z]*$"
     local parsed
+    local -r current="${cur//[\\\"\']}"
     mapfile -t names < <(command lxc-ls -1)
     # If `--name` or `-n` are present, do not complete with container names.
     for param in "${words[@]}"; do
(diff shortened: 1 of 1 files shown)


brauner:
@stgraber I'll leave this to you for review.

brauner:
jenkins: test this please

stgraber:
Thanks, that's a lot of improvements and it's looking very good!

PR #3984, opened 2021-09-29
lxc: core scheduling support

brauner:
Core scheduling defines if the container payload is marked as being
schedulable on the same core. Doing so will cause the kernel scheduler
to ensure that tasks that are not in the same group never run
simultaneously on a core. This can serve as an extra security measure to
prevent the container payload from using cross hyper thread attacks.

The only allowed values are 0 and 1. Set this to 1 to create a core
scheduling domain for the container or 0 to not create one. If not set
explicitly no core scheduling domain will be created for the container.


stgraber:
@brauner build seems to be failing on this one

brauner:
> @brauner build seems to be failing on this one

Fixing this!

PR #4003, opened 2021-10-19
conf: verify that rootfs is stable after setting up mounts

brauner:
Apparently some users changed their rootfs via their lxc.mount.entry
entries. Let's not allow that as that can cause confusion during
container setup. So lets verify that the rootfs is stable after setup.

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -1431,6 +1431,23 @@ static int lxc_mount_rootfs(struct lxc_rootfs *rootfs)
 	return log_trace(0, "Container uses separate rootfs. Opened container's rootfs");
 }
 
+static bool lxc_rootfs_overmounted(struct lxc_rootfs *rootfs)
+{
+	__do_close int fd_rootfs = -EBADF;
+
+	if (!rootfs->path)
+		fd_rootfs = open_at(-EBADF, "/", PROTECT_OPATH_DIRECTORY, PROTECT_LOOKUP_ABSOLUTE, 0);
+	else
+		fd_rootfs = open_at(-EBADF, rootfs->mount, PROTECT_OPATH_DIRECTORY, PROTECT_LOOKUP_ABSOLUTE_XDEV, 0);
+	if (fd_rootfs < 0)
+		return true;
+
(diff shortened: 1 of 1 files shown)


EinstokFair:
«that can cause confusion during container setup»

May I ask to tell what exactly was confusing? For me it was convinient to mount rootfs for each container, and now it is not - https://discuss.linuxcontainers.org/t/rootfs-overmounted-container-doesnt-start/15879

mihalicyn:
It may cause unpredictable problems during container start if the rootfs become overmounted. For example, if one some stage, we checked that some other mount (like cgroupfs/sysfs) exists on a particular place and assume that in the future. This guarantee may be broken if the rootfs is overmounted.

EinstokFair:
why not to make 'lxc.mount.rootfs.entry=...' to setup root in a predictable moment of time? (I understand that it require more work than just disable it, but was that the only reason?).

hallyn:
From your comment in the discuss:

>Will it help to mount vm/test2 volume with root filesystem into /var/lib/lxc/test2/rootfs before starting container (and remove that >mount from config)? In this case it will not change during container start. Just checked, yes, seems that it helps.

It sounds like maybe you're doing that mount outside of lxc altogether.  Which should be fine, but note you could also do this during the lxc pre-mount hook.

brauner:
After the rootfs is mounted LXC has deeply routed and by-design assumptions that the topmost mount is in fact the rootfs. If the rootfs is overmounted via an `lxc.mount.entry` these assumptions are broken. The problems become obvious when you consider operating on paths and stashed fds at the same time. When you use an fd that was opened before the rootfs was mounted it will refer to the overmounted rootfs. But when you look up a path it'll refer to the mount stacked on top of the rootfs. A while ago I changed LXC to always prefer fd-based and tightly scoped lookups from a well non-starting point. That well-known starting point is the rootfs. If the rootfs is mounted it's mounted. LXC will not allow overmounting it as that's just an invitation for subtle security bugs where files are created in the overmounted rootfs and the operating being performed on the topmost mount stacked on top of it. There's enough hooks to configure the rootfs.

PR #4019, opened 2021-10-28
Replace 'which' with 'command -v'

diederikdehaas:
The 'which' command is deprecated on Debian Sid as it is not POSIX
compliant and it's behavior is therefor not consistent, so replace it
with 'command -v' which is POSIX compliant.
See https://stackoverflow.com/a/677212 for details.

Also replaced a use of backticks (`) as that is deprecated as well.
See https://github.com/koalaman/shellcheck/wiki/SC2006 for details.

Signed-off-by: Diederik de Haas <didi.debian@cknow.org>

### DIFF
--- a/config/init/common/lxc-net.in
+++ b/config/init/common/lxc-net.in
@@ -31,7 +31,7 @@ use_nft() {
     [ -n "$NFT" ] && nft list ruleset > /dev/null 2>&1 && [ "$LXC_USE_NFT" = "true" ]
 }
 
-NFT="$(which nft)"
+NFT="$(command -v nft)"
 if ! use_nft; then
     use_iptables_lock="-w"
     iptables -w -L -n > /dev/null 2>&1 || use_iptables_lock=""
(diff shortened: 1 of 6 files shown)


diederikdehaas:
There is a related issue and due to me updating a different machine (and `apt-listchanges`), I got to see the message again:
```
  * The 'which' utility will be removed in the future.  Shell scripts
    often use it to check whether a command is available.  A more
    standard way to do this is with 'command -v'; for example:
      if command -v update-icon-caches >/dev/null; then
        update-icon-caches /usr/share/icons/...
      fi
    '2>/dev/null' is unnecessary when using 'command': POSIX says "no
    output shall be written" if the command isn't found.  It's also
    unnecessary for the debianutils version of 'which', and hides the
    deprecation warning.
```
I *think* that means that statements like 
`command -v apparmor_parser >/dev/null 2>&1` can/should be rewritten as 
`command -v apparmor_parser >/dev/null`

As I always have to think really hard how to use redirections and I'm not familiar with the code's history, I'll leave implementing the change to someone else.

PR #4032, opened 2021-11-07
autotools: Avoid multiple liblxc.so with --enable-pam

joanbm:
When installing LXC with the default options, a single non-symlink liblxc.so* (e.g. liblxc.so.1.7.0) file is created:

```
    $ ./autogen.sh && ./configure && make && \
    rm -rf "$HOME/lxci" && make DESTDIR="$HOME/lxci" install && \
    stat -c%N "$HOME/lxci/usr/local/lib/liblxc.so"*
    [...]
    '/home/someone/lxci/usr/local/lib/liblxc.so' -> 'liblxc.so.1'
    '/home/someone/lxci/usr/local/lib/liblxc.so.1' -> 'liblxc.so.1.7.0'
    '/home/someone/lxci/usr/local/lib/liblxc.so.1.7.0'
```

However, when automake>=1.16.5, and the `--enable-pam` option is used, two non-symlink liblxc.so* (e.g. liblxc.so.1.0.0 and liblxc.so.1.7.0) are erroneously created:

```
    $ ./autogen.sh && ./configure --enable-pam && make && \
    rm -rf "$HOME/lxci" && make DESTDIR="$HOME/lxci" install && \
    stat -c%N "$HOME/lxci/usr/local/lib/liblxc.so"*
    [...]
    '/home/someone/lxci/usr/local/lib/liblxc.so' -> 'liblxc.so.1.0.0'
    '/home/someone/lxci/usr/local/lib/liblxc.so.1' -> 'liblxc.so.1.0.0'
    '/home/someone/lxci/usr/local/lib/liblxc.so.1.0.0'
    '/home/someone/lxci/usr/local/lib/liblxc.so.1.7.0'
```

This is due to infighting between libtool's and LXC's versioning: libtool creates liblxc.so.1.0.0, then LXC's `install-exec-local` hook in `Makefile.am` moves it to liblxc.so.1.7.0. However, with `--enable-pam`, the `install-libLTLIBRARIES` target is re-triggered after `install-pamLTLIBRARIES`, which will create liblxc.so.1.0.0 again.

The bigger problem here is that the install for the pam_cgfs library is done on the `data` phase of the automake install process instead of the `exec` phase (https://www.gnu.org/software/automake/manual/html_node/The-Two-Parts-of-Install.html), which gives `install-libLTLIBRARIES` a chance to run again after the `install-exec-local` / `install-exec-hook` targets have already run.

To fix this, we add an "exec_" prefix to the pam_cgfs library to make it run during the `exec` phase (see link above). We also consolidate the various hooks in the `install-exec-hook` target, which runs after the whole install, avoiding needing to manually specify the dependencies like in `install-exec-local`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -636,19 +636,19 @@ AM_COND_IF([ENABLE_PAM],
 		[pamdir],
 		[AS_HELP_STRING([--with-pamdir=PATH],[Specify the directory where PAM modules are stored,
 							or "none" if PAM modules are not to be built])],
-		[pamdir=$withval],
+		[exec_pamdir=$withval],
 		[
 			if test "${prefix}" = "/usr"; then
-				pamdir="/lib${libdir##*/lib}/security"
+				exec_pamdir="/lib${libdir##*/lib}/security"
 			else
-				pamdir="\$(libdir)/security"
+				exec_pamdir="\$(libdir)/security"
 			fi
 		]
 	)])
 
-AM_CONDITIONAL([HAVE_PAM], [test x"$pamdir" != "xnone"])
+AM_CONDITIONAL([HAVE_PAM], [test x"$exec_pamdir" != "xnone"])
 AM_COND_IF([ENABLE_PAM],
-	[if test "z$pamdir" != "znone"; then
+	[if test "z$exec_pamdir" != "znone"; then
 		AC_ARG_VAR([PAM_CFLAGS], [C compiler flags for pam])
 		AC_ARG_VAR([PAM_LIBS], [linker flags for pam])
 		AC_CHECK_LIB(
@@ -659,7 +659,7 @@ AM_COND_IF([ENABLE_PAM],
 			])
 
 		AC_SUBST(PAM_LIBS)
-		AC_SUBST([pamdir])
+		AC_SUBST([exec_pamdir])
 	fi])
 
 # Some systems lack PR_CAPBSET_DROP definition => HAVE_DECL_PR_CAPBSET_DROP
@@ -1130,7 +1130,7 @@ Security features:
 
 PAM:
  - PAM module: $enable_pam
- - cgroup PAM module: $pamdir
+ - cgroup PAM module: $exec_pamdir
 
 Documentation:
  - examples: $enable_examples
--- a/src/lxc/Makefile.am
+++ b/src/lxc/Makefile.am
@@ -2631,7 +2631,7 @@ endif
 
 if ENABLE_PAM
 if HAVE_PAM
-pam_LTLIBRARIES = pam_cgfs.la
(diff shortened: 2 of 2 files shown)


joanbm:
PS: I noticed this because on my Arch Linux system, `pacman -Qkk lxc` was reporting a mismatch on the liblxc.so.1 symlink; it turns out that after installing the package, ldconfig causes further infighting by pointing liblxc.so.1 to liblxc.so.1.7.0 again, as it thinks it's more recent than liblxc.so.1.0.0.

joanbm:
Hmmm, I'm not sure what makes `install-libLTLIBRARIES` trigger after `install-pamLTLIBRARIES`, but it appears to only happen on very up-to-date systems... I can reproduce on the current Docker images for Arch Linux, Ubuntu 22.04 (dev), Debian Testing and Alpine Edge, but not on Ubuntu 20.04, Ubuntu 21.10, Debian Bullseye, Fedora 35 nor Alpine 3.14.

brauner:
jenkins: test this please

brauner:
> Hmmm, I'm not sure what makes `install-libLTLIBRARIES` trigger again after `install-pamLTLIBRARIES`, but it appears to only happen on very up-to-date systems... I can reproduce on the current Docker images for Arch Linux, Ubuntu 22.04 (dev), Debian Testing and Alpine Edge, but not on Ubuntu 20.04, Ubuntu 21.10, Debian Bullseye, Fedora 35 nor Alpine 3.14.
> 
> EDIT: Appears to be related to automake:
> 
> * automake 1.16.4 -> one liblxc.so*
> * automake 1.16.5 -> two liblxc.so*
> 
> EDIT: This is the commit in automake which makes the difference:
> 
> ```
> commit 4ffbab9337430131b21ce784a10aed67dbd35373
> Author: Jan Engelhardt <jengelh@inai.de>
> Date:   Fri Sep 10 15:45:29 2021 -0700
> 
>     automake: consistently depend on install-libLTLIBRARIES.
>     
>     Report and patch:
>     https://lists.gnu.org/archive/html/automake/2021-08/msg00016.html
>     
>     * bin/automake.in (generate_makefile): depend on
>     install-libLTLIBRARIES for all PROGRAMS and LTLIBRARIES,
>     such as install-pkglibLTLIBRARIES.
>     * NEWS: mention this.
> ```

Is this patch still safe to merge though?

joanbm:
Sorry if my last message wasn't entirely clear, yes, despite the oddness of this problem appearing on a automake patch version, I don't see any downside or risk to merging this, modulo hopefully non-existing bugs in my implementation.
It just makes the pam library build run on the 'exec' part of automake which is where it ought to be anyway, as it's generating platform-dependent files.

BTW, the Jenkins build failure looks spurious

PR #4039, opened 2021-11-29
api: ->save_config() doesn't need to create container dir

tych0:
If we're saving the config file to somewhere that's *not* the container
dir, we don't need to create the container dir. Let's not do this and
thus not require its parent to exist, which can be confusing, especially in
light of the sparse logging through these functions.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -2608,12 +2608,6 @@ static bool do_lxcapi_save_config(struct lxc_container *c, const char *alt_file)
 	int fd, lret;
 	bool ret = false, need_disklock = false;
 
-	if (!alt_file)
-		alt_file = c->configfile;
-
-	if (!alt_file)
-		return false;
-
 	/* If we haven't yet loaded a config, load the stock config. */
 	if (!c->lxc_conf) {
 		if (!do_lxcapi_load_config(c, lxc_global_config_value("lxc.default_config"))) {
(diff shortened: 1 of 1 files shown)


src/lxc/lxccontainer.c
```diff
@@ -2608,8 +2608,11 @@ static bool do_lxcapi_save_config(struct lxc_container *c, const char *alt_file)
 	int fd, lret;
 	bool ret = false, need_disklock = false;
 
-	if (!alt_file)
+	if (!alt_file) {
 		alt_file = c->configfile;
+		if (!create_container_dir(c))
+			return false;
+	}
```
brauner, line 2615: This unfortunately segfaults:
```
lxccontainer.c:1221:18: runtime error: member access within null pointer of type 'struct lxc_conf'
    #0 0x7ff4d9f09dd2 in do_create_container_dir /home/runner/work/lxc/lxc/src/lxc/lxccontainer.c:1221
+ echo ---
+ rm /tmp/tmp.cAeRLgG77M
    #1 0x7ff4d9f0a38c in create_container_dir /home/runner/work/lxc/lxc/src/lxc/lxccontainer.c:1246
    #2 0x7ff4d9f1bdb3 in do_lxcapi_save_config /home/runner/work/lxc/lxc/src/lxc/lxccontainer.c:2613
    #3 0x7ff4d9f1c3af in lxcapi_save_config /home/runner/work/lxc/lxc/src/lxc/lxccontainer.c:2666
    #4 0x559d8efdc0c0 in main /home/runner/work/lxc/lxc/src/tests/clonetest.c:60
    #5 0x7ff4d8e4c0b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #6 0x559d8efdbd4d in _start (/usr/bin/lxc-test-clonetest+0x2d4d)
    ```
It seem that `c` is not sufficiently initialized there.

tych0: Yeah, looks like I only tested my happy path :). I rearranged it, hopefully it'll work now...

brauner: :)

brauner:
Thanks!

PR #4041, opened 2021-12-02
add missing MS_BIND on bind-remount

Blub:
Without this the final remount fails with `EPERM`.

EDIT: I also just noticed that the commit introducing this isn't actually producing the intended result anyway, as it produces an `rw bind mount on an `ro` file system, so you still get `EROFS` when trying to write to it...
I'm not sure how to *actually* fix this without recreating the original reason for the commit cb4889abc830 ("conf: improve read-only /sys with read-write /sys/devices/virtual/net")


brauner:
> Without this the final remount fails with `EPERM`.
> 
> EDIT: I also just noticed that the commit introducing this isn't actually producing the intended result anyway, as it produces an `rw bind mount on an `ro`file system, so you still get`EROFS` when trying to write to it... I'm not sure how to _actually_ fix this 

Ugh...

> without recreating the original reason for the commit [cb4889a](https://github.com/lxc/lxc/commit/cb4889abc83057ed6568e543bdb0a89e3941ab54) ("conf: improve read-only /sys with read-write /sys/devices/virtual/net")

This is such an annoying issue. :D But yes, I figured sooner or later we end back here.

Blub:
Closing and opening a different PR for a new approach to this.

PR #4042, opened 2021-12-03
use 2 sysfs instances for sys:mixed

Blub:
In order to facilitate this, the default mount list's
'destination' may now be NULL to mean that the source should
be unmounted instead.

We now create two temporary staging directories when
mounting /sys, since we need to do the following:

1) Ensure the first sysfs mount point is writable.
2) Mount a read-only sysfs on /sys
3) Bind devices/virtual/net *writably* into /sys

This would be easier with the new mount API, but this way we
can support the old API and keep the general workflow in the
`default_mounts` array rather than writing custom code for
it, so at least it's easy to see what's going on.


brauner:
I like it but it regresses:
```
FAIL: lxc-tests: /usr/bin/lxc-test-rootfs (0s)
---
+ PHASE=setup
+ trap cleanup EXIT
+ lxc-destroy -n lxc-test-rootfs -f
lxc-destroy: lxc-test-rootfs: tools/lxc_destroy.c: main: 240 Container is not defined
+ true
+ lxc-create -t busybox -n lxc-test-rootfs
+ PHASE=ro_rootfs
+ echo 'Starting phase ro_rootfs'
Starting phase ro_rootfs
+ config=/var/lib/lxc/lxc-test-rootfs/config
+ sed -i /lxc.rootfs.options/d /var/lib/lxc/lxc-test-rootfs/config
+ echo 'lxc.rootfs.options = ro'
+ lxc-start -n lxc-test-rootfs
```
:)

Blub:
Yeah I noticed. A read-only rootfs probably doesn't allow creating `/sys/.ro` and `/sys/.rw`, that's annoying.
Not sure if a tmpfs on `/sys` would also re-create the original issue, or what other directory I could safely assume to be usable as a staging directory...

brauner:
> 

Something in /proc?

Blub:
Using `/proc/sys` as a staging directory for the r/w mount seems to work locally, let's see if the tests confirm.
Commit description needs updating though. hold on...

brauner:
jenkins: test this please

PR #4048, opened 2021-12-15
api: api-ify lxc-usernsexec

tych0:
We want to do something like lxc-usernsexec, but would rather not rely on
the binary being shipped since we're statically linking.

At first, I attempted to implement this with an lxc configuration file
(i.e. not open code the container bits like lxc-usernsexec does), but there
were a fair number of failures that seemed complicated to work around. So,
let's just hoist this code to an API call.

The lxc_usernsexec() API call takes exactly the same arguments as the
binary does. Ideally, we'd be able to expose this taking a list of idmaps
to do, but that would require making liblxc's list implementation a public
API, which is probably not great.

In addition, there are a bunch of small fixups, like adding a WIFEXITED()
check to the child process (vs. just assuming that via WEXITSTATUS()), some
additional logging on syscall failures, etc.


stgraber:
Tests are all getting stuck in lxc-test-capabilities (for over an hour)

stgraber:
Killing them all by hand now.

tych0:
Hmm... where were they hung? Not clear how it would relate to this patch.

stgraber:
Not sure, just saw it was the capabilities test, the sanitizers build on Github show the same behavior too.

tych0:
So it looks like it's hanging in a log write,

```
write(4, "capabilities capabilities 20211216160905.540 DEBUG    conf - conf.c:lxc_fill_autodev:1397 - Bind mounted host device 15(dev/full) to 17(full)\n", 142
```

fd 4 there is /dev/ptmx in the container. I'm not really sure why this patch would change that behavior though.

brauner:
> So it looks like it's hanging in a log write,
> 
> ```
> write(4, "capabilities capabilities 20211216160905.540 DEBUG    conf - conf.c:lxc_fill_autodev:1397 - Bind mounted host device 15(dev/full) to 17(full)\n", 142
> ```
> 
> fd 4 there is /dev/ptmx in the container. I'm not really sure why this patch would change that behavior though.

One thing to note is that your patch doesn't integrate `lxc-usernsexec` with the LXC logging system, You're calling `CMD_SYSINFO` but that's separate logging infra for these binaries. So that would need to change, e.g. to `SYSINFO`.

src/lxc/usernsexec.c
```diff
@@ -0,0 +1,451 @@
...
+			CMD_SYSINFO("Failed to open stdout. Continuing");
+			ttyname1[0] = '\0';
+		}
+
+		ret = readlink("/proc/self/fd/2", ttyname2, sizeof(ttyname2));
+		if (ret < 0) {
+			CMD_SYSINFO("Failed to open stderr. Continuing");
+			ttyname2[0] = '\0';
+		}
+	}
```
brauner, line 308: For example, I have zero idea why we actually need to do this and why this is even relevant for this binary. If we can't figure out why we should kick this.

src/lxc/usernsexec.c
```diff
@@ -0,0 +1,451 @@
...
+		return 1;
+	}
+
+	if (pid == 0) {
+		close(pipe_fds1[0]);
+		close(pipe_fds2[1]);
+
+		opentty(ttyname0, STDIN_FILENO);
+		opentty(ttyname1, STDOUT_FILENO);
+		opentty(ttyname2, STDERR_FILENO);
```
brauner, line 379: Same here. Why are we reopening the this? If there's no good reason we should kick it.

tych0:
I was wondering about that -- since we don't actually create a container, is it still appropriate to use the container's logging infra?

brauner (changes requested):
I think just exposing `lxc-usernsexec` with the same set of arguments as an api call is a bit weird.
I'd rather expose a variant of `usernsexec_*` that we have in `conf.c`.
One possibility how to pass idmaps might be to just expose a struct similar to the kernel one, i.e.:

```c
struct uid_gid_entry {
        u32 ns,
        u32 host,
        u32 range,
};

struct uid_gid_map {
        u32 nr_mappings;
        union {
                 struct uid_gid_entry simple_mappings[5],
                 struct uid_gid_entry *complex_mappings,
        };
}
```

and index the passed in array based on `nr_mappings`. For the most basic cases we provide the caller with the ability to specify up to 5 mappings without having to allocate any memory as a convenience. For more complex cases we require the caller to allocate memory.
This closely mirrors my old kernel patch which bumped the number of separate mappings one can set up from 5 to 340. It'll probably be worth to expose those limits as well, i.e. to add defines:

```c
#define LXC_NR_MAPPINGS_BASE 5
#define LXC_NR_MAPPINGS_MAX 340
```
and the verify that `LXC_NR_MAPPINGS_MAX` isn't exceeded in `->nr_mappings`. (Hm, it might actually be possible to use a flexible array member instead of a pointer in `struct uid_gid_map`. If we document that this might actually be nicer. Just a thought.)

So then two additional arguments for `lxc-usernsexec` as an api call would be:

```c
int lxc_userns_exec(const char *command, struct uid_gid_map *uid_map, struct uid_gid_map *gid_map, ...)
```

brauner:
> I was wondering about that -- since we don't actually create a container, is it still appropriate to use the container's logging infra?

Good point. Probably not.

tych0:
> For the most basic cases we provide the caller with the ability to specify up to 5 mappings without having to allocate any memory as a convenience. For more complex cases we require the caller to allocate memory.

The union bit seems unnecessarily complicated to me. What about just a:

```
struct uid_gid_map {
        u32 nr_mappings;
        struct uid_gid_entry *mappings;
}```

?

**Comment by brauner** [comment]
> > For the most basic cases we provide the caller with the ability to specify up to 5 mappings without having to allocate any memory as a convenience. For more complex cases we require the caller to allocate memory.
> 
> The union bit seems unnecessarily complicated to me. What about just a:
> 
> ```
> struct uid_gid_map {
>         u32 nr_mappings;
>         struct uid_gid_entry *mappings;
> }```
> 
> ?
> ```

Sure. I'm also happy to do that.

**Comment by brauner** [comment]
> > > For the most basic cases we provide the caller with the ability to specify up to 5 mappings without having to allocate any memory as a convenience. For more complex cases we require the caller to allocate memory.
> > 
> > 
> > The union bit seems unnecessarily complicated to me. What about just a:
> > ```
> > struct uid_gid_map {
> >         u32 nr_mappings;
> >         struct uid_gid_entry *mappings;
> > }```
> > 
> > ?
> > ```
> 
> Sure. I'm also happy to do that.

Meaning I'm happy if you do it that way. :)

**Comment by tych0** [comment]
Ok, sounds good. I'm still trying to track down the test failure which seems very strange, but I'll send updates when I get that figured out.

**Comment by hallyn** [comment]
Move
-__hidden extern int lxc_log_fd;
out of src/lxc/usernsexec.c `and` back to src/lxc/cmd/lxc_usernsexec.c

**Comment by tych0** [comment]
Huh, so it does. Actually we don't need it in either place now, as it turns out. So I've just deleted it. Very odd...

**Comment by tych0** [comment]
Seems like there might be some other weird dragons here...

```
FAIL: lxc-tests: /usr/bin/lxc-test-unpriv (1s)
+ DURATION=1
+ echo 'FAIL: lxc-tests: /usr/bin/lxc-test-unpriv (1s)'
+ '[' -f /tmp/tmp.gHFGU6b8nS ']'
+ echo ---
+ cat /tmp/tmp.gHFGU6b8nS
---
lxc-create: c1: tools/lxc_create.c: main: 317 Failed to create container c1
lxc-create c1 20211216211352.976 INFO     confile - confile.c:set_config_idmaps:2267 - Read uid map: type u nsid 0 hostid 910000 range 9999
lxc-create c1 20211216211352.976 INFO     confile - confile.c:set_config_idmaps:2267 - Read uid map: type g nsid 0 hostid 910000 range 9999
lxc-create c1 20211216211352.976 TRACE    conf - conf.c:userns_exec_mapped_root:5448 - Chowned 4(/home/lxcunpriv/.local/share/lxc/c1) to -1:1002
lxc-create c1 20211216211352.976 INFO     conf - conf.c:add_idmap_entry:5376 - Adding id map: type u nsid 0 hostid 910000 range 1
lxc-create c1 20211216211352.976 INFO     conf - conf.c:add_idmap_entry:5376 - Adding id map: type u nsid 1002 hostid 1002 range 1
lxc-create c1 20211216211352.976 INFO     conf - conf.c:add_idmap_entry:5376 - Adding id map: type g nsid 0 hostid 910000 range 1
lxc-create c1 20211216211352.976 INFO     conf - conf.c:add_idmap_entry:5376 - Adding id map: type g nsid 1002 hostid 1002 range 1
lxc-create c1 20211216211352.977 TRACE    conf - conf.c:userns_exec_mapped_root:5530 - Establishing uid mapping for "86222" in new user namespace: nsuid 0 - hostid 910000 - range 1
lxc-create c1 20211216211352.977 TRACE    conf - conf.c:userns_exec_mapped_root:5530 - Establishing uid mapping for "86222" in new user namespace: nsuid 1002 - hostid 1002 - range 1
lxc-create c1 20211216211352.977 TRACE    conf - conf.c:userns_exec_mapped_root:5530 - Establishing gid mapping for "86222" in new user namespace: nsuid 0 - hostid 910000 - range 1
lxc-create c1 20211216211352.977 TRACE    conf - conf.c:userns_exec_mapped_root:5530 - Establishing gid mapping for "86222" in new user namespace: nsuid 1002 - hostid 1002 - range 1
lxc-create c1 20211216211352.977 DEBUG    conf - conf.c:idmaptool_on_path_and_privileged:3516 - The binary "/usr/bin/newuidmap" does have the setuid bit set
lxc-create c1 20211216211352.978 DEBUG    conf - conf.c:idmaptool_on_path_and_privileged:3516 - The binary "/usr/bin/newgidmap" does have the setuid bit set
lxc-create c1 20211216211352.978 DEBUG    conf - conf.c:lxc_map_ids:3601 - Functional newuidmap and newgidmap binary found
lxc-create c1 20211216211352.981 TRACE    conf - conf.c:lxc_map_ids:3669 - newuidmap wrote mapping "newuidmap 86222 0 910000 1 1002 1002 1"
lxc-create c1 20211216211352.984 TRACE    conf - conf.c:lxc_map_ids:3669 - newgidmap wrote mapping "newgidmap 86222 0 910000 1 1002 1002 1"
lxc-create c1 20211216211352.985 NOTICE   utils - utils.c:lxc_switch_uid_gid:1323 - Switched to gid 0
lxc-create c1 20211216211352.985 NOTICE   utils - utils.c:lxc_switch_uid_gid:1332 - Switched to uid 0
lxc-create c1 20211216211352.985 NOTICE   utils - utils.c:lxc_drop_groups:1347 - Dropped supplimentary groups
lxc-create c1 20211216211352.985 TRACE    conf - conf.c:userns_exec_mapped_root:5520 - Chowned 4(/home/lxcunpriv/.local/share/lxc/c1) to 0:1002
lxc-create c1 20211216211352.986 ERROR    lxc_create - tools/lxc_create.c:main:317 - Failed to create container c1
```

**Comment by hallyn** [comment]
> Seems like there might be some other weird dragons here...

Not getting it - it passes for me locally, and jenkins said "testsuite passed".

**Comment by hallyn** [comment]
lxc-test-usernic fails bc it uses ifconfig.  passes after i install net-tools.

**Comment by tych0** [comment]
Yeah, I'm pretty sure there's something more sinister going on here, but it's not clear what. Dropping that `extern` should have had essentially no runtime effect if I understand correctly.

**Comment by brauner** [comment]
> Seems like there might be some other weird dragons here...
> 
> ```
> FAIL: lxc-tests: /usr/bin/lxc-test-unpriv (1s)
> + DURATION=1
> + echo 'FAIL: lxc-tests: /usr/bin/lxc-test-unpriv (1s)'
> + '[' -f /tmp/tmp.gHFGU6b8nS ']'
> + echo ---
> + cat /tmp/tmp.gHFGU6b8nS
> ---
> lxc-create: c1: tools/lxc_create.c: main: 317 Failed to create container c1
> lxc-create c1 20211216211352.976 INFO     confile - confile.c:set_config_idmaps:2267 - Read uid map: type u nsid 0 hostid 910000 range 9999
> lxc-create c1 20211216211352.976 INFO     confile - confile.c:set_config_idmaps:2267 - Read uid map: type g nsid 0 hostid 910000 range 9999
> lxc-create c1 20211216211352.976 TRACE    conf - conf.c:userns_exec_mapped_root:5448 - Chowned 4(/home/lxcunpriv/.local/share/lxc/c1) to -1:1002
> lxc-create c1 20211216211352.976 INFO     conf - conf.c:add_idmap_entry:5376 - Adding id map: type u nsid 0 hostid 910000 range 1
> lxc-create c1 20211216211352.976 INFO     conf - conf.c:add_idmap_entry:5376 - Adding id map: type u nsid 1002 hostid 1002 range 1
> lxc-create c1 20211216211352.976 INFO     conf - conf.c:add_idmap_entry:5376 - Adding id map: type g nsid 0 hostid 910000 range 1
> lxc-create c1 20211216211352.976 INFO     conf - conf.c:add_idmap_entry:5376 - Adding id map: type g nsid 1002 hostid 1002 range 1
> lxc-create c1 20211216211352.977 TRACE    conf - conf.c:userns_exec_mapped_root:5530 - Establishing uid mapping for "86222" in new user namespace: nsuid 0 - hostid 910000 - range 1
> lxc-create c1 20211216211352.977 TRACE    conf - conf.c:userns_exec_mapped_root:5530 - Establishing uid mapping for "86222" in new user namespace: nsuid 1002 - hostid 1002 - range 1
> lxc-create c1 20211216211352.977 TRACE    conf - conf.c:userns_exec_mapped_root:5530 - Establishing gid mapping for "86222" in new user namespace: nsuid 0 - hostid 910000 - range 1
> lxc-create c1 20211216211352.977 TRACE    conf - conf.c:userns_exec_mapped_root:5530 - Establishing gid mapping for "86222" in new user namespace: nsuid 1002 - hostid 1002 - range 1
> lxc-create c1 20211216211352.977 DEBUG    conf - conf.c:idmaptool_on_path_and_privileged:3516 - The binary "/usr/bin/newuidmap" does have the setuid bit set
> lxc-create c1 20211216211352.978 DEBUG    conf - conf.c:idmaptool_on_path_and_privileged:3516 - The binary "/usr/bin/newgidmap" does have the setuid bit set
> lxc-create c1 20211216211352.978 DEBUG    conf - conf.c:lxc_map_ids:3601 - Functional newuidmap and newgidmap binary found
> lxc-create c1 20211216211352.981 TRACE    conf - conf.c:lxc_map_ids:3669 - newuidmap wrote mapping "newuidmap 86222 0 910000 1 1002 1002 1"
> lxc-create c1 20211216211352.984 TRACE    conf - conf.c:lxc_map_ids:3669 - newgidmap wrote mapping "newgidmap 86222 0 910000 1 1002 1002 1"
> lxc-create c1 20211216211352.985 NOTICE   utils - utils.c:lxc_switch_uid_gid:1323 - Switched to gid 0
> lxc-create c1 20211216211352.985 NOTICE   utils - utils.c:lxc_switch_uid_gid:1332 - Switched to uid 0
> lxc-create c1 20211216211352.985 NOTICE   utils - utils.c:lxc_drop_groups:1347 - Dropped supplimentary groups
> lxc-create c1 20211216211352.985 TRACE    conf - conf.c:userns_exec_mapped_root:5520 - Chowned 4(/home/lxcunpriv/.local/share/lxc/c1) to 0:1002
> lxc-create c1 20211216211352.986 ERROR    lxc_create - tools/lxc_create.c:main:317 - Failed to create container c1
> ```

This is in the create codepath which is the only path that still calls `lxc-usernsexec`. All the other codepaths use a variant of `userns_exec_*` from `conf.c` so the failure seems at least related to `lxc-usernsexec`.

File: src/lxc/Makefile.am [hunk 3]
```diff
@@ -164,6 +164,7 @@ liblxc_la_SOURCES = af_unix.c af_unix.h \
 		    syscall_numbers.h \
 		    syscall_wrappers.h \
 		    terminal.c terminal.h \
+		    usernsexec.c \
```
brauner, line 167: You need to make a similar modification in the `src/lxc/meson.build` file:
```diff
diff --git a/src/lxc/meson.build b/src/lxc/meson.build
index 7c53646c2..9e36759df 100644
--- a/src/lxc/meson.build
+++ b/src/lxc/meson.build
@@ -119,6 +119,7 @@ liblxc_sources = files(
        'terminal.h',
        'utils.c',
        'utils.h',
+       'usernsexec.c',
        'uuid.c',
        'uuid.h')
```

brauner: Fwiw, you can build (most of LXC already) with meson:
```
meson setup -Ddebug=true -Dprefix=/usr build/
meson compile -C build/
```

src/lxc/usernsexec.c
```diff
@@ -0,0 +1,449 @@
+#include "config.h"
```
brauner, line 1: This needs to grow a SPDX line. :)
```
/* SPDX-License-Identifier: LGPL-2.1+ */
```

tych0:
> This is in the create codepath which is the only path that still calls lxc-usernsexec. All the other codepaths use a variant of userns_exec_* from conf.c so the failure seems at least related to lxc-usernsexec.

I'm not sure I follow. It doesn't look to me like this is calling lxc-usernsexec at all (lxc-create doesn't exec it, and it's not called in lxc-test-unpriv anywhere). It looks like it's using userns_exec_* from conf.c as you say. I must be missing something though.

brauner:
> > This is in the create codepath which is the only path that still calls lxc-usernsexec. All the other codepaths use a variant of userns_exec_* from conf.c so the failure seems at least related to lxc-usernsexec.
> 
> I'm not sure I follow. It doesn't look to me like this is calling lxc-usernsexec at all (lxc-create doesn't exec it, and it's not called in lxc-test-unpriv anywhere). It looks like it's using userns_exec_* from conf.c as you say. I must be missing something though.

Look at `create_run_template()` this is used when `-t <template>` is passed.

PR #4055, opened 2022-01-11
Update lxc-download.in

Dmole:
Better error message on bad domain (like Ubuntu 20.04 using pool.sks-keyservers.net)

### DIFF
--- a/templates/lxc-download.in
+++ b/templates/lxc-download.in
@@ -147,7 +147,13 @@ gpg_setup() {
   done
 
   if [ -z "${success}" ]; then
-    echo "ERROR: Unable to fetch GPG key from keyserver"
+    echo "ERROR: Unable to fetch GPG key from keyserver" >&2
+    TMP="$(basename "${DOWNLOAD_KEYSERVER}")"
+    if which dig > /dev/null \
+      && [ "$(dig "$TMP" +short)" == "" ] ; then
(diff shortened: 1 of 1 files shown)


stgraber:
DCO check failed, please make sure your commit message includes the required Signed-off-by line (`git commit --amend -s && git push --force` should sort that out). See the contributing document for details.

hallyn:
since you can override the keyserver with environment variables, perhaps pointing at a discuss url explaining what key server to use and how to do it would be better than suggesting overwriting the template. Especially as only root can do the latter.

Dmole:
@hallyn yes we could nicely ask the user to
```
export DOWNLOAD_KEYSERVER="hkp://keyserver.ubuntu.com"
```
but if the script is so old that it has a bad domain it likely has other issues to, and root is required to setup lxc anyway;
```
sudo apt install lxc
echo "$(id -un) veth lxcbr0 10" | sudo tee -a /etc/lxc/lxc-usernet
# and whatever undocumented action is required to fix systemd-run erroring out with "Failed to create bus connection: Permission denied"
```

Dmole:
@hallyn please let me know what you think of 2b66186

Dmole:
github seems to be ignoring my force push...

Dmole:
@hallyn https://github.com/lxc/lxc/pull/4060

PR #4090, opened 2022-02-23
cgroups: modify cgroup2 attach logic

brauner:
Recent kernels have seen various permission checking fixes when moving
processes into cgroups. So we're forced to modify how we attach to
containers.

Link: https://discuss.linuxcontainers.org/t/lxd-4-23-unable-to-start-nested-containers

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -2144,7 +2144,7 @@ static int cgroup_attach_leaf(const struct lxc_conf *conf, int unified_fd, pid_t
 }
 
 static int cgroup_attach_create_leaf(const struct lxc_conf *conf,
-				     int unified_fd, int *sk_fd)
+				     int unified_fd, int *sk_fd, bool unprivileged)
 {
 	__do_close int sk = *sk_fd, target_fd0 = -EBADF, target_fd1 = -EBADF;
 	int target_fds[2];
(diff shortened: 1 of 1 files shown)


stgraber:
@brauner failure looks legitimate

brauner:
This is a timeout for container shutdown so the failure is unrelated.

brauner:
jenkins: test this please

stgraber:
Did we change anything that'd explain us hitting that timeout?
Because we've now hit the exact same timeout in the exact same test 3 times out of 4 so that's looking odd...

brauner:
jenkins: test this please

brauner:
> Did we change anything that'd explain us hitting that timeout? Because we've now hit the exact same timeout in the exact same test 3 times out of 4 so that's looking odd...

Not really. My guess is that this can happen when the arm machines run both LXD and LXC tests at the same time there's a lot of stress on the machine. There's a long comment in the autostart tests:
```
# When shutting down, the container calls sync. If the machine running
# the test has a massive backlog, it can take minutes for the sync call to
# finish, causing a test failure.
# So try to reduce that by flushing everything to disk before we attempt
# a container shutdown.
```

brauner:
> > Did we change anything that'd explain us hitting that timeout? Because we've now hit the exact same timeout in the exact same test 3 times out of 4 so that's looking odd...
> 
> Not really. My guess is that this can happen when the arm machines run both LXD and LXC tests at the same time there's a lot of stress on the machine. There's a long comment in the autostart tests:
> 
> ```
> # When shutting down, the container calls sync. If the machine running
> # the test has a massive backlog, it can take minutes for the sync call to
> # finish, causing a test failure.
> # So try to reduce that by flushing everything to disk before we attempt
> # a container shutdown.
> ```

With Jenkins under no stress the test seems to pass without a problem.

PR #4096, opened 2022-03-23
build: add more tests to meson

brauner:


stgraber:
@brauner could you split the test code fixes from the meson changes in the first commit?

brauner:
> @brauner could you split the test code fixes from the meson changes in the first commit?

Yep, can do.

PR #4098, opened 2022-03-26
utils: add fastpath routine on mkdir_p function

memnoth:
Call 'access' to examine whether 'dir' is already existed or not instead
of directly calling 'mkdir' on each dir name separated by slash '/' even though
'dir' is existed.

Signed-off-by: Leesoo Ahn <lsahn@ooseel.net>


brauner:
jenkins: test this please

brauner:
Thank you!

PR #4099, opened 2022-03-27
tools: lxc-autostart: Reverse order on stop

corubba:
As it was discussed in lxc/lxd#2082, containers should be stopped in the opposite order they are started in.


stgraber:
jenkins: test this please

brauner (approved):

PR #4102, opened 2022-04-04
initutils: replace free function with __do_free on pointers from malloc

memnoth:
Use '__do_free' macro instead of calling 'free' to take the advange of any
concerns about memory leaks. Therefore developers could focus on the important
logic.


brauner (commented):
This specific codepaths is trick as there's a mixture of static and dynamically allocated memory here. So you can just call the __do_free helper here unconditonally as you might end up freeing static memory.

memnoth:
I got it. It would be good to let it go.

Thank you for the review!

PR #4103, opened 2022-04-08
confile: improve config key/value line parser

lsahn-gh:
trim quote marks ('/") in head and tail of the value of each line in container
config file.

This change takes care of the following cases.
1. Both marks are in head and tail of value part.
    lxc...='path/to/something'
2. Both marks are in head of value part.
    lxc...='path/to/something
3. Both marks are in tail of value part.
    lxc...=path/to/something'

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -3050,6 +3050,7 @@ static int parse_line(char *buffer, void *data)
 	bool empty_line;
 	struct lxc_config_t *config;
 	int ret;
+	size_t len;
 	char *dup = buffer;
 	struct parse_line_conf *plc = data;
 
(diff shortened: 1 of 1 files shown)


src/lxc/confile.c
```diff
@@ -3101,15 +3102,12 @@ static int parse_line(char *buffer, void *data)
 	value += lxc_char_left_gc(value, strlen(value));
 	value[lxc_char_right_gc(value, strlen(value))] = '\0';
 
-	if (*value == '\'' || *value == '\"') {
-		size_t len;
+	if (*value == '\'' || *value == '\"')
+		value ++;
```
brauner, line 3106: Why does the `value ++` become unconditional suddenly?

memnoth: The current codepath works for only that quote marks are in both head and tail like, `'value'` or `"value"` and also must be same value. But somehow if a value of config is as `'value`, `value"` or something like that, then I think we should take care of that cases as well.

mihalicyn:
Hi @lsahn-gh !

Sorry for so long delay with reviewing this.

I can't really understand the motivation behind this change.

The config parser in it's current state ensures that quotes are used as a pair. For example, if
may have `lxc.config.something='asdf'` either `lxc.config.something="asdf"`.
But if you have `lxc.config.something="asdf'` then parser must not skip the first and last quotes because they are different.

mihalicyn (changes requested):

stgraber:
Yeah, I agree that seems odd, we expect quotation marks to be in matching pair (beginning and end), missing marks at beginning or end or different marks at both ends aren't valid ways to quote a value, in which case the two valid behaviors would be to either fail or treat those marks as part of the value.

PR #4107, opened 2022-04-24
use systemd dbus StartTransientUnit for unpriv cgroup2

hallyn:

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -20,6 +20,7 @@
 #include <grp.h>
 #include <linux/kdev_t.h>
 #include <linux/types.h>
+#include <libgen.h>
 #include <poll.h>
 #include <signal.h>
 #include <stdint.h>
@@ -55,6 +56,11 @@
 
 #if !HAVE_STRLCAT
 #include "strlcat.h"
+#endif
+
+#ifdef HAVE_SYSTEMD_SD_EVENT_H
+#include <systemd/sd-bus.h>
+#include <systemd/sd-event.h>
 #endif
 
 lxc_log_define(cgfsng, cgroup);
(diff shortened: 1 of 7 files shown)


stgraber:
We're going to want to release LXC 5.0 before merging this as we can't add any new features to 5.0 (Ubuntu 22.04 has a pre-release snapshot of it and we can't SRU new features).

hallyn:
> We're going to want to release LXC 5.0 before merging this as we can't add any new features to 5.0 (Ubuntu 22.04 has a pre-release snapshot of it and we can't SRU new features).

This is nowhere near ready to merge anyway, sadly.

hallyn:
When you just do 'systemd-run ... -- lxc-start', then subtree_control is not set.
Similarly, if this patch just does StartTransientUnit, subtree_control is not set, and since there's a task in the cgroup, subtree_control *can't* be set.  So I'm creating a /init leaf cgroup, moving self to there, and then setting subtree_control on the parent, the original scope which systemd created.  Now we want lxc to create the init/../lxc.monitor.cgroup and init/../lxc.payload.cgroup and put tasks there, while using init/../{memory,pids,etc...}.* to set limits.  But I think I need a clearer head in order to get cgfsng to do the right thing for creating init/../lxc.monitor.cgroup etc.

src/lxc/conf.h
```diff
@@ -74,6 +74,13 @@ struct lxc_cgroup {
 			char *container_dir;
 			char *namespace_dir;
 			bool relative;
+			/* If an unpriv user in pure unified-only hierarchy
+			 * starts a container, then we ask systemd to create
+			 * a scope for us, and create the monitor and container
+			 * cgroups under that.
+			 * This will ignore the above things like monitor_dir
+			 */
+			char *systemd_scope;
```
tych0, line 83: Seems like we probably need to free this in lxc_conf_free() (or make a helper to free cgroup_meta and call that)?

hallyn: Oh that should be done in lxc_conf_free(), it looks like.  Thanks.

src/lxc/cgroups/cgfsng.c
```diff
@@ -947,6 +953,358 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+	if (r < 0) {
+		// bus won't clean up event since the attach failed
+		sd_event_unrefp(&event);
+		return log_error(SYSTEMD_SCOPE_FAILED, "Failed attaching event: %s\n", strerror(-r));
+	}
+
+	len = strlen("lxc-") + strlen(conf->name) + 4 + 6;
+	full_scope_name = malloc(len + 7);
+	if (!full_scope_name)
+		return log_error(-ENOMEM, "Out of memory");
```
brauner, line 1262: Fwiw, you can just use `syserror("Out of memory")` in such cases. It'll return a negative errno.

src/lxc/cgroups/cgfsng.c
```diff
@@ -947,6 +953,358 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+	if (r < 0)
+		return log_error(SYSTEMD_SCOPE_FAILED, "Failed allocating new event: %s\n", strerror(-r));
+	r = sd_bus_attach_event(bus, event, SD_EVENT_PRIORITY_NORMAL);
+	if (r < 0) {
+		// bus won't clean up event since the attach failed
+		sd_event_unrefp(&event);
+		return log_error(SYSTEMD_SCOPE_FAILED, "Failed attaching event: %s\n", strerror(-r));
+	}
+
+	len = strlen("lxc-") + strlen(conf->name) + 4 + 6;
```
brauner, line 1259: Uh, could you please comment what the `+4` and `+6` are for?

src/lxc/cgroups/cgfsng.c
```diff
@@ -947,6 +953,358 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+
+// Return a duplicate of cgroup path @cg without leading /, so
+// that caller can own+free it and be certain it's not abspath.
+static char *cgroup_relpath(char *cg)
+{
+	char *p;
+
+	if (!cg)
+		return NULL;
+	if (strcmp(cg, "/") == 0)
```
brauner, line 1297: I think this can be:
```c
static char *cgroup_relpath(char *cg)
{
	char *p;

	if (!cg || strequal(cg, "/"))
		return NULL;

	p = strdup(deabs(cg));
	if (!p)
		return ERR_PTR(-ENOMEM);

	return p;
}
```

src/lxc/cgroups/cgfsng.c
```diff
@@ -2949,8 +3333,10 @@ static bool __cgfsng_delegate_controllers(struct cgroup_ops *ops, const char *cg
 			return syserror_set(-EINVAL, "No upward walking paths allowed");
 
 		ret = lxc_writeat(dfd_cur, "cgroup.subtree_control", add_controllers, full_len);
-		if (ret < 0)
+		if (ret < 0) {
+			log_debug(0, "Could not enable \"%s\" controllers in the unified cgroup %d", add_controllers, dfd_cur);
 			return syserror("Could not enable \"%s\" controllers in the unified cgroup %d", add_controllers, dfd_cur);
+		}
```
brauner, line 3339: I think that `log_debug()` statement can be removed?

src/lxc/commands.c
```diff
@@ -89,6 +89,7 @@ static const char *lxc_cmd_str(lxc_cmd_t cmd)
 		[LXC_CMD_GET_CGROUP_CTX]		= "get_cgroup_ctx",
 		[LXC_CMD_GET_CGROUP_FD]			= "get_cgroup_fd",
 		[LXC_CMD_GET_LIMIT_CGROUP_FD]		= "get_limit_cgroup_fd",
+		[LXC_CMD_GET_SYSTEMD_SCOPE]		= "get_systemd_scope",
```
brauner, line 92: I tend to prefer to send file descriptors, not paths nowadays similar to `get_cgroup_fd()` and `get_limit_cgroup_fd()`..

brauner (changes requested):
Looks like a really good solution overall. I'm happy you did that work!

src/lxc/cgroups/cgfsng.c
```diff
@@ -947,6 +953,358 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+		if (first_line_read) // if >1 line, this is not pure unified
+			return false;
+		first_line_read = true;
+
+		if (strlen(p) > 3 && strncmp(p, "0:", 2) == 0)
+			return true;
+	}
+
+	return false;
+}
```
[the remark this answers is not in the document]
hallyn: Yeah, we may want to change the conditions under which we do the StartTransientUnit.  Maybe we should just try it (and ignore failure) any time we have a cgroup2 and are running systemd.  But my first priority was just making the common case work.  Fine - my case :)

src/lxc/cgroups/cgfsng.c
```diff
@@ -947,6 +953,358 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+	buf = read_file_at(EBADF, "/proc/self/cgroup", PROTECT_OPEN, 0);
+	if (!buf)
+		return NULL;
+
+	if (!string_pure_unified_system(buf))
+		return NULL;
+
+	// 0::/user.slice/user-1000.slice/session-136.scope
+	// Get past the "0::"
+	for (p = buf; p && *p && *p != '/'; p++);
```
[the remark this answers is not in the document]
hallyn: We *should* know it, yes.  But I'll leave the strnequal like you suggested for reasons of defense...

src/lxc/cgroups/cgfsng.c
```diff
@@ -947,6 +953,358 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+			first = false;
+		} else {
+			(void)strlcat(wbuf, " ", full_len + 1);
+		}
+		strlcat(wbuf, "+", full_len + 1);
+		strlcat(wbuf, controller, full_len + 1);
+	}
+	if (!wbuf) {
+		return log_debug(true, "No controllers to delegate!");
+	}
```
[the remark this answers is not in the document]
hallyn: OOps, I think I had a printf in there before.

src/lxc/cgroups/cgfsng.c
hallyn: Neat.

src/lxc/cgroups/cgfsng.c
```diff
@@ -947,6 +953,358 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+	if (!buf)
+		return false;
+	new_scope_pids = must_make_path(new_scope, "cgroup.procs", NULL);
+	ret = lxc_write_to_file(new_scope_pids, buf, strlen(buf), false, 0666);
+	if (ret != 0)
+		return log_error_errno(false, errno, "Failed to escape to cgroup \"%s\"", new_scope);
+
+	// enable controllers in parent_cgroup
+	return enable_controllers_delegation(pdir, parent_cgroup);
+}
```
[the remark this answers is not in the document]
hallyn: Yup, looks good.  (haven't yet rolled this into my changes, so leaving this comment unresolved until I do.)

src/lxc/commands.c
hallyn: We don't want an FD in this case though, we only want the actual systemd scope name, which I *think* is always just the last piece of the cgroup path.  So we could pass an fd along and do an fstat.  That *should* work.

brauner:
Putting this on blocked for a minute because we need to release 5.0 before.

hallyn:
> Putting this on blocked for a minute because we need to release 5.0 before.

Ok what's the timeline on that?

brauner:
> > Putting this on blocked for a minute because we need to release 5.0 before.
> 
> Ok what's the timeline on that?

@stgraber ?

brauner:
Merged in https://github.com/lxc/lxc/pull/4153 with fixes after we switched to meson as our new build system. :)

PR #4108, opened 2022-04-29
Meson improvements

stgraber:


stgraber:
@brauner this gets us a working dist tarball at least. There's still some more work to do to around the builds themselves to get the DESTDIR identical to pre-meson, but we're getting there.

brauner:
Thank you for the work. This is awesome!

PR #4110, opened 2022-05-02
More meson fixes

stgraber:


meson.build
```diff
@@ -207,7 +207,7 @@ threads = dependency('threads')
 libseccomp = dependency('libseccomp')
 conf.set10('HAVE_SECCOMP', libseccomp.found())
 if libseccomp.found()
-    pkgconfig_libs += '-lseccomp'
+    pkgconfig_libs += libseccomp
```
eli-schwartz, line 210: You don't need to protect this inside an if block. When the dependency is not-found, but added to pkg-config, it should do exactly as it does when added to a library or executable -- nothing. It is excluded from the generated pkg-config Requires field.

PR #4113, opened 2022-05-07
More meson tweaks

stgraber:


stgraber:
This gets us all of the examples. Left is:
 - init scripts (systemd, upstart, sysvinit)
 - apparmor profiles
 - selinux profiles
 - default config
 - pam plugin
 - lxc configs/includes
 - man pages

So slowly making it through. If anyone else's interested, let me know, the basics are all in place so it's mostly busy work adding meson.build files and comparing output.

PR #4115, opened 2022-05-11
bash: rename main bash completion file

terceiro:
Since the `lxc` binary is actually provided by lxd, the main
bash-completion file needs to be moved away to not conflict with a bash
completion file provided for the `lxc` binary by lxd.


terceiro:
tested with both autotools and meson and it DTRT in both cases

terceiro:
the usual convention for shared completion files is prefixing with a _, so I updated the patch accordingly.

stgraber:
Hey there, can you edit your commit message to include a `Signed-off-by` line?

nileshpatra:
@stgraber hi, could you merge this, if all looks good?
(We need this change at debian end as well actually)

terceiro:
On Wed, May 11, 2022 at 02:35:39PM -0700, Stéphane Graber wrote:
> Hey there, can you edit your commit message to include a `Signed-off-by` line?

Done.

PR #4120, opened 2022-05-18
build: add pam_cgfs to meson

brauner:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -87,6 +87,7 @@ lxctemplateconfcommondir = join_paths(lxctemplateconfdir, 'common.conf.d')
 lxctemplatedir = join_paths(lxcdatadir, 'templates')
 lxc_user_network_conf = join_paths(sysconfdir, user_network_conf_opt)
 lxc_user_network_db = join_paths(runtimepath, user_network_db_opt)
+pam_security = join_paths(libdir, 'security')
 
 conf.set_quoted('BINDIR', bindir)
 conf.set_quoted('DATADIR', datadir)
@@ -117,6 +118,7 @@ conf.set_quoted('LXC_USERNIC_DB', lxc_user_network_db)
 cgrouppattern = get_option('cgroup-pattern')
 wants_io_uring = get_option('io-uring-event-loop')
 want_examples = get_option('examples')
+want_pam_cgroup = get_option('pam-cgroup')
 want_mans = get_option('man')
 want_tests = get_option('tests')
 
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -279,10 +281,11 @@ endif
...
 ## PAM.
-pam = cc.find_library('pam',
-    has_headers: 'security/pam_modules.h')
-conf.set10('HAVE_PAM', pam.found())
-pkgconfig_libs += pam
+if want_pam_cgroup == true
+	pam = cc.find_library('pam', has_headers: 'security/pam_modules.h', required: true)
+	conf.set10('HAVE_PAM', pam.found())
+	pkgconfig_libs += pam
+endif
```
eli-schwartz, line 288: The `pam-cgroup` option is a boolean, and this block is only entered if it is force enabled. In this case, you make finding the library mandatory (the `required: true` kwarg is the implicit default), and then you set the HAVE_PAM define based on the value of `pam.found()`... which is impossible to ever be not found, because it was required on the line above and setup will simply error out if it is not found.

If you want to actually make this optional, you may want to switch the option to a feature option, and use:


```suggestion
pam = cc.find_library('pam', has_headers: 'security/pam_modules.h', required: get_option('pam-cgroup'))
conf.set10('HAVE_PAM', pam.found())
pkgconfig_libs += pam
```

src/lxc/pam/meson.build
```diff
@@ -0,0 +1,33 @@
...
+    include_directories: pam_cgfs_includes,
+    sources: pam_cgfs_sources,
+    dependencies: [
+        pkgconfig_libs,
+    ])
+
+pam_cgfs = shared_module(
+    'pam_cgfs',
+    pam_cgfs_sources,
+    dependencies: pam_cgfs_common_dependencies,
```
eli-schwartz, line 29: I'm a bit curious... what's the point of this declare_dependency? You use it in exactly one place, and add each source file twice to the shared_module.

src/lxc/pam/meson.build
```diff
@@ -0,0 +1,33 @@
...
+    name_prefix: '',
+    install_mode: 'rw-r--r--', # 644
+    install: want_pam_cgroup,
```
eli-schwartz, line 32: You always build this shared module by default, but only install it if it is enabled via an option? This doesn't seem correct... if it is not enabled, then the pam library won't be found and its headers may not be installed, and the shared_module will fail to link.

I think you should just have it always install, but just do not `subdir()` into this directory if the option is disabled (or more generally, if pam is not found).

meson.build
brauner: Thank you!

src/lxc/pam/meson.build
brauner: Sounds good!

brauner:
Thank you for the review, @eli-schwartz

PR #4124, opened 2022-05-28
More meson work

stgraber:
This tweaks the options a bit and add all the manpages.


stgraber:
With this one merged, we're down to just the systemd units, upstart config, apparmor profiles, sysconf files.

doc/meson.build
```diff
@@ -0,0 +1,81 @@
...
+    foreach page : PAGES
+        sgml = configure_file(
+            configuration: docconf,
+            input: page[0] + '.sgml.in',
+            output: page[0] + '.sgml')
+
+        configure_file(
+            input: sgml,
+            output: page[0] + '.' + page[1],
+            command: ['docbook2x-man', '--encoding=UTF-8', page[0] + '.sgml'],
```
eli-schwartz, line 77: https://github.com/lxc/lxc/blob/9647df5464acd2e71e7b40e185d6a6cb842dd89f/configure.ac#L228-L238

In order to have feature parity with the autotools build, you need to check all these alternative names too. e.g.

```meson
db2xman = find_program('docbook2X2man', 'docbook2x-man', 'db2x_docbook2man', 'docbook2man', 'docbook-to-man')

By the way, why use configure_file here instead of custom_target?

File: doc/meson.build [hunk 2]
```diff
@@ -0,0 +1,81 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+if want_mans == true
```
eli-schwartz, line 3: `if` expects a boolean, and `want_mans == true` is a boolean expression that evaluates to `true == true` -> `true`.

So you can also just do: `if want_mans`

PR #4125, opened 2022-05-30
meson: Tweak docs

stgraber:


meson.build
```diff
@@ -212,6 +212,20 @@ generate_date = run_command(date, '--utc', '--date=@' + time_epoch, '+%Y-%m-%d',
 time_epoch = time_epoch.to_int()
 conf.set('TIME_EPOCH', time_epoch)
 
+## Manpages.
+sgml2man = find_program('docbook2X2man', 'docbook2x-man', 'db2x_docbook2man', 'docbook2man', 'docbook-to-man', required: false)
```
eli-schwartz, line 216: If `want_docs` is true then this is actually required, right? So it should be `required: want_docs`.

PR #4126, opened 2022-05-30
Complete meson port

stgraber:


stgraber:
With this set, all the files in autotools are now in meson.

There are still some gaps and things I want to re-organized, mostly changing the way some variables and subdir calls are handled and fix a bunch of conf variables that are missing or wrong, but we're getting close.

I'm also yet to do a full diff of the generated trees to spot any issue there. I've done a limited number of diffs on things like manpages but haven't yet checked everything else.

stgraber:
Once I sort those things out, likely in a follow-up PR, then I can send another set I have pending which completely removes autoconf. We'll then need to deal with the Github and Jenkins breakage, effectively making them work with both autotools (for 3.0 and 4.0) as well as meson (for 5.0 onwards) and once that all clears, we can tag 5.0.

PR #4131, opened 2022-06-03
Remove autotools

stgraber:


stgraber:
@brauner I'm going to need your help to port the sanitizers test over as I have no idea what the various build flags are about and which ones we need new meson parameters for.

.github/workflows/build.yml
```diff
@@ -42,22 +42,17 @@ jobs:
...
-          make -j4
-          make DESTDIR="${BUILD}/install" install
-
-          cd "${SAN_BUILD}"
-          CFLAGS="$CFLAGS -fsanitize=address,undefined" ../configure --disable-no-undefined --enable-pam --enable-tests --with-distro=unknown
-          make -j4
-          make DESTDIR="${SAN_BUILD}/install" install
+          # Sanitized build
+          export CFLAGS="$CFLAGS -fsanitize=address,undefined"
+          meson setup san_build -Dtests=true -Dpam-cgroup=true
```
eli-schwartz, line 56: Meson has a `-Db_sanitize=address,undefined` builtin option.

.github/workflows/build.yml
```diff
@@ -42,22 +42,17 @@ jobs:
         env:
           CC: ${{ matrix.compiler }}
         run: |
-          # Configure
+          # Base flags
           export CFLAGS="-Wall -Werror"
           export LDFLAGS="-pthread -lpthread"
```
eli-schwartz, line 47:
```
meson setup -Dwerror=true
```

The -Wall is covered by meson.build default options for warning_level

I don't get the LDFLAGS here? The meson.build has `dependency('threads')` and I would presume an autotools build should be adding that in configure.ac or Makefile.am, not in the LDFLAGS...

eli-schwartz: And sure enough, AX_PTHREAD / PTHREAD_LIBS / PTHREAD_CFLAGS is present in the autotools build, so this was probably redundant all along?

stgraber:
@eli-schwartz any idea what's going on in https://github.com/lxc/lxc/runs/6732329110 ?

I would have expected `src` being in `liblxc_includes` to make an include of `lxc/attach_options.h` work as expected but it isn't...

eli-schwartz:
Looking at the logs, the command line in question doesn't actually have src in the includes. It does have src/lxc and src/include.

stgraber:
Yeah, I saw that too but can't figure out why meson is apparently skipping that directory despite it being directly added in meson.build...

stgraber:
Oh, I think I see what's going on.

eli-schwartz:
I'm assuming what you noticed is that this uses cmd_liblxc_includes, rather than liblxc_includes, and thus does not have that directory?

stgraber:
Yep and that those separate includes are quite wrong and pointless ;)
Validating a build locally with everything just using the same set of includes.

stgraber:
jenkins: test this please

stgraber:
jenkins: test this please

stgraber:
jenkins: test this please

stgraber:
jenkins: test this please

stgraber:
jenkins: test this please

stgraber:
@brauner ok, so I've got Jenkins mostly behaving now.

On the Github side, we need to fix the ASAN issue. It's currently just plain failing to link and I don't know why.
I'll need you to have a look.

Similarly, I'm not familiar with the oss-fuzz stuff and that needs fixing too.

With those two fixed, we should be good to go for the final switch to meson and the release of LXC 5.0.

stgraber:
I've now got this one building correctly on Android!

stgraber:
Clean Jenkins normal run: https://jenkins.linuxcontainers.org/job/lxc-github-pull-test/4610/
Clean Jenkins Android run: https://jenkins.linuxcontainers.org/job/lxc-build-android/9869/

@brauner over to you for ASAN and fuzz at which point this should go all green and we'll be ready for LXC 5.0!

PR #4137, opened 2022-06-08
build: remove autotools and finish meson port

brauner:
This is https://github.com/lxc/lxc/pull/4131 with very few patches on top.


meson.build
```diff
@@ -141,6 +141,11 @@ want_pam_cgroup = get_option('pam-cgroup')
 want_mans = get_option('man')
 want_tests = get_option('tests')
 want_tools = get_option('tools')
+want_commands = get_option('commands')
+want_capabilities = get_option('capabilities')
+want_apparmor = get_option('apparmor')
+want_openssl = get_option('openssl')
+want_selinux = get_option('selinux')
```
stgraber, line 148: Can you move those bits to their own commit?

meson.build
```diff
@@ -675,8 +698,12 @@ subdir('doc/ko')
...
 subdir('doc/rootfs')
 subdir('hooks')
-subdir('src/lxc/cmd')
-subdir('src/lxc/tools')
+if want_commands
+    subdir('src/lxc/cmd')
+endif
+if want_tools
+    subdir('src/lxc/tools')
+endif
```
stgraber, line 706: If we do that, we can then also change src/lxc/tools/meson.build to remove the `if want_tools` section from it.

src/tests/oss-fuzz.sh
```diff
@@ -37,7 +37,7 @@ meson setup san_build \
 	-Dapparmor=false \
 	-Dopenssl=false \
 	-Dselinux=false \
-	-Db_lto=false \
+	-Db_lto_mode=default \
 	-Db_sanitize=address,undefined
```
stgraber, line 41: Feels like this should go in another commit.

stgraber:
@brauner bit of rebasing and cleanup and we should be in good shape.

brauner:
> @brauner bit of rebasing and cleanup and we should be in good shape.

I did a rather fine-grained split now :)

PR #4139, opened 2022-06-08
Github workflow fixes

stgraber:


stgraber:
@brauner Looks like cifuzz is still pretty broken, you just didn't notice it because your PR broke the YAML so the test wasn't running anymore ;)

stgraber:
@brauner this should be good to merge as it's no worse than what we have currently, but there are still more issues for you to look into sadly :(

PR #4142, opened 2022-06-09
tests: fix oss-fuzz port to meson

brauner:

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -11,7 +11,7 @@ on:
       - master
 jobs:
   Fuzzing:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     if: github.repository == 'lxc/lxc'
     strategy:
       fail-fast: false
--- a/meson.build
+++ b/meson.build
@@ -146,6 +146,7 @@ want_capabilities = get_option('capabilities')
 want_apparmor = get_option('apparmor')
 want_openssl = get_option('openssl')
 want_selinux = get_option('selinux')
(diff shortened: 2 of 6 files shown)


brauner:
@evverx any chance you can take a look and help figure out what's going wrong here?

evverx:
Looks like `ASan` is linked statically somewhere. I'll try to take a closer look a bit later today.

FWIW `meson` isn't compatible with `fuzz-introspector` (https://github.com/systemd/systemd/pull/23158) so when this PR lands it will stop working.

brauner:
> @brauner I got it to work with [513e678](https://github.com/lxc/lxc/commit/513e6782251bd9d95f56028f1058638d76215c54) on top of your PR. The idea is to link the fuzz targets statically against liblxc to avoid dealing with shared libraries in any way because they tend to break fuzz-introspector, CIFuzz and coverage reports.

Thank you!

evverx:
@brauner it appears that `tests_common_sources` contains the liblxc sources so it seems it isn't necessary to link the fuzz targets against `liblxc` at all. I removed `link_with: [liblxc_static]` from the fuzz targets and they appear to be buildable locally with the OSS-Fuzz toolchain. I haven't tested it with CIFuzz though.

evverx:
> I can send a follow-up PR and see how we can improve this.

Makes sense. In the meantime I opened https://github.com/google/oss-fuzz/pull/7841 to trigger the CI there to make sure this PR covers all the bases. https://github.com/google/oss-fuzz/actions/runs/2471355690 looks promising :-)

PR #4149, opened 2022-06-15
tools: Provide multicall lxc binary

petris:
Create a binary, which embeds all lxc tools similar way as busybox
embeds its applets. This is handy for embedded systems as it saves
roughly 90% of the disk space.

To disable normal tools and use multicall binary exclusively use the
following meson setup options:
  -Dtools=false -Dtools-multicall=true

Signed-off-by: Petr Malat <oss@malat.biz>


stgraber:
LGTM but 5.0 is already locked feature wise so this will need to wait until after we've tagged 5.0

brauner (approved):
Very happy about this patch. I think that's a pretty sweet idea!
We would need to wait until 5.0 is released though. That should happen very soon!

PR #4159, opened 2022-07-01
add check for statvfs

Blub:
we use HAVE_STATVFS in the code but with meson the check got
lost causing mount_entry to fail to remount some things such
as a bind mount of /dev/fuse via

    lxc.mount.entry = /dev/fuse dev/fuse none bind,create=file 0 0

which would cause the following log messages:

    DEBUG    conf - ../src/lxc/conf.c:mount_entry:2416 - Remounting "/dev/fuse" on "/usr/lib/x86_64-linux-gnu/lxc/rootfs/dev/fuse" to respect bind or remount options
    ERROR    conf - ../src/lxc/conf.c:mount_entry:2459 - Operation not permitted - Failed to mount "/dev/fuse" on "/usr/lib/x86_64-linux-gnu/lxc/rootfs/dev/fuse"

note that the `Flags for ... were ...` line is not showing
up there, which depends on HAVE_STATVFS

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -538,6 +538,7 @@ foreach tuple: [
     ['sigdescr_np'],
     ['signalfd'],
     ['statx'],
+    ['statvfs'],
     ['strlcat'],
     ['strlcpy'],
     ['unshare'],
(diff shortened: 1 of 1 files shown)


brauner:
Thank you!

PR #4161, opened 2022-07-03
Fix errors constructing mount string when extra mount options supplied

srd424:
This fixes a really subtle off-by-one error constructing overlay mount options if rootfs options are provided and modern overlayfs (i.e. requiring a workdir) is used. We need to allow for the extra "," required to separate the extra options when computing the length!

Without the fix container creation fails with a "Failed to create string" error in the debug log.

### DIFF
--- a/src/lxc/storage/overlay.c
+++ b/src/lxc/storage/overlay.c
@@ -445,7 +445,7 @@ int ovl_mount(struct lxc_storage *bdev)
 			       upper, lower, mntdata);
 
 		len2 = strlen(lower) + strlen(upper) + strlen(work) +
-		       strlen("upperdir=,lowerdir=,workdir=") +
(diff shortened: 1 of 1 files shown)


srd424:
Hmm, I've just noticed in my dmesg:
```
[191458.696681] overlayfs: unrecognized mount option "O\xc8db\x05" or missing value
```
So fixing this looks like it's exposed something else.

srd424:
Hmm, with debug on in lxc-start:
```
lxc-start nuc-armtest 20220703170834.862 INFO     overlay - storage/overlay.c:ovl_mount:481 - Invalid argument - Failed to mount "/vol/contpool/lowermap" on "/
usr/lib/x86_64-linux-gnu/lxc" with options "?�Z". Retrying without workdir
```

brauner:
Thank you!

PR #4163, opened 2022-07-04
meson: add remaining still-in-use config checks

Blub:
These are all still in use in the code but have not been
added to meson.build when switching over from autoconf.


Blub:
ah, missing dependency changes in the oss fuzz meson targets, fixing...

brauner:
We're never switching build systems again! 😅🥲

PR #4166, opened 2022-07-14
src/lxc/log.h: fix STRERROR_R_CHAR_P

ffontaine:
`STRERROR_R_CHAR_P` is always defined to 0 or 1 depending on the value of `have_func_strerror_r_char_p` in `meson.build` so replace `#ifdef` by `#if` to avoid a redefinition build failure if `char *strerror_r` is not defined


brauner:
Thank you!

PR #4168, opened 2022-07-14
meson.build: fix build without stack-protector

ffontaine:
Move `-fstack-protector-strong` from `possible_cc_flags` to `possible_link_flags` to avoid a build failure on toolchains without ssp


brauner:
Thank you!

PR #4175, opened 2022-08-01
gitignore: Simplify

stgraber:
The move to meson has made it so that all rendered/built files are now
nicely self-contained. This lets us greatly simplify our gitignore,
effectively just ignoring release tarballs, the build directory and the
few usual temporary files we may deal with during development.


.gitignore
```diff
@@ -1,144 +1,11 @@
...
-patches
 *.orig
 *.rej
-tags
-TAGS
 
-doc/api/doxygen_sqlite3.db
-doc/api/*.tmp
+# Meson build directory.
+build/
```
eli-schwartz, line 8: Since Meson 0.57.0, `build/.gitignore` will be created by Meson itself with the contents "*", which means you don't need to have your own gitignore.

On the other hand, if people use "builddir/" instead (my personal favorite) or "builddir-release/" and "builddir-debug/", then this .gitignore from your PR doesn't cover it anyway.

Since you require a minimum of Meson 0.61, it's impossible for anyone to use a version of Meson too old to support automatic gitignore handling.

eli-schwartz (changes requested):

.gitignore
stgraber: ah, neat, fixed!

stgraber:
@brauner good to go

PR #4177, opened 2022-08-07
meson: fix docbook2x detection

CameronNemo:
docbook2man can sometimes be docbook2x and other times be docbook-utils.
Rather than compare paths, use version constraints to detect version.

Signed-off-by: Cameron Nemo <cam@nohom.org>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -324,9 +324,6 @@ endif
 generate_date = run_command(date, '--utc', '--date=@' + time_epoch, '+%Y-%m-%d', check: true).stdout().strip()
 
 ## Manpages.
-sgml2man = find_program('docbook2X2man', 'docbook2x-man', 'db2x_docbook2man', 'docbook2man', 'docbook-to-man', required: want_mans)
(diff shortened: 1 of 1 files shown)


eli-schwartz:
This seems very wrong, as you shouldn't need to pass an option in order to get basic sanity.

Instead, I suggest passing a `version: ` kwarg to the find_program (and looking it up twice, once for docbook2x, and once for docbook-utils), and not comparing paths at all. You could also use `sgml2man.version()` but that requires bumping the minimum version of Meson to 0.62.0.

CameronNemo:
So the version for the docbook2x command (`sgml2man.version()`) is `0.8.8` for me... I am not sure how to best express that constraint (via the find_program `version` arg) in a forward compatible way.

stgraber:
@eli-schwartz how does that look to you?

eli-schwartz:
LGTM, and this seems to implement what I suggested.

eli-schwartz (approved):

PR #4180, opened 2022-08-09
meson.build: allow explicit distrosysconfdir

### LINKED ISSUE #4178: Build fails when /etc/default or /etc/sysconfig do not exist
If neither of these directories exist, then LXC_DISTRO_SYSCONF is never set, and I get some warnings and an error like so:

```
config/init/common/meson.build:3: WARNING: The variable(s) 'LXC_DISTRO_SYSCONF' in the input file 'config/init/common/lxc-containers.in' are not present in the given configuration data.
config/init/common/meson.build:10: WARNING: The variable(s) 'LXC_DISTRO_SYSCONF' in the input file 'config/init/common/lxc-net.in' are not present in the given configuration data.
meson.build:900:9: ERROR: Entry LXC_DISTRO_SYSCONF not in configuration data.
```

The relevant section of code is here:

https://github.com/lxc/lxc/blob/cbbb09b56f0ab3a9054daf38987c4a5520480289/meson.build#L119-L128

sgn:
Allows either:

- Build inside minimal-and-clean chroot without neither
  /etc/sysconfdir nor /etc/default.
- Cross Compile lxc from foreign distro,
  let's say host distro uses /etc/sysconfdir and build distro
  uses /etc/default and vice versus.

Fix #4178

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -117,14 +117,18 @@ conf.set('SYSCONFDIR', sysconfdir)
 
 # Set sysconfdir
 fs = import('fs')
-if fs.is_dir('/etc/sysconfig')
+distrosysconfdir = get_option('distrosysconfdir')
+if distrosysconfdir != ''
+    distrosysconfdir = join_paths(sysconfdir, distrosysconfdir)
+    conf.set('LXC_DISTRO_SYSCONF', distrosysconfdir)
+elif fs.is_dir('/etc/sysconfig')
     distrosysconfdir = join_paths(sysconfdir, 'sysconfig')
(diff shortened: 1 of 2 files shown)


brauner (approved):
Thank you!

brauner (changes requested):
Thanks for the patch! Could you please sign-off your commit?

eli-schwartz:
Commit message is wrong -- it refers to "/etc/sysconfdir" but no such directory exists, it is rather "/etc/sysconfig".

*insert rant about legacy use of "distro-specific `--sysconfdir`"*

sgn:
@brauner  and @eli-schwartz  Both are done. Thanks.

PR #4187, opened 2022-08-25
use sd_bus_call_method_async to replace the asyncv one

ChenQi1989:
The sd_bus_call_method_asyncv's 10th parameter is of type
va_list and supplying NULL when invoking it causes compilation
error. Just replace it with the async one.

Signed-off-by: Chen Qi <Qi.Chen@windriver.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -295,9 +295,9 @@ if not want_sd_bus.disabled()
         has_sd_bus = false
     endif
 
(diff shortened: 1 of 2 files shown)


ulziibuyan:
Will this be merged soon?

brauner:
Sorry, I missed this PR.

PR #4200, opened 2022-09-14
conf: allow cross-device links

brauner:
Fixes: https://github.com/lxc/lxd/issues/10914


gene-vigelis:
I've install snap LXD so my lxc --version shows 5.5. How will i know that this has been back-ported? I have a workaround as shown above so would like to know when I can stop my workaround and have the container start properly with all disks.

Thank you so very much for such prompt response to my error!!!

stgraber:
`lxc info` will show you the liblxc version, currently `5.0.1`, `5.0.2` will have the fix.

PR #4210, opened 2022-10-19
tools: lxc-destroy: update help message for --force

Cypresslin:
Looks like the --force is a flag to stop a running container before destroying it.

Update the help message accordingly.


brauner:
Thank you!

PR #4211, opened 2022-10-19
tests: lxc-test-checkpoint-restore: use trap to do cleanup

Cypresslin:
This test will fail on Jammy 5.15, and because of the "set -e" it will never go through the lxc-stop and lxc-destroy code in the end of this script. Thus the lxc-test-criu container will not be removed.

Compose a cleanup() and use TRAP to solve this problem.


brauner:
Thank you!

PR #4219, opened 2022-11-08
Fix off-by-one error constructing mount options

smoser:
This fixes a really subtle off-by-one error constructing overlay mount options if rootfs options are provided and modern overlayfs (i.e. requiring a workdir) is used. We need to allow for the extra "," required to separate the extra options when computing the length!

Signed-off-by: srd424 <srd424@users.noreply.github.com>
(cherry picked from commit df3301046fc5f31881a2d4736cc5c381342ecc3d)


hallyn (approved):

hallyn:
No excuse for leaving this broken.  Thanks.

brauner:
Thanks for fixing this!

PR #4222, opened 2022-11-19
tests: lxc-test-reboot: Fix build on ia64

gibmat:
Add the prototype for `__clone2(...)` that is used on ia64, and adjust the code to use it via macro tests.

Verified that the code compiles properly on Debian's ia64 porterbox (yttrium), but was unable to actually run as `lxc-test-reboot` requires root privileges.

### DIFF
--- a/src/tests/reboot.c
+++ b/src/tests/reboot.c
@@ -34,7 +34,13 @@
 
 #include <linux/reboot.h>
 
+/*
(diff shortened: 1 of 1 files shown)


brauner (approved):
Classic `ia64` with its custom `clone()` implementation.

brauner:
Thanks for the patch!

PR #4225, opened 2022-11-22
src/lxc/meson.build: fix the static library path

smoser:
Since switching to meson, liblxc.a is being shipped as liblxc_static.a. Change it back to liblxc.a.

(cherry picked from commit 64eb31d02d4933ad414239130df3dff45a0f6f91)


brauner:
Thank you, looks good!

Issue #4223, opened 2022-11-21
stable-5.0 branch build using alpine:edge fails due to git output


rchincha:
The template below is mostly useful for bug reports and support questions.
Feel free to remove anything which doesn't apply to you and add more information where it makes sense.

# Required information

 * Distribution:
 * Distribution version:
 * The output of
   * `lxc-start --version`
   * `lxc-checkconfig`
   * `uname -a`
   * `cat /proc/self/cgroup`
   * `cat /proc/1/mounts`

# Issue description

A brief description of what failed or what could be improved.

On alpine:edge,

```

meson.build:319:0: ERROR: Command "/bin/date --utc --date=@1658982041
 +%Y-%m-%d" failed with status 1.
 
 ```

The issue appears to be that `git` cmd on alpine appears to be adding a newline at end of the output like so.

```
Running command: /usr/bin/git log --no-show-signature -1 --format=%at lxc-5.0.1
--- stdout ---
1658982041

--- stderr ---


Running command: /bin/date --utc --date=@1658982041
 +%Y-%m-%d
--- stdout ---

--- stderr ---
date: invalid date '@1658982041
'



meson.build:319:0: ERROR: Command "/bin/date --utc --date=@1658982041
 +%Y-%m-%d" failed with status 1.
 
 ```

# Steps to reproduce

 1. Step one
 2. Step two
 3. Step three

# Information to attach

 - [ ] any relevant kernel output (`dmesg`)
 - [ ] container log (The <log> file from running `lxc-start -n <c> -l TRACE -o <logfile> `)
 - [ ] the containers configuration file

rchincha:
Unfortunately, the merged fixes are not adequate.
At least on alpine:edge, it appears everything run_command(...).stdout() returns a newline appended and if it is assigned to a variable causes issues. The output needs to be stripped.

mihalicyn:
I can't understand what it is about to be honest.

Close for now. Feel free to reopen if you need any help.

PR #4227, opened 2022-11-23
meson.build: strip newline for variable assignments

rchincha:
Unfortunately, builds using alpine:edge still break! Apparently, run_command(...).stdout() must be strip()'ed for variable assignments

Addendum to d5600cf76a4d932a03ea75aea6dd6c997a4e2f35 Fixes issue #4223

Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -330,7 +330,7 @@ endif
 
 # Fallback to current epoch.
 if time_epoch == ''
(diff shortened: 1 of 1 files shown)


brauner:
Thanks!

PR #4229, opened 2022-11-24
conf: create separate peer group for container's root

brauner:
Finally, we turn the rootfs into a shared mount. Note, that this doesn't reestablish mount propagation with the hosts mount namespace. Instead we'll create a new peer group.

We're doing this because most workloads do rely on the rootfs being a shared mount. For example, systemd daemon like sytemd-udevd run in their own mount namespace. Their mount namespace has been made a dependent mount (MS_SLAVE) with the host rootfs as it's dominating mount. This means new mounts on the host propagate into the respective services.

This is broken if we leave the container's rootfs a dependent mount. In which case both the container's rootfs and the service's rootfs will be dependent mounts with the host's rootfs as their dominating mount. So if you were to mount over the rootfs from the host it would not just propagate into the container's mount namespace it would also propagate into the service.

### DIFF
--- a/config/apparmor/abstractions/start-container.in
+++ b/config/apparmor/abstractions/start-container.in
@@ -17,6 +17,8 @@
   mount options=bind /dev/pts/** -> /dev/**,
   mount options=(rw, make-slave) -> **,
   mount options=(rw, make-rslave) -> **,
+  mount options=(rw, make-shared) -> **,
+  mount options=(rw, make-rshared) -> **,
   mount fstype=debugfs,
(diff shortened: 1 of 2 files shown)


brauner:
Jenkins: test this please

brauner:
jenkins: test this please

brauner:
jenkins: test this please

stgraber:
jenkins: test this please

stgraber:
jenkins: test this please

PR #4244, opened 2022-12-12
lxc-oci template: support atomfs

hallyn:
Support OCIv1 images whose blobs are squashfs rather than tar.gz. We mount these using 'atomfs'.

This requires the start-container apparmor abstraction to have permission to do squashfuse mounts.


hallyn:
This is a draft right now - so that I don't lose the patch.

Need to figure out where atomfs will live before we can commit this.

PR #4246, opened 2022-12-29
src/lxc/meson.build: fix build without apparmor

ffontaine:
Don't build `lsm/apparmor.c` if apparmor is explicitly disabled by the user to avoid the following build failure with gcc 4.8:

```
/home/buildroot/autobuild/run/instance-3/output-1/host/arm-buildroot-linux-gnueabi/sysroot/usr/include/bits/fcntl2.h: In function '__apparmor_process_label_open.isra.0': /home/buildroot/autobuild/run/instance-3/output-1/host/arm-buildroot-linux-gnueabi/sysroot/usr/include/bits/fcntl2.h:50:24: error: call to '__open_missing_mode' declared with attribute error: open with O_CREAT in second argument needs 3 arguments
    __open_missing_mode ();
                        ^
```

Fixes:
 - http://autobuild.buildroot.org/results/c9f05ad264543adf429badb99310905427092772

### DIFF
--- a/src/lxc/meson.build
+++ b/src/lxc/meson.build
@@ -32,7 +32,6 @@ liblxc_sources = files(
     'cgroups/cgroup2_devices.h',
     'cgroups/cgroup_utils.c',
     'cgroups/cgroup_utils.h',
-    'lsm/apparmor.c',
     'lsm/lsm.c',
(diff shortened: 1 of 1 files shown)


brauner:
Thanks!

PR #4255, opened 2023-01-10
Fix build error on sparc64 caused by using the gold linker

gibmat:
When building on a sparc64 machine running Debian sid with gcc 12.2.0, linking fails with the following error:

`collect2: fatal error: ld terminated with signal 10 [Bus error]` ([build log](https://buildd.debian.org/status/fetch.php?pkg=lxc&arch=sparc64&ver=1%3A5.0.1-2&stamp=1672947476&raw=0))

It appears to be caused by using the gold linker, although I'm not exactly sure why. None of the other Linux-based Debian architectures exhibit this issue with the gold linker.

I made the proposed change and tested it on one of the GCC Compile Farm sparc64 machines (gcc202), which was then able to successfully compile lxc.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -225,10 +225,14 @@ possible_link_flags = [
     '-Wl,--gc-sections',
     '-Wl,-z,relro',
     '-Wl,-z,now',
-    '-Wl,-fuse-ld=gold',
     '-fstack-protector',
     '-fstack-protector-strong',
 ]
+
+# The gold linker fails with a bus error on sparc64
+if build_machine.cpu_family() != 'sparc64'
+    possible_link_flags += '-Wl,-fuse-ld=gold'
+endif
 
 if sanitize == 'none'
     possible_link_flags += '-Wl,--warn-common'


gibmat:
@stgraber I'm not sure where your comment went, as I got the email notification but when I click the link to reply it doesn't show up. So, I'll quote it here and reply below:

> @gibmat so your change is actually breaking our own builds now:
> https://launchpadlibrarian.net/645662600/buildlog_snap_ubuntu_focal_ppc64el_lxd-latest-edge_BUILDING.txt.gz
>
> Any idea of what to do to make everyone happy?

The only thing I can think of is that the order of the flags is now slightly different:

`-Wl,-z,now -Wl,-fuse-ld=gold -fstack-protector -fstack-protector-strong -Wl,--warn-common`

vs

`-Wl,-z,now -fstack-protector -fstack-protector-strong -Wl,-fuse-ld=gold -Wl,--warn-common`

Other than that there shouldn't be _any_ difference for any architectures other than sparc64. Locally I also did an amd64 build with those changes prior to submitting the pull request, but didn't test other architectures.

I'm not very familiar with meson; is there some sort of `list.remove(item)` that might be used to remove the gold linker flag conditionally just for sparc64?

stgraber:
I deleted the comment almost immediately because we were dealing with ppc64el and not sparc64 so couldn't have been your change. In the end, it was a linker bug. A simple rebuild got us past it.

gibmat:
:relieved: Phew -- not my fault! :smile:

Issue #4257, opened 2023-01-11
Provide a `state` liblxc function that has a timeout


tomponline:
Inspired by https://github.com/lxc/lxd/issues/11227

Similar to the `shutdown` function that already takes a timeout.

This way if a container's monitor isn't responding, LXD can timeout getting its state without leaking a Go routine.

mihalicyn:
It's pretty strange that we are facing this issue.

This is the lxcapi handler for `shutdown` (with timeout support):
```

static bool do_lxcapi_shutdown(struct lxc_container *c, int timeout)
{
// ... definitions and so on

	if (!c)
		return false;

	if (!do_lxcapi_is_running(c)) // <<=== IMPORTANT CALL
		return true;
```

Then,

```
static bool is_stopped(struct lxc_container *c)
{
	lxc_state_t s;

	s = lxc_getstate(c->name, c->config_path);
	return (s == STOPPED);
}

static bool do_lxcapi_is_running(struct lxc_container *c)
{
	if (!c)
		return false;

	return !is_stopped(c);
}
```

So, `do_lxcapi_is_running` just wraps `lxc_getstate`.

Now let's look on our liblxc `state` handler:
```
static const char *do_lxcapi_state(struct lxc_container *c)
{
	lxc_state_t s;

	if (!c)
		return NULL;

	s = lxc_getstate(c->name, c->config_path);
	return lxc_state2str(s);
}
```

It means that if the `state` API call freezes, then `shutdown` will freeze too. Am I correct?

All stuff around `timeout` in `shutdown` is about waiting for the container stopped state to be reached. I think we need to investigate this particular user case more in-depth.

tomponline:
@mihalicyn I've been investigating something which may (or may not) be related over in this PR:

https://github.com/lxc/lxd/pull/11278

In our LXD cluster evacuation tests we create some containers and then initiate a clean shutdown with a timeout of 5s (passed to liblxc via go-lxc), if the container doesn't response/shutdown in time (which is normal as we aren't waiting for it to fully start up so might not catch the signal) then LXD performs a forceful stop using go-lxc.

It seems that sometimes the forceful stop call either never returns or returns after a very long time (>30s).

The failure scenario we are seeing looks like this:
https://jenkins.linuxcontainers.org/job/lxd-github-pull-test/8609/arch=amd64,backend=zfs,compiler=golang-1.18,section=cluster/console#console-section-1

[...]

PR #4260, opened 2023-01-17
LXC API extension: set_timeout

mihalicyn:
lxccontainer set_timeout method allows to set LXC client
timeout for waiting monitor response.

Right now, it's implemented using the SO_RCVTIMEO client
socket option. (But it's the implementation detail that
can be changed in the future.)

See https://github.com/lxc/lxc/issues/4257

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -441,12 +441,12 @@ __access_r(3, 2) static int rsp_many_fds_reap(int fd, __u32 fds_len,
 }
 
 static int lxc_cmd_send(const char *name, struct lxc_cmd_rr *cmd,
-			const char *lxcpath, const char *hashed_sock_name)
+			const char *lxcpath, const char *hashed_sock_name, int rcv_timeout)
 {
 	__do_close int client_fd = -EBADF;
 	ssize_t ret = -1;
 
-	client_fd = lxc_cmd_connect(name, lxcpath, hashed_sock_name, "command");
+	client_fd = lxc_cmd_connect(name, lxcpath, hashed_sock_name, "command", rcv_timeout);
 	if (client_fd < 0)
 		return -1;
 
@@ -475,13 +475,15 @@ static int lxc_cmd_send(const char *name, struct lxc_cmd_rr *cmd,
 }
 
 /*
- * lxc_cmd: Connect to the specified running container, send it a command
- * request and collect the response
+ * lxc_cmd_timeout: Connect to the specified running container, send it a command
+ * request and collect the response with timeout
  *
- * @name           : name of container to connect to
- * @cmd            : command with initialized request to send
- * @stopped        : output indicator if the container was not running
- * @lxcpath        : the lxcpath in which the container is running
+ * @name             : name of container to connect to
+ * @cmd              : command with initialized request to send
+ * @stopped          : output indicator if the container was not running
+ * @lxcpath          : the lxcpath in which the container is running
+ * @hashed_sock_name : the hashed name of the socket (optional, can be NULL)
+ * @rcv_timeout      : SO_RCVTIMEO for LXC client_fd socket
  *
  * Returns the size of the response message on success, < 0 on failure
  *
@@ -493,8 +495,8 @@ static int lxc_cmd_send(const char *name, struct lxc_cmd_rr *cmd,
  * the slot with lxc_cmd_fd_cleanup(). The socket fd will be returned in the
  * cmd response structure.
  */
-static ssize_t lxc_cmd(const char *name, struct lxc_cmd_rr *cmd, bool *stopped,
-		       const char *lxcpath, const char *hashed_sock_name)
+static ssize_t lxc_cmd_timeout(const char *name, struct lxc_cmd_rr *cmd, bool *stopped,
+		       const char *lxcpath, const char *hashed_sock_name, int rcv_timeout)
 {
 	__do_close int client_fd = -EBADF;
 	bool stay_connected = false;
@@ -506,7 +508,16 @@ static ssize_t lxc_cmd(const char *name, struct lxc_cmd_rr *cmd, bool *stopped,
 
 	*stopped = 0;
 
-	client_fd = lxc_cmd_send(name, cmd, lxcpath, hashed_sock_name);
+	/*
+	 * We don't want to change anything for the case when the client
+	 * socket fd lifetime is longer than the lxc_cmd_timeout() execution.
+	 * So it's better not to set SO_RCVTIMEO for client_fd,
+	 * because it'll have an affect on the entire socket lifetime.
+	 */
+	if (stay_connected)
+		rcv_timeout = 0;
+
+	client_fd = lxc_cmd_send(name, cmd, lxcpath, hashed_sock_name, rcv_timeout);
 	if (client_fd < 0) {
 		if (errno == ECONNREFUSED || errno == EPIPE)
 			*stopped = 1;
@@ -525,6 +536,12 @@ static ssize_t lxc_cmd(const char *name, struct lxc_cmd_rr *cmd, bool *stopped,
 		cmd->rsp.ret = move_fd(client_fd);
 
 	return ret;
+}
+
+static ssize_t lxc_cmd(const char *name, struct lxc_cmd_rr *cmd, bool *stopped,
+		       const char *lxcpath, const char *hashed_sock_name)
+{
+	return lxc_cmd_timeout(name, cmd, stopped, lxcpath, hashed_sock_name, 0);
 }
 
 int lxc_try_cmd(const char *name, const char *lxcpath)
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -1059,7 +1059,7 @@ out:
  *
  * Returns the state on success, < 0 on failure
  */
-int lxc_cmd_get_state(const char *name, const char *lxcpath)
+int lxc_cmd_get_state(const char *name, const char *lxcpath, int timeout)
 {
 	bool stopped = false;
 	ssize_t ret;
@@ -1067,7 +1067,7 @@ int lxc_cmd_get_state(const char *name, const char *lxcpath)
 
 	lxc_cmd_init(&cmd, LXC_CMD_GET_STATE);
 
-	ret = lxc_cmd(name, &cmd, &stopped, lxcpath, NULL);
+	ret = lxc_cmd_timeout(name, &cmd, &stopped, lxcpath, NULL, timeout);
 	if (ret < 0 && stopped)
 		return STOPPED;
 
--- a/src/lxc/commands_utils.c
+++ b/src/lxc/commands_utils.c
@@ -144,9 +144,10 @@ int lxc_make_abstract_socket_name(char *path, size_t pathlen,
 }
 
 int lxc_cmd_connect(const char *name, const char *lxcpath,
-		    const char *hashed_sock_name, const char *suffix)
+		    const char *hashed_sock_name, const char *suffix, int rcv_timeout)
 {
-	int ret, client_fd;
+	int ret;
+	__do_close int client_fd = -EBADF;
 	char path[LXC_AUDS_ADDR_LEN] = {0};
 
 	ret = lxc_make_abstract_socket_name(path, sizeof(path), name, lxcpath,
@@ -159,7 +160,10 @@ int lxc_cmd_connect(const char *name, const char *lxcpath,
 	if (client_fd < 0)
 		return -1;
 
-	return client_fd;
+	if (lxc_socket_set_timeout(client_fd, rcv_timeout, 0))
+		return log_trace(-1, "Failed to set socket timeout");
+
+	return move_fd(client_fd);
 }
 
 int lxc_add_state_client(int state_client_fd, struct lxc_handler *handler,
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -491,7 +491,7 @@ static const char *do_lxcapi_state(struct lxc_container *c)
 	if (!c)
 		return NULL;
 
-	s = lxc_getstate(c->name, c->config_path);
+	s = lxc_getstate(c->name, c->config_path, c->rcv_timeout);
 	return lxc_state2str(s);
 }
 
@@ -501,7 +501,7 @@ static bool is_stopped(struct lxc_container *c)
 {
 	lxc_state_t s;
 
-	s = lxc_getstate(c->name, c->config_path);
+	s = lxc_getstate(c->name, c->config_path, c->rcv_timeout);
 	return (s == STOPPED);
 }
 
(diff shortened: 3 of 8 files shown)


mihalicyn:
https://github.com/lxc/lxd/pull/11293

src/lxc/commands.c
```diff
@@ -506,7 +508,18 @@ static ssize_t lxc_cmd(const char *name, struct lxc_cmd_rr *cmd, bool *stopped,
 
 	*stopped = 0;
 
-	client_fd = lxc_cmd_send(name, cmd, lxcpath, hashed_sock_name);
+	/*
+	 * We don't want to break anything for this special case,
```
tomponline, line 512: What is the special case here?

mihalicyn: we are setting `stay_connected` to `true` right now only for 2 command types `LXC_CMD_GET_TTY_FD` and `LXC_CMD_ADD_STATE_CLIENT`. In this case command socket file descriptor will not be closed after exiting from `lxc_cmd_timeout`. I'm afraid about possible side affects because we have to control how this socket will be used later and that our timeout setup will not break anything. That's why I think it's better not to set timeout for this special cases. Right now we don't have a plan to support timeout for this two command types anyway. But we can easily extend it in the future if needed.

tomponline: Right I see, makes sense. I think a comment there saying something about why those commands are special (what you've explained here) would be valuable for future readers.

src/lxc/commands_utils.c
```diff
@@ -144,7 +144,7 @@ int lxc_make_abstract_socket_name(char *path, size_t pathlen,
 }
 
 int lxc_cmd_connect(const char *name, const char *lxcpath,
-		    const char *hashed_sock_name, const char *suffix)
+		    const char *hashed_sock_name, const char *suffix, int rcv_timeout)
 {
 	int ret, client_fd;
```
brauner, line 149:
```
__do_close int client_fd = -EBADF;
.
.
.

return move_fd(client_fd);
```

src/lxc/commands_utils.c
```diff
@@ -159,6 +159,12 @@ int lxc_cmd_connect(const char *name, const char *lxcpath,
 	if (client_fd < 0)
 		return -1;
 
+	if (lxc_socket_set_timeout(client_fd, rcv_timeout, 0)) {
+		TRACE("Failed to set socket timeout");
+		close(client_fd);
+		return -1;
+	}
```
brauner, line 166: Not needed if you do like I suggested and use `__do_close`.

brauner: Should also use `log_trace(-1, "blurb")`.

src/lxc/lxccontainer.h
```diff
@@ -867,6 +873,17 @@ struct lxc_container {
 	 */
 	int (*seccomp_notify_fd_active)(struct lxc_container *c);
 
+	/*!
+	 * \brief Set response receive timeout for LXC commands
+	 *
+	 * \param c Container
+	 * \param timeout Seconds to wait before returning false.
+	 *  (0 to wait forever).
```
brauner, line 881: Should this be `-1` to wait forever and `0` to return immediately to keep in line with other socket APIs?

src/lxc/lxccontainer.c
```diff
@@ -5257,6 +5257,18 @@ static int do_lxcapi_seccomp_notify_fd_active(struct lxc_container *c)
 
 WRAP_API(int, lxcapi_seccomp_notify_fd_active)
 
+static bool do_lxcapi_set_timeout(struct lxc_container *c, int timeout)
```
brauner, line 5260: Why this generic approach instead of adding a new state api extension that waits with a timeout?

brauner (changes requested):

src/lxc/lxccontainer.h
mihalicyn: I thought about that too. But I can't imagine a use case for that here. Technically, when we are reading from LXC command socket `recvmsg` always getting blocked on waiting command result from the LXC container control daemon. I.e. if user set `0` to return immediately then all LXC commands that use this timeout setting will always return immediately without waiting for result. I understand why this is useful for `wait`/`reboot`/`shutdown` cases, but not wider.

brauner: Then block `0` but I still think that `-1` should indicate no timeout as is customary.

mihalicyn: > Then block `0` but I still think that `-1` should indicate no timeout as is customary.

sure!

brauner:
> yes, that was a first suggestion here #4257
> 
> But after analysis I've noticed that we have a global problem ([#4257 (comment)](https://github.com/lxc/lxc/issues/4257#issuecomment-1381896348)) with `lxc_getstate` function. It's called almost everywhere as a part of implementation of many different LXC commands, so it means that we have to add this "state timeout" to each LXC command. Second reason is that we can easily extend this and use the same timeout for other commands as a default without changing user visible API and keeping backward compatibility with an old `liblxc` users.

I really don't like this to be honest because the command socket is fundamentally an internal concept to LXC and I don't like that we expose internals such as altering the global socket timeout. That seems weird to me so if this isn't the only way to do this I would prefer that we expose a new state helper with a timeout.

I wonder if it wouldn't be neat to have a separate non-socket protocol for getting the state of a container. For example, similar to locks that we create in:

```c
        /* lockfile will be:
         * "/run" + "/lxc/lock/$lxcpath/$lxcname + '\0' if root
         * or
         * $XDG_RUNTIME_DIR + "/lxc/lock/$lxcpath/$lxcname + '\0' if non-root
         */
```

we could have sm like:

```c
         /*
         * "/run" + "/lxc/state/$lxcpath/$lxcname + '\0' if root
         * or
         * $XDG_RUNTIME_DIR + "/lxc/state/$lxcpath/$lxcname + '\0' if non-root
         */
```

and within that directory we create an mmaple file that the container updates everytime it changes state.

Then you can retrieve the state of the container without waiting. But, we can also allow clients to subscribe to the state file with an IN_ATTRIB inotify watch on that inode. Everytime the container manager has updated the state in the file we issue a fuitimensat() call allowing the client to be notified.

mihalicyn:
> I really don't like this to be honest because the command socket is fundamentally an internal concept to LXC and I don't like that we expose internals such as altering the global socket timeout. That seems weird to me so if this isn't the only way to do this I would prefer that we expose a new state helper with a timeout.

Thanks, Christian! That's why it was a RFC ;-) I agree that setting a global socket timeout is a little bit hacky approach.

> Then you can retrieve the state of the container without waiting. But, we can also allow clients to subscribe to the state file with an IN_ATTRIB inotify watch on that inode. Everytime the container manager has updated the state in the file we issue a fuitimensat() call allowing the client to be notified.

Sounds good. So, do we want to replace existing `lxc_getstate` function implementation to use this new approach?

brauner:
> > I really don't like this to be honest because the command socket is fundamentally an internal concept to LXC and I don't like that we expose internals such as altering the global socket timeout. That seems weird to me so if this isn't the only way to do this I would prefer that we expose a new state helper with a timeout.
> 
> Thanks, Christian! That's why it was a RFC ;-) I agree that setting a global socket timeout is a little bit hacky approach.
> 
> > Then you can retrieve the state of the container without waiting. But, we can also allow clients to subscribe to the state file with an IN_ATTRIB inotify watch on that inode. Everytime the container manager has updated the state in the file we issue a fuitimensat() call allowing the client to be notified.
> 
> Sounds good. So, do we want to replace existing `lxc_getstate` function implementation to use this new approach?

Hm, I think if we were to do this we would need a new state2 function so that clients can try to call state2() first and if that doesn't for whatever reason return an error they can choose to fallback to regular old state() function.

We should also briefly consult @stgraber and @hallyn whether they're ok with my concoction.

brauner:
One potential worry with my approach could be that we lose authentication against the unix socket. But I think for retrieving the state of the container that's perfectly fine. We don't need to authenticate and we can simply ensure that the state file is only accessible to the owner of the container.

mihalicyn:
> Hm, I think if we were to do this we would need a new state2 function so that clients can try to call state2() first and if that doesn't for whatever reason return an error they can choose to fallback to regular old state() function.

This works for direct LXC API `state` calls. If that was the only problem, then I'd just proposed to create like `state2` and forget about it. As I wrote there (https://github.com/lxc/lxc/issues/4257#issuecomment-1381896348) we have a global helper function called `lxc_getstate`. This function is used by LXC API `state` call, but it is also implicitly used by almost all other LXC API calls through another helpers `do_lxcapi_is_running` and `is_stopped`:
https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L500
https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2959
https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L1970
https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2007
https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2076
https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L4638
https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L4893
... and so on

So it means that `state2` won't help for such cases or we need to have 2nd version of all LXC API calls (to extend them with a new `timeout` argument). That's why I'm asking if we are ready to modify this "global" helper to use new approach.

brauner:
> > Hm, I think if we were to do this we would need a new state2 function so that clients can try to call state2() first and if that doesn't for whatever reason return an error they can choose to fallback to regular old state() function.
> 
> This works for direct LXC API `state` calls. If that was the only problem, then I'd just proposed to create like `state2` and forget about it. As I wrote there ([#4257 (comment)](https://github.com/lxc/lxc/issues/4257#issuecomment-1381896348)) we have a global helper function called `lxc_getstate`. This function is used by LXC API `state` call, but it is also implicitly used by almost all other LXC API calls through another helpers `do_lxcapi_is_running` and `is_stopped`:
> 
> https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L500
> 
> 
> https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2959
> 
> 
> https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L1970
> 
> 
> https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2007
> 
> 
> https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2076
> 
> 
> https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L4638
> 
> 
> https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L4893
> 
> 
> ... and so on
> So it means that `state2` won't help for such cases or we need to have 2nd version of all LXC API calls (to extend them with a new `timeout` argument). That's why I'm asking if we are ready to modify this "global" helper to use new approach.

No, I mean internally we can certainly just do the fallback thing. Realistically, it doesn't even matter because any fixed LXC version should be able to support the new method without fallbacking to the old method. But we need state2() for clients that really want to go through the old socket-based api.

brauner:
> > > Hm, I think if we were to do this we would need a new state2 function so that clients can try to call state2() first and if that doesn't for whatever reason return an error they can choose to fallback to regular old state() function.
> > 
> > 
> > This works for direct LXC API `state` calls. If that was the only problem, then I'd just proposed to create like `state2` and forget about it. As I wrote there ([#4257 (comment)](https://github.com/lxc/lxc/issues/4257#issuecomment-1381896348)) we have a global helper function called `lxc_getstate`. This function is used by LXC API `state` call, but it is also implicitly used by almost all other LXC API calls through another helpers `do_lxcapi_is_running` and `is_stopped`:
> > https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L500
> > 
> > https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2959
> > 
> > https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L1970
> > 
> > https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2007
> > 
> > https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L2076
> > 
> > https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L4638
> > 
> > https://github.com/lxc/lxc/blob/47630db7970022d7b174bbc631db4cd86f6a8fcb/src/lxc/lxccontainer.c#L4893
> > 
> > ... and so on
> > So it means that `state2` won't help for such cases or we need to have 2nd version of all LXC API calls (to extend them with a new `timeout` argument). That's why I'm asking if we are ready to modify this "global" helper to use new approach.
> 
> No, I mean internally we can certainly just do the fallback thing. Realistically, it doesn't even matter because any fixed LXC version should be able to support the new method without fallbacking to the old method. But we need state2() for clients that really want to go through the old socket-based api.

Because we're not using anything that depends on a super new kernel version. It's pretty boring, old stuff. And the file creation should always work because lxclock depends on /run being available and writable as well.

mihalicyn:
> No, I mean internally we can certainly just do the fallback thing. Realistically, it doesn't even matter because any fixed LXC version should be able to support the new method without fallbacking to the old method. But we need state2() for clients that really want to go through the old socket-based api.

so, we are changing old `state` API to go through the new approach (with file/mmap/inotify and fallback to the old method of course), then we are introducing `state2` which is just the old implementation of `state`. Correct?

brauner:
> > No, I mean internally we can certainly just do the fallback thing. Realistically, it doesn't even matter because any fixed LXC version should be able to support the new method without fallbacking to the old method. But we need state2() for clients that really want to go through the old socket-based api.
> 
> so, we are changing old `state` API to go through the new approach (with file/mmap/inotify and fallback to the old method of course), then we are introducing `state2` which is just the old implementation of `state`. Correct?

I guess you're right and we should just try and replace the old state api completely with the new one. So just try the approach were you don't introduce a new state2() function and instead just make the old one use the new approach.

stgraber:
jenkins: test this please

stgraber (approved):

mihalicyn:
gentle ping @brauner :-)

PR #4263, opened 2023-01-20
drop broken lxc-test-fuzzers

hallyn:


stgraber:
@brauner is that safe to do or is that script called from ossfuzz somehow?

brauner:
> @brauner is that safe to do or is that script called from ossfuzz somehow?

I'm not sure they're actually used.

Issue #4268, opened 2023-01-29
setproctitle doesn't work on arm64 musl


kdrag0n:
# Required information

 * Distribution: Alpine
 * Distribution version: edge 3.17_alpha20221110
 * `lxc-start --version`: 5.0.2
 * Linux kernel 6.1.2

# Issue description

setproctitle fails when running LXC on an arm64 system with musl libc, so the process title doesn't get changed to `[lxc monitor] ...`. I'm not sure if arm64 + glibc works, but it works on x86_64 with both glibc and musl.

# Steps to reproduce

 1. `lxc-create --name alpine --template download`
 2. `lxc-start alpine`
 3. Check `ps` output for process name

# Information to attach

Failed prctl call:
```
[pid 39373] openat(AT_FDCWD, "/proc/self/stat", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 39373] <... openat resumed>)       = 17
[pid 39373] fcntl(17, F_SETFD, FD_CLOEXEC) = 0
[pid 39373] mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0xffff812b2000
[pid 39373] read(17, "39373 (scon) R 39350 39347 1802 "..., 4095) = 291
[pid 39373] brk(NULL <unfinished ...>
[pid 39373] <... brk resumed>)          = 0x1f4a2000
[pid 39373] prctl(PR_SET_MM, PR_SET_MM_MAP, 0xffff59880800, 0x68, 0) = -1 EINVAL (Invalid argument)
```

LXC log:
```
lxc-start alpine 20230128081525.485 TRACE    commands - ../src/lxc/commands.c:lxc_cmd:514 - Connection refused - Command "get_init_pid" failed to connect command socket
lxc-start alpine 20230128081525.485 TRACE    commands - ../src/lxc/commands.c:lxc_cmd:514 - Connection refused - Command "get_state" failed to connect command socket
lxc-start alpine 20230128081525.485 TRACE    start - ../src/lxc/start.c:lxc_init_handler:738 - Created anonymous pair {4,5} of unix sockets
lxc-start alpine 20230128081525.485 TRACE    commands - ../src/lxc/commands.c:lxc_server_init:2129 - Created abstract unix socket "/var/lib/lxc/alpine/command"
[...]

PR #4274, opened 2023-02-02
initutils: setproctitle use sbrk syscall instead of brk

mihalicyn:
brk() is sets program break address, to get a current program break address we have to use sbrk() as `sbrk(0)`, because sbrk takes increment but not an address as an argument

Issue #4268

Signed-off-by: Alexander Mikhalitsyn <aleksandr.mikhalitsyn@canonical.com>


mihalicyn:
I wasn't right, cause `brk` implementation handles this case properly and just return an old value of `mm->brk`.

PR #4287, opened 2023-03-10
switch from libsystemd's dbus to dbus-1

hallyn:
This is purely so that we can do static linking.  Linking against libsystemd makes that a challenge because while it's perfectly simple to do, distros tend not to provide a libsystemd.a.

Tools that want to (a) link against liblxc and (b) have a statically linked binary to bind into a minimal container are ill served by this.  So link against libdbus-1.

.github/workflows/build.yml: switch to dbus-1.
src/lxc/cgroups/cgfsng.c: replace the unpriv_systemd_create_scope(),
   start_scope, and enter_scope() systemd code with dbus-1 code.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -59,9 +59,8 @@
 #include "strlcat.h"
 #endif
 
-#if HAVE_LIBSYSTEMD
-#include <systemd/sd-bus.h>
-#include <systemd/sd-event.h>
+#if HAVE_DBUS
+#include <dbus/dbus.h>
 #endif
 
 lxc_log_define(cgfsng, cgroup);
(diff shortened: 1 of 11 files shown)


brauner:
Seems fine to me. The project seems still maintained and if that helps with static linking for you then sure.

hallyn:
> Seems fine to me. The project seems still maintained and if that helps with static linking for you then sure.

Thanks.  It's more code and cruftier code, so if there were any other option, I'd prefer that...

I haven't figured out why the fuzzing script is failing to build.  I mean, it's simply not passing the pkgconf --cflags dbus-1 output to gcc for the cgroup/cgfsng.c file, while it *does* send that for other files like monitor.c ...  so it's funky, and probably some subtle meson usage...

brauner:
> > Seems fine to me. The project seems still maintained and if that helps with static linking for you then sure.
> 
> Thanks. It's more code and cruftier code, so if there were any other option, I'd prefer that...
> 
> I haven't figured out why the fuzzing script is failing to build. I mean, it's simply not passing the pkgconf --cflags dbus-1 output to gcc for the cgroup/cgfsng.c file, while it _does_ send that for other files like monitor.c ... so it's funky, and probably some subtle meson usage...

 Could

> > Seems fine to me. The project seems still maintained and if that helps with static linking for you then sure.
> 
> Thanks. It's more code and cruftier code, so if there were any other option, I'd prefer that...
> 
> I haven't figured out why the fuzzing script is failing to build. I mean, it's simply not passing the pkgconf --cflags dbus-1 output to gcc for the cgroup/cgfsng.c file, while it _does_ send that for other files like monitor.c ... so it's funky, and probably some subtle meson usage...

So in the code it looks like you have both `#if HAVE_DBUS` and `#Ifdef HAVE_DBUS` and I think that's the issue.

hallyn:
> So in the code it looks like you have both `#if HAVE_DBUS` and `#ifdef HAVE_DBUS` and I think that's the issue.

Hm, maybe, but really since I made that not optional (for a first cut) in meson.build, I should just have removed those altogether ...  which wouldn't help here.

To be precise, the above mistake would cause wrong behavior if config.h had `#define HAVE_DBUS 0`, but both evaluate to true for HAVE_DBUS 1, which is the case here.

So while this does need to be fixed (either by replacing #idef with #if or by removing those #ifs altogether), I don't think it explains what is happening here :(  /usr/include/dbus-1.0 existed, but ninja is simply not passing the full cflags on to make for files under src/lxc/cgroup/ but is for files under src/lxc - and only in the fuzz build.

While I'd like to get to the bottom of that, the simplest thing is probably to add a 'want_dbus' option to meson, fix the #ifdef, and allow HAVE_DBUS=0.

hallyn:
Thanks - I went that route, disabling dbus, which meant the bug you found definitely became relevant and would have confused me for hours :)

src/lxc/cgroups/cgfsng.c
```diff
@@ -963,189 +962,387 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+{
+	DBusMessageIter iter_array;
+	unsigned int i;
+
+	if (!dbus_message_iter_open_container(iter, DBUS_TYPE_ARRAY, DBUS_TYPE_UINT32_AS_STRING, &iter_array))
+		return log_debug(false, "Dbus error opening array container");
+
+	for (i = 0; i < len; i++) {
+		if (!dbus_message_iter_append_basic(&iter_array, DBUS_TYPE_UINT32, &(value[i])))
+			return log_debug(false, "Dbus error appending u32 to array");
```
mihalicyn, line 1040: Do we need to call `dbus_message_iter_close_container` on the error path too?

hallyn: Yes, probably should.

src/lxc/cgroups/cgfsng.c
```diff
@@ -963,189 +962,387 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+			DBUS_STRUCT_BEGIN_CHAR_AS_STRING
+			DBUS_TYPE_STRING_AS_STRING
+			DBUS_TYPE_VARIANT_AS_STRING
+			DBUS_STRUCT_END_CHAR_AS_STRING,
+			&array_iter))
+		return log_debug(false, "Dbus error...");
+
+	// "PIDs", "au", getpid()
+	if (!dbus_message_iter_open_container(&array_iter, DBUS_TYPE_STRUCT, NULL, &struct_iter))
+		return log_debug(false, "Dbus error...");
```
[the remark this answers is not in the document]
hallyn: Yeah, the complication with using the cleanup fn is you cannot run dbus_message_iter_close_container() twice, and the iterator itself is not a pointer so we can't set it to null to indicate not running it.  So probably best just to fix the structure of the fns to DTRT.  I'll take a look.  Thanks for pointing these out.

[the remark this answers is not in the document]
hallyn: We could write wrappers for the open, close, and abandon container fns which are passed a struct that indicates whether it needs to be freed or not.  That might be worthwhile.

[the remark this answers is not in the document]
hallyn: Ok that wasn't too bad.

mihalicyn:
LGTM. Thanks!

mihalicyn (approved):

src/lxc/cgroups/cgfsng.c
```diff
@@ -963,189 +962,420 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+
+	dbus_pending_call_unref(pending);
+
+	// Wait on a signal telling us the async scope request is handled
+	// TODO add a timeout
+	while (true) {
+		if (systemd_cgroup_scope_ready(connection, scope_name))
+			break;
+		nanosleep((const struct timespec[]){{0, 1000}}, NULL);
+		continue;
```
brauner, line 1220: That's a bit of a yucky construction but fine...

src/lxc/cgroups/cgfsng.c
```diff
@@ -963,189 +962,420 @@ static bool check_cgroup_dir_config(struct lxc_conf *conf)
...
+{
+	char *p;
+	bool first_line_read = false;
+
+	lxc_iterate_parts(p, contents, "\n") {
+		if (first_line_read) // if >1 line, this is not pure unified
+			return false;
+		first_line_read = true;
+
+		if (strlen(p) > 3 && strncmp(p, "0:", 2) == 0)
```
brauner, line 1372: We do have `strnequal()` for this. :)

brauner (approved):
This is fine to merge.

I would like to suggest that you file a feature request against systemd asking for a simpler protocol to be implemented that ideally doesn't rely on dbus to get a delegated cgroup fd. I talked with Lennart about this a while ago and he agreed that this would be a good idea.
You wouldn't be alone in wanting this. I know that the Podman people think that dbus takes too long to allocate a cgroup fd so they would like to have a simpler and faster method as well.

src/lxc/cgroups/cgfsng.c
hallyn: Oh - well that went in with the original libsystemd-dev patch last year.  May as well switch it over.

hallyn: Hm, while I suspect it would be fine, this check is doing more than strnequal does - it also makes sure there are *more* than 3 characters.  Using strnequal would return true for "0:", but we want to require more.

src/lxc/cgroups/cgfsng.c
hallyn: I'm torn on this.  Do we just give it a 10s timeout and hope for the best?  Or do we let it sleep forever if something malfunctions, and leave the caller to deal with it?  In general the caller - at some level above us - will know better than us what a good timeout is.

hallyn:
> This is fine to merge.
> 
> I would like to suggest that you file a feature request against systemd asking for a simpler protocol to be implemented that ideally doesn't rely on dbus to get a delegated cgroup fd. I talked with Lennart about this a while ago and he agreed that this would be a good idea. You wouldn't be alone in wanting this. I know that the Podman people think that dbus takes too long to allocate a cgroup fd so they would like to have a simpler and faster method as well.

https://github.com/systemd/systemd/issues/26885

PR #4288, opened 2023-03-12
fix non-root to up permission to run binary of root

duguhaotian:
use this rootfs to run container
```
# ls -l
total 660
-rwxr-x--- 1 root root 675200 2  28 15:04 pause
```

lxc-start with uid:gid 64600:64600, we expect to failed with **permission deny**. But we got success to run it....

we can get solve from runc:

https://github.com/opencontainers/runc/commit/957d97bcf43f41beef9670fe22ec78ccb5c5c101


hallyn:
Can you be more precise about 'lxc-start with uid:gid 64600:64600'?  Do you su to those permissions and then run lxc-start?

duguhaotian:
> Can you be more precise about 'lxc-start with uid:gid 64600:64600'? Do you su to those permissions and then run lxc-start?

i mean: use non-root user '64600', to start a binary which have 0750 permission successful. 
this is not right. user '64600'  should not have permission to run this binary.

hallyn:
> > Can you be more precise about 'lxc-start with uid:gid 64600:64600'? Do you su to those permissions and then run lxc-start?
> 
> i mean: use non-root user '64600', to start a binary which have 0750 permission successful. this is not right. user '64600' should not have permission to run this binary.

I'm sorry, that's not precise enough.  Because if I try to start a container as user 'ubuntu', and do not open up permissions to the full path along $HOME/.local/share/lxc/, then I can't start the container.  So I want to make sure I can reproduce.  So please show, starting from a root prompt, the full terminal log, exactly how you are starting the container.

duguhaotian:
> 
OK,  first, i give my step:
1.  a **static** compiled binary, (go binary or other static compiled);
2. chmod to 750, and owner is root:root;
3. use non-root to start container, and use **static** binary as init process of container;


i will add my log later, because my notebook donot have env to run lxc...

duguhaotian:
sorry, it's my mistake...

PR #4300, opened 2023-04-11
setproctitle(): Handle potential NULL return from strrchr()

solardiz:
Prior to the changes introduced in #4299 (on my suggestion), LXC's `setproctitle()` handled possible NULL returns from `strchr()` calls it made on contents of `/proc/self/stat`. With these changes, it does not handle a possible NULL return from the newly added `strrchr()`. This PR fixes that. CC: @tych0 

I've only tested this out-of-tree using the below program. Please also test in-tree.

```c
#include <string.h>
#include <stdlib.h>
#include <stdio.h>
#include <inttypes.h>
#include <unistd.h>
#include <sys/prctl.h>
#include <sys/syscall.h>

#include "lxc/src/include/strlcpy.c"

#define __do_fclose /* nothing */
#define fopen_cloexec fopen
#define lxc_read_nointr read
#define LXC_LINELEN 4096
#define prctl_arg(x) ((unsigned long)x)

/* setproctitle() body inserted here */

int main(void)
{
        setproctitle("changed");
        sleep(100);
        return 0;
}
```

### DIFF
--- a/src/lxc/initutils.c
+++ b/src/lxc/initutils.c
@@ -245,20 +245,19 @@ int setproctitle(char *title)
 	/*
 	 * executable names may contain spaces, so we search backwards for the
 	 * ), which is the kernel's marker for "end of executable name". this
-	 * skips the first two fields.
+	 * puts the pointer at the end of the second field.
 	 */
-	buf_ptr = strrchr(buf, ')')+2;
-
-	/* Skip the next 23 fields, column 26-28 are start_code, end_code,
-	 * and start_stack */
-	buf_ptr = strchr(buf_ptr, ' ');
-	for (i = 0; i < 22; i++) {
-		if (!buf_ptr)
-			return -1;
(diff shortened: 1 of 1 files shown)


tych0:
oops, thanks :⁠-⁠)

brauner:
On Tue, Apr 11, 2023 at 06:58:42AM -0700, Solar Designer wrote:
> @solardiz commented on this pull request.
> 
> 
> 
> >  	 */
> -	buf_ptr = strrchr(buf, ')')+2;
> +	buf_ptr = strrchr(buf, ')');
> 
> I can make this change if you like, along with a similar change to the next loop - would even save 2 lines of code there. Let me know if I should.

If you would be up for it or even have other ideas how to improve this
code then we'd be delighted to take them!

brauner:
jenkins: test this please

PR #4305, opened 2023-04-22
rename functions which clash with libsystemd's

hallyn:
If statically linking against both liblxc and libsystemd, some function names conflict:

mkdir_p fd_cloexec path_simplify is_dir is_fs_type

Rename those to lxc_\0, as:

for sym in mkdir_p fd_cloexec path_simplify is_dir is_fs_type; do
	git grep "$sym" | awk -F: '{ print $1 }' | sort | uniq | xargs sed -i "s/$sym/lxc_$sym/g"
done

(the above loop wrongly replaces is_dir in meson.build, but c'est la vie)


hallyn:
Only compile-tested so far.

hallyn:
built in ppa:serge-hallyn/jammy, seems to be working all right

PR #4313, opened 2023-05-24
explicitly convert *mainloop_handler to __u64

mosbrew-ag:
When lxc is built with io-uring-event-loop enabled, there is implicit pointer to int conversion in __io_uring_disarm.

### DIFF
--- a/src/lxc/mainloop.c
+++ b/src/lxc/mainloop.c
@@ -215,7 +215,7 @@ static int __io_uring_disarm(struct lxc_async_descr *descr,
 		return syserror_set(ENOENT,
 				    "Failed to get submission queue entry");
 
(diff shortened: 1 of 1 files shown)


mosbrew-ag:
GCC treats such conversion as warning ([gcc_warn.txt](https://github.com/lxc/lxc/files/11559059/gcc_warn.txt)), while Clang-15 aborts compilation ([clang_err.txt](https://github.com/lxc/lxc/files/11559058/clang_err.txt)).
Here I propose trivial fix to allow building with clang.
Another option would be add -Wno-int-conversion to meson.build

PS I never ever created pull requests before, so please excuse my mistakes.

brauner:
Thanks!

PR #4324, opened 2023-07-06
get_hierarchy: dont WARN about no usable controller

hallyn:
If I start a container with loglevel WARN, and (on a pretty stock ubuntu) do lxc-info -n $c, I get

lxc-start media 20230706233337.765 WARN     cgfsng - cgroups/cgfsng.c:get_hierarchy:142 - There is no useable cpuacct controller
lxc-start media 20230706233337.765 WARN     cgfsng - cgroups/cgfsng.c:get_hierarchy:142 - There is no useable blkio controller

I don't think that's worth WARNing about, so change it to INFO.


brauner:
Yep, sounds good!

PR #4336, opened 2023-08-22
Fix start api call to split quoted strings in execute or init command.

smoser:
If a user of the container.start api call provided NULL for the argv argument, then lxc would load either 'lxc.execute.cmd' or 'lxc.init.cmd' configuration items as the command.

Given a config like:

    lxc.execute.cmd = /usr/bin/touch "file one" "file 2"

lxc would just split the string on spaces and end up executing array:

  ['touch', 'file"', 'one"', '"file', '2"']

This differs from the experience with the `lxc-start` command which would use lxc_string_split_quoted and execute:

  ['touch', 'file one', 'file 2']

Note that as described in lxc_string_split_quoted, commands that include nested quotes and possibly other characters are still a problem.  In those cases, the caller of 'start' can provide an argv array.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -768,51 +768,6 @@ static bool am_single_threaded(void)
 	return count == 1;
 }
 
-static void push_arg(char ***argp, char *arg, int *nargs)
-{
-	char *copy;
-	char **argv;
-
-	copy = must_copy_string(arg);
-
-	do {
-		argv = realloc(*argp, (*nargs + 2) * sizeof(char *));
-	} while (!argv);
-
-	*argp = argv;
-	argv[*nargs] = copy;
-	(*nargs)++;
-	argv[*nargs] = NULL;
-}
-
-static char **split_init_cmd(const char *incmd)
-{
-	__do_free char *copy = NULL;
-	char *p;
-	char **argv;
-	int nargs = 0;
-
-	if (!incmd)
-		return NULL;
-
-	copy = must_copy_string(incmd);
-
-	do {
-		argv = malloc(sizeof(char *));
-	} while (!argv);
-
-	argv[0] = NULL;
-	lxc_iterate_parts (p, copy, " ")
-		push_arg(&argv, p, &nargs);
-
-	if (nargs == 0) {
-		free(argv);
-		return NULL;
-	}
-
-	return argv;
-}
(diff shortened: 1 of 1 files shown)


hallyn (changes requested):
Thanks, looks good.

Can you also remove the split_init_cmd() fn definition?

hallyn:
Hm, so one difference here is that while split_init_cmd() would return NULL if the argument is NULL, lxc_string_split_quoted will return "".  That may mess up fallback handling, since, as you can see, lxcapi_start() next goes on to see if argv is NULL again, to set the default_args.  So the next '!argv' test should also check for argv == ""?

smoser:
> Hm, so one difference here is that while split_init_cmd() would return NULL if the argument is NULL, lxc_string_split_quoted will return "". That may mess up fallback handling, since, as you can see, lxcapi_start() next goes on to see if argv is NULL again, to set the default_args. So the next '!argv' test should also check for argv == ""?

```c
	if (!argv) {
		char *cfgcmd = conf->init_cmd
		if (useinit && conf->execute_cmd)
			cfgcmd = conf->execute_cmd

		if (cfgcmd != NULL)
			argv = init_cmd = lxc_string_split_quoted(cfgcmd)
	}
```

how's that look?

hallyn:
Looks good, but you'll still need to drop the function, CI will fail due to its being unused.

smoser:
> Looks good, but you'll still need to drop the function, CI will fail due to its being unused.

```bash
$ git diff upstream/main  | diffstat -p1
 src/lxc/lxccontainer.c |   36 +++++-------------------------------
 1 file changed, 5 insertions(+), 31 deletions(-)
```

:partying_face: less software in the world!

src/lxc/lxccontainer.c
```diff
@@ -934,10 +906,12 @@ static bool do_lxcapi_start(struct lxc_container *c, int useinit, char * const a
 		return false;
 
 	if (!argv) {
+		char *cfgcmd = conf->init_cmd
```
hallyn, line 909: Someone's been doing golang :)

smoser: so much so that i did not even recognize any of the 3 times i did that in 4 new lines of code. ;;;;

hallyn (approved):

PR #4343, opened 2023-08-31
lxccontainer.h: Move new fields to the end

stgraber:
The recent PR to add set_timeout is adding new fields to the `lxc_container` struct, but as that struct is public API, in fact the main struct in our public API, additions should always be appended to it to avoid breaking ABI.


mihalicyn:
Ugh. My bad! Thanks, @stgraber for noticing and fixing!

PR #4344, opened 2023-09-01
lxc_storage_prepare(): Fix ephemeral copies

gibmat:
It appears that `rootfs->bdev_type` wasn't actually being set anywhere, other than if it comes from a config file.

Fixes: #4199

### DIFF
--- a/src/lxc/conf.c
+++ b/src/lxc/conf.c
@@ -536,16 +536,21 @@ int lxc_rootfs_init(struct lxc_conf *conf, bool userns)
 	struct stat st;
 	struct statfs stfs;
 	struct lxc_rootfs *rootfs = &conf->rootfs;
+	const char *type;
 
 	ret = lxc_storage_prepare(conf);
 	if (ret)
 		return syserror_set(-EINVAL, "Failed to prepare rootfs storage");
+	type = rootfs->storage->type;
+
+	if (!type)
+		return syserror_set(-EINVAL, "Storage type neither set nor automatically detected");
 
 	if (!is_empty_string(rootfs->mnt_opts.userns_path)) {
 		if (!rootfs->path)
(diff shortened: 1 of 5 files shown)


jcharaoui:
I've tested this change on my Debian 12 system and it works. Thank you!

hallyn:
ok to test

hallyn (approved):
lgtm

src/lxc/conf.c
```diff
@@ -513,6 +513,9 @@ int lxc_storage_prepare(struct lxc_conf *conf)
 				 rootfs->path, rootfs->mount,
 				 rootfs->mnt_opts.raw_options ? rootfs->mnt_opts.raw_options : "(null)");
 
+	if (!rootfs->bdev_type)
+		rootfs->bdev_type = strdup(rootfs->storage->type);
```
brauner, line 517: I don't understand. `lxc_rootfs_init()` explicitly allows for `rootfs->bdev_type` to be empty. If it's empty it means that it's supposed to be a directory.

brauner: What's the config file that produces this issue?

gibmat:
If possible I'd like to have a fix for creating ephemeral containers (either my change proposed in this pull request, or perhaps a better solution) included in the Debian 12.2 point release (scheduled for Oct 7). Working back from that date, I need to have the updated package basically ready by the 23rd so there's plenty of time for the update to be approved and included by the release team.

stgraber:
@brauner

gibmat:
@stgraber @brauner quick ping on this again -- Debian has another point release coming up on December 9.

stgraber:
jenkins: test this please

src/lxc/conf.c
brauner: Ok, I pushed a fix on top of your commit and added your Co-Developed-by and Signed-off-by.

PR #4346, opened 2023-09-13
containers in the FREEZING state also need to be unfreeze

zhongTao99:
When lxc is doing cgroup_freeze,：
```c
int cgroup_freeze(const char *name, const char *lxcpath, int timeout)
{
	__do_close int unified_fd = -EBADF;
	int ret;

	if (is_empty_string(name) || is_empty_string(lxcpath))
		return ret_errno(EINVAL);

	unified_fd = lxc_cmd_get_limit_cgroup2_fd(name, lxcpath);
	if (unified_fd < 0)
		return ret_errno(ENOCGROUP2);

	lxc_cmd_notify_state_listeners(name, lxcpath, FREEZING);
	ret = __cgroup_freeze(unified_fd, timeout);
	lxc_cmd_notify_state_listeners(name, lxcpath, !ret ? FROZEN : RUNNING);
	return ret;
}
```

if the lxc process is killed after setting the container status to FREEZING, the container status will be abnormal, unfreezing operation will not be possible, and normal life cycle management will not be possible. Therefore, containers in the FREEZING state also need to be unfreeze

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -544,7 +544,10 @@ static bool do_lxcapi_unfreeze(struct lxc_container *c)
 		return false;
 
 	s = lxc_getstate(c->name, c->config_path, 0);
-	if (s == FROZEN) {
+	// Prevent lxc from unexpectedly exiting when executing freeze,
+	// causing the container to be in the FREEZING state,
+	// making normal life cycle management impossible.
+	if (s == FROZEN || s == FREEZING) {
 		ret = cgroup_unfreeze(c->name, c->config_path, -1);
 		if (ret == -ENOCGROUP2)
 			ret = lxc_unfreeze(c->lxc_conf, c->name, c->config_path);


zhongTao99:
ok to test

zhongTao99:
test this please

zhongTao99:
jenkins: test this please

zhongTao99:
jenkins: test this please

brauner (changes requested):
Hm, I think unfreezing when FREEZING is racy. You need to wait for FROZEN and then unfreeze.

stgraber:
I've definitely seen containers stuck in FREEZING before, no amount of waiting would get them into FROZEN, but going back to THAWED would work fine to unstick them.

zhongTao99:
Thanks both for your replies.
If it is a normal scenario, the status will eventually be FROZEN. But what I am confused about is if the process exits abnormally after being killed, can the status still be FROZEN?

brauner:
> Thanks both for your replies. If it is a normal scenario, the status will eventually be FROZEN. But what I am confused about is if the process exits abnormally after being killed in the FREEZING state, can the state still change to FROZEN normally?

I'm not sure I follow: if the process exits abnormally after being killed in FREEZING state then lxc_getstate() would be returning STOPPED.

mihalicyn:
> My best guess is that it was freezing in the middle of some FUSE operations and found itself stuck there.

this is very-very important point. It explains why you have get this problem.
Actually, most of FUSE operations are uninterruptible (they use `wait_event` that puts task into `TASK_UNINTERRUPTIBLE` sleep). To be freezable task should be put to the `__refrigerator` (https://github.com/torvalds/linux/blob/1c8b86a3799f7e5be903c3f49fcdaee29fd385b5/kernel/freezer.c#L62) but we can't do that when `TASK_UNINTERRUPTIBLE` is set or if task is not sleeping in a proper state (`TASK_INTERRUPTIBLE|TASK_FREEZABLE` state, for example, use `wait_event_freezable`).

stgraber:
@brauner @mihalicyn so is there an actual concern with this change?

I agree that in general FREEZING shouldn't be a state that the kernel remains stuck on, but it's a thing that's be known to happen and when it does, you can absolutely move back to THAWED, so allowing someone to do so through liblxc still seems fine to me.

stgraber (approved):

PR #4370, opened 2023-11-24
Remove support for upstart

simondeziel:
AFAIK, upstart is a defunct project that isn't used by any still supported distribution. Maybe ChromeOS is still using it and would care?


hallyn:
Oh, boy, this will make me very sad :(

stgraber (approved):

PR #4374, opened 2023-12-18
confile_utils: fix incorrect multiply_overflow test

af-airbus:
The check of multiplication overflow to set time offsets (boottime and monotonic) is not correct and does not allow any valid negative value to be accepted.
This commit changes the detection of the multiplication overflow to accept valid negative values for time offsets.

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -2073,14 +2073,7 @@ int fix_stdio_permissions(uid_t uid)
 
 bool multiply_overflow(int64_t base, uint64_t mult, int64_t *res)
 {
-	if (base > 0 && base > (int64_t)(INT64_MAX / mult))
-		return false;
-
-	if (base < 0 && base < (int64_t)(INT64_MIN / mult))
-		return false;
-
-	*res = (int64_t)(base * mult);
-	return true;
+	return __builtin_mul_overflow(base, mult, res);
 }
 
 int print_r(int fd, const char *path)
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -2963,18 +2963,18 @@ static int set_config_time_offset_boot(const char *key, const char *value,
 
 	unit = lxc_trim_whitespace_in_place(buf);
 	if (strequal(unit, "h")) {
-		if (!multiply_overflow(offset, 3600, &lxc_conf->timens.s_boot))
+		if (multiply_overflow(offset, 3600, &lxc_conf->timens.s_boot))
 			return ret_errno(EOVERFLOW);
 	} else if (strequal(unit, "m")) {
-		if (!multiply_overflow(offset, 60, &lxc_conf->timens.s_boot))
+		if (multiply_overflow(offset, 60, &lxc_conf->timens.s_boot))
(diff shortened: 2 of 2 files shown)


stgraber:
Hello,

Your commit doesn't appear to be signed-off, can you please update it to include the required `Signed-off-by` line (see CONTRIBUTING). We'll then be able to review your proposed change.

Thanks!

af-airbus:
The required `Signed-off-by` line has been added.

stgraber:
jenkins: test this please

stgraber:
@mihalicyn @brauner this is ready for your review

src/lxc/utils.c
```diff
@@ -1868,13 +1868,15 @@ int fix_stdio_permissions(uid_t uid)
...
 bool multiply_overflow(int64_t base, uint64_t mult, int64_t *res)
 {
-	if (base > 0 && base > (int64_t)(INT64_MAX / mult))
-		return false;
+	int64_t tmp;
+
+	tmp = base * mult;
 
-	if (base < 0 && base < (int64_t)(INT64_MIN / mult))
+	if (base == (tmp / (int64_t)mult)) {
```
mihalicyn, line 1875: As far as I understand compiler can optimize-out this expression and decide that the whole condition is always true.

I would suggest easier fix, just replace `base < (int64_t)(INT64_MIN / mult)` with `base < -(int64_t)(INT64_MIN / mult)`. It takes into account that for any `mult`, always `(int64_t)(INT64_MIN / mult) > 0`. It's because `mult` is unsigned and wider than `int64_t`.

We can consider using `__builtin_mul_overflow`, but it's available only with GCC versions >= 5.

cc @stgraber @brauner

mihalicyn (changes requested):

mihalicyn:
Hi @af-airbus 

Sorry for so long delay with review and thanks for your report and fix suggested!

src/lxc/utils.c
stgraber: GCC 5 has been around for almost 10 years so should be fine.

af-airbus: Using `__builtin_mul_overflow` seems the best option.
Would it suits everyone if I replace the function body entirely with a single call to this builtin?

mihalicyn: > Would it suits everyone if I replace the function body entirely with a single call to this builtin?

we have special helper already (it's called `check_mul_overflow` and it uses this builtin function).

Dear Arnaud,

I would suggest you to wait a bit and I'll prepare a new PR with this fixed and I'll include your commit (with you authorship preserved ) as a part of this new PR.

The only question I have is that which email you want to use as a git author email? In your signed-off-by tag I see `151835536+af-airbus@users.noreply.github.com` which seems a bit weird github-specific email. Are you sure that you want to go with this address?

mihalicyn: Please, check https://github.com/lxc/lxc/pull/4418

I have added your commit and set you as an author of it (also added your signed-off-by tag)

mihalicyn:
Replaced by https://github.com/lxc/lxc/pull/4418

PR #4377, opened 2023-12-27
build: add more options for customizing install

adamcstephens:
Trying to get the nix lxc package up to 5.x and these options are helpful to avoid writing to places outside the nix sandbox. The default behavior should be unchanged.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -50,7 +50,7 @@ rootfsmount = get_option('rootfs-mount-path')
 user_network_db_opt = get_option('usernet-db-path')
 user_network_conf_opt = get_option('usernet-config-path')
 
-bashcompletiondir = join_paths('/', 'usr', 'share', 'bash-completion', 'completions')
+bashcompletiondir = join_paths(prefixdir, get_option('datadir'), 'bash-completion', 'completions')
 bindir = join_paths(prefixdir, get_option('bindir'))
 datadir = join_paths(prefixdir, get_option('datadir'))
 mandir = join_paths(prefixdir, get_option('mandir'))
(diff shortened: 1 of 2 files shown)


stgraber:
I'd need to do some tests to be sure, but I thought we couldn't do conditional subdir statements because that also had the side effect of not including things in the dist tarball.

stgraber:
(That's unless you pass all the right flags to cover every feature and so get everything in the dist tarball, but that kinda defeats the point as that target should be as reproducible as possible)

adamcstephens:
I’m open to adjusting if needed. I figured setting them to default enable should be sufficient but I’ll admit I don’t know meson very well. This won’t block the upgrade for us, but I do prefer to upstream instead of carrying patches if possible.

adamcstephens:
@stgraber Do you think this will get consideration, or should I plant to just maintain our own patch to enable building in nixpkgs?

stgraber:
It's been on my to-do to do a proper review for a little while. We should be able to accommodate Nix.

adamcstephens:
Thanks! No rush, I was just auditing my open PRs and figured it'd been long enough to check back in.

mihalicyn:
jenkins: ok to test

mihalicyn:
Looks safe and this does not change default behavior.

LGTM!

mihalicyn (approved):

stgraber:
Just did a quick diff of before/after tarballs to make sure this isn't causing a bunch of files to be missed and it looks fine:
```
stgraber@dakara:~/data/code/lxc/lxc/a (install-options)$ diff -Nrup orig new | diffstat
 meson.build       |   79 +++++++++++++++++++++++++++++++++++++++++++++----------------------------------
 meson_options.txt |    8 ++++++++
 2 files changed, 53 insertions(+), 34 deletions(-)
```

stgraber:
And also confirmed that our current dist tarball matches the git tree (as in, we're not missing any files outside of this change).

PR #4379, opened 2024-01-03
lxcmntent: cleanning the buffer before get data in getmntent_r

chenxy1988:
The buffer in getmntent_r used in other functions before (e.g. src/lxc/criu.c exec_criu()). We observed that on some platforms, criu cannot run succesfully due to parameters was wrong. After checking the parameters, we found that the the buffer contains old data and those old data also insert to criu's arguments. So add a cleanning the old data in buffer by memset before using fget.

### DIFF
--- a/src/include/lxcmntent.c
+++ b/src/include/lxcmntent.c
@@ -82,6 +82,8 @@ struct mntent *getmntent_r(FILE *stream, struct mntent *mp, char *buffer, int bu
 	do {
 		char *end_ptr;
 
+		memset((void *)buffer, 0x0, (size_t bufsiz));
+
 		if (!fgets(buffer, bufsiz, stream))
 			return NULL;
 


stgraber:
@brauner @mihalicyn @hallyn ready for review

hallyn:
It shouldn't break anything, but I don't understand how you can get to an error condition: the fgets will ensure that there is a terminating \0, so everything up to that must have been just read.  Re-using the buffer should not be possible.  So I'm worried this may be papering over another (actual) programming error which may now become even harder to trigger.

Can you show a simple reproducer?

chenxy1988:
> It shouldn't break anything, but I don't understand how you can get to an error condition: the fgets will ensure that there is a terminating \0, so everything up to that must have been just read. Re-using the buffer should not be possible. So I'm worried this may be papering over another (actual) programming error which may now become even harder to trigger.
> 
> Can you show a simple reproducer?

Hi @hallyn ,

Thanks for your reply, yes, this is not root cause, this is just lucky against the issue, so I closed this PR.

Currently I found that seems something wrong in exec_criu() on some platform(e.g. yocto ).
The behavior is criu parameter's array (args->argv) was broken, that cause criu cannot execute successfully.
On that platform, below warning message in exe_criu() always output in log file.
WARN("No cgroup controllers configured in container's cgroup %s", cgroup_base_path);

I'll turn on gdb and rebuild it again to try to find what happens on and keep updating the thread if there have more information.

PR #4386, opened 2024-01-15
fix uninitialized var in  lxccontainer.c

songtianlun:
lxccontainer.c: In function 'list_defined_containers': lxccontainer.c:5481:5: error: 'c' may be used uninitialized in this function [-Werror=maybe-uninitialized]
 5481 |     lxc_container_put(c);
      |


stgraber:
Closing as this PR is targeting the 4.0 branch which is in the security-only part of its maintenance and not expected to get any more releases except for critical security issues.

Other fixes are expected to always go through the main branch and then get included in relevant stable releases which still get bugfixes (5.0 for example).

PR #4403, opened 2024-02-19
github: master -> main

mihalicyn:
Our main branch is "main" for some time, but
in github scripts we still use "master" which is incorrect and prevent some stuff from working.


mihalicyn:
O! Now "Fuzzing with OSS-fuzz" runs on PRs too as it should be.

mihalicyn:
Ok. Safe to merge.

PR #4411, opened 2024-03-11
unshare: simplify lookup_name

hallyn:
pass the optarg straight to getpwnam_r, instead of first scanning the name out of it.


hallyn:
It's possible there's some reason why getpwnam_r requires the sscanf, but presumably if so the tests will catch it...

PR #4414, opened 2024-03-17
README: Correct 'armvl7' to 'armv7l'

cjavad:
Use the correct designation for the armv7l arch, as also reflected in src/lxc/confile.c.

Smaller nit that was the result of series of misconfigurations due to the mistake of using armvl7 in our configurations instead of armv7l.


stgraber (approved):
Nice catch!

PR #4438, opened 2024-05-01
sscanf solution with buffer overflow prevention

### LINKED ISSUE #4366: usage of sscanf with format %s is unsafe
if `oparg` length is bigger than `name` buffer length, it will cause buffer overflow because the width of %s is not provied.

insecure code:
```
if (sscanf(oparg, "%s", name) < 1) {
```

https://github.com/lxc/lxc/blob/9c1d70ef8b393d42fc089e89b16a3ec7d8769ff1/src/lxc/tools/lxc_unshare.c#L166C1-L166C1

maybe we should fix it like this:
```
if (sscanf(oparg, "%4095s", name) < 1)
```

DevonSchwartz:
This fix is related to issue https://github.com/lxc/lxc/issues/4366.

This issue points to unsafe call to sscanf to get a username in the case of an invalid uid. The sscanf call has a potential buffer overflow. However, this sscanf call is not implimented in main. We implement the sscanf call safely in main by limiting user input to the size of PATH_MAX. 

This commit creates a warning that fmt_str is not a string literal. However, fmt_str is equal to "%[PATH_MAX]s" 

fix #4366

### DIFF
--- a/src/lxc/tools/lxc_unshare.c
+++ b/src/lxc/tools/lxc_unshare.c
@@ -92,7 +92,6 @@ Options :\n\
 static int my_parser(struct lxc_arguments *args, int c, char *arg)
 {
 	struct lxc_list *tmplist;
-
 	switch (c) {
 	case 's':
 		args->flags = get_namespace_flags(arg);
@@ -144,6 +143,7 @@ static int get_namespace_flags(char *namespaces)
 
 static bool lookup_user(const char *oparg, uid_t *uid)
 {
+	char name[PATH_MAX];
 	struct passwd pwent;
 	struct passwd *pwentp = NULL;
 	char *buf;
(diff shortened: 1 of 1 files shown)


DevonSchwartz:
I saw that the sanitizer fails because of a warning produced by the string literal fmt_str. We make fmt_str a string literal by using [] rather than dynamically allocating, but we still get this error. Should we still change our code to avoid the error?

stgraber:
Yeah, pretty much everyone runs with those warnings enabled in production so that needs to be addressed.

DevonSchwartz:
Is the check for the size of PATH necessary before we read in the buffer?

DevonSchwartz:
Should I squash the latest commits to close the sscanf issue?

stgraber:
Probably good to squash. I'm really not sure about the fix though but I'm also not a C developer so I'll let @mihalicyn, @brauner or @hallyn look at it.

mihalicyn:
I don't understand this. We are not using `%s` in the original code in this place. Please can you elaborate on this a bit?

Blub:
As far as I can tell this was addressed by 046b63cd2034 ("unshare: simplify lookup_name"). It's simply not necessary to do any `sscanf` call there. The mentioned commit removes the `name` variable and instead passes the argument straight to `getpwnam_r`.

Issue #4337, opened 2023-08-28
lxc-top informs 0.00% of CPU usage on a container, even when its on 100%


paulocoghi:
## Environment info

 * Distribution: Ubuntu
 * Distribution version: 22.04.3
 * Kernel version: 6.4.12-zabbly

--

 * `lxc-start --version` : 5.0.0~git2209-g5a7b9ce67
 * `lxc-checkconfig`

<details>
  <summary>lxc-checkconfig</summary>

```sh
LXC version 5.0.0~git2209-g5a7b9ce67
Kernel configuration not found at /proc/config.gz; searching...
Kernel configuration found at /boot/config-6.4.12-zabbly+
--- Namespaces ---
Namespaces: enabled
Utsname namespace: enabled
Ipc namespace: enabled
Pid namespace: enabled
User namespace: enabled
Network namespace: enabled

--- Control groups ---
Cgroups: enabled
Cgroup namespace: enabled

Cgroup v1 mount points: 


Cgroup v2 mount points: 
/sys/fs/cgroup

Cgroup v1 systemd controller: missing
Cgroup v1 freezer controller: missing
Cgroup ns_cgroup: required
Cgroup device: enabled
Cgroup sched: enabled
Cgroup cpu account: enabled
Cgroup memory controller: enabled
Cgroup cpuset: enabled

--- Misc ---
Veth pair device: enabled, loaded
Macvlan: enabled, not loaded
Vlan: enabled, not loaded
Bridges: enabled, loaded
Advanced netfilter: enabled, loaded
CONFIG_IP_NF_TARGET_MASQUERADE: enabled, not loaded
CONFIG_IP6_NF_TARGET_MASQUERADE: enabled, not loaded
CONFIG_NETFILTER_XT_TARGET_CHECKSUM: enabled, loaded
CONFIG_NETFILTER_XT_MATCH_COMMENT: enabled, not loaded
FUSE (for use with lxcfs): enabled, not loaded

--- Checkpoint/Restore ---
checkpoint restore: enabled
CONFIG_FHANDLE: enabled
CONFIG_EVENTFD: enabled
CONFIG_EPOLL: enabled
CONFIG_UNIX_DIAG: enabled
CONFIG_INET_DIAG: enabled
CONFIG_PACKET_DIAG: enabled
CONFIG_NETLINK_DIAG: enabled
File capabilities: 

Note : Before booting a new kernel, you can check its configuration
[...]

PR #4426, opened 2024-04-03
tools: lxc-info: Print memory and cpu utilization on cgroup2 system

petris:
Emulate the cgroup1 behavior by calculating memory.kmem.usage_in_bytes and memory.usage_in_bytes using information provided in memory.stat and use cpu.stat/usage_usec instead of cpuacct.usage.

Fixes: https://github.com/lxc/lxc/issues/3389

PR #4439, opened 2024-05-01
Fixed cgroup2 path issue in LXC-Top

### LINKED ISSUE #3372: cgroupv2 support in lxc-top
# Required information

 * Distribution: Fedora
 * Distribution version: 31
   * `lxc-start --version`:  `4.0.1`
   * `uname -a`: `Linux [...] 5.5.15-200.fc31.x86_64 #1 SMP Thu Apr 2 19:16:17 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

# Issue description
```
$ lxc-top -n ubuntu19 -b
Unable to read cgroup item memory.usage_in_bytes
Unable to read cgroup item memory.limit_in_bytes
Unable to read cgroup item memory.memsw.usage_in_bytes
Unable to read cgroup item memory.memsw.limit_in_bytes
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item cpuacct.usage
Unable to read cgroup item cpuacct.stat
Unable to read cgroup item cpuacct.stat
Unable to read cgroup item blkio.throttle.io_service_bytes
Unable to read cgroup item blkio.throttle.io_serviced
1586610900483,ubuntu19,0,0,0,3617017222580694625,2665,0,0,0
```

The output seems not to be complete, and, in the case of BlkIO, numbers are unrealistic. Does lxc-top support cgroupv2?

### LINKED ISSUE #4376: lxc-top does not work with openrc and cgroup2 (Alpine Linux)
The template below is mostly useful for bug reports and support questions.
Feel free to remove anything which doesn't apply to you and add more information where it makes sense.

# Required information

 * Distribution: Alpine Linux
 * Distribution version: 3.19.0
 * The output of
   * `lxc-start --version`: 5.0.3
   * `lxc-checkconfig`:
```
LXC version 5.0.3                                                       
                                                                        
--- Namespaces ---                                                      
Namespaces: enabled                                                     
Utsname namespace: enabled                                              
Ipc namespace: enabled                                                  
Pid namespace: enabled                                                  
User namespace: enabled                                                 
newuidmap is not installed                                              
newgidmap is not installed                                              
Network namespace: enabled

--- Control groups ---                                                  
Cgroups: enabled                                                        
Cgroup namespace: enabled                                               
Cgroup v1 mount points:                                                 
Cgroup v2 mount points:                                                 
 - /sys/fs/cgroup                                                       
Cgroup device: enabled                                                  
Cgroup sched: enabled                                                   
Cgroup cpu account: enabled                                             
Cgroup memory controller: enabled                                       
Cgroup cpuset: enabled                                                  
                                                                        
--- Misc ---                                                            
Veth pair device: enabled, loaded                                       
Macvlan: enabled, not loaded                                            
Vlan: enabled, not loaded
Bridges: enabled, loaded
Advanced netfilter: enabled, loaded
CONFIG_IP_NF_TARGET_MASQUERADE: enabled, not loaded
CONFIG_IP6_NF_TARGET_MASQUERADE: enabled, not loaded
CONFIG_NETFILTER_XT_TARGET_CHECKSUM: enabled, loaded
CONFIG_NETFILTER_XT_MATCH_COMMENT: enabled, not loaded
FUSE (for use with lxcfs): enabled, not loaded

--- Checkpoint/Restore ---
checkpoint restore: enabled
CONFIG_FHANDLE: enabled
CONFIG_EVENTFD: enabled
CONFIG_EPOLL: enabled
CONFIG_UNIX_DIAG: enabled
CONFIG_INET_DIAG: enabled
CONFIG_PACKET_DIAG: enabled
CONFIG_NETLINK_DIAG: enabled
File capabilities: enabled

Note : Before booting a new kernel, you can check its configuration
usage : CONFIG=/path/to/config /usr/bin/lxc-checkconfig
```
   * `uname -a`: Linux alpine 6.6.4-1-virt #2-Alpine SMP PREEMPT_DYNAMIC Tue, 05 Dec 2023 20:25:50 +0000 x86_64 Linux

   * `cat /proc/self/cgroup`: 0::/
   * `cat /proc/1/mounts`: 
```
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
devtmpfs /dev devtmpfs rw,nosuid,noexec,relatime,size=10240k,nr_inodes=252290,mode=755,inode64 0 0
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000 0 0
shm /dev/shm tmpfs rw,nosuid,nodev,noexec,relatime,inode64 0 0
/dev/sr0 /media/cdrom iso9660 ro,relatime,nojoliet,check=s,map=n,blocksize=2048,iocharset=utf8 0 0
tmpfs / tmpfs rw,relatime,mode=755,inode64 0 0
tmpfs /run tmpfs rw,nosuid,nodev,size=405928k,nr_inodes=819200,mode=755,inode64 0 0
mqueue /dev/mqueue mqueue rw,nosuid,nodev,noexec,relatime 0 0
/dev/loop0 /.modloop squashfs ro,relatime,errors=continue 0 0
securityfs /sys/kernel/security securityfs rw,nosuid,nodev,noexec,relatime 0 0
debugfs /sys/kernel/debug debugfs rw,nosuid,nodev,noexec,relatime 0 0
pstore /sys/fs/pstore pstore rw,nosuid,nodev,noexec,relatime 0 0
tracefs /sys/kernel/debug/tracing tracefs rw,nosuid,nodev,noexec,relatime 0 0
none /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate 0 0
```
# Issue description

`lxc-top` stopped working after OpenRC switched to use unified cgroup (cgroup2). this comes as the default in alpine 3.19.0.

This is the output of `lxc-top -b`:
```
time_ms,container,cpu_nanos,cpu_sys_userhz,cpu_user_userhz,blkio_bytes,blkio_iops,mem_used_bytes,memsw_used_bytes,kernel_mem_used_bytes
Unable to read cgroup item memory.usage_in_bytes
Unable to read cgroup item memory.limit_in_bytes
Unable to read cgroup item memory.memsw.usage_in_bytes
Unable to read cgroup item memory.memsw.limit_in_bytes
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item cpuacct.usage
Unable to read cgroup item cpuacct.stat
Unable to read cgroup item cpuacct.stat
Unable to read cgroup item blkio.throttle.io_service_bytes
Unable to read cgroup item blkio.throttle.io_serviced
```

# Steps to reproduce

 1. Start alpine in qemu: `qemu-system-x86_64 -m 2048 -accel kvm -cdrom https://dl-cdn.alpinelinux.org/alpine/v3.19/releases/x86_64/alpine-virt-3.19.0-
x86_64.iso -nographic`
 2. login as root and execute `setup-alpine -q` to set up basic network apk repositories.
 3. install packages: `apk add iptables lxc lxc-bridge lxc-templates-legacy-alpine`
 4. start base services: `service cgroups start && service dnsmasq.lxcbr0 start`
 5. create a simple lxc container:  `lxc-create -t alpine -n a1`
 6. start the container: `lxc-start -n a1`
 7. run `lxc-top -b`

# Information to attach

 - [x] any relevant kernel output (`dmesg`)
 ```
[  136.939216] bridge: filtering via arp/ip/ip6tables is no longer available by default. Update your scripts to load br_netfilter if you need t.
[  190.764549] lxc-create[2436]: memfd_create() called without MFD_EXEC or MFD_NOEXEC_SEAL set
[  226.589710] lxcbr0: port 1(vethXOfaJf) entered blocking state
[  226.589712] lxcbr0: port 1(vethXOfaJf) entered disabled state
```
 - [x] container log (The <log> file from running `lxc-start -n <c> -l TRACE -o <logfile> `) [a1.trace.log](https://github.com/lxc/lxc/files/13753010/a1.trace.log)
 - [x] the containers configuration file  [a1.config.txt](https://github.com/lxc/lxc/files/13753035/a1.config.txt)

### PR DESCRIPTION
**Pull request by DevonSchwartz**
We implimented some of the functionality to convert the cgroup paths to cgroup2 paths. However, we do not know how to tell whether system is using cgroup or cgroup2. In addition, some of the paths for cgroup did not have cgroup2 equivalents in the incus directory. 

Fixes #4376
Fixes https://github.com/lxc/lxc/issues/3372

### DIFF
--- a/src/lxc/tools/lxc_top.c
+++ b/src/lxc/tools/lxc_top.c
@@ -41,10 +41,18 @@ struct cpu_stats {
 };
 
 struct mem_stats {
+	union {
+		struct {
+			uint64_t swap_used;
+			uint64_t swap_limit;
+		}; /* v2 only */
+		struct {
+			uint64_t memsw_used;
+			uint64_t memsw_limit;
+		}; /* v1 only */
+	};
 	uint64_t used;
 	uint64_t limit;
-	uint64_t swap_used;
-	uint64_t swap_limit;
 	uint64_t kmem_used;
 	uint64_t kmem_limit;
 };
@@ -362,12 +370,12 @@ out:
 
 static int cg1_mem_stats(struct lxc_container *c, struct mem_stats *mem)
 {
-	mem->used       = stat_get_int(c, "memory.usage_in_bytes");
-	mem->limit      = stat_get_int(c, "memory.limit_in_bytes");
-	mem->swap_used  = stat_get_int(c, "memory.memsw.usage_in_bytes");
-	mem->swap_limit = stat_get_int(c, "memory.memsw.limit_in_bytes");
-	mem->kmem_used  = stat_get_int(c, "memory.kmem.usage_in_bytes");
-	mem->kmem_limit = stat_get_int(c, "memory.kmem.limit_in_bytes");
+	mem->used        = stat_get_int(c, "memory.usage_in_bytes");
+	mem->limit       = stat_get_int(c, "memory.limit_in_bytes");
+	mem->memsw_used  = stat_get_int(c, "memory.memsw.usage_in_bytes");
+	mem->memsw_limit = stat_get_int(c, "memory.memsw.limit_in_bytes");
+	mem->kmem_used   = stat_get_int(c, "memory.kmem.usage_in_bytes");
+	mem->kmem_limit  = stat_get_int(c, "memory.kmem.limit_in_bytes");
 	return mem->used > 0 ? 0 : -1;
 }
 
@@ -423,6 +431,7 @@ static void stats_get(struct lxc_container *c, struct container_stats *ct, struc
 		/* only with cgroups v1 */
 		cg1_get_blk_stats(c, "blkio.throttle.io_serviced", &ct->stats->io_serviced);
 	}
+
 
 	if (total) {
 		total->mem.used       += ct->stats->mem.used;
--- a/src/lxc/tools/lxc_top.c
+++ b/src/lxc/tools/lxc_top.c
@@ -281,21 +281,21 @@ examples:
 	8:0 Total 149327872
 	Total 149327872
 */
-static void stat_get_blk_stats(struct lxc_container *c, const char *item,
+static int cg1_get_blk_stats(struct lxc_container *c, const char *item,
 			      struct blkio_stats *stats) {
 	char buf[4096];
 	int i, len;
 	char **lines, **cols;
+	int ret = -1;
 
 	len = c->get_cgroup_item(c, item, buf, sizeof(buf));
 	if (len <= 0 || (size_t)len >= sizeof(buf)) {
-		fprintf(stderr, "Unable to read cgroup item %s\n", item);
-		return;
+		return ret;
 	}
 
 	lines = lxc_string_split_and_trim(buf, '\n');
 	if (!lines)
-		return;
+		return ret;
 
 	memset(stats, 0, sizeof(struct blkio_stats));
 
@@ -314,10 +314,50 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
 
 		lxc_free_array((void **)cols, free);
 	}
-
+	ret = 0;
 out:
 	lxc_free_array((void **)lines, free);
-	return;
+	return ret;
+}
+
+static int cg2_get_blk_stats(struct lxc_container *c, const char *item,
+			      struct blkio_stats *stats) {
+	char buf[4096];
+	int i, j, len;
+	char **lines, **cols;
+	int ret = -1;
+
+	len = c->get_cgroup_item(c, item, buf, sizeof(buf));
+	if (len <= 0 || (size_t)len >= sizeof(buf)) {
+		return ret;
+	}
+
+	lines = lxc_string_split_and_trim(buf, '\n');
+	if (!lines)
+		return ret;
+
+	memset(stats, 0, sizeof(struct blkio_stats));
+
+	for (i = 0; lines[i]; i++) {
+		cols = lxc_string_split_and_trim(lines[i], ' ');
+		if (!cols)
+			goto out;
+
+		for (j = 0; cols[j]; j++) {
+			if (strncmp(cols[j], "rbytes=", 7) == 0) {
+				stats->read += strtoull(&cols[j][7], NULL, 0);
+			} else if (strncmp(cols[j], "wbytes=", 7) == 0) {
+				stats->write += strtoull(&cols[j][7], NULL, 0);
+			}
+		}
+
+		lxc_free_array((void **)cols, free);
+	}
+	stats->total = stats->read + stats->write;
+	ret = 0;
+out:
+	lxc_free_array((void **)lines, free);
+	return ret;
 }
 
 static int cg1_mem_stats(struct lxc_container *c, struct mem_stats *mem)
@@ -376,8 +416,14 @@ static void stats_get(struct lxc_container *c, struct container_stats *ct, struc
 		}
 	}
 
-	stat_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes);
-	stat_get_blk_stats(c, "blkio.throttle.io_serviced", &ct->stats->io_serviced);
+	if (cg1_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes) < 0) {
+		if (cg2_get_blk_stats(c, "io.stat", &ct->stats->io_service_bytes) < 0) {
+			fprintf(stderr, "Unable to read IO stats\n");
+		}
+	} else {
+		/* only with cgroups v1 */
+		cg1_get_blk_stats(c, "blkio.throttle.io_serviced", &ct->stats->io_serviced);
+	}
 
 	if (total) {
 		total->mem.used       += ct->stats->mem.used;
--- a/src/lxc/tools/lxc_top.c
+++ b/src/lxc/tools/lxc_top.c
@@ -373,14 +373,13 @@ static int cg1_mem_stats(struct lxc_container *c, struct mem_stats *mem)
 
 static int cg2_mem_stats(struct lxc_container *c, struct mem_stats *mem)
 {
-	mem->used       = stat_get_int(c, "memory.current");
-	mem->limit      = stat_get_int(c, "memory.max");
-	mem->swap_used  = stat_get_int(c, "memory.swap.current");
-	mem->swap_limit = stat_get_int(c, "memory.swap.max");
-	/* TODO: find the kernel usage */
-	mem->kmem_used  = 0;
+	mem->used          = stat_get_int(c, "memory.current");
+	mem->limit         = stat_get_int(c, "memory.max");
+	mem->swap_used     = stat_get_int(c, "memory.swap.current");
+	mem->swap_limit    = stat_get_int(c, "memory.swap.max");
+	mem->kmem_used     = stat_match_get_int(c, "memory.stat", "kernel", 1);
 	/* does not exist in cgroup v2 */
-	mem->kmem_limit = 0;
+	// mem->kmem_limit = 0;
 	return mem->used > 0 ? 0 : -1;
 }
 
--- a/src/lxc/tools/lxc_top.c
+++ b/src/lxc/tools/lxc_top.c
@@ -35,16 +35,24 @@ struct blkio_stats {
 	uint64_t total;
 };
 
-struct stats {
-	uint64_t mem_used;
-	uint64_t mem_limit;
-	uint64_t memsw_used;
-	uint64_t memsw_limit;
+struct cpu_stats {
+	uint64_t use_nanos;
+	uint64_t use_user;
+	uint64_t use_sys;
+};
+
+struct mem_stats {
+	uint64_t used;
+	uint64_t limit;
+	uint64_t swap_used;
+	uint64_t swap_limit;
 	uint64_t kmem_used;
 	uint64_t kmem_limit;
-	uint64_t cpu_use_nanos;
-	uint64_t cpu_use_user;
-	uint64_t cpu_use_sys;
+};
+
+struct stats {
+	struct mem_stats mem;
+	struct cpu_stats cpu;
 	struct blkio_stats io_service_bytes;
 	struct blkio_stats io_serviced;
 };
@@ -314,34 +322,45 @@ out:
 	return;
 }
 
+static void cg1_mem_stats(struct lxc_container *c, struct mem_stats *mem)
+{
+	mem->used       = stat_get_int(c, "memory.usage_in_bytes");
+	mem->limit      = stat_get_int(c, "memory.limit_in_bytes");
+	mem->swap_used  = stat_get_int(c, "memory.memsw.usage_in_bytes");
+	mem->swap_limit = stat_get_int(c, "memory.memsw.limit_in_bytes");
+	mem->kmem_used  = stat_get_int(c, "memory.kmem.usage_in_bytes");
+	mem->kmem_limit = stat_get_int(c, "memory.kmem.limit_in_bytes");
+}
+
+static void cg1_cpu_stats(struct lxc_container *c, struct cpu_stats *cpu)
+{
+	cpu->use_nanos = stat_get_int(c, "cpuacct.usage");
+	cpu->use_user  = stat_match_get_int(c, "cpuacct.stat", "user", 1);
+	cpu->use_sys   = stat_match_get_int(c, "cpuacct.stat", "system", 1);
+}
+
 static void stats_get(struct lxc_container *c, struct container_stats *ct, struct stats *total)
 {
 	ct->c = c;
-	ct->stats->mem_used      = stat_get_int(c, "memory.usage_in_bytes");
-	ct->stats->mem_limit     = stat_get_int(c, "memory.limit_in_bytes");
-	ct->stats->memsw_used    = stat_get_int(c, "memory.memsw.usage_in_bytes");
-	ct->stats->memsw_limit   = stat_get_int(c, "memory.memsw.limit_in_bytes");
-	ct->stats->kmem_used     = stat_get_int(c, "memory.kmem.usage_in_bytes");
-	ct->stats->kmem_limit    = stat_get_int(c, "memory.kmem.limit_in_bytes");
-	ct->stats->cpu_use_nanos = stat_get_int(c, "cpuacct.usage");
-	ct->stats->cpu_use_user  = stat_match_get_int(c, "cpuacct.stat", "user", 1);
-	ct->stats->cpu_use_sys   = stat_match_get_int(c, "cpuacct.stat", "system", 1);
-
+	cg1_mem_stats(c, &ct->stats->mem);
+	cg1_cpu_stats(c, &ct->stats->cpu);
 	stat_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes);
 	stat_get_blk_stats(c, "blkio.throttle.io_serviced", &ct->stats->io_serviced);
 
 	if (total) {
-		total->mem_used      = total->mem_used      + ct->stats->mem_used;
-		total->mem_limit     = total->mem_limit     + ct->stats->mem_limit;
-		total->memsw_used    = total->memsw_used    + ct->stats->memsw_used;
-		total->memsw_limit   = total->memsw_limit   + ct->stats->memsw_limit;
-		total->kmem_used     = total->kmem_used     + ct->stats->kmem_used;
-		total->kmem_limit    = total->kmem_limit    + ct->stats->kmem_limit;
-		total->cpu_use_nanos = total->cpu_use_nanos + ct->stats->cpu_use_nanos;
-		total->cpu_use_user  = total->cpu_use_user  + ct->stats->cpu_use_user;
-		total->cpu_use_sys   = total->cpu_use_sys   + ct->stats->cpu_use_sys;
+		total->mem.used       += ct->stats->mem.used;
+		total->mem.limit      += ct->stats->mem.limit;
+		total->mem.swap_used  += ct->stats->mem.swap_used;
+		total->mem.swap_limit += ct->stats->mem.swap_limit;
+		total->mem.kmem_used  += ct->stats->mem.kmem_used;
+		total->mem.kmem_limit += ct->stats->mem.kmem_limit;
+
+		total->cpu.use_nanos += ct->stats->cpu.use_nanos;
+		total->cpu.use_user  += ct->stats->cpu.use_user;
+		total->cpu.use_sys   += ct->stats->cpu.use_sys;
+
 		total->io_service_bytes.total += ct->stats->io_service_bytes.total;
-		total->io_service_bytes.read += ct->stats->io_service_bytes.read;
+		total->io_service_bytes.read  += ct->stats->io_service_bytes.read;
 		total->io_service_bytes.write += ct->stats->io_service_bytes.write;
 	}
 }
@@ -351,19 +370,19 @@ static void stats_print_header(struct stats *stats)
 	printf(TERMRVRS TERMBOLD);
 	printf("%-18s %12s %12s %12s %36s %10s", "Container", "CPU",  "CPU",  "CPU",  "BlkIO", "Mem");
 
-	if (stats->memsw_used > 0)
+	if (stats->mem.swap_used > 0)
 		printf(" %10s", "MemSw");
 
-	if (stats->kmem_used > 0)
+	if (stats->mem.kmem_used > 0)
 		printf(" %10s", "KMem");
 	printf("\n");
 
 	printf("%-18s %12s %12s %12s %36s %10s", "Name",      "Used", "Sys",  "User", "Total(Read/Write)", "Used");
 
-	if (stats->memsw_used > 0)
+	if (stats->mem.swap_used > 0)
 		printf(" %10s", "Used");
 
-	if (stats->kmem_used > 0)
+	if (stats->mem.kmem_used > 0)
 		printf(" %10s", "Used");
 
 	printf("\n");
@@ -388,7 +407,7 @@ static void stats_print(const char *name, const struct stats *stats,
 		size_humanize(stats->io_service_bytes.total, iosb_total_str, sizeof(iosb_total_str));
 		size_humanize(stats->io_service_bytes.read, iosb_read_str, sizeof(iosb_read_str));
 		size_humanize(stats->io_service_bytes.write, iosb_write_str, sizeof(iosb_write_str));
-		size_humanize(stats->mem_used, mem_used_str, sizeof(mem_used_str));
+		size_humanize(stats->mem.used, mem_used_str, sizeof(mem_used_str));
 
 		ret = snprintf(iosb_str, sizeof(iosb_str), "%s(%s/%s)", iosb_total_str, iosb_read_str, iosb_write_str);
 		if (ret < 0 || (size_t)ret >= sizeof(iosb_str))
@@ -396,18 +415,18 @@ static void stats_print(const char *name, const struct stats *stats,
 
 		printf("%-18.18s %12.2f %12.2f %12.2f %36s %10s",
 		       name,
-		       (float)stats->cpu_use_nanos / 1000000000,
-		       (float)stats->cpu_use_sys  / USER_HZ,
-		       (float)stats->cpu_use_user / USER_HZ,
+		       (float)stats->cpu.use_nanos / 1000000000,
+		       (float)stats->cpu.use_sys  / USER_HZ,
+		       (float)stats->cpu.use_user / USER_HZ,
 		       iosb_str,
 		       mem_used_str);
 
-		if (total->memsw_used > 0) {
-			size_humanize(stats->memsw_used, memsw_used_str, sizeof(memsw_used_str));
+		if (total->mem.swap_used > 0) {
+			size_humanize(stats->mem.swap_used, memsw_used_str, sizeof(memsw_used_str));
 			printf(" %10s", memsw_used_str);
 		}
-		if (total->kmem_used > 0) {
-			size_humanize(stats->kmem_used, kmem_used_str, sizeof(kmem_used_str));
+		if (total->mem.kmem_used > 0) {
+			size_humanize(stats->mem.kmem_used, kmem_used_str, sizeof(kmem_used_str));
 			printf(" %10s", kmem_used_str);
 		}
 	} else {
@@ -415,11 +434,11 @@ static void stats_print(const char *name, const struct stats *stats,
 		time_ms = (unsigned long long) (time_val.tv_sec) * 1000 + (unsigned long long) (time_val.tv_usec) / 1000;
 		printf("%" PRIu64 ",%s,%" PRIu64 ",%" PRIu64 ",%" PRIu64
 		       ",%" PRIu64 ",%" PRIu64 ",%" PRIu64 ",%" PRIu64 ",%" PRIu64,
-		       (uint64_t)time_ms, name, (uint64_t)stats->cpu_use_nanos,
-		       (uint64_t)stats->cpu_use_sys,
-		       (uint64_t)stats->cpu_use_user, (uint64_t)stats->io_service_bytes.total,
-		       (uint64_t)stats->io_serviced.total, (uint64_t)stats->mem_used,
-		       (uint64_t)stats->memsw_used, (uint64_t)stats->kmem_used);
+		       (uint64_t)time_ms, name, (uint64_t)stats->cpu.use_nanos,
+		       (uint64_t)stats->cpu.use_sys,
+		       (uint64_t)stats->cpu.use_user, (uint64_t)stats->io_service_bytes.total,
+		       (uint64_t)stats->io_serviced.total, (uint64_t)stats->mem.used,
+		       (uint64_t)stats->mem.swap_used, (uint64_t)stats->mem.kmem_used);
 	}
 
 }
@@ -441,9 +460,9 @@ static int cmp_cpuuse(const void *sct1, const void *sct2)
 	const struct container_stats *ct2 = sct2;
 
 	if (sort_reverse)
-		return ct2->stats->cpu_use_nanos < ct1->stats->cpu_use_nanos;
+		return ct2->stats->cpu.use_nanos < ct1->stats->cpu.use_nanos;
 
-	return ct1->stats->cpu_use_nanos < ct2->stats->cpu_use_nanos;
+	return ct1->stats->cpu.use_nanos < ct2->stats->cpu.use_nanos;
 }
 
 static int cmp_blkio(const void *sct1, const void *sct2)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by stgraber** [comment]
This pull request has a bunch of changes to unrelated files, you probably want to clean that up.

**Comment by DevonSchwartz** [comment]
Just cleaned up the file. I am now going to implement calling cgroup2 paths first and then cgroup afterwards.

**Comment by DevonSchwartz** [comment]
Added cgroup2 and cgroup functionality so that warnings are not encountered. If any of the cgroup2 stat calls return an error code, then we call the cgroup container statistics.

**Comment by stgraber** [comment]
The PR needs to be rebased as it's incorrectly reverting some changes now.

I also think it'd be best to have each individual metric attempt cgroup2 and fallback to cgroup1 as the current logic you have will fail if you're on a system that's cgroup2 but is somehow missing swap accounting, it will then assume it's cgroup1 and will fail even more as none of those files will be found.

**Comment by DevonSchwartz** [comment]
I'll go ahead and push a commit to resolve the conflict in network.c. I must have accidentally reverted to a change before the latest commit to main that fixed a network.c issue.

**Comment by DevonSchwartz** [comment]
Just rebased and unstaged irrelevant file changes.

File: src/lxc/tools/lxc_top.c [hunk 1]
```diff
@@ -314,21 +320,58 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
+	if (!success) {
+		stat_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes, &success);
+	}
+
+	// paths only exist in cgroup1
 	ct->stats->kmem_used     = stat_get_int(c, "memory.kmem.usage_in_bytes");
 	ct->stats->kmem_limit    = stat_get_int(c, "memory.kmem.limit_in_bytes");
-	ct->stats->cpu_use_nanos = stat_get_int(c, "cpuacct.usage");
-	ct->stats->cpu_use_user  = stat_match_get_int(c, "cpuacct.stat", "user", 1);
-	ct->stats->cpu_use_sys   = stat_match_get_int(c, "cpuacct.stat", "system", 1);
```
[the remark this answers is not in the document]
DevonSchwartz: The issue was related to how stats_match_get_int was called in try_cgroup2. I used "user" as the match for both calls. I fixed so that now the function takes a parameter for the type of match.

DevonSchwartz:
> Looks reasonable to me. Can you do a rebase to turn this whole thing into a single commit which also has the `Closes #4376` line so it closes the issue?
> 
> @mihalicyn can you review?

Awesome! I squashed into one commit and added the closing line.

ncopa:
I have applied this to lxc-6.0.0 in Alpine Linux but it does not seem to solve the issue.
```
# lxc-top
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
Unable to read cgroup item memory.usage_in_bytes
Segmentation fault (core dumped)
```

I have a backtrace:
```
[New LWP 24554]
Core was generated by `lxc-top'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f3667931573 in strlen (s=<optimized out>, s@entry=0x0) at src/string/strlen.c:17

warning: 17     src/string/strlen.c: No such file or directory
(gdb) 
#0  0x00007f3667931573 in strlen (s=<optimized out>, s@entry=0x0) at src/string/strlen.c:17
#1  0x000056275c5f4987 in stat_get_blk_stats (c=c@entry=0x7f366770fd60, item=item@entry=0x56275c5f9a27 "io.stat", stats=0x7f36676d1698, success=success@entry=0x7ffdbd023f1f)
    at ../src/lxc/tools/lxc_top.c:307
#2  0x000056275c5f706f in stats_get (total=0x7ffdbd023f40, ct=0x7f36676d1c30, c=0x7f366770fd60) at ../src/lxc/tools/lxc_top.c:361
#3  lxc_top_main (argc=<optimized out>, argv=<optimized out>) at ../src/lxc/tools/lxc_top.c:663
```

Segfault happens due to `cols[1]` is `NULL` apparently. 
```c
               if (strncmp(cols[1], "Read", strlen(cols[1])) == 0)
                        stats->read += strtoull(cols[2], NULL, 0);
```

DevonSchwartz:
> I have applied this to lxc-6.0.0 in Alpine Linux but it does not seem to solve the issue.
> 
> ```
> # lxc-top
> Unable to read cgroup item memory.kmem.usage_in_bytes
> Unable to read cgroup item memory.kmem.limit_in_bytes
> Unable to read cgroup item blkio.throttle.io_serviced
> Unable to read cgroup item memory.kmem.usage_in_bytes
> Unable to read cgroup item memory.kmem.limit_in_bytes
> Unable to read cgroup item blkio.throttle.io_serviced
> Unable to read cgroup item memory.kmem.usage_in_bytes
> Unable to read cgroup item memory.kmem.limit_in_bytes
> Unable to read cgroup item blkio.throttle.io_serviced
> Unable to read cgroup item memory.kmem.usage_in_bytes
> Unable to read cgroup item memory.kmem.limit_in_bytes
> Unable to read cgroup item blkio.throttle.io_serviced
> Unable to read cgroup item memory.kmem.usage_in_bytes
> Unable to read cgroup item memory.kmem.limit_in_bytes
> Unable to read cgroup item blkio.throttle.io_serviced
> Unable to read cgroup item memory.kmem.usage_in_bytes
> Unable to read cgroup item memory.kmem.limit_in_bytes
> Unable to read cgroup item blkio.throttle.io_serviced
> Unable to read cgroup item memory.kmem.usage_in_bytes
> Unable to read cgroup item memory.kmem.limit_in_bytes
> Unable to read cgroup item blkio.throttle.io_serviced
> Unable to read cgroup item memory.usage_in_bytes
> Segmentation fault (core dumped)
> ```
> 
> I have a backtrace:
> 
> ```
> [New LWP 24554]
> Core was generated by `lxc-top'.
> Program terminated with signal SIGSEGV, Segmentation fault.
> #0  0x00007f3667931573 in strlen (s=<optimized out>, s@entry=0x0) at src/string/strlen.c:17
> 
> warning: 17     src/string/strlen.c: No such file or directory
> (gdb) 
> #0  0x00007f3667931573 in strlen (s=<optimized out>, s@entry=0x0) at src/string/strlen.c:17
> #1  0x000056275c5f4987 in stat_get_blk_stats (c=c@entry=0x7f366770fd60, item=item@entry=0x56275c5f9a27 "io.stat", stats=0x7f36676d1698, success=success@entry=0x7ffdbd023f1f)
>     at ../src/lxc/tools/lxc_top.c:307
> #2  0x000056275c5f706f in stats_get (total=0x7ffdbd023f40, ct=0x7f36676d1c30, c=0x7f366770fd60) at ../src/lxc/tools/lxc_top.c:361
> #3  lxc_top_main (argc=<optimized out>, argv=<optimized out>) at ../src/lxc/tools/lxc_top.c:663
> ```
> 
> Segfault happens due to `cols[1]` is `NULL` apparently.
> 
> ```c
>                if (strncmp(cols[1], "Read", strlen(cols[1])) == 0)
>                         stats->read += strtoull(cols[2], NULL, 0);
> ```

I will try on my end

DevonSchwartz:
I do not get this on my end. What steps did you take on your end to reproduce this?

ncopa:
I was not able to reproduce the segfault on a clean/fresh system. It happened on a relatively new server with containers imported from old system.

That said, `lxc-top` does not show any running containers either. Here is how to reproduce in qemu on x86_64:

Boot alpine in qemu:
```
qemu-system-x86_64 --enable-kvm -m 2048 -cdrom https://dl-cdn.alpinelinux.org/alpine/v3.20/releases/x86_64/alpine-virt-3.20.2-x86_64.iso -nographic
```

Log in as root and run the following commands:
```
setup-alpine -q
apk add lxc lxc-dbg musl-dbg lxc-templates-legacy-alpine lxc-bridge iptables strace
rc-service cgroups start
rc-service dnsmasq.lxcbr0 start
lxc-create -t alpine -n a1
lxc-start -n a1
lxc-top -b
```

The output is:
```
alpine:~# lxc-top -b
time_ms,container,cpu_nanos,cpu_sys_userhz,cpu_user_userhz,blkio_bytes,blkio_iops,mem_used_bytes,mes
Unable to read cgroup item io.stat
Unable to read cgroup item blkio.throttle.io_service_bytes
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
```

To confirm that `a1` is running:
```
alpine:~# lxc-info -n a1
Name:           a1
State:          RUNNING
PID:            2596
IP:             10.0.3.86
Link:           vethGVKXS5
 TX bytes:      1.82 KiB
 RX bytes:      2.00 KiB
 Total bytes:   3.82 KiB
```

I would have expected that `lxc-top` would show that `a1` is running.

DevonSchwartz:
How do you run it on qemu using the change from this PR?

ncopa:
The patch was backported to the alpine 3.20 package, so it is included in `lxc-6.0.0-r3` and later: https://github.com/alpinelinux/aports/commit/1b2d32887a8dbb6a119f5c5d9078d60346bdffa7

DevonSchwartz:
Oh okay. I was testing it on alpine 3.17, so it could definitely be that my solution is not fit for later versions

DevonSchwartz:
I get the same thing you do on 3.20. For reference, in my forked repo I get this:

`time_ms,container,cpu_nanos,cpu_sys_userhz,cpu_user_userhz,blkio_bytes,blkio_iops,mem_used_bytes,memsw_used_bytes,kernel_mem_used_bytes
Unable to read cgroup item memory.kmem.usage_in_bytes
Unable to read cgroup item memory.kmem.limit_in_bytes
Unable to read cgroup item blkio.throttle.io_serviced
1721745840544,u2,0,5145694023161373743,8101259050997526833,0,0,5337088,0,0`

DevonSchwartz:
My guess is that the container stats are not being updated. If the paths were bad, then it would print "Unable to read cgroup item...". The container stats are set starting on line 371 (`if (total)... `). Could I use gdb to backtrace an lxc-top command?

DevonSchwartz:
I tried using `sudo ./lxc-create -t download a1` with alpine version 3.20, and I got the container statistics. However, when I use qemu (@ncopa 's setup), the statistics do not show up. Is it possible for there to be a discrepancy between an lxc template and running a released version of lxc on alpine?

src/lxc/tools/lxc_top.c
```diff
@@ -276,27 +276,33 @@ static uint64_t stat_match_get_int(struct lxc_container *c, const char *item,
 	Total 149327872
 */
 static void stat_get_blk_stats(struct lxc_container *c, const char *item,
-			      struct blkio_stats *stats) {
+			      struct blkio_stats *stats, bool *success) {
```
ncopa, line 279: Would it be an idea to do let the function return a `bool` for success?
```c
static bool stat_get_blk_stats(struct lxc_container *c, const char *item,
			       struct blkio_stats *stats) {
```

src/lxc/tools/lxc_top.c
```diff
@@ -276,27 +276,33 @@ static uint64_t stat_match_get_int(struct lxc_container *c, const char *item,
...
 	char buf[4096];
 	int i, len;
 	char **lines, **cols;
+	*success = true; 
 
 	len = c->get_cgroup_item(c, item, buf, sizeof(buf));
 	if (len <= 0 || (size_t)len >= sizeof(buf)) {
 		fprintf(stderr, "Unable to read cgroup item %s\n", item);
+		*success = false; 
 		return;
```
ncopa, line 289:
```suggestion
		return false;
```

src/lxc/tools/lxc_top.c
```diff
@@ -314,21 +320,53 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
+		ret_cgroup2 = stat_get_int(c, path_cgroup2); 
+		
+		if (ret_cgroup2 < 0) {
+			*stat = stat_get_int(c, path_cgroup1);
+		} else {
+			*stat = ret_cgroup2; 
+		}
+	}
+}
+
```
ncopa, line 344: We can easily [avoid using boolean argument](https://google.com/search?q=avoid+boolean+arguments) here.

```suggestion
static void try_match_cgroup2(struct lxc_container *c, u_int64_t *stat, const char* path_cgroup1, const char* path_cgroup2, 
	const char* match) {
	
	int ret_cgroup2; 
	ret_cgroup2 = stat_match_get_int(c, path_cgroup2, match, 1); 
	if (ret_cgroup2 < 0) {
		*stat = stat_match_get_int(c, path_cgroup1, match, 1);
	}
}		

static void try_cgroup2(struct lxc_container *c, u_int64_t *stat, const char* path_cgroup1, const char* path_cgroup2) {
	ret_cgroup2 = stat_get_int(c, path_cgroup2); 
		
	if (ret_cgroup2 < 0) {
		*stat = stat_get_int(c, path_cgroup1);
	} else {
		*stat = ret_cgroup2; 
	}
}

```

src/lxc/tools/lxc_top.c
```diff
@@ -314,21 +320,53 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
+	// paths only exist in cgroup1
 	ct->stats->kmem_used     = stat_get_int(c, "memory.kmem.usage_in_bytes");
 	ct->stats->kmem_limit    = stat_get_int(c, "memory.kmem.limit_in_bytes");
-	ct->stats->cpu_use_nanos = stat_get_int(c, "cpuacct.usage");
-	ct->stats->cpu_use_user  = stat_match_get_int(c, "cpuacct.stat", "user", 1);
-	ct->stats->cpu_use_sys   = stat_match_get_int(c, "cpuacct.stat", "system", 1);
-
-	stat_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes);
-	stat_get_blk_stats(c, "blkio.throttle.io_serviced", &ct->stats->io_serviced);
+	stat_get_blk_stats(c, "blkio.throttle.io_serviced", &ct->stats->io_serviced, &success);
```
ncopa, line 369:
```suggestion
	success = stat_get_blk_stats(c, "blkio.throttle.io_serviced", &ct->stats->io_serviced);
```

src/lxc/tools/lxc_top.c
```diff
@@ -314,21 +320,53 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
+	try_cgroup2(c, &(ct->stats->cpu_use_nanos), "cpuacct.usage", "cpu.stat", NULL, false);
+	try_cgroup2(c, &(ct->stats->cpu_use_user), "cpuacct.stat", "cpu.stat", "user", true); 
+	try_cgroup2(c, &(ct->stats->cpu_use_sys), "cpuacct.stat", "cpu.stat", "system", true); 
+
+	// singular cgroup2 case for get blk stats
+	bool success; 
+	stat_get_blk_stats(c, "io.stat", &ct->stats->io_service_bytes, &success);
+	if (!success) {
+		stat_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes, &success);
+	}
```
ncopa, line 364:
```suggestion
	success = stat_get_blk_stats(c, "io.stat", &ct->stats->io_service_bytes);
	if (!success) {
		success = stat_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes);
	}
```

Could even be done as:
```c
	success = stat_get_blk_stats(c, "io.stat", &ct->stats->io_service_bytes) 
		|| stat_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes);
```

src/lxc/tools/lxc_top.c
```diff
@@ -314,21 +320,53 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
 }
 
+static void try_cgroup2(struct lxc_container *c, u_int64_t *stat, const char* path_cgroup1, const char* path_cgroup2, 
+	const char* match, bool call_match) {
+	
+	int ret_cgroup2; 
+	
+	if (call_match) {
+		ret_cgroup2 = stat_match_get_int(c, path_cgroup2, match, 1); 
+		if (ret_cgroup2 < 0) {
```
ncopa, line 330: This wont work. `stat_match_get_int` returns a `uint64_t`, which is _unsigned_  int, meaning it will never be less than zero.

src/lxc/tools/lxc_top.c
```diff
@@ -314,21 +320,53 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
+			*stat = stat_match_get_int(c, path_cgroup1, match, 1);
+		}
+		
+	} else {
+		ret_cgroup2 = stat_get_int(c, path_cgroup2); 
+		
+		if (ret_cgroup2 < 0) {
```
ncopa, line 337: `uint64_t` will never be less than zero.

src/lxc/tools/lxc_top.c
```diff
@@ -314,21 +320,53 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
-	ct->stats->memsw_limit   = stat_get_int(c, "memory.memsw.limit_in_bytes");
+
+	// handle stat_get_int cases
+	try_cgroup2(c, &(ct->stats->mem_used), "memory.usage_in_bytes", "memory.current", NULL, false);
+	try_cgroup2(c, &(ct->stats->mem_limit), "memory.limit_in_bytes", "memory.max", NULL, false);
+	try_cgroup2(c, &(ct->stats->memsw_used), "memory.memsw.usage_in_bytes", "memory.swap.current", NULL, false);
+	try_cgroup2(c, &(ct->stats->memsw_limit), "memory.memsw.limit_in_bytes", "memory.swap.max", NULL, false);
+	try_cgroup2(c, &(ct->stats->cpu_use_nanos), "cpuacct.usage", "cpu.stat", NULL, false);
+	try_cgroup2(c, &(ct->stats->cpu_use_user), "cpuacct.stat", "cpu.stat", "user", true); 
+	try_cgroup2(c, &(ct->stats->cpu_use_sys), "cpuacct.stat", "cpu.stat", "system", true);
```
ncopa, line 357: for cgroups2 the match strings to look for in `cpu.stat` are:
- `usage_usec`
- `user_usec`
- `system_usec`

DevonSchwartz: Thanks. I’ll put those and then send the result.

@ncopa are there any other match strings that I have incorrectly?

ncopa: It's more complicated than that. the `io.stat` may look like this:
```
259:0 rbytes=0 wbytes=5094400000 rios=0 wios=4276 dbytes=0 dios=0
253:0 rbytes=0 wbytes=10491523072 rios=0 wios=1656 dbytes=0 dios=0
```
So you cannot simply look for the key as coloumn 1.

ncopa:
Please also note that the units for CPU stats are different. In cgroups v1 it is nanoseconds and USER_HZ. In v2 it is usec.

And there is currently a bug. The `USER_HZ` is hardcoded to 100, but it depends on the running kernel, so it needs to be detected at run-time.

What do you think about something like this:
https://github.com/ncopa/lxc/commits/lxc-top-2/

DevonSchwartz:
No worries @ncopa. As a student, I am always appreciate of whatever help I can get. I am going to try your changes on my end and merge them. I'll also try to see if I can find a way to detect USER_HZ at runtime.

DevonSchwartz:
@ncopa, is it okay if I merge your branch into mine and fix the cpu stat issue?

DevonSchwartz:
I found this link in documentation: https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst

ncopa:
> @ncopa, is it okay if I merge your branch into mine and fix the cpu stat issue?

Sure!

ncopa:
Good catch on the kernel mem fix! I missed that in the docs. The [lxc-info PR](https://github.com/lxc/lxc/pull/4426) calculates it differently and can probably be simplified.

You have the kernel docs into nicely rendered html as well: https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html

One comment on your commit message. It is usually a good idea to write the commit message in [imperative mood](https://google.com/search?q=git+commit+message+imperative).

src/lxc/tools/lxc_top.c
```diff
@@ -308,40 +314,130 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
+	}
+	if (cg1_cpu_stats(c, &ct->stats->cpu) < 0) {
+		if (cg2_cpu_stats(c, &ct->stats->cpu) < 0) {
+			fprintf(stderr, "Unable to read CPU stats\n");
+		}
+	}
+
+	if (cg1_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes) < 0) {
+		if (cg2_get_blk_stats(c, "io.stat", &ct->stats->io_service_bytes) < 0) {
+			fprintf(stderr, "Unable to read CPU stats\n");
```
ncopa, line 420: I did a copy/pasta error:
```suggestion
			fprintf(stderr, "Unable to read IO stats\n");
```

ncopa:
this also fixes https://github.com/lxc/lxc/issues/4337

ncopa:
I found another bug in `lxc-top` with `-b`. The `ct_print_cnt` should probably be ignored with `batch` mode.
```diff
diff --git a/src/lxc/tools/lxc_top.c b/src/lxc/tools/lxc_top.c
index 6cb38bb90..7498331ee 100644
--- a/src/lxc/tools/lxc_top.c
+++ b/src/lxc/tools/lxc_top.c
@@ -732,7 +732,7 @@ int lxc_top_main(int argc, char *argv[])
                        stats_print_header(&total);
                }
 
-               for (i = 0; i < active_cnt && i < ct_print_cnt; i++) {
+               for (i = 0; i < active_cnt && (i < ct_print_cnt || batch); i++) {
                        stats_print(container_stats[i].c->name, container_stats[i].stats, &total);
                        printf("\n");
                }
```

DevonSchwartz:
@ncopa have we figured out why container stats wont print on a alpine distro?

ncopa:
Yes, I believe it is due to: https://github.com/lxc/lxc/pull/4439#issuecomment-2250154083

ncopa:
Also, when running in qemu, diskless, there is no disk io, which gave "Unable to get CPU stats", where "CPU" was a copy/paste typo of "IO".

You can do: `git rebase -i main`, then set `e` or `edit`in front of the `tools: lxc-top: add cgroups2 IO stats`. Then you edit the file and fix the typo, and do `git add src/lxc/tools/lxc_top.c && git commit --amend -v`. Finally `git rebase --continue`

ncopa:
> @ncopa have we figured out why container stats wont print on a alpine 3.20 distro?

correction. No, I still don't know why it does not print on alpine 3.20 distro (without `-b`) . I suspect it is related the `ct_print_cnt`.

ncopa:
> > @ncopa have we figured out why container stats wont print on a alpine 3.20 distro?
> 
> correction. No, I still don't know why it does not print on alpine 3.20 distro (without `-b`) . I suspect it is related the `ct_print_cnt`.

I believe that the serial console does not give the window size of the terminal. 

But it also makes me think that we should probably get the window size in each iteration in the loop, because it is possible to resize the window while its running.

That is different issue though.

DevonSchwartz:
I pushed the commits. Should we wait to squash commits until it is backported to alpine?

ncopa:
I think this is ready for merge. Personally I don't think the commits needs to be squashed, they give a good history in small steps. (and I have already backported this to alpine, even to stable branches)

@mihalicyn can you please review?

ncopa (approved):

stgraber:
Looks like not everything is signed-off properly.

src/lxc/tools/lxc_top.c
```diff
@@ -631,7 +732,7 @@ int lxc_top_main(int argc, char *argv[])
 			stats_print_header(&total);
 		}
 
-		for (i = 0; i < active_cnt && i < ct_print_cnt; i++) {
+		for (i = 0; i < active_cnt && i < (ct_print_cnt || batch); i++) {
```
mihalicyn, line 735: Please, can you explain this change?

src/lxc/tools/lxc_top.c
```diff
@@ -35,16 +34,24 @@ struct blkio_stats {
...
+struct cpu_stats {
+	uint64_t use_nanos;
+	uint64_t use_user;
+	uint64_t use_sys;
+};
+
+struct mem_stats {
+	uint64_t used;
+	uint64_t limit;
+	uint64_t swap_used;
```
mihalicyn, line 46: it's not a swap, really. It's mem+swap. In cgroupv1 we have a combined page counter for memory+swap:
https://github.com/torvalds/linux/blob/1722389b0d863056d78287a120a1d6cadb8d4f7b/mm/memcontrol.c#L1536

for cgroup-v2, yes, we have a separate `swap` page counter.

I would go the same way, as it's done in the kernel:
https://github.com/torvalds/linux/blob/1722389b0d863056d78287a120a1d6cadb8d4f7b/include/linux/memcontrol.h#L191

and use union:
```
	union {
                struct {
                        uint64_t swap_used;
	                uint64_t swap_limit;
                }; /* v2 only */
                struct {
                        uint64_t memsw_used;
	                uint64_t memsw_limit;
                }; /* v1 only */
	};
```

it makes it a bit cleaner.

src/lxc/tools/lxc_top.c
```diff
@@ -204,7 +212,6 @@ static uint64_t stat_get_int(struct lxc_container *c, const char *item)
 
 	len = c->get_cgroup_item(c, item, buf, sizeof(buf));
 	if (len <= 0) {
-		fprintf(stderr, "Unable to read cgroup item %s\n", item);
```
mihalicyn, line 207: can we keep these error messages? I think it might be very-very useful in case if something goes wrong.

src/lxc/tools/lxc_top.c
```diff
@@ -308,40 +314,130 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
+		}
+	}
+	if (cg1_cpu_stats(c, &ct->stats->cpu) < 0) {
+		if (cg2_cpu_stats(c, &ct->stats->cpu) < 0) {
+			fprintf(stderr, "Unable to read CPU stats\n");
+		}
+	}
+
+	if (cg1_get_blk_stats(c, "blkio.throttle.io_service_bytes", &ct->stats->io_service_bytes) < 0) {
+		if (cg2_get_blk_stats(c, "io.stat", &ct->stats->io_service_bytes) < 0) {
```
mihalicyn, line 419: ah, now I understand why you have removed these error messages :-)

I think you can detect if cgroup1/cgroup2 is used one time, save it in a variable and then call the right function instead of trying each time.

mihalicyn:
@DevonSchwartz @ncopa Huge thanks for your work on this!

src/lxc/tools/lxc_top.c
DevonSchwartz: You're right. I tried on my end and it only prints the container statistics of one container. Thanks for catching that @mihalicyn 

```
Name:           a2
State:          RUNNING
PID:            4916
Link:           vethq4uXbP
 TX bytes:      9.20 KiB
 RX bytes:      0 bytes
 Total bytes:   9.20 KiB
```

```
time_ms,container,cpu_nanos,cpu_sys_userhz,cpu_user_userhz,blkio_bytes,blkio_iops,mem_used_bytes,memsw_used_bytes,kernel_mem_used_bytes
1722022106544,a1,280167000,17,10,2785280,0,4681728,0,1327104
``` 

In the above example, a2 is running, but only a1 prints

src/lxc/tools/lxc_top.c
DevonSchwartz: Is it possible that a system could be using cgorup1 and cgroup2 if the rc cgroup mode is set to "hybrid"?

src/lxc/tools/lxc_top.c
DevonSchwartz: In cgroup2, should we keep the `memsw_used_bytes` column in `lxc-top -b` and calculate that value by adding `memory.current + memory.swap.current`?

src/lxc/tools/lxc_top.c
```diff
@@ -308,40 +314,130 @@ static void stat_get_blk_stats(struct lxc_container *c, const char *item,
...
+			fprintf(stderr, "Unable to read IO stats\n");
+		}
+	} else {
+		/* only with cgroups v1 */
+		cg1_get_blk_stats(c, "blkio.throttle.io_serviced", &ct->stats->io_serviced);
```
DevonSchwartz, line 424: Is it possible to calculate the fields in `ct->stats->io_serviced` by taking all the fields from `ct->stats->io_serviced_bytes` and dividing by `sizeof char`? This would be for cgroup2 since `io_serviced` is not a path in that cgroup

paulocoghi:
Hello @DevonSchwartz! Is there any pending doubt you would like to ask? If yes, I would like to cordially ask you to tag the correspondent contributor related to your doubt, so we can have your PR merged :slightly_smiling_face:

DevonSchwartz:
There is no pending doubt.

paulocoghi:
@DevonSchwartz thanks for the fast response!

Also, I would like your opinion (also @ncopa's and @mihalicyn's opinion) on using this simple command to check which cgroup version is being used.

As per https://github.com/admiracloud/admira-containers/blob/develop/src/cli/requisites.cr#L5, by using:

```
stat -fc %T /sys/fs/cgroup/
```

You will get `cgroup2fs` when cgroups v2 is being used and `tmpfs` when v1 is being used.

Wouldn't this strategy be simpler than "trial and error" checks?

paulocoghi:
Thanks for the important clarification Stéphane @stgraber

mihalicyn (approved):
LGTM

Huge thanks to @DevonSchwartz and @ncopa!

PR #4456, opened 2024-06-05
AppArmor fixup

mihalicyn:
It turned out, that old (and incorrect) rule:
mount options=(rw,make-slave) -> **,

is NOT equivalent to:
mount options=(rw,make-slave) -> /**,

Let's use:
mount options=(rw,make-slave) -> /{,**},

Fixes https://github.com/lxc/lxc/pull/4452


tomponline:
Would it be viable to add some end-to-end testing to this repo to catch regressions, similar to what the LXD tests caught here:

https://github.com/canonical/lxd/actions/runs/9380038088/job/25826201261

```
lxc-create -P /home/runner/work/lxd/lxd/test/tmp.hKK/lxc -n c1 -B dir -t busybox
+ lxc-start -P /home/runner/work/lxd/lxd/test/tmp.hKK/lxc -n c1
lxc-start: c1: ../src/lxc/lxccontainer.c: wait_on_daemonized_start: 837 Received container state "ABORTING" instead of "RUNNING"
lxc-start: c1: ../src/lxc/tools/lxc_start.c: lxc_start_main: 307 The container failed to start
lxc-start: c1: ../src/lxc/tools/lxc_start.c: lxc_start_main: 310 To get more details, run the container in foreground mode
lxc-start: c1: ../src/lxc/tools/lxc_start.c: lxc_start_main: 312 Additional information can be obtained by setting the --logfile and --logpriority options
```

mihalicyn:
> Would it be viable to add some end-to-end testing to this repo to catch regressions

we have tests in Jenkins. We were too fast yesterday merging my fix because it looked trivial and correct. Moreover, we already have an analogical change in Incus/LXD.

Issue #4427, opened 2024-04-05
Unable to build v6.0.0 with nix


adamcstephens:
# Issue description

Hi, I'm trying to upgrade the nixpkgs package to v6.0.0 but I'm currently stuck due to changes which have broken our build's ability to link. Whether using multi-binary or tools-multicall, the only way I've found to build v6.0 is by reverting https://github.com/lxc/lxc/pull/4401. 

```
[135/139] Linking target src/lxc/cmd/lxc-monitord
FAILED: src/lxc/cmd/lxc-monitord
gcc  -o src/lxc/cmd/lxc-monitord src/lxc/cmd/lxc-monitord.p/lxc_monitord.c.o src/lxc/cmd/lxc-monitord.p/.._.._include_netns_ifaddrs.c.o -flto -Wl,--no-undefined -pie -Wl,--no-as-needed -Wl,--gc-sections -Wl,-z,relro -Wl,-z,now -fstack-protector -fstack-protector-strong -Wl,-fuse-ld=gold -Wl,--warn-common '-Wl,-rpath,$ORIGIN/../../..' -Wl,-rpath-link,/home/adam/git/lxc/build/ -Wl,--start-group src/lxc/liblxc.a liblxc.so.1.8.0 /nix/store/jnlxl2ry9y03vq4skhm26igfd84xwa93-dbus-1.14.10-lib/lib/libdbus-1.so -pthread /nix/store/iglb2fmla149fbprj6fb18swfls033q4-libseccomp-2.5.4-lib/lib/libseccomp.so /nix/store/5mfk83ikgk62gv0xj95xylnkf0f52chv-libselinux-3.3/lib/libselinux.so /nix/store/wp5c5sjlvm6nxavvs1prdnw5n41mgciq-libapparmor-3.1.6/lib/libapparmor.so /nix/store/c7rf6w8xvxc4423m9rsgq5nsqhknshvy-openssl-3.0.12/lib/libssl.so /nix/store/c7rf6w8xvxc4423m9rsgq5nsqhknshvy-openssl-3.0.12/lib/libcrypto.so /nix/store/lmj8lfkd8h448grf1nf92lsm5gm2fnf6-libcap-2.69-lib/lib/libcap.so -lutil -Wl,--end-group
../src/lxc/cmd/lxc_monitord.c:67: error: undefined reference to 'lxc_monitor_fifo_name'
[...]

PR #4428, opened 2024-04-05
meson: fix build on NixOS

mihalicyn:
It was reported recently that on NixOS LXC fails to build with the following errors:
```
[135/139] Linking target src/lxc/cmd/lxc-monitord
FAILED: src/lxc/cmd/lxc-monitord
gcc  -o src/lxc/cmd/lxc-monitord src/lxc/cmd/lxc-monitord.p/lxc_monitord.c.o src/lxc/cmd/lxc-monitord.p/.._.._include_netns_ifaddrs.c.o -flto -Wl,--no-undefined -pie -Wl,--no-as-needed -Wl,--gc-sections -Wl,-z,relro -Wl,-z,now -fstack-protector -fstack-protector-strong -Wl,-fuse-ld=gold -Wl,--warn-common '-Wl,-rpath,$ORIGIN/../../..' -Wl,-rpath-link,/home/adam/git/lxc/build/ -Wl,--start-group src/lxc/liblxc.a liblxc.so.1.8.0 /nix/store/jnlxl2ry9y03vq4skhm26igfd84xwa93-dbus-1.14.10-lib/lib/libdbus-1.so -pthread /nix/store/iglb2fmla149fbprj6fb18swfls033q4-libseccomp-2.5.4-lib/lib/libseccomp.so /nix/store/5mfk83ikgk62gv0xj95xylnkf0f52chv-libselinux-3.3/lib/libselinux.so /nix/store/wp5c5sjlvm6nxavvs1prdnw5n41mgciq-libapparmor-3.1.6/lib/libapparmor.so /nix/store/c7rf6w8xvxc4423m9rsgq5nsqhknshvy-openssl-3.0.12/lib/libssl.so /nix/store/c7rf6w8xvxc4423m9rsgq5nsqhknshvy-openssl-3.0.12/lib/libcrypto.so /nix/store/lmj8lfkd8h448grf1nf92lsm5gm2fnf6-libcap-2.69-lib/lib/libcap.so -lutil -Wl,--end-group
../src/lxc/cmd/lxc_monitord.c:67: error: undefined reference to 'lxc_monitor_fifo_name'
../src/lxc/cmd/lxc_monitord.c:104: error: undefined reference to 'lxc_monitor_fifo_name'
../src/lxc/cmd/lxc_monitord.c:140: error: undefined reference to 'lxc_read_nointr'
../src/lxc/cmd/lxc_monitord.c:193: error: undefined reference to 'lxc_mainloop_add_handler'
[...]

PR #4459, opened 2024-06-26
meson: fix build with -Dtools-multicall=true on NixOS

mihalicyn:
See also:
https://github.com/lxc/lxc/pull/4428

Fixes: #4427


mihalicyn:
cc @adamcstephens

PR #4462, opened 2024-07-09
Exit 0 when there's no error

Jip-Hop:
Running `bash lxc-download.in --help` and `bash lxc-download.in --list` currently results in a non-zero exit code, indicating an error occurred (e.g. wget is not installed to download the image index). This PR changes the exit code for `--help` and `--list` to 0 so that a successful run with these flags can be distinguished from runs where a failure occurred.

### DIFF
--- a/templates/lxc-download.in
+++ b/templates/lxc-download.in
@@ -147,7 +147,7 @@ eval set -- "$options"
 
 while :; do
   case "$1" in
-    -h|--help)     usage && exit 1;;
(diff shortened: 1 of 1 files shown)


hallyn (approved):

hallyn:
Looks good, but please sign off your commit:  https://github.com/lxc/lxc/blob/main/CONTRIBUTING#L52

Jip-Hop:
Signed off!

hallyn:
Thank you.

PR #4466, opened 2024-07-25
apparmor: allow nosymfollow remounts

mihalicyn:
This should fix distros with modern versions of systemd.

It's worth mentioning that `nosymfollow` flag is supported only in fresh enough versions of AppArmor (`>= 3`):
https://gitlab.com/apparmor/apparmor/-/merge_requests/1005


hallyn:
Hi @mihalicyn why was this marked as draft?  Is it not actually needed?

mihalicyn:
> why was this marked as draft? Is it not actually needed?

I was quite slow with this, because this change assume AppArmor version >= 3.0, and can break older users.
But I think it is safe to land this now, because starting from 7.0 we can require fresh version of AppArmor.
Debian Bookworm (old stable as of now), for example has 3.0.x and Debian Trixie (current stable as of now) - 4.1.x

PR #4472, opened 2024-08-07
doc: Fix definitions of get_config_path and set_config_path

stgraber:


src/lxc/lxccontainer.h
```diff
@@ -496,7 +496,7 @@ struct lxc_container {
 	bool (*set_cgroup_item)(struct lxc_container *c, const char *subsys, const char *value);
 
 	/*!
-	 * \brief Determine full path to the containers configuration file.
+	 * \brief Determine full path to the containers configuration directory.
```
ftankovic1, line 499: Don't want to be nitpicky, but could you maybe add also an example (like `e.g. /var/lib/lxc`)?

stgraber: I believe the paragraph below mentioning it uses the LXCPATH build time variable is sufficient. Providing an exact path when it's dependent on how it's built and shipped doesn't help all that much.

ftankovic1 (approved):

PR #4482, opened 2024-09-16
meson.build: add -ffat-lto-objects

hallyn:
Otherwise, if we generate a static library, lintian warns that it has no code sections.  See

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=977596

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -204,7 +204,6 @@ possible_cc_flags = [
     '-Wunused-but-set-variable',
     '-Wno-unused-parameter',
     '-Wfloat-equal',
-    '-Wsuggest-attribute=noreturn',
     '-Werror=return-type',
     '-Werror=incompatible-pointer-types',
     '-Wformat=2',


mihalicyn:
Looks good to me.

Hmm, it's weird but with this change we start to get compiler errors like:
```
error: function might be candidate for attribute ‘noreturn’ [-Werror=suggest-attribute=noreturn]
```

hallyn:
Oddly I don't see these errors when I build in my own clean noble container.

hallyn:
@brauner @mihalicyn ok, so if i just add the __noreturn attribute to each of those src/lxc/tools/*.c:*_main functions, then it complains e.g.

```
../src/lxc/tools/lxc_attach.c:320:53: warning: ‘main’ specifies less restrictive attribute than its target ‘lxc_attach_main’: ‘noreturn’ [-Wmissi
ng-attributes]
  320 | int __attribute__((weak, alias("lxc_attach_main"))) main(int argc, char *argv[]);
```
Now I see that we explicitly buy into this warning in meson.build.  So one way to silence this is to just remove

```
-    '-Wsuggest-attribute=noreturn',
```
from meson.build.  IMO this is not an important warning.  Any objections to dropping it?

Blub:
I find it curious that are we calling `exit` specifically instead of returning?
If we do that, we could use `void` instead of `int` for the main functions.

The 2nd warning should be solved by including `noreturn` in the alias declaration as well.
Curiously it needs to be within the `__attribute__`:

```c
void __attribute__((noreturn, weak, alias("lxc_attach_main"))) main(int argc, char *argv[]);
_Noreturn void lxc_attach_main(int argc, char *argv[])
```

works, while using `_Noreturn` in both cases does not.

Issue #4484, opened 2024-09-25
lxc-attach changes permissions on file passed as standard input


terceiro:
# Required information

 * Distribution: Debian
 * Distribution version: trixie (testing)
 * `lxc-start --version`:  6.0.2

# Issue description

A file passed as standard input to lxc-attach, via shell redirection, gets their permissions changed to 600, and owner changed to root.

# Steps to reproduce

```
$ echo foo > /tmp/data
$ chmod 755 /tmp/data
$ ls -lh /tmp/data
-rwxr-xr-x 1 terceiro terceiro 4 set 25 11:43 /tmp/data
$ sudo lxc-attach -n test cat < /tmp/data
foo
$ ls -lh /tmp/data
-rw------- 1 root terceiro 4 set 25 11:43 /tmp/data
```

mbiebl:
For completeness sake, I'll link the the downstream change which triggered this
https://salsa.debian.org/ci-team/autopkgtest/-/commit/813d7c1ce2158c730c991fc8f09e9681ebb3efed

paride:
Maybe related: https://github.com/lxc/lxc/issues/4338

When I fist dug into this I concluded that the ownership/permission change was made on purpose: https://github.com/lxc/lxc/commit/6aff5157220de749afc559901d990c5088ca8143

hallyn:
Did not reproduce on 4.0.12 on focal.

*did* reproduce on t.0.1 on oracular.

So this is definitely a regression.

PR #4486, opened 2024-09-26
fix_stdio_permissions: only change owner/permissions on ttys

terceiro:
This avoids changing permissions on files passed in as shell redirections, e.g. `lxc-attach [...] < FILE`.

Fixes: #4484

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -1958,18 +1958,20 @@ int fix_stdio_permissions(uid_t uid)
 		if (st.st_rdev == st_null.st_rdev)
 			continue;
 
-		ret = fchown(std_fds[i], uid, st.st_gid);
-		if (ret) {
-			SYSTRACE("Failed to chown standard I/O file descriptor %d to uid %d and gid %d",
-			         std_fds[i], uid, st.st_gid);
-			fret = -1;
-			continue;
-		}
+		if (isatty(std_fds[i])) {
+			ret = fchown(std_fds[i], uid, st.st_gid);
+			if (ret) {
+				SYSTRACE("Failed to chown standard I/O file descriptor %d to uid %d and gid %d",
+				         std_fds[i], uid, st.st_gid);
+				fret = -1;
+				continue;
+			}
 
-		ret = fchmod(std_fds[i], 0600);
-		if (ret) {
-			SYSTRACE("Failed to chmod standard I/O file descriptor %d", std_fds[i]);
-			fret = -1;
+			ret = fchmod(std_fds[i], 0600);
+			if (ret) {
+				SYSTRACE("Failed to chmod standard I/O file descriptor %d", std_fds[i]);
+				fret = -1;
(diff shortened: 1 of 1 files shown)


mihalicyn:
jenkins: ok to test

mihalicyn:
Hi @terceiro,

thanks for your patch!

I'm trying to figure out a story behind the original change https://github.com/lxc/lxc/pull/3344.
It seems to me that your patch is, effectively, disabling a behavior introduced in #3344,
because you add a `isatty` check which means that permissions/owner will be changed only for a terminal devices.

Unfortunately, there is no reproducer or any additional info in #3344 to figure out what was the use case. At the same time,
it seems a bit weird to me that we are changing an ownership/permission information on the file passed as a stdin/stdout/stderr,
because usually, Linux kernel does all the permission checks during `open()` syscall and if container's init process has an FD
then it should be fine to perform `read()`/`write()` on it.

Let's ask Christian's (@brauner) opinion on this story in general.

brauner:
`fix_stdio_permissions()` is only called when the namespace uid/gid that is used to setup the container is different from the namespace uid/gid the container actually runs as. Usually both will be namespace uid/gid 0 mapped to some value on the host. But there are cases where the container workload switches to a non-zero namespace uid/gid. For that case we adjust the permissions on stdio/stdout/stderr to ensure that the non-zero namespace uid/gid can open them.

brauner:
From the reproducer @terceiro pasted it looks like the `lxc-attach` uid/gid isn't different from the namesapce setuid uid/gid. So that would be a bug in `attach.c` somewhere.

hallyn:
> From the reproducer @terceiro pasted it looks like the `lxc-attach` uid/gid isn't different from the namesapce setuid uid/gid. So that would be a bug in `attach.c` somewhere.

A bit of debug is giving me:

after alloc_attach_context ctx->setup_ns_uid = -1 ctx->target_ns_uid = -1
after get_attach_context ctx->setup_ns_uid = -1 ctx->target_ns_uid = 0

Doh!  It just occurred to me I was wrong in my previous comment: my 4.0.12 test was with
a root owned container that is uid mapped.  My 6.0.1 test was without a user namespace.

hallyn:
("previous comment" being the one in the related issue)

hallyn:
So on the one hand, if root in the container is root on the host, then we shouldn't need to chown for the container to be able to read/write the contents.  But, the container might have sensitive info which through some convoluted manipulation of a system using lxc, allow an unprivileged user to arrange to exfiltrate data.

My biggest worry with what we're currently doing would be if it were possible for a container to create a root-owned executable file, but we prevent that by chmoding (unless the chmod can fail, in which case perhaps we should work harder to undo our chown?).  So finally there could be the inconvenience for an unprivileged sudo-capable user who gets their file owned by root this way.

So right now my recommendation would be to not merge this pr, and close the issue as wontfix.

PR #4491, opened 2024-10-23
fix return code of recursive all of cgroup_tree_prune

gjaekel:
IMHO, the value to return here is `ret` as it is the value got from recursion. Because this is not a call to a I/O-function, the value of `errno` is unrelated to this call, u.e. undefined.


stgraber:
jenkins: test this please

Issue #4496, opened 2024-12-06
network config of unprivileged containers with an inherited network namespace is not shown


ElJeffe:
When an unprivileged container is running with an inheritted network namespace, then the ip addresses of the container are not shown with 'lxc-ls -f'.
on main branch, following error is shown:

```
# lxc-ls -f
lxc-ls: ../src/lxc/utils.c: switch_to_ns: 900 Operation not permitted - Failed to set process 6305 to "net" of 3
lxc-ls: ../src/lxc/lxccontainer.c: do_lxcapi_get_interfaces: 2335 Operation not permitted - Failed to enter network namespace
lxc-ls: ../src/lxc/utils.c: switch_to_ns: 900 Operation not permitted - Failed to set process 6305 to "net" of 3
lxc-ls: ../src/lxc/lxccontainer.c: do_lxcapi_get_ips: 2432 Operation not permitted - Failed to attach to network namespace
lxc-ls: ../src/lxc/utils.c: switch_to_ns: 900 Operation not permitted - Failed to set process 6305 to "net" of 3
lxc-ls: ../src/lxc/lxccontainer.c: do_lxcapi_get_ips: 2432 Operation not permitted - Failed to attach to network namespace
NAME                                 STATE   AUTOSTART GROUPS IPV4 IPV6 UNPRIVILEGED 
ddb3dfcf-1c9a-5332-8f59-0c0b3c2a72fa RUNNING 0         -      -    -    true   
```
This happens because in enter_net_ns, lxc first enters the user namespace before entering the net namespace. But since the user namespace is inherited, the process does not have permission to enter the net namespace after entering the user namespace.Reversing the order solves the issue.

mihalicyn:
Fixed in https://github.com/lxc/lxc/pull/4497

PR #4497, opened 2024-12-06
network config of unprivileged containers is not shown

ElJeffe:
When an unprivileged container is running with an inheritted network namespace, then the ip addresses of the container are not shown with 'lxc-ls -f'.

This MR solves  #4496

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -2219,17 +2219,23 @@ WRAP_API_1(bool, lxcapi_clear_config_item, const char *)
 
 static inline bool enter_net_ns(struct lxc_container *c)
 {
+	bool net_ns_entered;
 	pid_t pid = do_lxcapi_init_pid(c);
 
 	if (pid < 0)
 		return false;
+
+	net_ns_entered = switch_to_ns(pid, "net");
 
 	if ((geteuid() != 0 || (c->lxc_conf && !list_empty(&c->lxc_conf->id_map))) &&
 	    (access("/proc/self/ns/user", F_OK) == 0))
 		if (!switch_to_ns(pid, "user"))
 			return false;
 
-	return switch_to_ns(pid, "net");
+	if (!net_ns_entered)
+		return switch_to_ns(pid, "net");
+
+	return true;
 }
 
 /* Used by qsort and bsearch functions for comparing names. */


stgraber:
jenkins: test this please

src/lxc/lxccontainer.c
```diff
@@ -2224,12 +2224,15 @@ static inline bool enter_net_ns(struct lxc_container *c)
 	if (pid < 0)
 		return false;
 
+	if (!switch_to_ns(pid, "net"))
```
mihalicyn, line 2227: I think we should be a bit more careful here, as it's not a random choice that we do switch to network namespace after user namespace here. For example, you can do `unshare -U -n` as an unprivileged user, but you can not do `unshare -n`. Strictly speaking, you can change user namespace if your current user has `CAP_SYS_ADMIN` in the target user namespace, while for network namespace you need to have `CAP_SYS_ADMIN` in current user namespace **and** also user namespace that owns target network namespace.

I'm pretty sure that with this change we should get a degradation for a case when you run unprivileged container from an unprivileged user and try to list IP address/network interfaces. (Please, correct me if I'm wrong.)

mihalicyn (changes requested):

mihalicyn:
Hi @ElJeffe,

thanks for your PR!

I've left a small comment about it. Please, can you also provide me with your container configuration example? I just want to make sure that I understand your use-case right. And also, it's important to know how you run your unprivileged container, specifically, if you do `lxc-start` as a root user on the host machine or as an unprivileged user.

ElJeffe:
Hi @mihalicyn 

I start the unprivileged containers as root. To reproduce one can create two container:
```
lxc-create --name parent --template download -- --dist alpine --release 3.19 --arch amd64
lxc-create --name inherrited --template download -- --dist alpine --release 3.19 --arch amd64
```
and modify the config of the second one to
```
lxc.tty.dir = lxc
lxc.pty.max = 1024
lxc.tty.max = 4

lxc.cap.drop = mac_admin mac_override sys_time sys_module sys_rawio
lxc.mount.auto = proc:mixed
lxc.seccomp.profile = /usr/share/lxc/config/common.seccomp

lxc.rootfs.path = dir:/lxc/rootfs
lxc.uts.name = inherrited

# unprivileged
lxc.idmap = g 0 100000 65536
lxc.idmap = u 0 100000 65536

# inherrit namespace
lxc.namespace.share.net=9074
```
where lxc.namespace.share.net is the PID of `lxc-info -p parent`

the result is:
```
root@netnstest:/var/lib/lxc# lxc-start inherrited
root@netnstest:/var/lib/lxc# lxc-ls -f
NAME       STATE   AUTOSTART GROUPS IPV4          IPV6 UNPRIVILEGED 
inherrited RUNNING 0         -      -             -    true         
parent     RUNNING 0         -      192.168.5.116 -    false        
root@netnstest:/var/lib/lxc# lxc-attach inherrited
/ # ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
2: eth0@if6: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP qlen 1000
    link/ether 00:16:3e:0c:e4:94 brd ff:ff:ff:ff:ff:ff
    inet 192.168.5.116/24 brd 192.168.5.255 scope global eth0
       valid_lft forever preferred_lft forever
```

src/lxc/lxccontainer.c
ElJeffe: Yes I see that there may be a degradation when starting as an unprivileged user. I see that nsenter enters the namespaces in two passes for this reason (https://github.com/util-linux/util-linux/commit/854d0fef7f2bbae6e13a2b2ec8116b7e29541d45), I can modify the code to do something similar.

mihalicyn:
jenkins: ok to test

mihalicyn:
Hi @ElJeffe!

Please, can you squash your commits into one?

Also, please, adjust your commit message head line to be like `lxccontainer: fix enter_net_ns helper to work when netns is inherited`.

mihalicyn (approved):

PR #4502, opened 2024-12-16
let lxc_container_init forward SIGHUP to child processes

neoaggelos:
This commit reverts commit "186dfb166aeddb92ef2e52a3bcad17ae05fa2030".

This makes the handling of SIGHUP similar to that of other signals like SIGUSR1 and SIGUSR2. This is especially useful for OCI containers, as it is currently tricky to send a SIGHUP to an OCI container, as the injected init process will consume and ignore it.

Also preserved the note that a SIGTERM is sent automatically by the monitor's handler in case of a proper SIGHUP.

Original commit is at https://github.com/lxc/lxc/commit/186dfb166aeddb92ef2e52a3bcad17ae05fa2030

### DIFF
--- a/src/lxc/initutils.c
+++ b/src/lxc/initutils.c
@@ -632,17 +632,6 @@ __noreturn int lxc_container_init(int argc, char *const *argv, bool quiet)
 
 		switch (was_interrupted) {
 		case 0:
-		/* Some applications send SIGHUP in order to get init to reload
-		 * its configuration. We don't want to forward this onto the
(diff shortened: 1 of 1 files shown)


stgraber:
After some discussions with the other maintainers, I sent what should be a safer variation of this https://github.com/lxc/lxc/pull/4503

Basically adding something that should take care of the reason for SIGHUP being ignored in the first place.

neoaggelos:
great, thanks for #4503 !

Issue #4507, opened 2024-12-20
lxc-attach can not exit when active detached sub-processes exists


asainkujovic:
* Distribution: any, as far as I know this behaviour did not changed for years and years
 
# issue description:

If you run lxc-attach and inside it if you run some detached process like for example "tail -f somelog.txt &", and than continue to work inside lxc-attach shell, and then you try to exit ... you are blocked, you can not use that terminal anymore. Shell has subprocess and on exit shell enters some zombie 'defunct' state, child subprocess goes to parent 'init (1)', but parent of parent, our lxc-attach does nothing about it, just wait and wait.

# Steps to reproduce

1. run `lxc-attach name`
2. run `sleep 25 &` inside it
3. run `exit` and you will be blocked for25 seconds

# Information to attach

After inspecting this problem when I tried to make minor code changes to test it, i found out that good thing is that lxc-attach in 'exit' moment recive 'SIGCHLD' well. So if we add somewhere at 1734 line of file attach.c line like: `signal(SIGCHLD, newhandler)`,
problem would be solved if we just also define this `void newhandler(int)` as something like: `{ if waitpid((-1,NULL,WNOHANG)>0) exit(0); }`. In this code, waitpid(-1) used only to get childpid number, and if it is 0 - avoid killing lxc-attach process in the beginning (0 initial fork) ... just ignore pid0, and exit for other (other is the only child process that lxc-attach has, shell itself. When shell sends signal 'child is killed', then lxc-attach may also exit.

Not sure that this is ideal solution but it shows me that problem could be solved with ease for sure. So hope that someone from creators of lxc can check this, and i am also wondering why this problem was not listed before, because it is very annoying in manual usage but also block us often to use lxc-attach integrated inside something else.

hallyn:
Hm.  But is that not the behavior we want?

@stgraber @brauner I'm curious of your opinion here.  I think this behavior is actually preferable to not waiting for all processes to exit.

asainkujovic:
@hallyn, yes, it should be behaviour that any user want and expect, to exit when typing 'exit' without hanging in blocked program forever (or for 25 seconds in above example). Current version without the patch from [connected PR](https://github.com/lxc/lxc/pull/4509) and all versions for last 5 years that I checked will be blocked for sure.

asainkujovic:
closing issue because PR [#4509](https://github.com/lxc/lxc/pull/4509) that should solve this problem is now merged well

hallyn:
[...]

PR #4509, opened 2024-12-24
- LXC attach should exit on SIGCHLD

asainkujovic:
problem described in the issue 169, https://github.com/lxc/lxc/issues/4507

lxc-attach hangs if there are active sbprocesses.

this changes should fix it.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1430,6 +1430,15 @@ static inline void lxc_attach_terminal_close_log(struct lxc_terminal *terminal)
 	close_prot_errno_disarm(terminal->log_fd);
 }
 
+void lxc_attach_sig_handler(int signum) {
+	// exit on SIGCHLD, but not if sender-pid is 0 (main process)
+	if (signum == SIGCHLD) {
+		int sender_pid = waitpid(-1, NULL, WNOHANG);
+		if (sender_pid > 0)
+			exit(EXIT_SUCCESS);
+	}
+}
+
 int lxc_attach(struct lxc_container *container, lxc_attach_exec_t exec_function,
 	       void *exec_payload, lxc_attach_options_t *options,
 	       pid_t *attached_process)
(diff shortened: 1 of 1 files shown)


hallyn:
How easily could you implement and integrate an automated testcase for this?  It would be great if that were part of this PR.

asainkujovic:
@hallyn thanks for the comment, I will try to understand other src/test examples and to make some kind of a testcase. ... I should also maybe try to test some other 'less agressive' way of exit than direct `exit(0)` because on some terminals it quits program well but TTY stay in some 'cursor not visible' state and 'reset' command is needed. So functionaly is OK, just that minor problem sometimes. Maybe to resend some 'SIGKILL' instead of exit or something like that.

hallyn:
Note that the patch as is won't compile - the signum argument isn't named in the function declaration.

Other than that, looks ok to me.

hallyn:
Hi,

thanks for your patience, just a few more things:

1. please squash all of these into one commit, as they are not separate ideas
2. please add signed-off-by
3. please rename sig_num back to signum.  yes we want snake_case rather than camelcase, but we also want to avoid extraneous snakes :)
4. since the handler function is nice and short, there is no doubt what the sender pid is, so please call it just sender_pid, not sig_sender_pid.

Thanks!

hallyn (approved):
Thanks!

hallyn:
Hm, so if you wget https://github.com/lxc/lxc/pull/4509.patch , you'll see the 'From' and signed-off-by don't match, and that's upsetting the github helper.  I'm ok just ignoring that and merging it, but if you want this to be from your gmail address, you could do a git commit --amend --author="Asain Kujovic <asainnp@gmail.com>"

hallyn:
@asainkujovic  please let me know your preference.

asainkujovic:
@hallyn sorry I did not realise that, my preference is what ever, and I update now again with git --author like you explained ... as that is mail registered with github account.

hallyn (approved):

PR #4512, opened 2025-01-09
conf: useful logging for capabilities

sdanailo-42:


src/lxc/conf.c
```diff
@@ -2944,6 +2944,9 @@ int parse_cap(const char *cap_name, __u32 *cap)
 	if (strequal(cap_name, "none"))
 		return -2;
 
+#if HAVE_LIBCAP == 0
+	WARN("capabilities are disabled or libcap not found");
+#endif
```
stgraber, line 2949: Tiny nit but I'd put a line break after this line :)

sdanailo-42: Done, force pushed

PR #4518, opened 2025-01-27
LXC attach exit on SIGCHLD corrections and ignore-param

asainkujovic:
TRY2: 

- not using exit() nor _exit()  but raise(SIGTERM) as it works ok, cause clean deleting and restoring PTY
- setting signal handler moved down in code after verified transient process exit (so transient process can not fire sigchld anymore that caused race condition)
- added option flag 'LXC_ATTACH_SIGCHLD_EXIT' that is OFF by default, so API function `->attach()` and code other than lxc-attach tool command will behave as before.
- tool **lxc-attach** puts this flag to ON by default, and for lxc-attach new command line parameter introduced `--sigchld-ignore (-G)`, which can overide this new behaviour by leaving flag LXC_ATTACH_SIGCHLD_EXIT to value OFF.

original issue: https://github.com/lxc/lxc/issues/4507 
than code is [reverted back](https://github.com/lxc/lxc/pull/4517) because of regressions - that I hope are solved here. @mihalicyn should check if this is acceptable

Not sure will this new parameter be used in reality, maybe already existent 'LXC_ATTACH_TERMINAL' could be used and enough (api will not have it by default and only lxc-attach tool set it), so maybe just this few lines could be enough:
https://github.com/lxc/lxc/compare/main...asainkujovic:lxc:sigchld_try2

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1430,6 +1430,19 @@ static inline void lxc_attach_terminal_close_log(struct lxc_terminal *terminal)
 	close_prot_errno_disarm(terminal->log_fd);
 }
 
+void lxc_attach_sig_handler(int signum)
+{
+	if (signum == SIGCHLD) {
+		/* SIG_IGN for SIGTERM will:
+		 *   - prevent shell from printing 'pid terminated' message
+		 *   - makes the program exit with code 0 (not 143)
+		 *   - NOT really disable SIGTERM from terminating the program
+		 */
+		signal(SIGTERM, SIG_IGN);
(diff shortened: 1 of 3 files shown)


asainkujovic:
closing because it is working great, but still not sure why this:
`signal(SIGTERM, SIG_IGN); raise(SIGTERM);` works well. It probably doing nearly nothing, but successfully breaks epoll_wait where lxc_attach() is waiting... and it is fine, still not 100% clear will it work with io_ring and will it possible have some problems because it is not so logical.

PR #4519, opened 2025-01-29
LXC attach exit on SIGCHLD, corrections

asainkujovic:
These corrections resolve all mentioned regressions from [reverting-commit](https://github.com/lxc/lxc/pull/4517)

I have also prepared possible branches [(1)](https://github.com/lxc/lxc/compare/main...asainkujovic:lxc:sigchld_with_flag2) and [(2)](https://github.com/lxc/lxc/compare/main...asainkujovic:lxc:sigchld_with_parexit) for introducing the suggested new cmd-line-parameter, but this version still remains significantly cleaner and easier to understand.

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1430,6 +1430,11 @@ static inline void lxc_attach_terminal_close_log(struct lxc_terminal *terminal)
 	close_prot_errno_disarm(terminal->log_fd);
 }
 
+void SIGCHLD_handler(int signum) {
+	signal(SIGTERM, SIG_IGN); /* not terminating directly but: */
+	raise(SIGTERM);           /* leads to LXC_MAINLOOP_CLOSE and a clean exit */
+}
+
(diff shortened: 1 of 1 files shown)


hallyn:
iiuc it doesn't address @mihalicyn 's 4th point, that this changes the API.  So i think one of your other branches is required.

asainkujovic:
@hallyn hi just to mention now when changes are reviewed again, as far as i remember API should stay the same, and program behavior differs only when LXC_ATTACH_TERMINAL flag explicitly turned ON (by default is OFF). Currently it is only used as ON inside lxc_attach.c  (lxc_attach program itself where we need the change). And other callers of the API will be affected only if they for some reason attaches terminal too, and if they are setting this flag.

asainkujovic:
Closing this PR now since it seems this isn't something that's currently needed after all. If it ever becomes relevant, feel free to reuse any part of this code or check out the other two branches I mentioned.

PR #4536, opened 2025-04-05
start: Re-introduce first SET_DUMPABLE call

stgraber:
Without it, we're running into issues with complex hooks like nvidia.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -1125,6 +1125,11 @@ static int do_start(void *data)
 		if (!lxc_switch_uid_gid(nsuid, nsgid))
 			goto out_warn_father;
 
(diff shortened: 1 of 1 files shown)


mihalicyn (approved):
LGTM. Thanks, Stéphane!

I'm stupid and haven't noticed that in between these `lxc_switch_uid_gid` calls we have `run_lxc_hooks(handler->name, "start", handler->conf, NULL)` hooks call. Obviously, without `PR_SET_DUMPABLE` we lose ability to access `/proc/<container_init>/..` things from hooks.

mihalicyn:
I'll add some extra comments in the code a bit later and also try to extend our tests suite to catch sadly regressions like this one.

brauner (approved):

Issue #4525, opened 2025-02-27
loading seccomp after processing LSM labels causes the LSM profiles to require `sys_admin` capability


imerin:
Because `ret = handler->lsm_ops->process_label_set(handler->lsm_ops, NULL, handler->conf, true);` runs before `ret = lxc_seccomp_load(handler->conf);` any profile defined in the configuration file will require the `sys_admin` capability to load the seccomp policy. 

`process_label_set` should be as close to the last thing called during container initialization to avoid polluting LSM profiles with permissions required by lxc-start itself.  

I do not know if there is any technical reason why it must be done the way it is currently done.

imerin:
The relevant change is here: https://github.com/lxc/lxc/commit/b2fe91c7d4e9dd26903dca9801a32a03a65bd2a4

imerin:
After further examination, I think the current functionality is incorrect and should be reverted to the previous implementation.  According to the discussion at the time on the lxc-devel mailing list, this was done prior to the existence of `process_label_set_at` when `process_label_set` was used for both attach and start.

hallyn:
This is odd... But I can't find the lxc-devel discussions you are referring to, either in my mailbox, or online.  Can you send a url to that discussion?  In particular I'd like to see what was breaking at the time, and make sure that has been fixed by now (presumably in apparmor kernel module)

imerin:
@hallyn  https://groups.google.com/a/lists.linuxcontainers.org/g/lxc-devel/c/K-3UrnAGYdM/m/ThmY_UIeDwAJ

hallyn:
> @hallyn  https://groups.google.com/a/lists.linuxcontainers.org/g/lxc-devel/c/K-3UrnAGYdM/m/ThmY_UIeDwAJ

Thanks, so per Stéphane's email, overlay containers were not getting
apparmor onexec transitions.  So that should be easy enough to test
with and without this patch.  Even add a testcase to the testsuite.
I'm 99% sure that must have been fixed by now: there were quite a
few overlay weirdnesses back then.

mihalicyn:
See also:
- https://github.com/lxc/lxc/pull/4539
- https://github.com/lxc/lxc/issues/4553

Issue #3393, opened 2020-04-23
lxc-attach: Permission denied for container with AppArmor profile and no_new_privs flag set


alivenets:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 18.04
 * The output of
   * `lxc-start --version`: 3.1.0
   * `lxc-checkconfig`
```
Kernel configuration not found at /proc/config.gz; searching...
Kernel configuration found at /boot/config-4.15.0-30-generic
--- Namespaces ---
Namespaces: enabled
Utsname namespace: enabled
Ipc namespace: enabled
Pid namespace: enabled
User namespace: enabled
Network namespace: enabled

--- Control groups ---
Cgroups: enabled

Cgroup v1 mount points: 
/sys/fs/cgroup/systemd
/sys/fs/cgroup/net_cls,net_prio
/sys/fs/cgroup/blkio
/sys/fs/cgroup/devices
/sys/fs/cgroup/perf_event
/sys/fs/cgroup/cpuset
/sys/fs/cgroup/memory
/sys/fs/cgroup/rdma
/sys/fs/cgroup/cpu,cpuacct
/sys/fs/cgroup/pids
/sys/fs/cgroup/freezer
/sys/fs/cgroup/hugetlb

Cgroup v2 mount points: 
/sys/fs/cgroup/unified

Cgroup v1 clone_children flag: enabled
Cgroup device: enabled
Cgroup sched: enabled
Cgroup cpu account: enabled
Cgroup memory controller: enabled
Cgroup cpuset: enabled

--- Misc ---
Veth pair device: enabled, not loaded
Macvlan: enabled, not loaded
Vlan: enabled, not loaded
Bridges: enabled, not loaded
Advanced netfilter: enabled, not loaded
CONFIG_NF_NAT_IPV4: enabled, not loaded
CONFIG_NF_NAT_IPV6: enabled, not loaded
CONFIG_IP_NF_TARGET_MASQUERADE: enabled, not loaded
CONFIG_IP6_NF_TARGET_MASQUERADE: enabled, not loaded
CONFIG_NETFILTER_XT_TARGET_CHECKSUM: enabled, not loaded
CONFIG_NETFILTER_XT_MATCH_COMMENT: enabled, not loaded
FUSE (for use with lxcfs): enabled, not loaded

--- Checkpoint/Restore ---
checkpoint restore: enabled
CONFIG_FHANDLE: enabled
CONFIG_EVENTFD: enabled
[...]

PR #4539, opened 2025-04-08
delay assumption of apparmor labels

ianmerin:
attach and start were assuming the container's apparmor profile too early, causing the profile to require elevated capabilities, including CAP_SYS_ADMIN to finish the container setup (e.g. load seccomp profiles).  This change re-adds the `onexec` usage to apparmor LSM module and moves the label assumption to later in the attach code.

relevant issue is https://github.com/lxc/lxc/issues/4525

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1261,19 +1261,6 @@ __noreturn static void do_attach(struct attach_payload *ap)
 		if (!lxc_switch_uid_gid(ctx->setup_ns_uid, ctx->setup_ns_gid))
 			goto on_error;
 
-	if (attach_lsm(options) && ctx->lsm_label) {
-		bool on_exec;
-
-		/* Change into our new LSM profile. */
-		on_exec = options->attach_flags & LXC_ATTACH_LSM_EXEC ? true : false;
-		ret = ctx->lsm_ops->process_label_set_at(ctx->lsm_ops, fd_lsm, ctx->lsm_label, on_exec);
-		close_prot_errno_disarm(fd_lsm);
-		if (ret < 0)
-			goto on_error;
-
-		TRACE("Set %s LSM label to \"%s\"", ctx->lsm_ops->name, ctx->lsm_label);
-	}
-
 	if (conf->no_new_privs || (options->attach_flags & LXC_ATTACH_NO_NEW_PRIVS)) {
 		ret = prctl(PR_SET_NO_NEW_PRIVS, prctl_arg(1), prctl_arg(0),
 			    prctl_arg(0), prctl_arg(0));
@@ -1366,6 +1353,19 @@ __noreturn static void do_attach(struct attach_payload *ap)
 
 	if (!lxc_switch_uid_gid(ctx->target_ns_uid, ctx->target_ns_gid))
 		goto on_error;
+
+	if (attach_lsm(options) && ctx->lsm_label) {
+		bool on_exec;
+
+		/* Change into our new LSM profile. */
+		on_exec = options->attach_flags & LXC_ATTACH_LSM_EXEC ? true : false;
+		ret = ctx->lsm_ops->process_label_set_at(ctx->lsm_ops, fd_lsm, ctx->lsm_label, on_exec);
+		close_prot_errno_disarm(fd_lsm);
+		if (ret < 0)
+			goto on_error;
+
+		TRACE("Set %s LSM label to \"%s\"", ctx->lsm_ops->name, ctx->lsm_label);
+	}
 
 	put_attach_payload(ap);
 
--- a/src/lxc/lsm/apparmor.c
+++ b/src/lxc/lsm/apparmor.c
@@ -5,6 +5,7 @@
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/apparmor.h>
 #include <sys/mount.h>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -1212,15 +1213,17 @@ static int apparmor_process_label_set(struct lsm_ops *ops, const char *inlabel,
 	if (strequal(label, "unconfined") && apparmor_am_unconfined(ops))
 		return log_info(0, "AppArmor profile unchanged");
 
-	label_fd = apparmor_process_label_fd_get(ops, lxc_raw_gettid(), on_exec);
-	if (label_fd < 0)
-		return log_error_errno(-EINVAL, EINVAL, "Failed to change AppArmor profile to %s", label);
-
-	ret = apparmor_process_label_set_at(ops, label_fd, label, on_exec);
-	if (ret < 0)
-		return log_error_errno(-EINVAL, EINVAL, "Failed to change AppArmor profile to %s", label);
-
-	return log_info(0, "Changed AppArmor profile to %s", label);
+	if (on_exec) {
+		ret = aa_change_onexec(label);
+	} else {
+		ret = aa_change_profile(label);
+	}
+	
+ 	if (ret < 0)
+		return log_error_errno(-1, errno, "Failed to set AppArmor%s context to \"%s\"",
+				       on_exec ? " exec" : "", label);
+	
+	return log_info(0, "Changed AppArmor%s profile to \"%s\"", on_exec ? " exec" : "", label);
 }
 
 static struct lsm_ops apparmor_ops = {
--- a/src/tests/aa.c
+++ b/src/tests/aa.c
@@ -15,15 +15,45 @@
 #include <string.h>
 
 #define MYNAME "test-aa"
+#define MYNAME2 "test-aa-o1"
+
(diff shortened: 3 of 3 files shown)


stgraber:
Given the previous placement of the apparmor logic, can you do a test to confirm that no-new-priv being set to TRUE doesn't cause issues?

ianmerin:
you started me down a bit of a rabbit hole, as I found https://github.com/lxc/lxc/issues/3393 which implies that 
no-new-priv should occur after setting the label, which the documentation confirms: https://linuxcontainers.org/lxc/manpages/man5/lxc.container.conf.5.html

However, the diagram in the code clearly has no-new-priv happening prior to setting the lsm label: https://github.com/lxc/lxc/blob/main/src/lxc/attach.c#L1523

FWIW I did run with no-new-privs = 1 and was able to attach successfully, but my environment may not be properly setup for that test.

ianmerin:
Its also important to note that prior to this change apparmor and selinux behaved differently.  SELinux worked with onexec, so the label was not set until init was execve'd, while apparmor did not so it assumed the label immediately.

So the docs are wrong in both instances because when the label is set depends on which LSM you are using!

stgraber:
@brauner @hallyn @mihalicyn can you take a look?

hallyn:
FWIW a manual build and test shows lxc-attach to overlay containers getting the right profile:

```
root@lxctest:~# cat /proc/39974/attr/current
lxc-u2_</var/lib/lxc>//&:lxc-u2_<-var-lib-lxc>:unconfined (enforce)
```

An automated test for this would be nice.  Perhaps in src/tests/aa.c, after test_aa_policy()
and stop()ing, create an overlay snapshot, start it, and make sure that the init pid is
confined.

mihalicyn:
I'll take a closer look into this during the week.

I also support Serge's point that we should have a testcase for this.
@ianmerin would you mind trying to make one?

hallyn:
Here is a test that appears to work, and appears to show that this branch is good.


```
From 51c66ac5ab32c7e31dd3a5b8ead27008f0e0a3be Mon Sep 17 00:00:00 2001
From: Serge Hallyn <serge@hallyn.com>
Date: Fri, 18 Apr 2025 03:51:45 +0000
Subject: [PATCH 1/1] apparmor test: add an overlay container start

make sure that when we start an overlay container, the
init pid is aa-confined.

---
 src/tests/aa.c | 71 ++++++++++++++++++++++++++++++++++++++++++++++++--
 1 file changed, 69 insertions(+), 2 deletions(-)

diff --git a/src/tests/aa.c b/src/tests/aa.c
index 5002beeaf..68d532b4f 100644
--- a/src/tests/aa.c
+++ b/src/tests/aa.c
@@ -15,16 +15,46 @@
 #include <string.h>
 
 #define MYNAME "test-aa"
+#define MYNAME2 "test-aa-o1"
+
+static char *read_file(char *fname) {
+	FILE *f = fopen(fname, "r");
+	ssize_t ret;
+	size_t len = 0;
+	char *v = NULL;
+
+	if (!f) {
+		fprintf(stderr, "Failed to open %s: %m\n", fname);
+		return NULL;
+	}
+	ret = getline(&v, &len, f);
+	if (ret == -1) {
+		fprintf(stderr, "Failed reading a line from %s: %m\n", fname);
+		free(v);
+		if (f)
+			fclose(f);
+		return NULL;
+	}
+	if (f)
+		fclose(f);
+	return v;
+}
 
 static void try_to_remove(void)
 {
-	struct lxc_container *c;
+	struct lxc_container *c, *c2;
 	c = lxc_container_new(MYNAME, NULL);
 	if (c) {
 		if (c->is_defined(c))
 			c->destroy(c);
 		lxc_container_put(c);
 	}
+	c2 = lxc_container_new(MYNAME2, NULL);
+	if (c2) {
+		if (c2->is_defined(c2))
+			c2->destroy(c2);
+		lxc_container_put(c2);
+	}
 }
 
 static int test_attach_write_file(void* payload)
@@ -138,7 +168,9 @@ static bool test_aa_policy(struct lxc_container *c)
 
 int main(int argc, char *argv[])
 {
-	struct lxc_container *c;
+	struct lxc_container *c, *c2= NULL;
+	char *v = NULL;
+
 	try_to_remove();
 	c = lxc_container_new(MYNAME, NULL);
 	if (!c) {
@@ -178,10 +210,45 @@ int main(int argc, char *argv[])
 
 	c->stop(c);
 
+	c2 = c->clone(c, MYNAME2, NULL, LXC_CLONE_SNAPSHOT, "overlayfs", NULL, 0, NULL);
+	if (!c2) {
+		fprintf(stderr, "Error cloning " MYNAME " to " MYNAME2 "\n");
+		goto err;
+	}
+
+	c2->want_daemonize(c2, true);
+	if (!c2->startl(c2, 0, NULL)) {
+		fprintf(stderr, "Error starting container\n");
+		goto err;
+	}
+
+	char pidstr[50];
+	snprintf(pidstr, 50, "/proc/%d/attr/current", c2->init_pid(c2));
+	v = read_file(pidstr);
+	if (!v) {
+		fprintf(stderr, "Failed to read the apparmor profile name from '%s'\n", pidstr);
+		goto err;
+	}
+
+	fprintf(stderr, "apparmor policy of clone: %s\n", v);
+	char *exp1 = "lxc-";
+	if (strncmp(v, exp1, strlen(exp1)) != 0) {
+		fprintf(stderr, "Wrong profile: %s", v);
+		goto err;
+	}
+	if (strstr(v, "(enforce)") == NULL) {
+		fprintf(stderr, "Wrong profile: %s", v);
+		goto err;
+	}
+
+	c2->stop(c2);
+
+	free(v);
 	try_to_remove();
 	exit(EXIT_SUCCESS);
 
 err:
+	free(v);
 	try_to_remove();
 	exit(EXIT_FAILURE);
 }
-- 
2.43.0

```

ianmerin:
@hallyn thanks for the patch, I was out on holiday last week.  My only concern is the previously mentioned NO_NEW_PRIVS timing issue.  My container environment does not work well with NO_NEW_PRIVS option.

ianmerin:
Are there any other concerns to address?

stgraber:
@hallyn

hallyn:
Oh, sorry - I'm good with the pr, except src/tests/meson.build should really get the new testcase added, right?  Can you do that as one more step? And then +1 from me.

hallyn:
Oh, duh, it's patching aa.c not making a new test, sorry.

ianmerin:
I can put together a simple container, but I can also describe a simple example very quickly.

Create any container with a seccomp profile.  Create a custom apparmor profile (not inheriting from container-base) WITHOUT sys_admin capability.  The container will fail to start or attach due to apparmor blocking the loading of the seccomp profile.

mihalicyn:
Hi @ianmerin,

>I can put together a simple container, but I can also describe a simple example very quickly.
> Create any container with a seccomp profile. Create a custom apparmor profile (not inheriting from container-base) WITHOUT sys_admin capability. The container will fail to start or attach due to apparmor blocking the loading of the seccomp profile.

Cool, then please, add a test case, you can take some inspiration from https://github.com/lxc/lxc/blob/main/src/tests/lxc-test-unpriv or https://github.com/lxc/lxc/blob/main/src/tests/lxc-test-apparmor-generated or https://github.com/lxc/lxc/blob/main/src/tests/lxc-test-apparmor-mount

Test code is better than 1000 words ;-)

You can submit your test as a PR. We expect that:
- test will fail on a current `main` branch of LXC
- test will pass if we apply https://github.com/lxc/lxc/pull/4539/commits/50dee37cfe3201ed51f477356f81941c960a5511

Then I will take you branch with test and try to figure out what we can do on the LXC side to support your case and at the same time don't break anything for others.

hallyn:
Ok, so this caused lxc-test-no-new-privs to fail.  That's because once the process has no-new-privs set, it can only switch to an apparmor domain which "is a subset" of its current domain, according to the kernel code, and this patch makes the domain transition happen after nnp is set.  So persumably we can simply make the container profiles be subset of the lxc-start profile?

mihalicyn:
> So persumably we can simply make the container profiles be subset of the lxc-start profile?

I like the idea, but what about cases when AppArmor profile is not under LXC control? Like in Incus/LXD when they are generating their own profile and set it to the instance?

Another option might be to introduce a new option like `lxc.lsm.delay_setting_label` or something like that (with default to false, of course). That would help in cases like @ianmerin has, but also ensures that we don't change behavior existed for years.

mihalicyn:
It feels like we should make a new issue with label "feature" and discuss everything in there. @ianmerin do you want to work on this? We are ready to help with reviews and consultations along the way.

ianmerin:
Sure.  This has lost a bit of my focus recently but I'll put some time toward making those changes later this week

mihalicyn:
Great! Let's then put all the necessary info/discussion in https://github.com/lxc/lxc/issues/4525

PR #4562, opened 2025-07-03
Ensure lxc config is saved atomically

### LINKED ISSUE #4561: LXC config file empty after calling save_config API
From time to time I've experienced issues where the lxc config file is empty after calling the save_config. This is especially flimsy when using network storage. From what I was able to discern in the save_config handler the config file is opened and truncated immediately with the `open()` syscall. After that write is called but not immediately. If for some reason some issue occurs on the filesystem or the `write()` syscall fails, the file will end up empty.

emirbuljubasic:
This PR fixes an issue where save_config could leave the container config file empty if a write fails after truncation. The current implementation opens the config file with `O_TRUNC`, which clears its contents before writing. In cases of filesystem issues or write errors, this results in an empty config.

This change writes the updated configuration to a temporary file first, and then performs an atomic `rename()` to replace the original config file. This ensures that the config file is only replaced if the write succeeds, preventing accidental data loss.

Fixes #4561


stgraber:
Hmm, what happens if I have custom ownership (user/group), ACLs or xattrs on the config file?

src/lxc/lxccontainer.c
```diff
@@ -2630,16 +2630,40 @@ static bool do_lxcapi_save_config(struct lxc_container *c, const char *alt_file)
 	if (lret)
 		return log_error(false, "Failed to acquire lock");
 
-	fd_config = open(alt_file, O_WRONLY | O_CREAT | O_TRUNC | O_CLOEXEC,
-			 S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
-	if (fd_config < 0) {
-		SYSERROR("Failed to open config file \"%s\"", alt_file);
+	char tmpfile[PATH_MAX];
```
hallyn, line 2633: We don't really know where/how lxc was compiled, so PATH_MAX could be really short.  Worth adding an assert that it will be big enough for the tempfile path.

emirbuljubasic:
I completely forgot about this PR. Although this is a "concern" it's almost impossible to replicate. The issue ended up being with our infra. Sorry for the hold up here and for any confusion. Bug was burried deep in our infra and I suspected this to be the primary issue.

Closing this PR, once again sorry for the hassle :) <3

PR #4565, opened 2025-07-28
lxc/process_utils.h: use strsignal() or sys_siglist[] for Non-GNU dis…

DreamConnected:
…tros

use strsignal() for Non-GNU and sys_siglist[] for nothing, even if sys_siglist[] has been marked as deprecated by Glibc


mihalicyn (approved):
LGTM

PR #4568, opened 2025-07-30
lxc/lxccontainer.c: try fixing #4561 and improving on #4562

DreamConnected:
Added support for restoring file ownership, ACLs, and attributes.

### DIFF
--- a/src/lxc/lxccontainer.c
+++ b/src/lxc/lxccontainer.c
@@ -69,11 +69,8 @@
 #endif
 
 #if HAVE_ACL
-#include <acl/libacl.h>
-#endif
-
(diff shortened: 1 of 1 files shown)


DreamConnected (commented):
Seems like I forgot to judge the `setxattr()` call, maybe I should use the more convenient `libattr` or other implementations?

stgraber:
@mihalicyn can you take a look?

DreamConnected:
> IMHO, introducing so much code and an additional ~30KiB library dependency for this makes no sense.

I thought the same thing. Besides, the ACL was just added for this part.

PR #4569, opened 2025-08-02
lxc/utils.c: add lxc_check_kernel_met() to check kernel version at ru…

DreamConnected:
…ntime

for unavailable or downstream kernels has been merged, we can use KERNEL_VERSION_OVERRIDE to specify the version


src/lxc/conf.c
```diff
@@ -2599,10 +2599,13 @@ FILE *make_anonymous_mount_file(const struct list_head *mount_entries,
 {
 	__do_close int fd = -EBADF;
 	FILE *f;
-	int ret;
+	int ret, flags = MFD_CLOEXEC;
 	struct string_entry *entry;
 
-	fd = memfd_create(".lxc_mount_file", MFD_CLOEXEC);
+	if (lxc_check_kernel_met("6.3.0"))
```
mihalicyn, line 2605: I don't think that we need a kernel version-based feature detection. For example RHEL kernel 3.10 can have some features from 5.x line. So it is better to do a syscall-based check, basically, just try to call `memfd_create` with a flag you want to check and see what it returns. And judge if this flags is supported or not.

mihalicyn (changes requested):

src/lxc/conf.c
DreamConnected: If RHEL has this feature, we should check if return from `memfd_create` is `EINVAL` or try adding flags with `fcntl`.

src/lxc/rexec.c
```diff
@@ -98,7 +101,13 @@ static void lxc_rexec_as_memfd(char **argv, char **envp, const char *memfd_name)
 	ssize_t bytes_sent = 0;
 	struct stat st = {0};
 
-	memfd = memfd_create(memfd_name, MFD_ALLOW_SEALING | MFD_CLOEXEC);
+	memfd = memfd_create(memfd_name, MFD_ALLOW_SEALING | MFD_CLOEXEC | MFD_EXEC);
+
+	if (memfd < 0 && errno == EINVAL) {
+		TRACE("MFD_EXEC may unsupported, using MFD_CLOEXEC nad MFD_CLOEXEC");
```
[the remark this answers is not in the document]
DreamConnected: OK.

mihalicyn:
Ah, I haven't noticed that you have two commits and one is partially reverting changes from another :)

@DreamConnected, please, can you do `git rebase -i HEAD~2` and squash your commits into one (then you'll need to do `git push -f`)?

DreamConnected:
> Ah, I haven't noticed that you have two commits and one is partially reverting changes from another :)
> 
> @DreamConnected, please, can you do `git rebase -i HEAD~2` and squash your commits into one (then you'll need to do `git push -f`)?

Is it correct now?

Issue #4520, opened 2025-02-20
Allow write access to binfmt_misc when using Linux kernel >= 6.7


psoftware:
Hi,
I am not sure if I found the right place where my issue lies.
Right now it seems that it is not possible to use binfmt_misc in unprivileged lxc containers, except for (read-only) mounting. I am able to mount binfmt in a lxc container, but if i write into /proc/sys/fs/binfmt_misc/register I get a permission denied error.

Since Kernel v6.7 binfmt supports namespaces, so it should be safe to use this in containers without affecting the entire machine. Is it possible to enable permissions to use binfmt for newer kernels only?

I'm actually testing this on a Proxmox machine with lxc 6.0.0 installed.

Thank you very much (:

stgraber:
It should be possible to extend the dynamic apparmor profile generator for that.
For the static profiles, not so much as those are obviously static and so can't change based on kernel version.

psoftware:
Thank you for your answer (:

If I understood correctly, by static profiles you mention files like this
https://github.com/lxc/lxc/blob/fc77e20953b87bc1a3f2f07a079647bec990da57/config/apparmor/abstractions/container-base#L66-L68
in which there can't be conditional rules based on the kernel version, right?

How does the dynamic apparmor profile generator work? Is it only used to generated static configuration files to be commited in this lxc repo, or is it supposed to be used also on the deployed systems (either at configuration or container startup time)?

Sorry for the basic questions (: I was asking this to see if I need to file a feature request to Proxmox directly rather than here :D
In their case they know the kernel version on which lxc is deployed, so maybe they can have a custom patch for their lxc to adjust the binfmt_misc rules.

Thank you gain

kevans1124:
Hi! We are UT Austin students doing open-source contributions for a project in a class. Are there any recent details about this issue that would be helpful? Thank you!

stgraber:
I think what's mentioned above should still be accurate.

Basically we have logic to generate AppArmor profiles on-demand in `src/lxc/lsm/apparmor.c` which currently prevents accessing the binfmt filesystem.

https://github.com/lxc/lxc/blob/main/src/lxc/lsm/apparmor.c#L89

This would need to become conditional on the kernel version and then if on a recent enough kernel, block everything under `@{PROC}/sys/fs/**` EXCEPT for `@{PROC}/sys/binfmt_misc`.
[...]

PR #4576, opened 2025-08-23
Add read-write access to /proc/sys/fs/binfmt_misc in kernels >= 6.7

FernandoPicazo:
This pull request addresses #4520. 
 
This merge includes a yet unmerged feature from #4569 that gives an easy way to check current kernel version. 

This pull request, includes logic for checking current kernel version at runtime while generating apparmor profile. Largest change is that AA_PROFILE_BASE is now a template string with %s placeholders where the binfmt_misc changes need to be done. 

There are two cases handled: 
1. Kernel version less than 6.7
We generate the same apparmor profile that was previously generated. 

2. Kernel version greater than or equal to 6.7
Added a mount rule that allows for rw mount to /proc/sys/fs/binfmt_misc
Added progressive rules to exclude binfmt_misc from denied write access. 

The following config was used to verify write access on kernel 6.14 Ubuntu 25.04

```
lxc.include = /usr/share/lxc/config/ubuntu.common.conf
lxc.uts.name = noble-container
lxc.rootfs.path = dir:/home/pixel/git/lxc/test-container/rootfs
lxc.apparmor.profile = generated
lxc.init.cmd = /bin/bash
lxc.mount.entry = /proc/sys/fs/binfmt_misc proc/sys/fs/binfmt_misc none bind,private,rw,create=dir 0 0
```

Please let me know if any changes need to be done.


DreamConnected:
Please do not refer to this PR(#4569 ), which is not perfect at present. For the fragmented kernel, we should not only use the kernel version to check. I'm looking for a better implementation to complete it.

FernandoPicazo:
@DreamConnected Got it, will remove the reference to the other merge request. I'll see if I can develop a more accurate test for checking  namespace mount capabilities. 

Will close this pull request shortly.

FernandoPicazo:
I removed all the commits from #4569. I kept the check_kernel_version_met logic but expanded the kernel version parsing logic in a separate function to utilize strtoul for more robust parsing with overflow checks.

The minimum kernel version check function and the kernel parsing function are both exposed in the utils header as well.

stgraber:
@mihalicyn

stgraber:
@mihalicyn

stgraber:
The update code doesn't appear to build.

FernandoPicazo:
Hi @stgraber I fixed and hardened the logic a bit more regarding the template string. It should compile correctly now. Please take a look whenever you get a chance.

stgraber:
Tests still look rather unhappy :)

DreamConnected:
Maybe we could find a more elegant implementation.

FernandoPicazo:
@DreamConnected happy to receive some guidance. 
I had not found a different way to profile this capability, other than analyzing current kernel version. 

For now I have fixed the failure I cause by not applying rules correctly to the NO__NESTING profile extension. Tests look a little more pleased now.

DreamConnected:
Is there an API to probe for kernel features? I haven't dug into the Linux kernel source, but for non-mainline kernels, version number alone isn't reliable :)

FernandoPicazo:
@DreamConnected did not identify a dedicated kernel API for querying this feature directly.

The most reliable method is to probe for the capability at runtime by attempting the operation in isolation.

Proposed approach:

1. fork() to avoid polluting the caller’s process state
2. unshare() a fresh user and mount namespace
3. attempt a binfmt_misc mount on a temporary mountpoint under /tmp
4. clean up and exit: unmount, remove the temporary directory, and report success/failure through the child process’s exit status

I'm thinking this would be the most correct way of probing for the existence of this feature. Also playing with the fsopen api to see if that is a little friendlier.

stgraber:
I'm not sure I really want every container startup to go through all that though.

That's what we do within Incus but with Incus we have a management daemon that can do the check once on startup and then never again.

For LXC 7.0, the minimum supported kernel version is going to be 6.12, so I suspect the easiest fix for this one is to edit README.md to make it clear that we expect a 6.12 or higher kernel, then just change the AppArmor profile to always allow binfmt_misc.

DreamConnected:
It looks like LXC 7.0 will drop support for some older devices. We really should put out an announcement, otherwise, some commercial companies using LTS kernel might encounter issues with the new default configuration, especially since the 6.12 LTS kernel has only been out for less than two years :)

stgraber:
@DreamConnected it will be mentioned in the announcement. LXC 5.0 and 6.0 are both still supported so folks running older platforms can keep using those versions.

src/lxc/lsm/apparmor.c
```diff
@@ -755,12 +802,48 @@ static char *get_apparmor_profile_content(struct lsm_ops *ops, struct lxc_conf *
 	                      NULL);
 	size = strlen(profile);
 
-	must_append_sized(&profile, &size, AA_PROFILE_BASE,
-	                  STRARRAYLEN(AA_PROFILE_BASE));
+	if (apparmor_binfmt_misc_ns_allowed()) {
+		log_info(0, "AppArmor Binfmt Misc Namespace is allowed");
+		binfmt_misc_mount_rule = AA_PROFILE_BINFMT_MISC_MOUNT_RW_6_7;
```
mihalicyn, line 807: We should only allow binfmt_misc access for unprivileged containers, i.e `!is_privileged(conf)`. Otherwise, we just break isolation here for privileged containers and introduce a security vulnerability.

mihalicyn (changes requested):

mihalicyn:
Hi @FernandoPicazo,

thanks for working on this. My plan is to take your changes and fix it a bit and then we can merge it (no need to mention that I'll keep your authorship on the patches ;-) ). We are preparing for 7.0 release, that's why I want to go this faster way.

stgraber:
The change should be much smaller once all the detection logic is gone since we can rely on >= 6.12 kernel :)

PR #4579, opened 2025-08-30
Implement initial protection of LXC monitor using Landlock

stgraber:
When LXC spawns a container, the parent process is called the monitor and is a process that will remain running in the host namespaces as long as the container is running.

That process has three main roles:
 - Provide a basic API over an abstract UNIX socket to interact with the container
 - Handle container restarts by re-spawning PID1 if it exits due to a requested reboot
 - Run the post-stop functions if the container gets stopped

The latter two of those tasks require full privileges as any restriction on the monitor would then be passed on either to stop/post-stop hooks (shutdown case) or onto the container itself (restart case).

The first case however doesn't really require much in the way of privileges.
The monitor mostly asks as a forwarder of already open file descriptors, then the API client is the one actually performing the system calls needed to perform the desired action (typically attaching to the container's namespaces).

We found that it only really ever needs to open new PTS devices (/dev/ptmx, /dev/pts/XYZ) as well as perform read and writes of existing cgroup files.

With that in mind, we built a minimal Landlock policy allowing exactly that.

Enforcement was a bit trickier because the same process has to handle all 3 tasks above.
Landlock restrictions cannot be undone (for obvious reasons), so we couldn't just apply the Landlock policy to the entire process.

Instead, we've had to move that part of the monitor process into a dedicated thread.
Landlock policies apply on a per-thread basis, allowing us to then restrict that particular thread while keeping the rest of the process unrestricted and so able to handle the restart and shutdown tasks.

This really is just an initial attempt at using Landlock in the project, to get to see how it works and where else it may be a good fit. As part of this, we've identified a few places in Incus where it would be a superior fit to our current AppArmor based approach and look forward to expanding its use over there too.

As this is a very new feature in a pretty critical part of LXC, it's made optional and defaults to being disabled. We're expecting to turn it on for the LXC builds we use in the Zabbly Incus packages as well as within Incus OS since those are pretty well controlled environments where we can quickly get and address user feedback should something go wrong.


stgraber:
@mihalicyn clang is reporting an issue with the threading commit

mihalicyn (approved):
LGTM

src/lxc/start.c
```diff
@@ -572,6 +603,163 @@ int lxc_set_state(const char *name, struct lxc_handler *handler,
...
+	}
+
+	struct landlock_ruleset_attr ruleset_attr = {
+		.handled_access_fs =
+			LANDLOCK_ACCESS_FS_WRITE_FILE |
+			LANDLOCK_ACCESS_FS_READ_FILE,
+	};
+
+	// Define a new ruleset.
+	int ruleset_fd;
```
mihalicyn, line 627: nit: all variables we tend to declare at the beginning of the function

stgraber:
Thanks for the review, I've fixed the reported issues.

hallyn (approved):
Thanks,  looks good!

PR #4581, opened 2025-09-03
lxccontainer: check if target exists before remove in create_mount_target()

kadinsayani:
This pull request updates create_mount_target() to check the type of the existing mount target before removing it. Previously, the function would unconditionally call remove(dest), which could result in unintended deletions.

Relevant thread: https://github.com/canonical/lxd/pull/16280#discussion_r2301692498.


tomponline:
Thanks @kadinsayani !

PR #4582, opened 2025-09-15
conf: split `lxc.environment` into `runtime` and `hooks`

Filiprogrammer:
Introduce `lxc.environment.runtime` to set environment variables only
for the container init process and `lxc.environment.hooks` to set
environment variables only for hooks. Leave the original
`lxc.environment` unchanged. It still applies to everything.

This is needed in Proxmox Virtual Environment for containers created from OCI images that specify custom environment variables.
See: https://lore.proxmox.com/pve-devel/20250908150224.155373-6-f.schauer@proxmox.com/

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -160,3 +160,10 @@ This adds the new options `cgroup2`, `cgroup2:ro`, `cgroup2:force`,
 `cgroup2:ro:force` for the `lxc.mount.auto` configuration key. For example, if
 a user specifies `cgroup2:force` LXC will pre-mount a pure `cgroup2` layout for
 the container even if the host is running with a hybrid layout.
+
+## environment\_runtime\_hooks
+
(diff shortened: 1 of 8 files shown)


stgraber:
Ah, nice, sounds like the approach I discussed with @Blub a while back (I think?).

@mihalicyn can you take a look? General approach LGTM.

stgraber:
Ah, @Filiprogrammer we need an extra commit in here to add an API extension so API users can detect if the config key is supported or not.

Just a one-line change to `./src/lxc/api_extensions.h`.

Filiprogrammer:
> Ah, nice, sounds like the approach I discussed with @Blub a while back (I think?).

indeed

> Ah, @Filiprogrammer we need an extra commit in here to add an API extension so API users can detect if the config key is supported or not.

done. I also added some documentation.

mihalicyn:
Thanks for the contribution, Filip!

LGTM

Issue #4588, opened 2025-10-03
Adjust mode of console log file to 0660


rsyring:
Will you adjust the permissions on the console log file to 0660 (or 0640):

https://github.com/lxc/lxc/blob/48b1e82a4ff4c7376dba8b7860b1313364544b1c/src/lxc/terminal.c#L872

Which matches the mode on the main log file:

https://github.com/lxc/lxc/blob/48b1e82a4ff4c7376dba8b7860b1313364544b1c/src/lxc/log.c#L528

The current `0600` mode hinders giving additional read-only permissions on log files to groups through extended ACLs as it prevents more permissive defaults inherited from the parent directory from taking effect.

stgraber:
0660 seems like a potential issue waiting to happen, but 0640 seems like it should be fine.
Either way, I'd want things to be consistent, so if going with 0640 we'd want all of the log files to use that mode rather.

@rsyring feel like opening a PR for that?

PR #4589, opened 2025-10-03
Standardize log file create mode to 0640

rsyring:
I read through the contributing guidelines and it didn't seem like there was much required except the sign-off line.

I did look at some of the tests but it seems like this kind of change probably doesn't warrant a new test?

I'm not a C developer so this was a manual code-only change.  It makes sense to me but I didn't run locally.

Let me know if you need changes.

Thanks.

Fixes: https://github.com/lxc/lxc/issues/4588


stgraber:
@brauner @hallyn @mihalicyn any of you having an issue with the change in mode here?

hallyn (approved):
I think it should be fine.

PR #4590, opened 2025-10-13
Automatically detect compression format in the lxc-local template

stribika:
Distrobuilder is able to generate archives with various types of compression but the local template only supports xz. This patch uses `tar -a` instead of `tar -J` to autodetect the compression format.

### DIFF
--- a/templates/lxc-local.in
+++ b/templates/lxc-local.in
@@ -285,7 +285,7 @@ unpack_metadata() {
 
   echo "Using metadata file: ${LXC_METADATA}"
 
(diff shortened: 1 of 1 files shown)


stgraber:
@stribika looks like your commit message is missing the required `Signed-off-by` line.

stribika:
Can I just add myself there, or is this some kind of approval?

stgraber:
No approval or anything, just the one line Signed-off-by, same as for Linux kernel contributions and other projects using the DCO (see CONTRIBUTING for details).

Issue #4580, opened 2025-09-03
lxc fails to compile with clang 21 with "initializer-string for character array is too long, array size is 16 but initializer has size 17"


evverx:
In https://github.com/google/oss-fuzz/pull/13915 LLVM was bumped and lxc failed to build with
```sh
Step #23 - "compile-libfuzzer-address-x86_64": [1m../src/lxc/uuid.c:77:32: [0m[0;1;31merror: [0m[1minitializer-string for character array is too long, array size is 16 but initializer has size 17 (including the null terminating character); did you mean to use the 'nonstring' attribute? [-Werror,-Wunterminated-string-initialization][0m
Step #23 - "compile-libfuzzer-address-x86_64":    77 |         [0;34mstatic[0m [0;34mconst[0m [0;34mchar[0m table[[0;32m16[0m] = [0;32m"0123456789abcdef"[0m;[0m
Step #23 - "compile-libfuzzer-address-x86_64":       | [0;1;32m                                      ^~~~~~~~~~~~~~~~~~
Step #23 - "compile-libfuzzer-address-x86_64": [0m1 error generated.
```
https://oss-fuzz-gcb-logs.storage.googleapis.com/log-99befb7c-feaf-4479-8f1a-4a3f56812e94.txt

It points to https://github.com/lxc/lxc/blob/c8fd40f6e54a8bdd721cc84de6da1f37a5988a40/src/lxc/uuid.c#L75-L80

maflcko:
The compiler assumes that the null-terminator is required and has been forgotten. The corresponding GCC warning is https://gcc.gnu.org/pipermail/gcc-patches/2024-June/656014.html.

An easy workaround would be to just enumerate all elements in the array:

```c
static const char table[] = {
    '0', '1', '2', '3', '4', '5', '6', '7', 
    '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'
};
```

(with or without the explicit `[16]`)

evverx:
@maflcko thank you for the link. I took a look and ended up in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115185#c14 (where the same code in strace was discussed). There it was fixed by turning it into an array as well: https://github.com/strace/strace/commit/1ee921b962fd2156eb1b5867f4dc20c0fc115469.

[...]

PR #4592, opened 2025-10-15
start: Only include linux/landlock.h when landlock is enabled

### LINKED ISSUE #4591: lxc fails to build when the Landlock header is missing and `landlock-monitor` is `false`
For example it fails to compile on OSS-Fuzz because of that: https://github.com/lxc/lxc/actions/runs/18479871665/job/52652294373:
```sh
...
clang -Isrc/lxc/liblxc.a.p -Isrc/lxc -I../src/lxc -I. -I.. -Isrc -I../src -Isrc/include -I../src/include -I../src/lxc/cgroups -I../src/lxc/storage -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=gnu11 -O0 -g -Wno-format-signedness -Wno-missing-field-initializers -Wno-unused-parameter -Wvla -Wcast-align -Wstrict-prototypes -fno-strict-aliasing -fstack-clash-protection --param=ssp-buffer-size=4 -Werror=implicit-function-declaration -Wmissing-include-dirs -Wold-style-definition -Winit-self -Wunused-but-set-variable -Wno-unused-parameter -Wfloat-equal -Werror=return-type -Werror=incompatible-pointer-types -Wformat=2 -Wshadow -Wendif-labels -Werror=overflow -fdiagnostics-show-option -Werror=shift-count-overflow -Wdate-time -Wnested-externs -fasynchronous-unwind-tables -fexceptions -Warray-bounds -Wreturn-local-addr -Wno-typedef-redefinition -Wno-gnu-variable-sized-type-not-at-end -ffat-lto-objects -include config.h -O1 -fno-omit-frame-pointer -gline-tables-only -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion -Wno-error=deprecated-declarations -Wno-error=implicit-function-declaration -Wno-error=implicit-int -Wno-error=unknown-warning-option -Wno-error=vla-cxx-extension -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -fPIC -pthread -fvisibility=default -DIN_LIBLXC -MD -MQ src/lxc/liblxc.a.p/start.c.o -MF src/lxc/liblxc.a.p/start.c.o.d -o src/lxc/liblxc.a.p/start.c.o -c ../src/lxc/start.c
../src/lxc/start.c:9:10: fatal error: 'linux/landlock.h' file not found
    9 | #include <linux/landlock.h>
      |          ^~~~~~~~~~~~~~~~~~
1 error generated.
```
In a couple of weeks It should be possible to bump the base builder images from Ubuntu 20.04 to Ubuntu 24.04 (where `linux-libc-dev` is new enough and comes with the header)) but it seems it shouldn't fail to compile when `landlock-monitor` is `false` regardless of whether the header is present or not.

stgraber:
Closes #4591

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -6,7 +6,6 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <grp.h>
-#include <linux/landlock.h>
 #include <poll.h>
 #include <pthread.h>
 #include <signal.h>
@@ -69,6 +68,8 @@
 #endif
 
 #if HAVE_LANDLOCK_MONITOR
+#include <linux/landlock.h>
+
 #ifndef landlock_create_ruleset
 static inline int
 landlock_create_ruleset(const struct landlock_ruleset_attr *const attr,


evverx:
I can confirm that the OSS-Fuzz build script can get past that error with this patch applied.

(It still fails because of https://github.com/lxc/lxc/issues/4580 and that's weird because that warning was supposed to be turned off on OSS-Fuzz. I'll try to figure out why it isn't off)

hallyn (approved):

PR #4597, opened 2025-10-26
lxc/file_utils: ensure complete data write in lxc_write_nointr

### LINKED ISSUE #4546: lxc-attach drops output with lots of console output
i'm experiencing a problem that when there is a lot of terminal output in a lxc-attach session, it appears to drop characters. It could be some kind of flow control problem. Unfortunately, this makes TUI applications like vim unusable without refreshing a lot (e.g. they stop drawing halfway the screen).

This happens on version 5.0.3 on Ubuntu 24.04. Have not yet been able to test it with 6.x. Not sure when this started happening (possibly on upgrade to Ubuntu 24.04) but i've been using lxc for ages and it's not always been an issue.

Other people have reported similar problems on the forum: https://discuss.linuxcontainers.org/t/lxc-attach-long-output-stops-suddenly-possible-bug/22031

DreamConnected:
Previously, lxc_write_nointr could return without writing all data when write() returned EAGAIN/EWOULDBLOCK due to buffer full conditions. ﻿

This change:
- Implements a loop to continue writing until all data is sent
- Handles EINTR, EAGAIN, and EWOULDBLOCK errors appropriately
- Uses poll() to wait for fd to become ready when blocked

Fixed #4546

### DIFF
--- a/src/lxc/file_utils.c
+++ b/src/lxc/file_utils.c
@@ -11,6 +11,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <time.h>
+#include <poll.h>
 
 #include "file_utils.h"
 #include "macro.h"
@@ -147,14 +148,91 @@ ssize_t lxc_read_try_buf_at(int dfd, const char *path, void *buf, size_t count)
 	return ret;
 }
 
+static int lxc_wait_for_io_ready(int fd, int event, int timeout)
+{
+	struct pollfd pfd = {
+		.fd = fd,
+		.events = event,
+		.revents = 0
+	};
+	int ret;
+
+	do {
+		ret = poll(&pfd, 1, timeout);
+	} while (ret < 0 && errno == EINTR);
+
+	if (ret < 0)
+		return -errno;
+
+	if (ret == 0)
+		return -ETIMEDOUT;
+
+	if (pfd.revents & (POLLERR | POLLHUP | POLLNVAL)) {
+		if (pfd.revents & POLLERR)
+			return -EPIPE;
+		if (pfd.revents & POLLHUP)
+			return -EPIPE;
+		if (pfd.revents & POLLNVAL)
+			return -EBADF;
+	}
+
+	if (!(pfd.revents & event))
+		return -EAGAIN;
+
+	return ret;
+}
+
 ssize_t lxc_write_nointr(int fd, const void *buf, size_t count)
 {
 	ssize_t ret;
+	size_t written = 0;
+	const char *ptr = (const char *)buf;
+	int flags, pret, is_blocking = 1;
 
-	do {
-		ret = write(fd, buf, count);
-	} while (ret < 0 && errno == EINTR);
+	flags = fcntl(fd, F_GETFL);
+	if (flags != -1 && (flags & O_NONBLOCK))
+		is_blocking = 0;
 
+	while (written < count) {
+		ret = write(fd, ptr + written, count - written);
+		
+		if (ret > 0) {
+			written += ret;
+			continue;
+		}
+		
+		if (ret == 0)
+			break;
+		
+		if (errno == EINTR)
+			continue;
+
+		if (!is_blocking && errno == EAGAIN) {
+			pret = lxc_wait_for_io_ready(fd, POLLOUT, 5000);
+
+			if (pret > 0)
+				continue;
+
+			if (pret == -ETIMEDOUT) {
+				if (written > 0)
+					return (ssize_t)written;
+				return 0;
+			}
+
+			if (pret < 0) {
+				errno = -pret;
+				break;
+			}
+
+			continue;
+		}
+		
+		break;
+	}
+	
+	if (written > 0)
+		return (ssize_t)written;
+	
 	return ret;
 }
 


DreamConnected:
@stgraber Can anyone review it? @mihalicyn seems too busy to handle.

src/lxc/file_utils.c
```diff
@@ -147,15 +148,50 @@ ssize_t lxc_read_try_buf_at(int dfd, const char *path, void *buf, size_t count)
...
+			continue;
+		}
+		
+		if (ret == 0)
+			break;
+		
+		if (errno == EINTR)
+			continue;
+		
+		if (errno == EWOULDBLOCK) {
```
mihalicyn: Please can you explain how you figured that there is a problem.
Do you have a pointer to a specific `lxc_write_nointr()` call site where this error occurs?
AFAIK, `EWOULDBLOCK` can only happen with sockets.

From commit message:
>Previously, lxc_write_nointr could return without writing all data
when write() returned EAGAIN/EWOULDBLOCK due to buffer full conditions.

I guess this is about sockets, right? We have only a few cases when we call this function on a socket fd and buffer is tiny. Also, in case of EAGAIN/EWOULDBLOCK error, function returns an error, i.e. there is no silent data corruption.

src/lxc/file_utils.c
```diff
@@ -147,15 +148,50 @@ ssize_t lxc_read_try_buf_at(int dfd, const char *path, void *buf, size_t count)
...
 
+static int lxc_wait_for_io_ready(int fd, int event, int timeout)
+{
+	struct pollfd pfd = {
+		.fd = fd,
+		.events = event,
+		.revents = 0
+	};
+	
+	return poll(&pfd, 1, timeout);
```
mihalicyn: I guess we need a proper error handling on `poll()` syscall too. It returns negative value in case of error (including EINTR btw.). In current approach, we don't distinguish between error and "normal" scenarios.

mihalicyn:
Hi @DreamConnected,

thanks for working on this. AFAIU, this is an attempt to fix #4546.
Can you confirm that you was able to reproduce this specific issue from #4546?
Can you confirm that with this patches applied, issue is not reproducible anymore?

(I'm very sorry for a super long delay with review.)

Hi @laanwj,

thanks for reporting an issue. As far as I see, you expressed interest to test this change here:
https://github.com/lxc/lxc/issues/4546#issuecomment-3450574414

Please, can you clone this branch, compile LXC and test it to confirm if this change helps to solve the problem for you.

Kind regards,
Alex

src/lxc/file_utils.c
mihalicyn: Amazing! I think we should try to make a test for this scenario. Let's think about this a bit.

mihalicyn: Hi @DreamConnected,

I've spend more time with this issue and the more I look, the more I feel that I'm that bad guy who introduced this regression in this commit https://github.com/lxc/lxc/pull/4304/commits/715fb4effaa7f35e642056d2a10d5b8c393bfac0.

Back then, there was an issue with io_uring and there were several attempts to properly handle this stuff. I believe, that clean and correct solution is https://github.com/lxc/lxc/pull/4304/commits/7fd671dbce98d139e52e8c4266f1050ef49ea8af, while I claimed in the comment that this is a workaround, in fact, using oneshot-polling mode is the only option there. If we use oneshot mode with io_uring, then making FD-s non-blocking seems unnecessary.

Another point is that the logic you've introduced in `lxc_write_nointr` basically implements *blocking* write on a *non-blocking* FD, right? But why we can't just make FD blocking if we really need this blocking semantic anyways?

My proposal is to take https://github.com/lxc/lxc/commit/715fb4effaa7f35e642056d2a10d5b8c393bfac0 and revert it, then check if this helps with the issue we are trying to fix and at the same time doesn't break io_uring scenario (when LXC is built with `-Dio-uring-event-loop=true`). It would be awesome if you try this, cause you've already spend a lot of time researching this issue and preparing a fix, and I don't want to take this initiative from you. If this revert helps and doesn't break anything else, then you can submit this revert as your fix. ;)

Feel free to disagree or correct me if I'm wrong ;)

mihalicyn: >But I need some time to make sure this revert won't affect https://github.com/lxc/lxc/pull/4304

No worries about that part. I gonna re-introduce CI tests for io-uring case, so we will be able to validate this part. I wonder if this revert will help with data corruption issue (it should _theoretically_, but Software Engineering as Physics are experimental sciences ;-) ).

laanwj:
Sorry for the late response.

i've tested this PR (commit 1e1ed6e4a7efad7efb3b05a39186d8a1bb6f61e7) and can confirm it fixes the issue i reported in #4546.

i've checked in two ways, inside a `lxc-attach` session:
- Open neovim on various large text files. The problem i reported that the output gets broken off halfway does no longer happen.
- Check using `dd if=/dev/urandom bs=1 count=5000000 status=none | base64`. No lines are cut off anymore.

Thank you!

laanwj (approved):

mihalicyn:
Hey @laanwj,

thanks for testing this!

I was busy with this code and also have prepared an automated test that can recreate a data corruption scenario on CI.
What I've found is that current PR fixes data corruption, *but* there is something else that makes test to stuck sometimes. It looks like another one and very close bug. I'm digging further.

mihalicyn:
Replaced by https://github.com/lxc/lxc/pull/4633

DreamConnected:
> Replaced by https://github.com/lxc/lxc/pull/4633

OK, there seems to be a problem with the GitHub API. My login token is banned every few minutes (401).

PR #4598, opened 2025-10-26
Fix meson build generation of apparmor container-base

gibmat:
While tracking down a bug, I noticed that the `meson.build` file is incorrectly configured for generating the `container-base` file, and that shipping a generated `container-base` file was masking that issue.


stgraber:
Looks like the apparmor test doesn't like this change much :)

gibmat:
> Looks like the apparmor test doesn't like this change much :)

Yeah, haven't had a chance to investigate that yet. :)

PR #4609, opened 2025-11-20
apparmor: skip /proc and /sys restrictions if nesting is enabled

ThomasLamprecht:
If nesting is enabled, it's already possible to mount your own instance of both procfs and sysfs inside the container, so protecting the "original" ones at /proc and /sys makes no sense, but breaks certain nested container setups.

This was recommended by @Blub and already rolled out for PVE and also Incus [0] a few weeks before.
Submitting this patch from a colleague before we forget again, as it might help others.

[0]: https://github.com/lxc/incus/pull/2624/commits/1fbe4bffb9748cc3b07aaf5db310d463c1e827d0


stgraber:
Thanks!

PR #4611, opened 2025-11-26
process_utils: Remove P_PIDFD macro

jaeyoonjung:
It is already defined as enum in sys/wait.h. Redefining it as a macro would end up a build error if !HAVE_SYS_PIDFD_H, eg.
```
../git/src/lxc/process_utils.h:144:17: error: expected identifier before numeric constant
  144 | #define P_PIDFD 3
      |          
```       ^

### DIFF
--- a/src/lxc/process_utils.h
+++ b/src/lxc/process_utils.h
@@ -138,13 +138,6 @@
 #define CLONE_NEWTIME 0x00000080 /* New time namespace */
 #endif
 
-/* waitid */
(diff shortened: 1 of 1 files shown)


stgraber:
Clearly this is still needed as all the Ubuntu 22.04 tests are failing to build ;)

jaeyoonjung:
v2 sent in https://github.com/lxc/lxc/pull/4614

PR #4619, opened 2025-12-04
checkonfig: Fixed compatible with toybox/gunzip

yangh:
gunzip in Android/toybox has no -q option.


stgraber:
We don't take pull requests against the stable branches. Instead we have an automatic process to consider every commit for backporting to stable once it's in the main branch.

PR #4620, opened 2025-12-04
Fallback to XDG_RUNTIME_DIR when /run not found

yangh:
Instead of return null immediately when RUNTIME_PATH not found, fallback to XDG_RUNTIME_DIR or HOME.

### DIFF
--- a/src/lxc/utils.c
+++ b/src/lxc/utils.c
@@ -253,10 +253,8 @@ char *get_rundir(void)
 	const char *homedir;
 	struct stat sb;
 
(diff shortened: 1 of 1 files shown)


stgraber:
@yangh can you rebase your branch on main to get rid of that "Merge branch" commit that you have in here?

yangh:
> @yangh can you rebase your branch on main to get rid of that "Merge branch" commit that you have in here?

OK, I'll rebase my branch. I tried "update branch" in the web which didn't work as I hope.

Issue #3767, opened 2021-04-04
Missing "--rbduser" option in "lxc-create -B rbd" and perhaps in config files as well


creatica-soft:
# Required information

 * Distribution: ubuntu
 * Distribution version: focal
 * The output of
   * `lxc-start --version`
   4.0.2 (tried 4.0.6 as well)

# Issue description

"lxc-create -B rbd" is missing "--rbduser" option (it seems to rely on default admin for ceph.client.admin.keyring file)

admin is not the only user in ceph, so the ability to provide the ceph username to authenticate to ceph cluster is needed.

Rahik-Sikder:
I believe I have a good idea of how to implement this. Could I get this issue assigned to me?

stgraber:
done!

PR #4622, opened 2025-12-09
added "--rbduser" option in "lxc-create -B rbd"

Rahik-Sikder:
Added a flag "--rbduser" to "lxc-create -B rbd" to allow lxc-create to specify a ceph username when executing rbd commands . Also added a default config RBDUSER that is set as NULL. During the execution of rbd commands if rbduser is NULL, such as with the default config, it is assumed to be ceph admin (aka prior behavior). 

Resolves issue #3767

Sidenote: The RBDUSER config is used for all rbd commands (create, map, unmap, destroy) but the user is only able to pass in --rbduser for lxc-create. During lxc-destroy, the rbduser will always be NULL, and thus admin, which I assumed to be the intended behavior.

### DIFF
--- a/doc/lxc-create.sgml.in
+++ b/doc/lxc-create.sgml.in
@@ -139,6 +139,7 @@
 	    You can specify the following options :
 	    <replaceable>--rbdname RBDNAME</replaceable> will create a blockdevice named RBDNAME rather than the default, which is the container name.
 	    <replaceable>--rbdpool POOL</replaceable> will create the blockdevice in the pool named POOL, rather than the default, which is 'lxc'.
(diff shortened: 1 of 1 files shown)


Rahik-Sikder:
Seems like the PR failed the last check due to a missing character in the commit message. Just updated it, sorry about that!

stgraber:
Should be added to `doc/lxc-create.sgml.in` too I think.

Rahik-Sikder:
I've added "--rbduser" to docs, let me know if I should squash both commits into one.

stgraber:
@Rahik-Sikder missing Signed-off-by line on that latest change.

Rahik-Sikder:
sorry again, I seem to keep forgetting! let me fix

PR #4633, opened 2026-01-22
lxc-attach: fix data corruption during heavy IO on PTS

mihalicyn:
Reworked version of https://github.com/lxc/lxc/pull/4597 with testcase.

Authorship of @DreamConnected is preserved, of course.

Fixes: #4546

### DIFF
--- a/src/tests/lxc-test-lxc-attach
+++ b/src/tests/lxc-test-lxc-attach
@@ -58,6 +58,7 @@ for i in $(seq 1 100); do
 	if [ "$attach" != "busy" ]; then
 		FAIL "lxc-attach -n busy -- hostname"
 	fi
+	rm -f "${ATTACH_LOG}"
 done
 
 # stdin  --> /dev/null
@@ -67,6 +68,7 @@ attach=$(lxc-attach -n busy -l trace -o "${ATTACH_LOG}" -- hostname < /dev/null 
 if [ "$attach" != "busy" ]; then
         FAIL "lxc-attach -n busy -- hostname < /dev/null"
 fi
+rm -f "${ATTACH_LOG}"
 
 # stdin  --> attached to pty
 # stdout --> /dev/null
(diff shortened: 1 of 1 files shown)


DreamConnected:
It seems that the `cmp` has detected that the files are different and there are no short write in the log yet.

mihalicyn:
> It seems that the `cmp` has detected that the files are different and there are no short write in the log yet.

yep, the reason was that depending on at which point of time `dd if=/dev/urandom bs=$BS count=1 status=none | hexdump | tee large-data.txt` command reaches busybox, it will change (or not change) current directory and file can land to `/root` or to `/`. Solution is to use an absolute path `/root/large-data.txt`.

mihalicyn:
Okay, now everything looks green :)

mihalicyn:
Just for the history. In a new test case I mentioned [a bug in Busybox](https://github.com/lxc/lxc/pull/4633/commits/97bd7699c79493f32fcee0f976b6c397cf0f0ad5#diff-394e07319f8b32cb322ae6050d944d7f327477841ee52c7d8d2cdd95d0f6a232R210) this is a reproducer (of course, not involving LXC anyhow):
```
#!/bin/sh
set -e

BS=1000000

out=$(mktemp /tmp/out_XXXX)
( sleep 0.25; echo "echo DATASTART ; dd if=/dev/urandom bs=$BS count=1 status=none | hexdump | tee large-data.txt ; echo DATAEND"; ) | \
	script -q -e -c "busybox sh" | \
	sed -n '/DATASTART/,/DATAEND/{/DATASTART/d;/DATAEND/d;s/[\r\n]*$//;p}' > $out


[ $(stat -c%s $out) -gt $BS ] || ( echo FAILSIZE && false )
cmp -s large-data.txt $out || (  echo FAILDATA && false )
rm -f large-data.txt $out
```

this will stuck forever. If we add `sleep 1` after `echo "echo DATASTART ; dd if=/dev/urandom bs=$BS count=1 status=none | hexdump | tee large-data.txt ; echo DATAEND";` it brings this back to life. It is something I can look into once time permits.

stgraber:
That'd be one for @gibmat as he's the packager for Debian.
On our end, this will make it into LXC 6.0.6 when we publish that bugfix release over the coming few weeks, but for Debian stable, that'd then need to be cherry-picked in their package.

gibmat:
> do you see a possibility to get this fix into current debian stable?

Staged at https://salsa.debian.org/lxc-team/lxc/-/commit/4615978f082ebe26dc4908f1bd235d48a341ba83 for the 13.4 stable point release, [planned for March 14](https://release.debian.org/).

Issue #4641, opened 2026-02-14
Build failure with glibc-2.43 (../lxc-6.0.5/src/lxc/open_utils.h:80:19: error: conflicting types for 'openat2'; have 'int(int,  const char *, struct open_how *, size_t)' {aka 'int(int,  const char *, struct open_how *, long unsigned int)'})


juippis:
With glibc-2.43 building the latest release 6.0.5 fails to:
```
In file included from ../lxc-6.0.5/src/lxc/file_utils.c:18:
../lxc-6.0.5/src/lxc/open_utils.h:80:19: error: conflicting types for 'openat2'; have 'int(int,  const char *, struct open_how *, size_t)' {aka 'int(int,  const char *, struct open_how *, long unsigned int)'}
   80 | static inline int openat2(int dfd, const char *filename, struct open_how *how, size_t size)
      |                   ^~~~~~~
In file included from /usr/include/bits/fcntl-linux.h:492,
                 from /usr/include/bits/fcntl.h:61,
                 from /usr/include/fcntl.h:35,
                 from ../lxc-6.0.5/src/lxc/file_utils.c:6:
/usr/include/bits/fcntl-linux-fortify.h:36:1: note: previous definition of 'openat2' with type 'int(int,  const char *, const struct open_how *, long unsigned int)'
   36 | openat2 (int __dfd, const char *__filename, const struct open_how *__how,
      | ^~~~~~~
```

Downstream bug: https://bugs.gentoo.org/970021

Full build.log attached.

[lxc-6.0.5:20260214-100916.log](https://github.com/user-attachments/files/25314254/lxc-6.0.5.20260214-100916.log)

DreamConnected:
It seems that the `meson` script did not correctly detect `openat2()` has already been defined. You can try adding disabling cflags `- U_FORTIFY_SOURCE`, or check `meson.build# L572-L574`

juippis:
Actually looking at the **meson.build** file, it has:
```
    ['openat2',           '''#include <sys/types.h>
                             #include <sys/stat.h>
                             #include <fctnl.h>'''],
```
And **fctnl.h** doesn't exist in my system, but **fcntl.h** does. If I swap the library,
```patch
diff '--color=auto' -Naur a/meson.build b/meson.build
--- a/meson.build       2025-08-15 07:44:19.000000000 +0300
+++ b/meson.build       2026-02-15 09:33:23.968819561 +0200
@@ -546,7 +546,7 @@
[...]

PR #4642, opened 2026-02-15
meson.build: fix openat2 include typo, fix with glibc-2.43 +FORTIFY

juippis:
Closes: https://github.com/lxc/lxc/issues/4641

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -571,7 +571,7 @@ foreach ident: [
     ['move_mount',        '''#include <sys/mount.h>'''],
     ['openat2',           '''#include <sys/types.h>
                              #include <sys/stat.h>
-                             #include <fctnl.h>'''],
+                             #include <fcntl.h>'''],
     ['open_tree',         '''#include <sys/mount.h>'''],
     ['personality',       '''#include <sys/personality.h>'''],
     ['pidfd_open',        '''#include <stdlib.h>


vortexilation:
@juippis 
Did you still using -U_FORTIFY_SOURCE cflags?, because even tough using this PR, I am still getting that compilation error.

juippis:
It should be enabled by default, but I don't see them in compile flags indeed. Wonder if lxc strips them out?

I do get warnings such as `/usr/include/bits/fcntl-linux-fortify.h:46:10: warning: argument 3 null where non-null expected [-Wnonnull]` so do you have `-Werror` enabled?

```

In function 'openat2',
    inlined from 'can_use_mount_api' at ../lxc-6.0.5/src/lxc/mount_utils.c:545:8:
/usr/include/bits/fcntl-linux-fortify.h:46:10: warning: argument 3 null where non-null expected [-Wnonnull]
   46 |   return __openat2_alias (__dfd, __filename, __how, __usize);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/features.h:540,
                 from /usr/include/fcntl.h:25:
/usr/include/bits/fcntl-linux-fortify.h: In function 'can_use_mount_api':
/usr/include/bits/fcntl-linux-fortify.h:25:12: note: in a call to function '__openat2_alias' declared 'nonnull'
   25 | extern int __REDIRECT (__openat2_alias, (int __dfd, const char *__filename,
      |            ^~~~~~~~~~
```

Find the full build.log attached here.
[lxc-6.0.5:20260218-081251.log](https://github.com/user-attachments/files/25385328/lxc-6.0.5.20260218-081251.log)

Anyway if I take the patch out, it fails again to the error I reported at #4641 so this typo fix fixes the original issue that I had :) you may want to open a new issue?

vortexilation:
@juippis 

My issue with glibc 2.43 and above has been fixed, thanks for your replies.

PR #4660, opened 2026-03-10
utils: Update buffer size to account for quotes

stgraber:


skozina:
Thanks for quick follow up! This is not sufficient fix I'm afraid. If `lxc.hook.version = 1` is set in the config, the arguments can be already in the quotes in the hook, eg.:
```
lxc.hook.version = 1
lxc.hook.pre-start = /proc/13699/exe callhook /var/lib/lxd 'default' 'c1' start
lxc.hook.stop = /root/go/bin/lxd callhook /var/lib/lxd 'default' 'c1' stopns
lxc.hook.post-stop = /root/go/bin/lxd callhook /var/lib/lxd 'default' 'c1' stop
```

In this case the code, even with the increased size of buffers, will put the whole hook in quotes, which will result in something like:
```
exec '/proc/13699/exe callhook /var/lib/lxd 'default' 'c1' start'
```
(note the inner quotes)

skozina:
This is the specific piece of code I'm referring to (when `hook_version == 1`):
https://github.com/lxc/lxc/blob/main/src/lxc/utils.c#L619-L622

stgraber:
There, that should provide the right middle ground where the hook command itself is left unchanged but quotes are put around the arguments that are managed by LXC itself.

skozina:
That works (if it's not necessary to escape the script itself). Thanks!

stgraber:
> That works (if it's not necessary to escape the script itself). Thanks!

None of the escaping is strictly necessary. If we were a daemon running as root accepting input from a non-root user, that'd be a different matter, but here we basically just take some user controlled input and put it together into an exec command, so any issue this may cause is entirely of the user's doing ;)

skozina:
I agree;-) Thanks for the quick fix!

PR #4663, opened 2026-03-14
Fix issue where lxc-start takes a long time to start up on a cgroup v2 system without systemd

### LINKED ISSUE #4657: lxc-start 'hangs' on a cgroup v2 system without systemd
By 'hangs', I mean it takes about 10 minutes for the command to succeed. The problem is that lxc is trying to communicate with a non-existent DBUS systemd controller.

alex14641:
Fixes https://github.com/lxc/lxc/issues/4657

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -811,7 +811,7 @@ found_headers = []
 missing_headers = []
 foreach tuple: [
     ['sys/resource.h'],
-    ['sys/memfd.h'],
+    ['linux/memfd.h'],
     ['sys/personality.h'],
     ['sys/pidfd.h'],
     ['sys/signalfd.h'],
--- a/meson.build
+++ b/meson.build
@@ -865,6 +865,11 @@ foreach tuple: [
     endif
 endforeach
 
+if want_install_init
+    if 'systemd' in init_script
+        srcconf.set10('HAVE_SYSTEMD', 1)
+    endif
+endif
 # Generate config.h
 config_h = configure_file(
         output: 'config.h',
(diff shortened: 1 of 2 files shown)


stgraber:
Hmm, not sure why there's a revert in this branch.

Also for the systemd part, it feels like the problem has more to do with a lack of systemd at runtime rather than at compile time. So we'd want the function to maybe check for /run/systemd at run time rather than check if LXC was built with the systemd units included.

alex14641:
I'm going to delete this branch and start over, but I want to think about your statement some more. 

Yahoo Mail: Search, Organize, Conquer 
 
  On Fri, Mar 13, 2026 at 11:12 PM, Stéphane ***@***.***> wrote:   stgraber left a comment (lxc/lxc#4663)
Hmm, not sure why there's a revert in this branch.

Also for the systemd part, it feels like the problem has more to do with a lack of systemd at runtime rather than at compile time. So we'd want the function to maybe check for /run/systemd at run time rather than check if LXC was built with the systemd units included.

alex14641:
> I'm going to delete this branch and start over, but I want to think about your statement some more.  Yahoo Mail: Search, Organize, Conquer On Fri, Mar 13, 2026 at 11:12 PM, Stéphane ***@***.***> wrote: stgraber left a comment ([lxc/lxc#4663](https://github.com/lxc/lxc/pull/4663)) Hmm, not sure why there's a revert in this branch. Also for the systemd part, it feels like the problem has more to do with a lack of systemd at runtime rather than at compile time. So we'd want the function to maybe check for /run/systemd at run time rather than check if LXC was built with the systemd units included. — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you authored the thread.Message ID: ***@***.***>

I think this has to be a compile-time thing:  on a non-systemd system, the systemd dependencies will not be present, and compilation will fail.

PR #4666, opened 2026-03-15
Fix issue where lxc-start takes a long time to start up on a cgroup v2 system without systemd.

### LINKED ISSUE #4657: lxc-start 'hangs' on a cgroup v2 system without systemd

alex14641:
Fixes https://github.com/lxc/lxc/issues/4657

I believe this has to be a compile-time fix: non-systemd systems won't have the systemd dependencies, causing a compile failure.

### DIFF
--- a/src/lxc/cgroups/cgfsng.c
+++ b/src/lxc/cgroups/cgfsng.c
@@ -1521,6 +1521,9 @@ static int unpriv_systemd_create_scope(struct cgroup_ops *ops, struct lxc_conf *
 	__attribute__((__cleanup__(_dbus_connection_free))) DBusConnection *connection = NULL;
 	unsigned int len;
 
+	if (!file_exists("/run/systemd/system"))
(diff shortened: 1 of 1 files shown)


stgraber:
I still think this needs to be a runtime check. 

If you take the Debian packages as an example, they are built with dbus and systemd but can be run in environments where the init system isn't systemd.

alex14641:
> I still think this needs to be a runtime check.
> 
> If you take the Debian packages as an example, they are built with dbus and systemd but can be run in environments where the init system isn't systemd.

It's now a runtime check. I guess I confused a setup error for a compile time error.

stgraber:
Looks good to me. I'll rebase the branch to just have the one commit and we can merge that.

stgraber (approved):

Issue #4655, opened 2026-03-03
Compatibility of `lxc_check_inherited` logic with Valgrind instrumentation


anibalportero:
When using wrapping Valgrind around a container management system that uses LXC C API, I get some trouble when LXC tries to close an fd that Valgrind uses for instrumentation. This gets `lxc_check_inherited` into an infinite loop as Valgrind keeps running and fd can't be closed.

To solve this, I made this [patch](https://github.com/pantavisor/meta-pantavisor/blob/084f540895048d7ea9daa0d71ef3472826564f3c/recipes-pv/lxc6-pv/files/0001-lxc-fix-infinite-loop-on-closeall-lxc6.patch). The patch would be only used for testing purposes:

```
From d8500b3b451aca9f55712579d63293b9c000ffe5 Mon Sep 17 00:00:00 2001
From: Anibal Portero Hermida <anibal.portero@pantacor.com>
Date: Tue, 3 Mar 2026 09:56:09 +0000
Subject: [PATCH] feat(lxc6-pv): add lxc6 patch to fix infinite loop on closeall

---
 src/lxc/start.c | 8 +++++++-
 1 file changed, 7 insertions(+), 1 deletion(-)

diff --git a/src/lxc/start.c b/src/lxc/start.c
index e0ec0b6f4..c4ad10e56 100644
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -238,7 +238,7 @@ static unsigned int get_listen_fds_max(void)
 int lxc_check_inherited(struct lxc_conf *conf, bool closeall,
 			int *fds_to_ignore, size_t len_fds)
 {
-	int fd, fddir;
+	int fd, fddir, prev_fd = -1;
 	size_t i;
 	DIR *dir;
 	struct dirent *direntp;
@@ -284,6 +284,11 @@ restart:
 			if (fds_to_ignore[i] == fd)
 				break;
 
+		if (prev_fd == fd) {
+			WARN("Tried already to close inherited fd %d. Ignoring...", fd);
+			break;
+		}
+
 		if (fd == fddir || fd == lxc_log_fd || fd == lxc_log_out_fd ||
 		    (i < len_fds && fd == fds_to_ignore[i]))
 			continue;
@@ -320,6 +325,7 @@ restart:
 		}
 
 		if (closeall) {
+			prev_fd = fd;
 			if (close(fd))
 				SYSINFO("Closed inherited fd %d", fd);
 			else
```

[...]

PR #4667, opened 2026-03-17
Fix infinite loop caused by failure to close inherited file descriptors.

anibalportero:
This patch modifies the lxc_check_inherited function to avoid hangs when closing inherited file descriptors. In case it finds a close failure, it continues with the next file descriptor following a best effort strategy.

Issue is tracked in https://github.com/lxc/lxc/issues/4655.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -352,9 +352,10 @@ restart:
 		}
 
 		if (closeall) {
-			if (close(fd))
-				SYSINFO("Closed inherited fd %d", fd);
-			else
+			if (close(fd)) {
+				WARN("Could not close fd %d. Continuing...", fd);
+				continue;
+			} else
 				INFO("Closed inherited fd %d", fd);
 			closedir(dir);
 			goto restart;


src/lxc/start.c
```diff
@@ -352,9 +352,10 @@ int lxc_check_inherited(struct lxc_conf *conf, bool closeall,
...
-			if (close(fd))
-				SYSINFO("Closed inherited fd %d", fd);
-			else
+			if (close(fd)) {
+				WARN("Could not close fd %d. Continuing...", fd);
+				continue;
+			} else
 				INFO("Closed inherited fd %d", fd);
 			closedir(dir);
 			goto restart;
```
DreamConnected, line 361: This patch still cannot solve the loop problem. After `continue`, you immediately follow with `goto restart`, causing the loop again.

anibalportero: Maybe it's not the most elegant solution, but I think it would still work.

We are doing `goto restart` only when we are able to close the fd. There will be a point when `/proc/self/fd` is either empty or just inhabited by non closeable fds. In the second case, we are going to continue till the end of the while loop and exit.

PR #4669, opened 2026-03-23
commands: Enhance validation for config item requests and console log commands

nidu-ninja:
I’ve improved validation in the command socket handling to make it safer against malformed inputs.


stgraber:
@nidu-ninja Could you edit your commit message to include the required Signed-off-by line?

PR #4678, opened 2026-04-29
Fix security issue with lxc-user-nic and OpenVswitch networks

stgraber:


stgraber:
https://github.com/lxc/lxc/security/advisories/GHSA-3m9j-g9gc-vcvq

PR #4683, opened 2026-05-09
simplify cleanup in storage_copy

hallyn:
struct lxc_rootfs has an lxc_storage pointer, and vice versa. put_lxc_rootfs(r) will free the storage it points to, but freeing the storage won't free the lxc_rootfs.

By setting new_rootfs->storage and calling put_lxc_rootfs(new_rootfs), which we weren't doing, we can avoid having to manually close the dfd_idmapped.


hallyn:
@brauner I'd appreciate a review here to make sure I've got the struct relations right in my head.

hallyn:
After looking some more, I'm confident this is correct.

PR #4685, opened 2026-05-19
lxc/caps: fix cap_to_text() memory cleanup

### LINKED ISSUE #4670: lxc_ambient_caps_up uses free() on libcap-allocated memory
# lxc_ambient_caps_up uses free() on libcap-allocated memory

## Required info
- Ubuntu 24.04.4 LTS
  - I have also reproduced this on Ubuntu Jammy
- LXC version: 5.0.3
  - I have also reproduced this on LXC6 
- `Linux ip-172-31-50-223 6.17.0-1007-aws #7~24.04.1-Ubuntu SMP Thu Jan 22 21:04:49 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux`
- cat /proc/self/cgroup
  ```
  > cat /proc/self/cgroup
  0::/user.slice/user-1000.slice/session-21.scope
  ```
- lxc-checkconfig:
  ```
  LXC version 5.0.3
  Kernel configuration not found at /proc/config.gz; searching...
  Kernel configuration found at /boot/config-6.17.0-1007-aws

  --- Namespaces ---
  Namespaces: enabled
  Utsname namespace: enabled
  Ipc namespace: enabled
  Pid namespace: enabled
  User namespace: enabled
  Network namespace: enabled

  --- Control groups ---
  Cgroups: enabled
  Cgroup namespace: enabled
  Cgroup v1 mount points: 
  Cgroup v2 mount points: 
   - /sys/fs/cgroup
  Cgroup device: enabled
  Cgroup sched: enabled
  Cgroup cpu account: enabled
  Cgroup memory controller: enabled
  Cgroup cpuset: enabled

  --- Misc ---
  Veth pair device: enabled, not loaded
  Macvlan: enabled, not loaded
  Vlan: enabled, loaded
  Bridges: enabled, loaded
  Advanced netfilter: enabled, loaded
  CONFIG_IP_NF_TARGET_MASQUERADE: enabled, not loaded
  CONFIG_IP6_NF_TARGET_MASQUERADE: enabled, not loaded
  CONFIG_NETFILTER_XT_TARGET_CHECKSUM: enabled, not loaded
  CONFIG_NETFILTER_XT_MATCH_COMMENT: enabled, not loaded
  FUSE (for use with lxcfs): enabled, not loaded

  --- Checkpoint/Restore ---
  checkpoint restore: enabled
  CONFIG_FHANDLE: enabled
  CONFIG_EVENTFD: enabled
  CONFIG_EPOLL: enabled
  CONFIG_UNIX_DIAG: enabled
  CONFIG_INET_DIAG: enabled
  CONFIG_PACKET_DIAG: enabled
  CONFIG_NETLINK_DIAG: enabled
  File capabilities: enabled

  Note : Before booting a new kernel, you can check its configuration
  usage : CONFIG=/path/to/config /usr/bin/lxc-checkconfig

  ```
- cat /proc/1/mounts
  ```
  root@ip-172-31-50-223:/home/ubuntu# cat /proc/1/mounts
  /dev/root / ext4 rw,relatime,discard,errors=remount-ro,commit=30 0 0
  devtmpfs /dev devtmpfs rw,nosuid,noexec,relatime,size=3920216k,nr_inodes=980054,mode=755,inode64 0 0
  proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
  sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
  securityfs /sys/kernel/security securityfs rw,nosuid,nodev,noexec,relatime 0 0
  tmpfs /dev/shm tmpfs rw,nosuid,nodev,inode64 0 0
  devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000 0 0
  tmpfs /run tmpfs rw,nosuid,nodev,size=1572768k,nr_inodes=819200,mode=755,inode64 0 0
  tmpfs /run/lock tmpfs rw,nosuid,nodev,noexec,relatime,size=5120k,inode64 0 0
  cgroup2 /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot 0 0
  none /sys/fs/pstore pstore rw,nosuid,nodev,noexec,relatime 0 0
  efivarfs /sys/firmware/efi/efivars efivarfs rw,nosuid,nodev,noexec,relatime 0 0
  bpf /sys/fs/bpf bpf rw,nosuid,nodev,noexec,relatime,mode=700 0 0
  systemd-1 /proc/sys/fs/binfmt_misc autofs rw,relatime,fd=32,pgrp=1,timeout=0,minproto=5,maxproto=5,direct,pipe_ino=3229 0 0
  hugetlbfs /dev/hugepages hugetlbfs rw,nosuid,nodev,relatime,pagesize=2M 0 0
  debugfs /sys/kernel/debug debugfs rw,nosuid,nodev,noexec,relatime 0 0
  mqueue /dev/mqueue mqueue rw,nosuid,nodev,noexec,relatime 0 0
  tracefs /sys/kernel/tracing tracefs rw,nosuid,nodev,noexec,relatime 0 0
  fusectl /sys/fs/fuse/connections fusectl rw,nosuid,nodev,noexec,relatime 0 0
  configfs /sys/kernel/config configfs rw,nosuid,nodev,noexec,relatime 0 0
  /dev/loop0 /snap/amazon-ssm-agent/12322 squashfs ro,nodev,relatime,errors=continue,threads=single 0 0
  /dev/loop2 /snap/snapd/25935 squashfs ro,nodev,relatime,errors=continue,threads=single 0 0
  /dev/loop1 /snap/core22/2339 squashfs ro,nodev,relatime,errors=continue,threads=single 0 0
  /dev/nvme1n1p16 /boot ext4 rw,relatime 0 0
  /dev/nvme1n1p15 /boot/efi vfat rw,relatime,fmask=0077,dmask=0077,codepage=437,iocharset=iso8859-1,shortname=mixed,errors=remount-ro 0 0
  binfmt_misc /proc/sys/fs/binfmt_misc binfmt_misc rw,nosuid,nodev,noexec,relatime 0 0
  tmpfs /run/qemu tmpfs rw,nosuid,nodev,relatime,mode=755,inode64 0 0
  tmpfs /run/user/1000 tmpfs rw,nosuid,nodev,relatime,size=786380k,nr_inodes=196595,mode=700,uid=1000,gid=1000,inode64 0 0
  ```
## Issue Description
I work on a product that uses LXC for container execution and during the development of some new functionality I've run into an issue with container starting and id remapping. The issue appears as a failed container start with a call to `free()` failing. This occurs somewhat low percentage of the time, roughly 1-2% of runs, when starting containers with user namespace mappings. Specifically it seems like mapping a container uid into the 0-65535 range on the host can trigger the issue, while remapping a uid into a high range (e.g. 400000) does not. Running a loop of container starts will typically hit the crash within 200, though often under 100, iterations.

Here is an example of the error message that occurs when this is triggered:

```
free(): invalid pointer
lxc-start: bug-repro-10046-17: ../src/lxc/sync.c: sync_wait: 25 Connection reset by peer - Sync wait failure
lxc-start: bug-repro-10046-17: ../src/lxc/start.c: __lxc_start: 2107 Failed to spawn container "bug-repro-10046-17"
```

> [!NOTE]
> Disclosure: This script was built by an attended AI assistant
 
I've included a reproduction script, `repro.sh`, I've found helpful for validating this behavior:

```
# repro.sh accepts an optional argument to pre-existing rootfs
./repro.sh
/tmp/lxc-bug-repro/run-test.sh
```

I've been running the script as root on a test EC2 instance. Running as non-root adds more complexity. It does the following:
- Checks `/etc/subuid` and `/etc/subgid` for the existence of necessary mappings to reproduce
- Makes a temp dir: `/tmp/lxc-bug-repro`
- If a rootfs path was not provided, runs `debootstrap --variant=minbase noble /tmp/lxc-bug-repro/rootfs` to create one
- Generates an LXC config `/tmp/lxc-bug-repro/config`
- Generates a run script `/tmp/lxc-bug-repro/run-test.sh` (Accepts an argument for number of runs, defaults to 200)


Minimal config that reproduces. Remapping into the original host range (0-65535) seems to be the trigger:

```
lxc.idmap = u 0 0 1000
lxc.idmap = g 0 0 1000

lxc.rootfs.path = dir:/tmp/lxc-bug-repro/rootfs
lxc.uts.name = bug-repro
lxc.autodev = 1
lxc.pty.max = 1024
lxc.mount.auto = proc:mixed sys:ro
lxc.init.cmd = /bin/true
```

If I instead add the necessary ids to `/etc/subuid` and `/etc/subgid` to for the following high-range remap the issue does not seem to trigger

```
# Mapping 0-999 to 500000-500999 (Does not trigger)

lxc.idmap = u 0 500000 1000
lxc.idmap = g 0 500000 1000
```

*I should note that the actual config I am attempting to use in prod is a bit more coherent than map 0-999 to 0-999, but this very simplified case also reproduces the bug.*

Here's what I think is going on:

In [src/lxc/caps.c](https://github.com/lxc/lxc/blob/cb8e38aca27a23964941f0f011a8919aab8bebab/src/lxc/caps.c#L88), the function `lxc_ambient_caps_up()` uses `__do_free` for memory returned by `cap_to_text()`. This causes `free()` to be called on memory that was allocated by libcap, which should be freed with `cap_free()`.

The __do_free: [line 91](https://github.com/lxc/lxc/blob/cb8e38aca27a23964941f0f011a8919aab8bebab/src/lxc/caps.c#L91):

```c
__do_free char *cap_names = NULL;
```

The `__do_free` cleanup attribute causes `free()` to be called when the variable goes out of scope. However, `cap_names` is assigned from `cap_to_text()` on [line 133](https://github.com/lxc/lxc/blob/cb8e38aca27a23964941f0f011a8919aab8bebab/src/lxc/caps.c#L133):

```c
cap_names = cap_to_text(caps, NULL);
```

The libcap [documentation](https://linux.die.net/man/3/cap_to_text#:~:text=When%20the%20capability%20state%20in%20working%20storage%20is%20no%20longer%20required%2C%20the%20caller%20should%20free%20any%20releasable%20memory%20by%20calling%20cap_free()%20with%20the%20returned%20string%20pointer%20as%20an%20argument.) specifies that memory returned by `cap_to_text()` must be freed with `cap_free()`, not `free()`. Calling `free()` on this memory can result in a crash.

The trace log shows that `lxc_ambient_caps_up` completes successfully (the TRACE on line 137 prints), then the child process crashes shortly after during a subsequent memory operation, causing the parent to see "Connection reset by peer" on the sync socket.

## Debugging

I have tried making the following changes and have found them stop the issue from occurring. I have not yet tried to create any kind of unit or functional test in lxc/lxc for this.

Change line 91 from:
```c
__do_free char *cap_names = NULL;
```

to:
```c
char *cap_names = NULL;
```

And add before the `return 0` on line 138:
```c
cap_free(cap_names);
```

I've done some additional debugging and testing and I have discovered some additional confusion. In theory we shouldn't really even hit this code but it seems like occasionally getuid() returns an unexpected value. I added some additional logging that looks like:

```                                                                                                                                                        
  uid_t uid, euid;                                                                                                                                                              
                                                                                                                                                                                                                            
  uid = getuid();
  euid = geteuid();                                                                                                                                                             
                                                                                                                                                                                
  TRACE("CAPS_DEBUG: lxc_ambient_caps_up() entered - uid=%d euid=%d", uid, euid);                                                                                               
                                                                                                                                                                                
  if (!uid || euid) {                                                                                                                                                           
      TRACE("CAPS_DEBUG: returning early - !uid=%d euid=%d", !uid, euid);                                                                                                     
      return 0;                                                                                                                                                                 
  }                                                                                                                                                                             
                                                                                                                                                                                
  TRACE("CAPS_DEBUG: past uid check (uid=%d euid=%d) - will execute buggy code path", uid, euid);
```

I would normally expect the conditional to pass and use to `return 0` in my specific test case, but when running my reproduction I occasionally see the behavior change and then trigger this bug:

```
# Run Output
  free(): invalid pointer                                                                                                                                                       
  lxc-start: bug-repro-189248-62: ../src/lxc/sync.c: sync_wait: 25 Connection reset by peer - Sync wait failure                                                               
  lxc-start: bug-repro-189248-62: ../src/lxc/start.c: __lxc_start: 2107 Failed to spawn container "bug-repro-189248-62" 

# trace.log
     lxc-start bug-repro-189248-56 20260416195210.479 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:104 - CAPS_DEBUG: returning early - !uid=1 euid=0
     lxc-start bug-repro-189248-57 20260416195210.486 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:101 - CAPS_DEBUG: lxc_ambient_caps_up() entered - uid=0 euid=0
     lxc-start bug-repro-189248-57 20260416195210.486 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:104 - CAPS_DEBUG: returning early - !uid=1 euid=0
     lxc-start bug-repro-189248-58 20260416195210.493 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:101 - CAPS_DEBUG: lxc_ambient_caps_up() entered - uid=0 euid=0
     lxc-start bug-repro-189248-58 20260416195210.493 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:104 - CAPS_DEBUG: returning early - !uid=1 euid=0
     lxc-start bug-repro-189248-59 20260416195210.500 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:101 - CAPS_DEBUG: lxc_ambient_caps_up() entered - uid=0 euid=0
     lxc-start bug-repro-189248-59 20260416195210.500 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:104 - CAPS_DEBUG: returning early - !uid=1 euid=0
     lxc-start bug-repro-189248-60 20260416195210.508 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:101 - CAPS_DEBUG: lxc_ambient_caps_up() entered - uid=0 euid=0
     lxc-start bug-repro-189248-60 20260416195210.508 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:104 - CAPS_DEBUG: returning early - !uid=1 euid=0
     lxc-start bug-repro-189248-61 20260416195210.515 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:101 - CAPS_DEBUG: lxc_ambient_caps_up() entered - uid=0 euid=0
     lxc-start bug-repro-189248-61 20260416195210.515 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:104 - CAPS_DEBUG: returning early - !uid=1 euid=0
     lxc-start bug-repro-189248-62 20260416195210.522 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:101 - CAPS_DEBUG: lxc_ambient_caps_up() entered - uid=65534 euid=0
     lxc-start bug-repro-189248-62 20260416195210.522 TRACE    caps - ../src/lxc/caps.c:lxc_ambient_caps_up:108 - CAPS_DEBUG: past uid check (uid=65534 euid=0) - will execute buggy codepath
```

I think these are technically separate bugs but one seems to enable the other.

[repro.sh](https://github.com/user-attachments/files/26760853/repro.sh)
[runs.log](https://github.com/user-attachments/files/26760854/runs.log)
[trace.log](https://github.com/user-attachments/files/26760855/trace.log)

Same files in [gist](https://gist.github.com/thomas-holmes/5b85f60584fce09910d6c8baeba87dcb)

DreamConnected:
Fixed #4670

### DIFF
--- a/src/lxc/caps.c
+++ b/src/lxc/caps.c
@@ -88,7 +88,7 @@ int lxc_caps_up(void)
 int lxc_ambient_caps_up(void)
 {
 	call_cleaner(cap_free) cap_t caps = NULL;
-	__do_free char *cap_names = NULL;
+	char *cap_names = NULL;
 	int ret;
 	cap_value_t cap;
 	cap_value_t last_cap = CAP_LAST_CAP;
@@ -135,6 +135,7 @@ int lxc_ambient_caps_up(void)
 		return log_warn_errno(0, errno, "Failed to convert capabilities %d", cap);
 
 	TRACE("Raised %s in inheritable and ambient capability set", cap_names);
+	cap_free(cap_names);
 	return 0;
 }
 


mihalicyn (approved):
Thanks, @DreamConnected!

PR #4686, opened 2026-05-19
lxc-autostart: fix exit code logic for partial failure

### LINKED ISSUE #4662: lxc-autostart should return proper exit code
Right now lxc-autostart counts every `lxc.start.auto == 0` as success and thus even if all containers with `lxc.start.auto == 1` fail to start, still returns `2` (partially failed) instead of `1` (failed). So  https://github.com/lxc/lxc/blob/stable-6.0/src/lxc/tools/lxc_autostart.c#L509 should be changed and take "disabled" containers into account.

DreamConnected:
Try to fix #4662 

The previous logic compared 'failed' with 'count' (total containers) to determine if it was a total failure. This is incorrect when only some containers have `lxc.start.auto=1` set.
﻿
Now uses 'attempted' which correctly tracks the number of containers that were actually processed.


mihalicyn (approved):
Thanks, @DreamConnected!

PR #4689, opened 2026-05-26
lxc/start: fix possible race condition sockets leak

GermanAizek:
I think can protect yourself from an socket data leak in advance.


mihalicyn (changes requested):
We already have `__do_close` for `data_sock0` and `data_sock1`.

PR #4690, opened 2026-05-26
lxc/start: sanitize user lxcpath string parameter

GermanAizek:
Simple protection against symbolic links.

### DIFF
--- a/src/lxc/start.c
+++ b/src/lxc/start.c
@@ -23,6 +23,7 @@
 #include <sys/un.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <limits.h>
 
 #include "lxc.h"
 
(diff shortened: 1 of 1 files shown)


src/lxc/start.c
```diff
@@ -1841,6 +1833,13 @@ static int lxc_spawn(struct lxc_handler *handler)
 	const char *name = handler->name;
 	struct lxc_conf *conf = handler->conf;
 	struct cgroup_ops *cgroup_ops = handler->cgroup_ops;
+	char real_lxcpath[PATH_MAX];
+	
+	if (!realpath(handler->lxcpath, real_lxcpath)) {
+		SYSERROR("Failed to resolve realpath for lxcpath");
+		return -1;
+	}
+	handler->lxcpath = real_lxcpath;
```
mihalicyn, line 1842: you are writing an address that belongs to the stack to the object with a lifetime that is much longer than a function call...

mihalicyn (changes requested):
Can you explain in detail why this change is needed?

PR #4693, opened 2026-05-27
lxc-net: Even more random IPv6 subnet

steelman:
Generate a stable IPv6 prefix based on the content of /etc/machin-id. This way the subnet on every machine will be different but stable between reboots and software updates.

Note: Although in section 3.1 of RFC 4193[1] the Prefix is declared to be FC00::/7, the L bit declared to be 1. Hence ULA prefixes always start with FD.

[1] https://datatracker.ietf.org/doc/html/rfc4193#section-3.1

Follow-up-to: 31012d49a ("lxc-net: Replace random IPv6 subnet")

### DIFF
--- a/config/init/common/lxc-net.in
+++ b/config/init/common/lxc-net.in
@@ -21,10 +21,11 @@ LXC_DOMAIN=""
 LXC_USE_NFT="true"
 
 # IPv6 connectivity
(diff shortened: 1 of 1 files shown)


stgraber:
There is no guarantee that /etc/machine-id will exist on a given system, we support more than just systemd systems.

steelman:
If the file is missing a valid prefix is still generated (`echo LXC_IPV6_PREFIX | sha256sum`) it's just not random anymore, which is how things work today.

stgraber (approved):

stgraber:
That appears to be breaking all the tests.

stgraber:
Thanks!

PR #4695, opened 2026-05-29
lxc/criu: restore cgroup limits and freezer on restore

### LINKED ISSUE #4643: Containers restored from `lxc-checkpoint` cannot subsequently be checkpointed or frozen
### Background Context
I've built a system that does periodic checkpoints of user-interactive LXCs (plus coordinated disk snapshots), to support duplicating user-interactive workspaces including their running program state (by restoring the checkpoint with disk snapshot cloning) into new LXCs.

I had this fully working on 2022-era Ubuntu/LXC/CRIU, but a lot has changed in Linux since (inc. cgroupsv2). On contemporary 2026 LXC/CRIU, this is only partially working: first generation LXCs can be checkpointed and restored into second generations, but second generations cannot be checkpointed (or frozen).

I would greatly appreciate help getting to the bottom of this.

### Symptoms/Reproduction Steps:
1. `lxc-checkpoint -n [first-generation] -D [first-dump-dir] --action-script [post-dump/disk-snapshot-hook]` succeeds without issue.

The `dump.log`:
```
Warn  (compel/arch/x86/src/lib/infect.c:418): Will restore 5224 with interrupted system call
Warn  (criu/net.c:534): net: The eth0 veth is in an external netns
```
2. (following config creation and disk snapshot cloning) `lxc-checkpoint -n [second-generation] -r -D [first-dump-dir]` succeeds and generates a running LXC named [second-generation], with functional/responsive copied running processes from [first-generation]. To be clear: [first-generation] and [second-generation] are separate LXC names; I am _not_ attempting to restore [first-generation] to an earlier version of itself.

The `restore.log`:
```
Warn  (criu/cr-restore.c:1043): Set CLONE_PARENT | CLONE_NEWPID but it might cause restore problem,because not all kernels support such clone flags combinations!
RTNETLINK answers: File exists
RTNETLINK answers: File exists
RTNETLINK answers: File exists
RTNETLINK answers: File exists
```
3. Invoking `lxc-checkpoint -n [second-generation] -D [second-dump-dir] ...` always fails. No `dump.log` is created, but it outputs:
```
lxc-checkpoint: [second-generation]: criu.c: exec_criu: 402 No such file or directory - Failed getting freezer path
lxc-checkpoint: [second-generation]: criu.c: do_dump: 1265 No such file or directory - Failed to execute criu
lxc-checkpoint: [second-generation]: criu.c: do_dump: 1297 dump failed with 1
lxc-checkpoint: [second-generation]: criu.c: do_dump: 1311 criu output: 
lxc-checkpoint: [second-generation]: tools/lxc_checkpoint.c: checkpoint: 182 Checkpointing [second-generation] failed
```
4. Since the error message mentioned "freezer path", I tried playing with `lxc-freeze -n [second-generation]` and discovered it has no effect on the [second-generation] container (but, no output and exit status=0). `lxc-freeze -n [first-generation]` works as expected (processes frozen/unresponsive).
5. This made me suspect a cgroups problem, so I ran:
```
A=[first-generation]
B=[second-generation]
for n in $A $B; do
  echo "== $n =="
  PID=$(lxc-info -n "$n" -pH || true)
  echo "PID=$PID"
  [ -n "$PID" ] && cat /proc/$PID/cgroup || true
done
== [first-generation] ==
PID=5213
0::/lxc.payload.[first-generation]
== [second-generation] ==
PID=5421
0::/lxc.payload.[first-generation]
```
_Both [first-generation] and [second-generation] are in [first-generation]'s cgroup_, which doesn't look right at all.

I've tried adding `manage-cgroups none` to `/etc/criu/default.conf`, but it doesn't change this behavior.

### Host Environment
```
# uname -a
Linux <host> 5.15.0-170-generic #180-Ubuntu SMP Fri Jan 9 16:10:31 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
# lxc-info --version
5.0.0
# criu --version
Version: 4.2
# cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.5 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy

# stat -fc %T /sys/fs/cgroup 
cgroup2fs
# mount | grep -E 'cgroup2|/sys/fs/cgroup'
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime)
```

### Container Environment
The container is based off the LXC ubuntu-jammy template (downloaded `lxc-create -t download ...`), but modified to run minimal `tini` init (v0.19.0) and a custom minimal SSH server as PID 2 for interactive sessions. Contemporary systemd init created too many checkpointing problems (between nested UTS namespaces, etc.), so I simplified.
```
$ ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 06:05 ?        00:00:00 /sbin/tini -- /sbin/chronitond
root           2       1  0 06:05 ?        00:00:01 /sbin/chronitond
root           9       2  0 06:05 pts/0    00:00:01 /bin/login -p -f
ubuntu        19       9  0 06:05 pts/0    00:00:00 -bash
ubuntu        32      19  0 06:18 pts/0    00:00:00 ps -ef
```

Thanks in advance for help and I'm happy to provide any additional information/details.

emirbuljubasic:
When performing `lxc-checkpoint -r` there were two bugs I observed.

1. The freezer cgroup wasn't restored properly, causing subsequent `lxc-checkpoint` calls to fail during the checkpoint faze due to missing freezer cgroup. Meaning after restoring a container, it needs to be shutdown before checkpointing it again.
2. cgroup limits were not applied after restore, for example, if `/sys/fs/cgroup/memory.max` was set to `2147483648`, after restoring it would just show the value `max`.

Fixes #4643 

This issue was also present for LXC 6.0.5 (where I originally wrote the patch for). If you plan to release hotfixes for 6.x let me know. I can submit another PR on that branch, since one more line needs to be added for `setup_limits_legacy`, which is missing here since 7.x has dropped cgroup v1 support.

I have also observed one more issue, but I assume that has nothing to do with this project. However, I would still like to leave a question here. If the container is using OpenRC as its init system. After restoring, PID 1 (init) will spike the CPU to 100%. Some errors due to epoll. Mainly what happens is that `/run/openrc/init.ctl` is RW before checkpoint. After checkpointing in `files.img` (inside the dump), it is set to RO.

I have manually fixed this in my workflow by:
1. walking `files.img`
2. setting `/run/openrc/init.ctl` to RW

Then after performing restore, container behaves normally, no CPU pinning.

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -939,6 +939,21 @@ static void do_restore(struct lxc_container *c, int status_pipe, struct migrate_
 		goto out_fini_handler;
 	}
 
+	if (!cgroup_ops->payload_delegate_controllers(cgroup_ops)) {
+		ERROR("Failed to delegate controllers to payload cgroup");
+		goto out_fini_handler;
+	}
+
+	if (!cgroup_ops->setup_limits(cgroup_ops, handler)) {
+		ERROR("Failed to setup cgroup limits");
+		goto out_fini_handler;
+	}
+
+	if (!cgroup_ops->chown(cgroup_ops, handler->conf))
+		goto out_fini_handler;
+
+	cgroup_ops->finalize(cgroup_ops);
+
 	if (!restore_net_info(c)) {
 		ERROR("failed restoring network info");
 		goto out_fini_handler;


mihalicyn (approved):
LGTM

Thanks, @emirbuljubasic!

mihalicyn:
>After restoring, PID 1 (init) will spike the CPU to 100%. Some errors due to epoll. Mainly what happens is that /run/openrc/init.ctl is RW before checkpoint. After checkpointing in files.img (inside the dump), it is set to RO.

I've tried to run a Gentoo container, and it looks like this file is now located in `/run/initctl` and:
```
stat /run/initctl 
  File: /run/initctl
  Size: 0         	Blocks: 0          IO Block: 4096   fifo
Device: 0,117	Inode: 30          Links: 1
Access: (0600/prw-------)  Uid: (    0/    root)   Gid: (    0/    root)
```

it is not regular file, but FIFO. What you are describing looks like a bug in CRIU. Please, can you report this to us in CRIU project https://github.com/checkpoint-restore/criu/issues (ideally, with full reproducer from creating an LXC container, downloading an image and up to `lxc-checkpoint` invokation) and ping me in the issue there? I'll take a look into this.

emirbuljubasic:
> > After restoring, PID 1 (init) will spike the CPU to 100%. Some errors due to epoll. Mainly what happens is that /run/openrc/init.ctl is RW before checkpoint. After checkpointing in files.img (inside the dump), it is set to RO.
> 
> I've tried to run a Gentoo container, and it looks like this file is now located in `/run/initctl` and:
> 
> ```
> stat /run/initctl 
>   File: /run/initctl
>   Size: 0         	Blocks: 0          IO Block: 4096   fifo
> Device: 0,117	Inode: 30          Links: 1
> Access: (0600/prw-------)  Uid: (    0/    root)   Gid: (    0/    root)
> ```
> 
> it is not regular file, but FIFO. What you are describing looks like a bug in CRIU. Please, can you report this to us in CRIU project https://github.com/checkpoint-restore/criu/issues (ideally, with full reproducer from creating an LXC container, downloading an image and up to `lxc-checkpoint` invokation) and ping me in the issue there? I'll take a look into this.

@mihalicyn I will do that.

I observed the bug in an Artix Linux container. That might be why the path differs for you and me, but I will check this out for Gentoo containers as well.

PR #4699, opened 2026-06-08
criu: use strnprintf for the --cgroup-root argument

netliomax25-code:
1. exec_criu() builds the --cgroup-root argument with sprintf() into a fixed char buf[4096] from the joined controller list and the container's cgroup path.
2. The `ret >= sizeof(buf)` guard runs only after sprintf() has written, so a cgroup path long enough to push controllers + ":" + path past 4096 overflows the stack buffer before the check can reject it.

Switched both calls to strnprintf(buf, sizeof(buf), ...), already used for this buffer elsewhere in the same function, so over-long input truncates and returns < 0 and the existing guard rejects it before any out-of-bounds write.


netliomax25-code:
Pushed the missing DCO sign-off for the Signed-off-by check; the amended commit is otherwise unchanged.

PR #4701, opened 2026-06-09
commands: validate string in lxc_cmd_get_config_item_callback

netliomax25-code:
1. lxc_cmd_get_config_item_callback() passes the raw req->data to lxc_get_config(), but the command server stores req->data as exactly datalen bytes read off the socket, so a client can send data that is not NUL-terminated.
2. match_config_item() then runs strcmp()/strncmp() against the config names and reads past the end of the heap buffer once the bytes match a name prefix (datalen=1, data="l" is enough).

Added a datalen check and reused validate_string_request(), the same guard the get_cgroup callback already applies, so unterminated or empty input is rejected before the lookup.

### DIFF
--- a/src/lxc/commands.c
+++ b/src/lxc/commands.c
@@ -1028,8 +1028,19 @@ static int lxc_cmd_get_config_item_callback(int fd, struct lxc_cmd_req *req,
 	int cilen;
 	struct lxc_config_t *item;
 	struct lxc_cmd_rsp rsp;
+	ssize_t ret;
 
 	memset(&rsp, 0, sizeof(rsp));
+
+	if (req->datalen <= 0) {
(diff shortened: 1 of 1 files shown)


netliomax25-code:
gentle ping

mihalicyn:
Hi @netliomax25-code,

please read through https://github.com/lxc/lxc/blob/main/CONTRIBUTING.md

You need to sign-off your commit ;-)

mihalicyn:
I've reviewed the change itself and if looks correct to me.

netliomax25-code:
Done, amended the commit with the sign-off and force-pushed. Same change otherwise, and thanks for taking a look at the fix itself.

mihalicyn (approved):
LGTM, thanks!

PR #4711, opened 2026-07-18
tree-wide: fix const-correctness issues exposed by glibc 2.43

### LINKED ISSUE #4710: Const propagation build failure with latest glibc and clang
# Summary

glibc 2.43 implements `strchr(3)` et co as C23-style _Generic macros which propagate the const qualifier of the input string to the return type: calling them on a `const char *` now returns `const char *` instead of `char *`.

Several call sites in the tree assign such results to plain `char *` variables. Since meson.build adds `-Werror=incompatible-pointer-types` unconditionally, every clang build against glibc 2.43 fails out of the box:

```
../src/lxc/confile.c:2690:4: error: assigning to 'char *' from 'const char *'
    discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
 2690 |         p = strchr(line, '=');
      |             ^ ~~~~~~~~~~~~~~~~~
```

GCC builds are not affected because GCC diagnoses this as `-Wdiscarded-qualifiers`, which is not part of `-Wincompatible-pointer-types` and is not promoted to an error, so this only affects Clang builds.

Affected sites on current main:

 * `src/lxc/confile.c:2690` (`update_hwaddr()`)
 * `src/lxc/confile.c:4221` (`get_config_hooks()`)
 * `src/lxc/confile_utils.c:912` (`lxc_inherit_namespace()`)
 * `src/lxc/pam/pam_cgfs.c:365` (`is_lxcfs()`)
 * `src/lxc/pam/pam_cgfs.c:1917` (`cgv1_handle_cpuset_hierarchy()`)
 * `src/lxc/storage/nbd.c:294` (`nbd_get_partition()`)
 * `src/lxc/storage/zfs.c:464` (`zfs_clonepaths()`)

This is the same glibc 2.43 fallout family as the already-merged meson detection fixes 5e68a7a63 and 540f9e2bc, which addressed the `openat2()`/`struct open_how` side of the uprev, but not the string.h const propagation.

# Steps to reproduce

 1. Use a toolchain with glibc 2.43 and clang (e.g. any recent Gentoo/ChromeOS SDK cross toolchain, that's what I use).
 2. `meson setup build && meson compile -C build`
 3. Compilation of `liblxc` fails at the sites listed above.

# Proposed solution

I have a fix ready and will attach a PR: it constifies the result pointers where they are only read and casts back to `char *` at the two sites which intentionally write the result into caller-owned writable storage.

10ne1:
glibc 2.43 implements strchr(3), strrchr(3) and strstr(3) as C23-style _Generic macros which propagate the const qualifier of the input string to the return type.

For example, a 'const char *' input now produces 'const char *' output types and vice-versa (non-const -> non-const).

Building with clang and -Werror=incompatible-pointer-types fails:

  src/lxc/confile.c:2690:4: error: assigning to 'char *' from
      'const char *' discards qualifiers
      [-Werror,-Wincompatible-pointer-types-discards-qualifiers]

Constify the result pointers where they are only read and cast the result back to char* where the code intentionally writes through it into caller-owned writable storage.

This is just a build fix, the runtime behavior is unchanged.

Fixes: #4710

### DIFF
--- a/src/lxc/confile.c
+++ b/src/lxc/confile.c
@@ -2675,7 +2675,7 @@ static int set_config_console_size(const char *key, const char *value,
  * lxc.include.
  * 'x' and 'X' are substituted in-place.
  */
-static void update_hwaddr(const char *line)
+static void update_hwaddr(char *line)
 {
 	char *p;
 
@@ -2701,7 +2701,7 @@ static void update_hwaddr(const char *line)
 	rand_complete_hwaddr(p);
 }
 
-int append_unexp_config_line(const char *line, struct lxc_conf *conf)
+int append_unexp_config_line(char *line, struct lxc_conf *conf)
 {
 	size_t linelen;
 	size_t len = conf->unexpanded_len;
@@ -4213,7 +4213,7 @@ static int get_config_uts_name(const char *key, char *retv, int inlen,
 static int get_config_hooks(const char *key, char *retv, int inlen,
 			    struct lxc_conf *c, void *data)
 {
-	char *subkey;
+	const char *subkey;
 	int len, fulllen = 0, found = -1;
 	struct string_entry *entry;
 	int i;
--- a/src/lxc/pam/pam_cgfs.c
+++ b/src/lxc/pam/pam_cgfs.c
@@ -362,7 +362,7 @@ static char *copy_to_eol(char *s)
 /* Check if given entry under /proc/<pid>/mountinfo is a fuse.lxcfs mount. */
 static bool is_lxcfs(const char *line)
 {
-	char *p = strstr(line, " - ");
+	const char *p = strstr(line, " - ");
 	if (!p)
 		return false;
 
@@ -1914,7 +1914,7 @@ static bool cgv1_handle_cpuset_hierarchy(struct cgv1_hierarchy *h,
 
 	if (*cgroup == '/')
 		cgroup++;
-	slash = strchr(cgroup, '/');
+	slash = (char *)strchr(cgroup, '/');
 	if (slash)
 		*slash = '\0';
 
--- a/src/lxc/confile.h
+++ b/src/lxc/confile.h
@@ -75,7 +75,7 @@ __hidden extern int lxc_list_net(struct lxc_conf *c, const char *key, char *retv
 
 __hidden extern int lxc_config_read(const char *file, struct lxc_conf *conf, bool from_include);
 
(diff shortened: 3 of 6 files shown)


src/lxc/confile.c
```diff
@@ -2687,7 +2687,7 @@ static void update_hwaddr(const char *line)
 		return;
 
 	/* Let config_net_hwaddr raise the error. */
-	p = strchr(line, '=');
+	p = (char *)strchr(line, '=');
```
mihalicyn: It seems to me, that instead we should change `update_hwaddr(const char *line)` to `update_hwaddr(char *line)` here, because in fact, later `rand_complete_hwaddr(p)` call modifies memory pointed by `line`. WDYT?

src/lxc/pam/pam_cgfs.c
```diff
@@ -1914,7 +1914,7 @@ static bool cgv1_handle_cpuset_hierarchy(struct cgv1_hierarchy *h,
 
 	if (*cgroup == '/')
 		cgroup++;
-	slash = strchr(cgroup, '/');
+	slash = (char *)strchr(cgroup, '/');
```
mihalicyn, line 1917: A few lines below we do `*slash = '\0';` and modify memory behind `cgroup` while it is declared as `const char *cgroup`. I traversed a `cgv1_handle_cpuset_hierarchy` callstack up and stopped in `handle_login()` function, where:
```
	char cg[PATH_MAX];
...
		if (!cgv1_create(cg, uid, gid, &existed)) { // << HERE cg pointer is safe to be written to
                                                                        // while cgv1_create prototype says const char *cgroup
```

What do you think about this?

Another point here is that this code is cgroup1 code and we will definetely nuke it soon ;-) So, maybe, we can just go with this simple patch, cause in fact there is no real issue with memory handling, it's just `const` is being misused on our old code.

mihalicyn:
Hi Adrian,

thank you very much for working on this.

I've left a few small comments. But overall, this LGTM ;-)

src/lxc/confile.c
10ne1: I think this works, yes.

src/lxc/pam/pam_cgfs.c
10ne1: Agreed, since it's cgroup1 that's on its way out and the buffer is writable, I'll keep the minimal cast here rather than churn. Thanks!

mihalicyn (approved):
LGTM

Thanks! ;-)

PR #4712, opened 2026-07-22
criu: Finalize cgroups after restore

fivetime:
## Summary

Finalize the container cgroup setup after a successful CRIU restore.

Normal startup finalizes cgroup discovery after the payload cgroup exists. The CRIU restore path marked the container as running but skipped that finalization. As a result, utility cgroup v2 controllers, including freezer, may not be available for a subsequent checkpoint or live migration of the restored container.

## Testing

- Built `liblxc.so.1.9.0` from the current `main` base in Alpine 3.24 using Meson and Ninja.

### DIFF
--- a/src/lxc/criu.c
+++ b/src/lxc/criu.c
@@ -1088,6 +1088,15 @@ static void do_restore(struct lxc_container *c, int status_pipe, struct migrate_
 					ERROR("error setting running state after restore");
 					goto out_fini_handler;
 				}
+
+				/*
+				 * Normal startup finalizes cgroup discovery after the
+				 * payload cgroup exists. Do the same after CRIU restore
+				 * so utility controllers such as cgroup v2 freezer are
+				 * available to a later checkpoint of this container.
+				 */
+				cgroup_ops->finalize(cgroup_ops);
+				TRACE("Finished setting up restored cgroups");
 			}
 		} else {
 			ERROR("CRIU was killed with signal %d", WTERMSIG(status));


src/lxc/criu.c
```diff
@@ -1088,6 +1088,15 @@ static void do_restore(struct lxc_container *c, int status_pipe, struct migrate_
 					ERROR("error setting running state after restore");
 					goto out_fini_handler;
 				}
+
+				/*
```
mihalicyn, line 1092: shouldn't we do this *before* `lxc_set_state(c->name, handler, RUNNING)`?

mihalicyn:
Hi @fivetime,

thank you for preparing this PR!

> Normal startup finalizes cgroup discovery after the payload cgroup exists. The CRIU restore path marked the container as running but skipped that finalization. As a result, utility cgroup v2 controllers, including freezer, may not be available for a subsequent checkpoint or live migration of the restored container.

yeah, I agree. AFAIU the problem is only with `freezer` controller, right?

P.S. Please, can you share how did you find this issue?

fivetime:
Hi @mihalicyn,

Yes, the concrete failure I hit is with the cgroup v2 freezer utility controller.

The reason I used `finalize()` rather than a freezer-only helper is that this mirrors the normal startup path: after the payload cgroup exists, `start.c` calls `cgroup_ops->finalize(cgroup_ops)`, and `cgfsng_finalize()` then discovers whether the unified hierarchy supports the freezer controller. The CRIU restore path also creates/attaches the restored payload cgroup and marks the container as running, but it never ran that same finalization step. So after a successful restore, a later checkpoint of that same container could fail because LXC still had not marked freezer as available.

How I found it: I was testing repeated stateful migration/checkpoint-restore of Incus system containers on cgroup v2. The first restore completed, but a second checkpoint/live migration of the restored container failed on the freezer path. Comparing the normal start path with the CRIU restore path showed that normal startup finalizes cgroup discovery after the container is running, while restore skipped it. Adding the same finalization after `lxc_set_state(..., RUNNING)` fixed the second checkpoint/migration case in that test.

So the observed problem is freezer, but the patch intentionally reuses the existing lifecycle hook instead of adding CRIU-specific freezer handling.

emirbuljubasic:
@fivetime @mihalicyn Freezer isn't the only issue here, cgroup limits are also not restored, take a look at the values in `/sys/fs/cgroup/lxc.<container>.payload/*.{max,high} ` after performing a restore.

Also, this PR is a duplicate of an existing one: #4695

mihalicyn:
Thanks @emirbuljubasic.

@fivetime If this is a duplicate let's then close this one and focus on #4695.

PR #4713, opened 2026-07-26
Replace heap alloc with stack VLA for temp buffer in lxc_rsync_exec

GermanAizek:
Refs:
- https://stackoverflow.com/a/56798629
- https://www.bitesizedengineering.com/p/why-is-stack-faster-than-heap

### DIFF
--- a/src/lxc/storage/rsync.c
+++ b/src/lxc/storage/rsync.c
@@ -46,21 +46,22 @@ int lxc_rsync_exec(const char *src, const char *dest)
 	char *s;
 
 	l = strlen(src) + 2;
-	s = malloc(l);
-	if (!s)
-		return -1;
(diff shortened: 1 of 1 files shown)


paulocoghi:
Not a maintainer here, but just some doubts.

The `CODING_STYLE.md` says to not use VLA and to never allocate memory on the stack (rules 12 and 25).

I'm concerned that the VLA approach introduces an unbounded stack growth proportional to the path length, and this has the potential to create a security problem (including privileged execution, and I can explain more, if desired).

My understanding is that, with `malloc`, if the path is too long for the heap, it fails and the function returns `-1`. With VLA this safety if completely removed.

mihalicyn:
Hi @GermanAizek,

sorry but I'm not sure that this optimization makes any sense given that right after we have very costly `execlp()` syscall ;-)

PR #4716, opened 2026-08-19
Integrate WolfSSL give choice to user

GermanAizek:
OpenSSL has become too CPU resource‑intensive, and many MCU, SoC don’t have capability to use OpenSSL — that’s why other cryptographic libraries are being developed. For example, WolfSSL has lower memory consumption for a TLS connection, and it also scales best in terms of the number of threads.

References:
- https://news.ycombinator.com/item?id=43906194
- https://www.haproxy.com/blog/state-of-ssl-stacks
- https://blog.st.com/wolfssl/

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -141,7 +141,12 @@ want_commands = get_option('commands')
 want_capabilities = get_option('capabilities')
 want_apparmor = get_option('apparmor')
 want_openssl = get_option('openssl')
+want_wolfssl = get_option('wolfssl')
 want_selinux = get_option('selinux')
+
+if want_openssl and want_wolfssl
+    error('Cannot enable both OpenSSL and WolfSSL. Please choose one: -Dopenssl=true -Dwolfssl=false or -Dopenssl=false -Dwolfssl=true')
+endif
 want_oss_fuzz = get_option('oss-fuzz')
 want_seccomp = get_option('seccomp')
 want_spec = get_option('specfile')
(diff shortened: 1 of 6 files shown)


stgraber:
Hmm, none of the performance bits really matter given that we use it purely to generate a one time hash. So this is a bunch more logic and another test permutation for effectively zero benefit from my point of view.

@hallyn @brauner @mihalicyn

stgraber:
The only place we use this logic is to generate a comment in the configuration about the hash of the template which was used to generate a container.

This logic is almost never used given the vast majority of users use pre-built images these days and that logic is already completely optional, simplify skipping that comment if the system doesn't have openssl.

mihalicyn:
I agree with Stéphane. Only if you have a real use-case for this stuff @GermanAizek, we can consider this.

[IMHO begin]
In general, in Linux kernel, LXC, QEMU and in many other critically important open-source projects everything starts from a use-case, not from the code. You need to have some real world problem, then you analyze it, then you try to solve it without writing code and if this does not work => you write some code. Because code is not only a valuable asset (with AI being able to generate it I tend to think that not anymore), but a thing that you have to **maintain**.
[/IMHO end]

>OpenSSL has become too CPU resource‑intensive, and many MCU, SoC don’t have capability to use OpenSSL

so my question would be... are you trying to run LXC on MCU and stuck on _this specific_ issue?

stgraber:
Closing as the user can't possibly be stuck on this particular issue given we support building without libssl and the only change in that scenario is a missing comment line in a text file.

Issue #4719, opened 2026-08-25
keep capabilities if init process is not run as root


ElJeffe:
### Issue
When a container is configured to start the init cmd as a non-root user, all capabilities are dropped.
### Use case
We want to give capabilities to the container, but let the processes inside the container not run as root. This can be useful for instance to create a web server inside a container that listens on port 80 of the host.
### How to reproduce
Create a container
`lxc-create --name mycontainer2 --template download -- --dist alpine --release 3.23 --arch amd64`
modify the config with
```
# Network configuration
lxc.net.0.type = none

# keep caps to listen on port 90
lxc.cap.drop=
lxc.cap.keep=net_bind_service

# set init uid and gid to 1000
lxc.init.uid=1000
lxc.init.gid=1000

# listen on port 90
lxc.init.cmd=nc -l -p 90 localhost
```
This container stops immediately. When looking at the capabilities, one sees:
```
CapInh:	0000000000000000
CapPrm:	0000000000000000
CapEff:	0000000000000000
CapBnd:	0000000000000400
CapAmb:	0000000000000000
```

### proposed solution
Add an option to let processes with lxc.init.uid!=0 also keep capabilities

ElJeffe:
fixed in https://github.com/lxc/lxc/pull/4720

PR #4720, opened 2026-08-26
keep capabilities if init process is not run as root

ElJeffe:
Proposed solution for https://github.com/lxc/lxc/issues/4719

If the init process is not run as root, then all capabilities will be dropped.

This adds an option to let the container inherit the requested capabilities when the init process is not root:

lxc.cap.inheritance=1

### DIFF
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1368,7 +1368,7 @@ __noreturn static void do_attach(struct attach_payload *ap)
 		lxc_seccomp_close_notifier_fd(&conf->seccomp);
 	}
 
-	if (conf->cap_inheritance) {
+	if (conf->nonroot_keepcaps) {
 		ret = lxc_set_keepcaps();
 		if (ret < 0)
 			goto on_error;
--- a/src/lxc/attach.c
+++ b/src/lxc/attach.c
@@ -1368,8 +1368,21 @@ __noreturn static void do_attach(struct attach_payload *ap)
 		lxc_seccomp_close_notifier_fd(&conf->seccomp);
 	}
 
-	if (!lxc_switch_uid_gid(ctx->target_ns_uid, ctx->target_ns_gid))
-		goto on_error;
+	if (conf->cap_inheritance) {
+		ret = lxc_set_keepcaps();
+		if (ret < 0)
+			goto on_error;
+
+		if (!lxc_switch_uid_gid(ctx->target_ns_uid, ctx->target_ns_gid))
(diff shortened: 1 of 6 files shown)


stgraber:
@hallyn I see capabilities and I think of you, can you take a look?

hallyn:
Interesting.  I'm good with the goal, and probably the patch, but
let me think about the ambient caps just a bit.

src/lxc/caps.c
```diff
@@ -323,4 +323,46 @@ bool lxc_proc_cap_is_set(cap_value_t cap, cap_flag_t flag)
...
+	call_cleaner(cap_free) cap_t caps = NULL;
+	int ret;
+	cap_value_t cap;
+
+	caps = cap_get_proc();
+	if (!caps)
+		return log_error_errno(-1, errno, "Failed to retrieve capabilities");
+
+	for (cap = 0; cap <= CAP_LAST_CAP; cap++) {
+		if (cap_get_bound(cap) <= 0)
```
hallyn, line 338: For your purposes, is there any reason to set the whole bounding set, rather than just the requested lxc.caps.keep set, as ambient?

src/lxc/attach.c
```diff
@@ -1368,8 +1368,21 @@ __noreturn static void do_attach(struct attach_payload *ap)
 		lxc_seccomp_close_notifier_fd(&conf->seccomp);
 	}
 
-	if (!lxc_switch_uid_gid(ctx->target_ns_uid, ctx->target_ns_gid))
-		goto on_error;
+	if (conf->cap_inheritance) {
```
hallyn: Is there a better name we could use for this?

maybe lxc_conf member is `conf->nonroot_keepcaps`, and the lxc.conf keyword is `lxc.cap.nonroot = true`?

Open to better names.  But someone coming in to look at the code without this in mind will mis-interpret `cap_inheritence`.

src/lxc/caps.c
ElJeffe: Only the requested caps (in lxc.cap.keep) will be bound. All other caps will be dropped from the bounding set [here](https://github.com/lxc/lxc/blob/main/src/lxc/conf.c#L2938). I added logs locally to verify this.

src/lxc/attach.c
ElJeffe: updated

hallyn (approved):

hallyn:
Thanks.

